PR #80, opened 2013-08-18
Suppresses erroneous Corrupt Jpeg messages sent from the jpeg library

knight-of-ni:
Certain Foscam cameras and their clones stream jpeg images that appear fine visually but have some kind of minor issue that causes the jpeg library to continually report a warning.  

Prior to this patch, Zoneminder would fill the log with events like this:
zmc_m1 10017 WAR Corrupt JPEG data: 1 extraneous bytes before marker 0xd9 zm_jpeg.cpp 63

This patch suppresses these warnings

### DIFF
--- a/src/zm_jpeg.cpp
+++ b/src/zm_jpeg.cpp
@@ -59,7 +59,8 @@ void zm_jpeg_emit_message( j_common_ptr cinfo, int msg_level )
 		 */
 		if ( zmerr->pub.num_warnings == 0 || zmerr->pub.trace_level >= 3 )
 		{
-			(zmerr->pub.format_message)( cinfo, buffer ); 
+			(zmerr->pub.format_message)( cinfo, buffer );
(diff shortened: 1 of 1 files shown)


src/zm_jpeg.cpp
```diff
@@ -59,7 +59,8 @@ void zm_jpeg_emit_message( j_common_ptr cinfo, int msg_level )
 		 */
 		if ( zmerr->pub.num_warnings == 0 || zmerr->pub.trace_level >= 3 )
 		{
-			(zmerr->pub.format_message)( cinfo, buffer );
```
kylejohnson, line 62: Isn't this line the same as the one below it?

knight-of-ni, line 62: That was not intentional, but now that I take a closer look at the two I see that the proposed deletion has an extra whitespace character at the end while the new line does not.

PR #154, opened 2013-09-18
Add support to change order of monitors

WDKevin:
Added ability to drag and drop monitors on the dashboard to reorder them

### DIFF
--- a/web/app/Controller/MonitorsController.php
+++ b/web/app/Controller/MonitorsController.php
@@ -4,14 +4,14 @@
   
 		public function index() {
       $zmBandwidth = $this->Cookie->read('zmBandwidth');
-	  $this->set('width', Configure::read('ZM_WEB_LIST_THUMB_WIDTH'));
+      $this->set('width', Configure::read('ZM_WEB_LIST_THUMB_WIDTH'));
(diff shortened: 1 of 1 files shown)


web/app/Controller/MonitorsController.php
```diff
@@ -172,7 +172,16 @@ public function add() {
 			}
 		}
 	}
+    public function reorder() {
+      $this->log(print_r($this->data['Monitor'],true));
+      foreach ($this->data['Monitor'] as $key => $value) {
```
kylejohnson, line 177: Since you can only resort one monitor at a time, shouldn't this array always only contain one item?

WDKevin, line 177: The positions of 2 different monitors changes

web/app/Controller/MonitorsController.php
```diff
@@ -161,18 +161,26 @@ public function edit($id = null) {
...
+        }
+      }
+    }
 
-	}
+    public function reorder() {
+      foreach ($this->data['Monitor'] as $key => $value) {
+        $this->log($value);
+        $this->Monitor->id = $value;
+        $this->Monitor->saveField('Sequence', $key + 1);
```
kylejohnson, line 180: This isn't the right function to use - that is saving a new row each and every time instead of updating old rows.

PR #213, opened 2013-10-18
Initial zone editing in html5 canvas

kylejohnson:
If you navigate to /zones, you'll get a table of monitors with their respective zones.  If you click on the zone, you'll be taken to the zone edit page.   This page is still in its infancy and has a couple of bugs (you need to click on reset for the image to show).  You can click around on the image and then click 'Done' to fill in the 'zone'.  Eventually 'Done' will save these values to the database as the legacy skin does.  Also need to add the rest of the zone configuration form elements.

This is mostly a proof-of-concept that zones can be done in html5 canvas.  Please criticize.
screen shot 2013-10-18 at 10 12 30 am
screen shot 2013-10-18 at 10 12 40 am
screen shot 2013-10-18 at 10 13 05 am
screen shot 2013-10-18 at 10 13 12 am

### DIFF
--- a/web/app/Controller/ZonesController.php
+++ b/web/app/Controller/ZonesController.php
@@ -5,6 +5,14 @@ class ZonesController extends AppController {
 	$this->set('zones', $this->Zone->find('all'));
     }
 
+	public function edit($zid = null) {
+		$zone = $this->Zone->find('first', array('conditions' => array('Zone.Id' => $zid)));
+		$mid = $zone['Monitor']['Id'];
+		$this->Zone->createSnapshot($mid, $zid);
+		$this->set('zoneImage', '/img'.'/Zones'.$mid.'.jpg?'.time());
+
+	}
+
 }
 
 ?>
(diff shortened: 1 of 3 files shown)


web/app/Controller/ZonesController.php
```diff
@@ -5,6 +5,14 @@ public function index() {
 	$this->set('zones', $this->Zone->find('all'));
     }
 
+	public function edit($zid = null) {
+		$zone = $this->Zone->find('first', array('conditions' => array('Zone.Id' => $zid)));
+		$mid = $zone['Monitor']['Id'];
+		$this->Zone->createSnapshot($mid, $zid);
+		$this->set('zoneImage', '/img'.'/Zones'.$mid.'.jpg?'.time());
```
kylejohnson, line 12: Just noticed that I'm setting the path here hard.  The functionality will still work but until I fix this to have the Model function return the path to the image (or create it here), you won't be able to view the image in the zone edit page.  My bad.

PR #240, opened 2013-11-03
remove extra stuff that I don't think we need because we are the source.  Opinions?

connortechnology:
Also unconditionally call zmupdate.pl since it should do all the checking of versions to see if anything needs doing.
We don't need to be doing that work in a postinst bash script.

This also means that we don't have to update postinst anymore when we release a new version.

### DIFF
--- a/distros/ubuntu1204/changelog
+++ b/distros/ubuntu1204/changelog
@@ -1,4 +1,4 @@
-zoneminder (1.26.4+1-SNAPSHOT2013110301) precise; urgency=high
+zoneminder (1.26.4+1-SNAPSHOT2013110302) precise; urgency=high
 
   * Today's snapshot of master
 


distros/ubuntu1204/postinst
```diff
@@ -2,44 +2,32 @@
...
-                        mysqladmin --defaults-file=/etc/mysql/debian.cnf -f reload
-                        # test if database if already present...
-                        if ! $(echo quit | mysql --defaults-file=/etc/mysql/debian.cnf zm > /dev/null 2> /dev/null) ; then
-                            cat /usr/share/zoneminder/db/zm_create.sql | mysql --defaults-file=/etc/mysql/debian.cnf
-                            echo 'grant lock tables, alter,select,insert,update,delete on zm.* to 'zmuser'@localhost identified by "zmpass";' | mysql --defaults-file=/etc/mysql/debian.cnf mysql
-                        fi
+	#
+	# Get mysql started if it isn't
+	#
+	if ! $(/etc/init.d/mysql status >/dev/null 2>&1); then
```
kylejohnson, line 9: I would argue for calling 'server mysql status' over the init.d script

connortechnology, line 9: Doesn't exist on ubuntu

PR #351, opened 2014-03-21
Stability fixes for ffmpeg streams

Sune1337:
The commit messages really say it all, but to summarize it; I had some issues when using ffmpeg sources. My camera went red at random times. These commits fix the cause of those problems.

Its my first pull request and participation in a open source project. Just let me know if I'm doing something wrong.

### DIFF
--- a/src/zm_ffmpeg_camera.cpp
+++ b/src/zm_ffmpeg_camera.cpp
@@ -46,6 +46,7 @@ FfmpegCamera::FfmpegCamera( int p_id, const std::string &p_path, int p_width, in
     mIsOpening = false;
     mCanCapture = false;
     mOpenStart = 0;
+    mReopenThread = 0;
 	
 #if HAVE_LIBSWSCALE    
 	mConvertContext = NULL;
@@ -110,6 +111,20 @@ int FfmpegCamera::Capture( Image &image )
 {
     if (!mCanCapture){
         return -1;
+    }
+    
+    // If the reopen thread has a value, but mCanCapture != 0, then we have just reopened the connection to the ffmpeg device, and we can clean up the thread.
+    if (mReopenThread != 0) {
+        void *retval = 0;
+        int ret;
+        
+        ret = pthread_tryjoin_np(mReopenThread, &retval);
+        if (ret != 0){
+            Error("Could not join reopen thread.");
+        }
+        
+        Info( "Successfully reopened stream." );
+        mReopenThread = 0;
     }
 
 	AVPacket packet;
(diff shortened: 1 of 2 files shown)


src/zm_ffmpeg_camera.cpp
```diff
@@ -212,7 +128,20 @@ int FfmpegCamera::Capture( Image &image )
...
+                (avResult == AVERROR_EOF || (mFormatContext->pb && mFormatContext->pb->eof_reached)) ||
+                // Check for Connection failure.
+                (avResult == -110)
+            )
+            {
+                Info( "av_read_frame returned \"%s\". Reopening stream.", errbuf);
+                ReopenFfmpeg();
+            }
+
+            Error( "Unable to read packet from stream %d: error %d \"%s\".", packet.stream_index, avResult, errbuf );
```
mastertheknife, line 159: Whats the point of showing this error and returning, if we just re-opened ffmpeg? Shouldn't we retry instead?

src/zm_ffmpeg_camera.cpp
```diff
@@ -260,4 +189,207 @@ int FfmpegCamera::PostCapture()
...
+    return 0;
+}
+
+int FfmpegCamera::ReopenFfmpeg() {
+
+    Debug(2, "ReopenFfmpeg called.");
+
+    mCanCapture = false;
+    pthread_t thread1;
+    if (pthread_create( &thread1, NULL, ReopenFfmpegThreadCallback, (void*) this) != 0){
```
mastertheknife: Is doing this in another thread absolutely required? It adds complications and requires exclusive access to shared data to protect against race conditions. e.g. FfmpegCamera::Capture() starts the ffmpeg reopen thread and returns because reading packet failed. In the meanwhile, the new thread is reopening ffmpeg. What happens if zm's main thread is calling Ffmpeg::Capture() while the other thread is busy reopening ffmpeg? It can happen if reopening ffmpeg takes too long. Without proper care, race conditions can occur and they are difficult to debug.

src/zm_ffmpeg_camera.cpp
```diff
@@ -260,4 +189,207 @@ int FfmpegCamera::PostCapture()
...
+        // Log a fatal error and exit the process.
+        Fatal( "ReopenFfmpeg failed to create worker thread." );
+    }
+
```
mastertheknife, line 333: A call to pthread_join() is nowhere to be found. Its required for freeing resources used by the new thread.
Also, adding a call to pthread_join() here will solve the race condition issue because this thread (zm's main thread) will block until the reopen ffmpeg thread is done.

src/zm_ffmpeg_camera.cpp
Sune1337, line 159: well i see your point; I agree that it's redundant to log errors while trying to reconnect; it should only log an error for the frame that triggers the reconnect.

src/zm_ffmpeg_camera.cpp
Sune1337, line 333: I think the mCanCapture member was supposed to keep track of this. But i cant see that variable beeing used anywhere now. Ill checkout the code and get back.

vajonam:
I am still seeing this as an issue w/ ffmpeg, it doesn't help with the disconnects.

Issue #271, opened 2013-12-16
incomplete path to arp in zoneminder/skins/classic/views/monitorprobe.php causes probe code 127 failures


rinchen:
Hi,
In zoneminder/skins/classic/views/monitorprobe.php  the arp probe command is simply "arp" and on many systems it fails with return code 127. The fix is to put in the full path , e.g. /usr/sbin/arp.  Usually most systems will have arp in /usr/sbin/arp so this may be  ok to permanently change. Another option would be to add a configure option to specify it's location.

Thanks,

Joey

ERRVV1N:
Change in monitorprobe.php $result = exec( escapeshellcmd($command), $output, $status ); to $result = shell_exec( escapeshellcmd($command), $output, $status ); and it works (Ubuntu)

knnniggett:
@ERRVV1N You have triple posted to an issue that has already been solved.

aplocher:
This still causes error 127 using version 1.28.1 on Ubuntu Desktop 15.10 (latest version) - fresh install, apt-get installation of zoneminder + all dependencies it prompted for.  

The command 'type' is POSIX command, but the -p argument is internal to **BASH**.  By default on this version of Ubuntu, it's using /bin/sh and it returns this result:

Array ( [0] => -p: not found [1] => arp is /usr/sbin/arp )

It's trying to use "-p" as a command, and it's still getting a bad exit code.  Just for fun, I changed the line of code in monitorprobe.php to this and it worked fine:

$command = "/bin/bash -c \"type -p arp\"";

But changing it to this broke it again with the same result as before:

$command = "/bin/sh -c \"type -p arp\"";

I don't think this is a good fix, since you're relying on bash to exist.  I think you should probably use 'which' which is a pretty standard command on any Linux system I believe.  

Also, take a look at the SECOND answer here: 
http://stackoverflow.com/questions/12424787/how-to-check-if-a-shell-command-exists-from-php

knnniggett:
@aplocher Your late to the party. :-)

We can't use "which" as described in the second answer because that causes similar problems for a different set of users, as described somewhere by another issue.

There have been a few commits after this to further improve the code such as: 13c6be548d7852a15eb2c0a1010a2dd6b9b688b8 13c6be548d7852a15eb2c0a1010a2dd6b9b688b8 and 7ef503ff591a5fd15e23552dc83bc1cbcb51bcce

Rather than look at it piecemeal, you should look at the source code here:
[...]

PR #391, opened 2014-04-18
Supply full path to arp binary

hamiltont:
Addressing #271 - this fixes an issue with detecting remote camera's by using arp to probe the network


ERRVV1N:
Change $result = exec( escapeshellcmd($command), $output, $status ); to $result = shell_exec( escapeshellcmd($command), $output, $status ); and it works (Ubuntu)

connortechnology:
Um... shell_exec may be the answer, but it takes different arguments and has other issues..

PR #404, opened 2014-04-25
Zms/videostream improvements

Sune1337:
Support streaming with zms (using ffmpeg) better …

```
* ZoneMinder compiles with ffmpeg 0.5..master
* Make VideoStream class able to stream through ffmpeg better.
* Ability to use fixed quality instead of fixed bitrate (by specifying bitrate 0...100)
* Format url parameter supports <format>/<encoder> syntax
```

See commit message for details.

### DIFF
--- a/src/zm_mpeg.cpp
+++ b/src/zm_mpeg.cpp
@@ -491,7 +491,7 @@ VideoStream::~VideoStream( )
     if (packet_buffers) {
         delete packet_buffers[0];
         delete packet_buffers[1];
-        delete packet_buffers;
+        delete[] packet_buffers;
     }
 	
 	/* close each codec */


src/zm_mpeg.cpp
```diff
@@ -130,139 +163,178 @@ void VideoStream::SetupCodec( int colours, int subpixelorder, int width, int hei
...
+		opicture = avcodec_alloc_frame( );
 		if ( !opicture )
 		{
 			Panic( "Could not allocate opicture" );
 		}
-		int size = avpicture_get_size( c->pix_fmt, c->width, c->height);
-		uint8_t *opicture_buf = (uint8_t *)av_malloc(size);
+		
+		int size = avpicture_get_size( c->pix_fmt, c->width, c->height );
+		uint8_t *opicture_buf = (uint8_t *)av_malloc( size );
```
mastertheknife, line 333: Shouldn't av_mallocz be used here? 
Oops, looks not, my bad.

src/zm_mpeg.cpp
```diff
@@ -291,127 +363,250 @@ void VideoStream::OpenStream()
...
+		{
+			Error( "pthread_mutex_destroy failed" );
+		}
+		delete buffer_copy_lock;
+	}
+    
+    if (packet_buffers) {
+        delete packet_buffers[0];
+        delete packet_buffers[1];
+        delete packet_buffers;
```
mastertheknife: `delete packet_buffers;`
This is an array, it should be deleted like an array:
`delete[] packet_buffers;`

PR #421, opened 2014-05-05
Ffmpegoptions

m-bene:
ffmpeg and also libVLC support lots of settings to fine-tune communications with the camera, verbosity of debug output etc.
This enhancement adds an "Options" Field to the source tab of the monitor definition for ffmpeg and libVLC types.
image
The options entered here are passed on the the libraries on opening
libVLC Example: `--rtsp-tcp,--verbose=2` use tcp transport mode for rtsp and increase verbosity to 2. debug messages show up if you run the monitor from the command line.
ffmpeg example:`rtsp_transport=tcp` to use tcp transport mode instead of udp (it used to be possible to set this by appending ?tcp to the rtsp url, has been disabled in newer ffmpeg versions)
for both libraries, multiple options can be set separated by ",".
Does not support ffmpeg with LIBAVFORMAT_VERSION_INT < AV_VERSION_INT(53, 4, 0), option / parameter structure is different there.

### DIFF
--- a/web/skins/classic/views/monitor.php
+++ b/web/skins/classic/views/monitor.php
@@ -1,7 +1,7 @@
 <?php
 //
 // ZoneMinder web monitor view file, $Date$, $Revision$
-// Copyright (C) 2001-2008 Philip CoombesMethod
+// Copyright (C) 2001-2008 Philip Coombes
 //
 // This program is free software; you can redistribute it and/or
 // modify it under the terms of the GNU General Public License


src/zm_ffmpeg_camera.cpp
```diff
@@ -110,12 +111,23 @@ void FfmpegCamera::Terminate()
...
 #else
-    if ( avformat_open_input( &mFormatContext, mPath.c_str(), NULL, NULL ) !=0 )
+    // Handle options
+    AVDictionary *opts = 0;
+    StringVector opVect = split(Options(), ",");
+    for (size_t i=0; i<opVect.size(); i++)
+    {
+    	StringVector parts = split(opVect[i],"=");
+    	if (parts.size() > 1)
+    		av_dict_set(&opts, trimSpaces(parts[0]).c_str(), trimSpaces(parts[1]).c_str(), 0);
```
mastertheknife: Line #128
I think should we should log (warning) any rejected options by ffmpeg, by checking the return value of av_dict_set()

mastertheknife: Also, adding a Debug() call here, just like for the one you are doing for libvlc will be probably useful here for troubleshooting purposes.

src/zm_ffmpeg_camera.cpp
```diff
@@ -110,12 +111,32 @@ void FfmpegCamera::Terminate()
...
-    if ( avformat_open_input( &mFormatContext, mPath.c_str(), NULL, NULL ) !=0 )
+    // Handle options
+    AVDictionary *opts = 0;
+    StringVector opVect = split(Options(), ",");
+  	Debug(2, "Number of Options: %d",opVect.size());
+    for (size_t i=0; i<opVect.size(); i++)
+    {
+    	StringVector parts = split(opVect[i],"=");
+    	if (parts.size() > 1) {
+    	    if ( av_dict_set(&opts, trimSpaces(parts[0]).c_str(), trimSpaces(parts[1]).c_str(), 0) == 0 ) {
```
m-bene: @mastertheknife Yes, Makes sense to check the return value and add debug message. Check of return value is not as usefull as I'd wish though: setting option "foo" to "bar" is accepted without error. Doesn't break anything so no big problem, but feedback would have been nice

PR #447, opened 2014-06-27
Introduction of CI tests

kylejohnson:
First shot at adding some tests to travis.  This involved rearranging the travis.yml file and then adding a script (run_tests.sh) that gets run at the end of the build.  Let's see what happens!


SteveGilvarry:
#611 has all these changes included I will add some basic tests for the web server and cameras. Then we should be able to implement. Closing this in preference to #611

PR #458, opened 2014-07-05
Determine full path to arp

### LINKED ISSUE #271: incomplete path to arp in zoneminder/skins/classic/views/monitorprobe.php causes probe code 127 failures

knight-of-ni:
Fixes #271

### DIFF
--- a/web/skins/classic/views/monitorprobe.php
+++ b/web/skins/classic/views/monitorprobe.php
@@ -287,7 +287,15 @@ $macBases = array(
 );
 
 unset($output);
-$command = "arp -a";
+// Calling arp without the full path was reported to fail on some systems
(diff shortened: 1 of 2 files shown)


ERRVV1N:
Change in monitorprobe.php $result = exec( escapeshellcmd($command), $output, $status ); to $result = shell_exec( escapeshellcmd($command), $output, $status ); and it works (Ubuntu)

PR #459, opened 2014-07-06
Update rules

whopperg:
There was a missing tab-indent. So dpkg-buildpackage under debian didnt work:

Error:
debian/rules:77: **\* missing separator.  Stop.

### DIFF
--- a/distros/debian/rules
+++ b/distros/debian/rules
@@ -74,4 +74,4 @@ override_dh_auto_test:
 
 .PHONY: override_dh_strip
 override_dh_strip:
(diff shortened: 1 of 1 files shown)


distros/debian/rules
```diff
@@ -74,4 +74,4 @@ override_dh_auto_test:
 
 .PHONY: override_dh_strip
 override_dh_strip:
-        dh_strip --dbg-package=zoneminder-dbg
```
whopperg, line 77: Spaces-Indent, no Tab-Indent

Issue #434, opened 2014-05-27
Zoneminder Readthedocs Documentation


knnniggett:
This issue has been created as a place holder to discuss what content should go into the documentation that is viewable from http://zoneminder.readthedocs.org

For starters there are two main topics:
Installation Guide
User Guide

I'm not entirely sure how much Installation documentation we should add since it will be distro-specific and change more frequently.

PR #475, opened 2014-07-13
Initial attempt to migrate wiki to readthedocs #434

SteveGilvarry:
Off the back of #434 I made some place holders for Installation Guide brought some info from readme in, but need to document PPA install process, and a full cmake build for Ubuntu for example. For the User Guide I started migrating the wiki content from http://www.zoneminder.com/wiki/index.php/Documentation. This is mainly to give a starting point for a new version of the documentation. No images included yet. Now starting to review and update to 1.27+

### DIFF
--- a/src/zmu.cpp
+++ b/src/zmu.cpp
@@ -416,6 +416,11 @@ int main( int argc, char *argv[] )
 			{
 				printf( "Monitor %d(%s)\n", monitor->Id(), monitor->Name() );
 			}
+			if ( ! monitor->connect() ) {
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -50,9 +50,9 @@ This command will add a new http monitor.
 Edit monitor 1
 ^^^^^^^^^^^^^^
 
-This command will change the 'Name' field of Monitor 1 to 'tits'
```
kylejohnson, line 53: haha

PR #526, opened 2014-10-06
Added preset/autostop/IR support to FI8908W, new module for all Foscam IP cameras

sjthespian:
The update to FI8908W.pm adds support for presets, autostop, and IR on/off support. This fixes a bug in the pull request I submitted the other day.

This pull request also contains a new file, FoscamIPCam.pm. This is an all-purpose Foscam module that should work for any camera supporting the commands at http://www.foscam.es/descarga/ipcam_cgi_sdk.pdf. I don't have a camera to test with that supports iris, focus, and zoom, so it's possible I have the values reversed for those (i.e. zoom in instead of zoom out).  I attempted to simplify the original module by getting rid of duplicated code and creating a hash of the decode command values instead of having them hard-coded into the methods. I also added code to handle inverted cameras, it swaps the up/down and left/right commands if Orientation is set to inverted.

FYI, there appears to be a bug in the Foscam command doc. On my cameras left and right are reversed from what is in the spec and this corresponds to the values that are in the initial FI8908W module. It's possible that it is just an issue with these cameras, or it may be true for all Foscam cameras. If anyone can verify this with an additional Foscam model, I can fix the module appropriately.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Control/FI8908W.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Control/FI8908W.pm
@@ -241,7 +241,7 @@ sub presetSet
 	my $params = shift;
 	my $preset = $self->getParam( $params, 'preset' );
 	Debug( "Set Preset $preset" );
-	my $cmdnum = 30 + ($preset*2);
+	my $cmdnum = 30 + (($preset-1)*2);
 	$self->sendCmd( 'decoder_control.cgi?command=$cmdnum&' );
 }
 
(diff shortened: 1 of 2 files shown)


knnniggett:
Closing due to inactivity.  Adding ptz support requires the necessary sql dB entries, along with the control script.

connortechnology:
I may have to re-open.

knnniggett:
Are we done with this one?  It can no longer be auto-merged.

sjthespian:
Sorry, I haven't had any time to work on this. Go ahead and close the pull request and I'll resubmit when I find some time to fork off the current release and rework my changes.

PR #546, opened 2014-10-31
Bootstrap skin

kylejohnson:
I started to fix bugs in the flat skin and decided to start over... ish.

Reusing most of the existing ZM framework - just changing client side to look better and be easier to maintain.  That means using bootstrap for css.

This PR is not ready to merge, but I would like feedback as I go through and make my changes.  This PR should however work fine if it were merged - it just isn't complete.

2014-10-31-163845_1894x692_scrot


kylejohnson:
It is about time that this PR could use some testing.

You /should/ be able to clone my fork, change to the 'bootstrap-skin' branch,  and then `cp` the web/skins/bootstrap directory into your existing ZoneMinder install.

Wait, there are also a ton of API changes, so you'd need to do the same for web/api, making sure to preserve /web/api/app/Config/database.php

newburns:
Whoa.... I get an all text-based window when selecting bootstrap skin.
The Apache error is:
`File does not exist: /var/www/html/skins, referer: http://10.12.22.180/zm/index.php?view=options&tab=skins`
I am using the RPM from connortechnology at zmrepo.
The Zoneminder RPM installs into /usr/share/zoneminder/
There are two paths, one for SQL db changes and the www
/usr/share/zoneminder/www
/usr/share/zoneminder/db

On top of that, I am using a remote MySQL DB. I have adjusted my 
/usr/share/zoneminder/www/api/app/Config/database.php
to match my remote MySQL Settings. The default settings were there.

kylejohnson:
@newburns Try changing view= from 'options' to 'console' and see if that makes a difference.

newburns:
Nope. That does not work either:
`File does not exist: /var/www/html/skins, referer: http://10.12.22.180/zm/index.php?view=console&tab=skins`
I can try and `ln -s` the web directory from /usr/share/zoneminder/ to /var/www/html/

newburns:
Sorry. That did not work either.
And I'm assuming I would have to do a lot of Apache configuration to get it to follow that symlink into the /usr/share/zoneminder/www directory
`Symbolic link not allowed or link target not accessible: /var/www/html/skins, referer`

ndobbs:
Can you set options directive in your apache vhosts to allow symlinks?

Options FollowSymlinks

newburns:
Options are already set. They were set previously.
The /usr/share/zoneminder/www was owned by root:root. I adjusted the chmod to 775 and change ownership recursively to apache:root. Not sure what else needs to be done, but technically, the apache www user can access the files, and the apache directive should allow the symlink. Do I need to symlink the complete directory or just the skins directory. I only did the skins directory

/etc/httpd/conf/httpd.conf
`DocumentRoot "/var/www/html"

#
# Each directory to which Apache has access can be configured with respect
# to which services and features are allowed and/or disabled in that
# directory (and its subdirectories).
#
# First, we configure the "default" to be a very restrictive set of
# features.
#
<Directory />
    Options FollowSymLinks
    AllowOverride None
</Directory>`

/etc/httpd/conf.d/zoneminder.conf
`Alias /zm "/usr/share/zoneminder/www"
<Directory "/usr/share/zoneminder/www">
    Options -Indexes MultiViews FollowSymLinks
    AllowOverride All
    Order allow,deny
    Allow from all
    # The code unfortunately uses short tags in many places
    php_value short_open_tag 1

</Directory>`

justinlawrence:
Hi @kylejohnson 

"I started to fix bugs in the flat skin and decided to start over... ish."

Could you elaborate on why you decided to start over?

kylejohnson:
@newburns I'm not sure of your issue right now.  I'm able to spin it up on my machines repeatably (Ubuntu, with ZM from source).

kylejohnson:
@justinlawrence Sure can.

TL, DR: The web frontend is old, and it is showing its age.  I'm not talking about `<table>`s, I'm talking about the code under the web/ dir.  It is difficult to maintain.  Code is with content, there are /tons/ of vulnerabilities. 

When Phil wrote ZoneMinder, he did a fantastic job and essentially created a web framework in PHP, because there wasn't much out there.  Nowadays, we have tons of options.  Phil 'created' the wheel, if you will.

One specific reason for starting over is to separate code from content, and the frontend from the backend.  I decided to move all of the server-side logic (php) to a RESTful API backend, and to build a web client that is modern, fast, and easy to maintain.  I chose the CakePHP framwork for the backend, and the AngularJS framework for the frontend, because they both seemed like (and still do) intelligent options for what I wanted to accomplish.

These frameworks allow me to 'start over', while not having to recreate the wheel.

Why do you ask?

kylejohnson:
@justinlawrence And with that said, simply slapping in a new skin/ would only address a subset of the actual problems.  It would allow one to make the UI prettier, but that is about it.

newburns:
I'm assuming you compile into /var/www/html
I'm using the zmrepo install on Centos.
http://localhost/zm has Apache directives to resolve to /use/share/ZoneMinder/www

kylejohnson:
@newburns Actually, `/var/www/zm`

Here you can see that I symlink `api/` and `skins/bootstrap` right into my git clone of the code.

justinlawrence:
> Why do you ask?

@kylejohnson Thanks for the detailed response. I just wanted to get a sense of the bigger picture. I like all that you've suggested. I would also back AngularJS for the frontend.

Once you have this functionality split, we will happily have another go at creating a skin. I've always felt that this would assist in widespread adoption of Zoneminder.

AngularJS will also allow us to add a bit of delight to the way things work.

kylejohnson:
I'm going to close this PR, and create a second one which targets a branch off of master, instead of master itself.  That way I can get this into the hands of more people earlier, without having to worry about screwing things up in master.

kylejohnson:
This code is now tracked in the angular-ui branch.

PR #611, opened 2014-11-30
Reorganise Travis and add Saucelabs for testing

SteveGilvarry:
Takes what Kyle started in #447 and what I was doing in #610, I went away and cleaned up some of the excessive commits needed to test options in Travis. This PR will fail to build as it has my repo details and sauce lab account in the secure variables in the travis.yml file. Here is the build from my Travis account showing it works, https://travis-ci.org/SteveGilvarry/ZoneMinder/jobs/42546173, as soon as we swap those secure variables it will pass Travis and Saucelabs using the zone minder account.
- Splits Travis into scripts to clean up
- Sets up tunnel between Saucelabs and Travis VM to allow browser based tests
- Used Selenium Builder to build a basic test that ZM was running on web interface
- Uses SE Interpreter NPM to convert and run json tests built by Selenium Builder
- Also calls JS to send results of Selenium tests back to Saucelabs, and updates the test status badge displaying on Readme
- SE Interpreter also allows me to run the same tests on multiple browsers in the same run, so far setup firefox and chrome.


.travis.yml
```diff
@@ -1,52 +1,33 @@
 language: cpp
 notifications:
-  irc: "chat.freenode.net#zoneminder-dev"
+  irc: chat.freenode.net#zoneminder-dev
 branches:
   except:
-    - modern
-addons:
-  sauce_connect:
```
kylejohnson, line 8: Do we not need the sauce_connect credentials?

.travis.yml
```diff
@@ -1,52 +1,33 @@
...
-    username: "zoneminder"
-    access_key: "046ec7c1-c598-4e7e-949a-f86e725d1722"
+  - modern
 env:
   global:
-    - LD_LIBRARY_PATH="/usr/local/lib:/opt/libjpeg-turbo/lib:$LD_LIBRARY_PATH"
```
kylejohnson, line 13: Picky, but I'd rather any indentation fixes be in a separate PR.  They're really only noise in the larger focus of this PR.

utils/travis/run-tests.sh
```diff
@@ -10,14 +10,15 @@ with_timestamps() {
...
-	sudo zmu -l
-	sudo grep ERR /var/log/syslog
+	mysql -uzmuser -pzmpass < ${TRAVIS_BUILD_DIR}/db/zm_create.sql
+	mysql -uzmuser -pzmpass zm < ${TRAVIS_BUILD_DIR}/db/test.monitor.sql
 	sudo zmpkg.pl start
 	sudo zmfilter.pl -f purgewhenfull
+	sudo cp -f utils/travis/apache-vhost /etc/apache2/sites-enabled/000-default
+	sudo service apache2 restart
+	npm install -g se-interpreter
+	se-interpreter --listener=utils/tests/sauce_listener.js utils/tests/interpreter_config.json
```
kylejohnson: Would you mind giving me a high-level overflow of the flow of how this works?
Between firefox being spun up, and the selenium test being run, how does the listener fit in?

.travis.yml
SteveGilvarry, line 8: No they instead live in the se-interpreter config, it was a while ago that I got it all working the first time so I forget exactly why we don't use them with the add-on setup. But the tunnel creation doesn't use the logon, only the running of tests

.travis.yml
SteveGilvarry, line 13: Noted, did you want me to try and fix? I think it is here because -secure is in that block, so I fixed it as a section.

utils/travis/run-tests.sh
SteveGilvarry: Yep I am going to document it.
se-interpreter, allows you to create Selenium tests with SE Builder, or I think it is now called Selenium Builder, Firefox add-on that lets you build the tests in page as a recording. If you give me some examples you want to test I will try and write them, initial test was just proof of concept.
Using the interpreter config json for se-interpreter, that controls the tunnel setup, under Driver, and then the Suacelabs connection and test setup using the Browser section.
Away it goes and kicks off the tests and runs them, and writes the results to the command line in Travis. But Selenium has no way to update Saucelabs with the results of the test, so you end up with a bunch of tests that just say Finished, but don't evaluate if the test works. The final piece is that listener, which is part of se-interpreter, it listens for results of selenium tests and uses the API to update SauceLabs with Pass or Fail. If you look at http://saucelabs/u/SteveGilvarry, and go back through the pages of tests you will see when it changed from Finished to Pass. Needed that to make the badge work, and the Browser status matrix as well.

knnniggett:
Let me know what we want to do with this pr.

SteveGilvarry:
#447 could be closed as this builds on that as a baseline and makes it work. I just added some code so that it will only run Selenium tests on SauceLabs on a merge not PR requests as commented about above. Need @kylejohnson to check it out or spend some time explaining it to someone so we can setup Selenium testing of the GUI. If someone lays down some simple GUI tests they want me to build I can add them in here also.

SteveGilvarry:
Can believe I forget our 1 year anniversary pull request, I will make it up to you.
@kylejohnson so other than updating master what else would you like here so we can put some basic tests in and have a cleaner travis setup?

connortechnology:
Hey is this still wanted/relevant?  Maybe we should finish it up and merge? or delete it.

PR #646, opened 2014-12-21
Split the debian package into several packages

manupap1:
Following discussions on topic https://github.com/ZoneMinder/ZoneMinder/pull/605, this pull request splits the debian package into several packages:
- zoneminder (a metapackage to install all packages)
- libzoneminder-perl (the perl libraries and associated man files)
- zoneminder-database (this package installs and maintains the database, it can also be used alone to install the database on other computer)
- zoneminder-core  (the core binaries and perl scripts)
- zoneminder-ui-base (this package installs the essential web files and configures the web server)
- zoneminder-ui-classic (the classic web user interface)
- zoneminder-ui-mobile (the mobile web user interface)
- zoneminder-ui-xml (the xml interface)

I have made several changes to the dependencies in the control file:
- Removal of compiled libraries because this stuff is added automatically at build time by ${shlibs:Depends} 
- Removal of libpolkit-gobject-1-0 (policykit-1 brings this dependency) 
- Update of perl libraries :
 - Added: libdbi-perl, libimage-info-perl
 - Replaced: libmodule-load-perl -> libmodule-load-conditional-perl (which also brings dependency to libmodule-load-perl)
 - Removed: libio-pty-perl, libexpect-perl (I have not found call to modules provided by these packages)


connortechnology:
Just a heads up... if there is a - in the version it won't build.  I suggest using something like zoneminder-1.28.0+1

pluses are good, minuses bad.  

Havn't been able to get this to build yet. Will merge when I succeed.

connortechnology:
why are we using quilt instead of native?

manupap1:
I updated the branch to fix a merge conflict and to add a dependency to apache2 in zoneminder-core package (apache2 is required by systemd script). 

I did not encountered build issues with `-` in version.
According [debian wiki](https://wiki.debian.org/DebianMentorsFaq#What_is_the_difference_between_a_native_Debian_package_and_a_non-native_package.3F) , `-` is used to set the debian source package to `non-native`.
This should not be an issue as long as the `orig.tar.xz` file is up-to-date.

quilt is still used because of new patch added (see https://github.com/ZoneMinder/ZoneMinder/pull/646#issuecomment-67769630).
I think the only clean way to address this issue without `quilt` would be to keep the service file name `zoneminder.service` and manage this file manually in maintainer scripts.
Currently `dh_systemd` manages the file automatically by adding some instructions in maintainer scripts at build time, but this requires file name  `zoneminder-core.service` and file `zmsystemctl.pl.in` needs to be patched accordingly to call `zoneminder-core`.
And natively merge the patch does not sound reasonable to me because distro's restrictions should not affect ZM sources.

connortechnology:
My problem is that I don't want to have to generate .orig.tar.gz everytime I edit something.  I want it auto-generated.

manupap1:
So indeed the native source format is required for this.

But can we consider the source format as 'native'?

By reading the debian wiki, I don't think so because ZM is not purely developed for Debian and a fix on the upstream source is currently required to package it cleanly.

This is maybe a question for the debian guys.

connortechnology:
I think debian will always use their own version of the packaging scripts.  I want us to make it easier for them by using ours as a template, but they will likely always have other things going on.

 So they would use quilt, but we would use native.  (I think).

manupap1:
I will check if we can override the default behaviour of dh_systemd to keep the name of service file to 'zoneminder.sercice'.
If we can, the debian patch is no longer required and we can easily switch to native
If we can not, this requires some extra modifications in files rules, zoneminder-core.postinst, zoneminder-core.postrm zoneminder-core.preinst and zoneminder.install (at least) to get ride of the debian patch. Debian's guys will probably ask questions about these modifications.

manupap1:
Debian developers are smart people!
I switched the debian source to native.
I added 'nmu' in the version number as advised in debian documentation (this removes a lintian warning for native packages without a debian revision number).

connortechnology:
Dang, I merged this cuz it seemed to work, but now... the zoneminder package depends on zoneminder-perl, but we don't seem to be building it.

connortechnology:
ok, nevermind, it's libzoneminder.

connortechnology:
I've been testing this on ubuntu, without systemd.  The only real issue is the db config stuff.  I refuses to do a remote config, so it gets tricky to get the job done.   Other than that, seems to be working well.  I will be doing more with it to play with zm on pi

connortechnology:
also... installing zm should never overwrite /etc/zm/zm.conf

manupap1:
Yes the installation can be tricky but this is due to the possibility to choose the database location and some design choice in dbconfig.

I added some information in debconf screens to help users to overcome those issues but I think I missed a point when there is no local database available and a user wants to install the database locally. In this case reconfiguring zoneminder-database package is not enough because the user has to fix installation of broken packages with 'sudo apt-get -f install'. I will send a commit to fix this.

About zm.conf: From my point of view, the criterion that must prevail is the security. One of debconf goals is exactly to address this issue by offering the possibilty to replace default/static passwords and yes this may lead to modify some files installed by the package itself but this case is allowed and documented. So I will not remove this feature.

PR #651, opened 2015-01-03
Remove no longer needed patch because applied to master

manupap1:


knnniggett:
Interesting.
On the RPM packaging side of things we would set INSTALLDIRS in the rpm spec file, when performing make install:
`make install DESTDIR=%{buildroot} INSTALLDIRS=vendor`

PR #654, opened 2015-01-03
replace getLoad() / getDiskPercent() with PHP native functions

lifeofguenter:
instead of allowing dangerous "system-exec" calls (which additionally spawn an additional process, so not very good performance wise) we should rely on native PHP calls where we can.

Ideally future releases would communicate via TCP or similar to zmc (or other parts) instead of doing exec calls.

### DIFF
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -1611,19 +1611,14 @@ function sortTag( $field )
 
 function getLoad()
 {
-    $uptime = shell_exec( 'uptime' );
-    $load = '';
-    if ( preg_match( '/load average: ([\d.]+)/', $uptime, $matches ) )
-        $load = $matches[1];
-    return( $load );
+    $load = sys_getloadavg();
+    return( $load[0] );
 }
 
 function getDiskPercent()
 {
-    $df = shell_exec( 'df '.ZM_DIR_EVENTS );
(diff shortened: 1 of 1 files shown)


connortechnology:
Please note that sys_getloadaverage is only available for php > 5.1.3
I don't have a problem with that... our use of PDO probably means php5 anyways.

connortechnology:
yeah PDO is php5.  So I'm happy with this PR.

kylejohnson:
This is what I'm doing in my new skin, and I haven't seen any issues.

lifeofguenter:
Thanks @kylejohnson for the merge.

Just a note on PHP versions: PHP4 is deprecated and everything below 5.4 is not supported anymore officially and contains currently _known_ vulnerabilities.
PHP versions <= 5.3 should not be supported anymore and it makes no sense to offer backward compatibility - _especially_ for version 4x. PHP 5.1.4 itself has been released 2006.

Short: especially as we are dealing with sensitive data, lets try to use a "sane" PHP version as minimum and not try to support legacy versions. Let that be a job of commercial support services (perhaps?).

connortechnology:
agreed

knnniggett:
CentOS 6 ships with PHP 5.3 so we need to support that.

PR #657, opened 2015-01-03
Make the log export use ZM_PATH_SWAP and report the full path on error

connortechnology:


knnniggett:
Oh, look at that.  It was hard coded.

Issue #626, opened 2014-12-10
Version detection


connortechnology:
Since we removed the ZM_VERSON for /etc/zm/zm.conf it is slightly more difficult to determine what version of ZM is running.

I propose that all zm programs should take a --version argument and report their version.

PR #664, opened 2015-01-07
Versions command line args

connortechnology:
This implements a -v and a --version or just a version command as appropriate to the various programs that make up ZoneMinder.

Why?  Because we no longer have the version in the config file and so it is slightly harder to get the current version in a script or whatnot.


connortechnology:
Solves #626

knnniggett:
The Travis failure looks bogus so I restarted the job.

PR #665, opened 2015-01-07
Don't fail if an unexpected rtp packet type is received

knnniggett:


knnniggett:
I want to test this one more time before we merge

PR #671, opened 2015-01-11
Fix arp can not be found problem in Ubuntu 14.04

liucougar:
(this fix should work for other distros/versions too)


knnniggett:
No, this doesn't work for all systems.  We changed the code to use "type", rather than "which", several months ago because "which" wasn't working.  This pull request would cause a regression.  If you think the current code is not working, then create an issue and describe in detail.  Note that the current code was tested on Ubuntu.

connortechnology:
type works on ubuntu 14.10 too.

PR #675, opened 2015-01-13
small performance improvement when streaming.

connortechnology:
This may not be the way to go.

From what I read, we really should be using std::string instead of doing it the C way.

In any event, at least not calling strlen(swap_path) many times is good.  
PATH_MAX is 4096 on Linux, and is actually filesystem dependent, so it is silly to even worry about it.

So this saves us 4k-whatever length the swap_path actually is on the stack.  

comments?

### DIFF
--- a/src/zm_monitor.cpp
+++ b/src/zm_monitor.cpp
@@ -3940,7 +3940,7 @@ void MonitorStream::runStream()
     temp_read_index = temp_image_buffer_count;
     temp_write_index = temp_image_buffer_count;
 
-    char *swap_path;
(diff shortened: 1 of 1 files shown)


knnniggett:
What you've got looks fine to me, but admittedly this is an area of zoneminder where I have not taken the time to get a deep understanding of what is going on.

connortechnology:
what I have here has a memleak of course.  I have the fix for it, just havn't committed it yet.  

The debate is guess is the use of malloc to allocate such a small amount of mem vs using and dealing with 4k on the stack

PR #682, opened 2015-01-22
Monitorprobe - User selectable ARP tool

cspenceiv:
The arp command is considered a legacy utility now. This replaces arp with
iproute2's ip neigh command.

One side effect: The $host variable is just set to $ip since ip neigh does not
use DNS. DNS resolution can be added back in with several different utilities
if desired. But DNS resolution tends to slow down response time.

### DIFF
--- a/db/zm_update-1.28.99.sql
+++ b/db/zm_update-1.28.99.sql
@@ -1,0 +1,6 @@
+-- 
+-- New values for the Config table
+-- 
+insert into Config set Id = 107, Name = 'ZM_PATH_ARP', Value = '/sbin/ip neigh', Type = 'string', DefaultValue = '/sbin/ip neigh', Hint = '/absolute/path/to/somewhere', Pattern = '(?-xism:^((?:/[^/]*)+?)/?$)', Format = ' $1 ', Prompt = 'Path to a supported ARP tool', Help = 'The camera probe function uses Address Resolution Protocol in order to find known devices on the network. Supply the full path to \"ip neigh\", \"arp -a\", or any other tool on your system that returns ip/mac address pairs.', Category = 'paths', Readonly = '0', Requires = '';
+
+
--- a/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
@@ -655,6 +655,14 @@ our @options =
 		category => "paths",
 	},
 	{
+		name => "ZM_PATH_ARP",
+		default => "/sbin/ip neigh",
+		description => "Path to a supported ARP tool",
+		help => "The camera probe function uses Address Resolution Protocol in order to find known devices on the network. Supply the full path to \"ip neigh\", \"arp -a\", or any other tool on your system that returns ip/mac address pairs.",
+		type => $types{abs_path},
+		category => "paths",
+	},
+	{
 		name => "ZM_WEB_TITLE_PREFIX",
 		default => "ZM",
 		description => "The title prefix displayed on each window",
--- a/web/skins/classic/views/monitorprobe.php
+++ b/web/skins/classic/views/monitorprobe.php
@@ -287,27 +287,26 @@ $macBases = array(
 );
 
 unset($output);
-// Calling arp without the full path was reported to fail on some systems
+// Calling ip without the full path was reported to fail on some systems
 // Use the builtin unix command "type" to tell us where the command is
-$command = "type -p arp";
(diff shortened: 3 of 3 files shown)


cspenceiv:
This change will reduce the number of required dependencies needed for full functionality whereas iproute2 is used by most major distributions as default (Debian, RHEL, and Arch to name a few).

knnniggett:
We need to make sure we don't exclude freebsd.
I've got an old copy of 8.3 running in a vm and it does not appear to have this command.  

Looking on the FreeBSd ports page, the closest I could find is hs-iproute, but that does not appear to be the same project.

I use FreeBSD only rarely so there maybe there is something I'm missing.

cspenceiv:
You would be correct. They continued to maintain the tools that the Linux variants (legacy net tools) were based on.

I was unaware that ZoneMinder was supported on FreeBSD. The installation page made no mention of it. I did find that the main website still has a <a href="http://www.zoneminder.com/wiki/index.php/FreeBSD">page</a> for FreeBSD, but the tar ball is a broken link and the "documentation" is a thread from 2007 for version 1.2.3.

There were some patches for porting to BSD, but those too are no longer hosted. If those patches are still maintained somewhere, I would gladly work to push a fix since FreeBSD still maintains what are legacy tools for Linux.

cspenceiv:
See <a href="https://github.com/ZoneMinder/ZoneMinder/issues/683">issue 683</a>. The code in monitorprobe.php doesn't work on FreeBSD 10.1 anyway.

We can continue using tools on the Linux side that haven't been updated in over ten years (and does not support IPv6), or we can go this direction and write in new logic to fix the BSD issues while supporting the latest tools for both OSes. I personally think it would be beneficial to primarily update for Linux and come back to fix the FreeBSD shortfall for this utility that is not required for working-functionality.

knnniggett:
As shown in our pull request history, we have had requests from the freebsd community.  We know there is an interest.
This morning @connortechnology mentioned someone was developing zoneminder to run on freenas in a jail, for example.  

I'm not in favor of intentionally causing breakage in any operating system.  As you pointed out in #683, it looks like we do need to make a change, but we need to find a solution that works for everyone.  For starters we need to dig a little deeper into the freebsd world and find out what tools are reasonably available to us.

cspenceiv:
Sounds good to me. I can expand on this fix to include FreeBSD then.

SteveGilvarry:
MonitorProbe.php seems to be limited to a small selection cameras with default admin passwords. So if I can be the devil's advocate do we actually need it still? I have never used it and my IP cameras are not on the list so interested if we still have a use case for probing?

knnniggett:
I'll do some digging to see if perhaps we could implement the functionality natively in php.  No idea if that is even possible, but I do seem to need an excuse to learn php better.  I can't seem to get through my php book w/o a specific reason to do it/problem to solve, etc.

@SteveGilvarry 
You are right that monitorprobe is very limited.  @connortechnology and I were talking about this the other day.  In a perfect world where we could spend all our time on zoneminder, I think we would design some kind of hybrid probe that would scan local video devices, compare an arp request against a static table of known values (e.g. what monitorprobe does), and use the new ONVIF functionality to find ONVIF compliant cameras.  

Realistically, I think we ought to just tweak monitorprobe for now and then point our focus on other big ticket items like video container, etc.  Of course we welcome pull requests from anyone willing to do the work.

kylejohnson:
@knnniggett Is the `ip` command in 8.4?  8.3 is no longer maintained.

knnniggett:
It is not.  However, I'm going to make the relevant command a user changeable option that lives in the dB.

cspenceiv:
@knnniggett Got it. I'll look out for it.

I like the proposed approach. It's much easier than trying to guess commands for each distribution/OS.

cspenceiv:
@kylejohnson History/context: BSD doesn't use the <a href="http://en.wikipedia.org/wiki/Iproute2">iproute2 suite</a> at all since they have been maintaining the original net-tools the Linux variant was based upon. Linux development stopped with net-tools over ten years ago and the focus turned to iproute2.

knnniggett:
@cspenceiv Just sent a new pull request to you.  I'm sure it will need additional tweaking.

cspenceiv:
Merged. I'll probably be able to double check everything tomorrow.

knnniggett:
We made some changes in #709 that affects this.  I still plan to create another pr that adds the option for the user to pick the correct tool

knnniggett:
This pull request has been superseded by #723 
Consequently, I am going to close this one out.  Please test #723 and let me know your comments.

Issue #685, opened 2015-01-28
RTSP fails when hostname is used for source


dgilmour77:
ZM console allows an RTSP source URL to be entered as a host name, and will send the host name in the URL when performing an RTSP DESCRIBE.  However, when the source provides a control url, it may do so with an IP address, not a hostname.  This causes the logic added in the following fix:

https://github.com/ZoneMinder/ZoneMinder/commit/e764938a0d652a0e24db0fa6dde416b583a281df

...to fail, generating a 404 error in response to a SETUP.  This is shown in the following zmc log file based on 1.28+.  Close to the end of this log excerpt, it is clear that the URL passed along with the SETUP is malformed because the logic in the fix above does not support a hostname/IP mismatch:

[begin log excerpt]

01/27/15 19:01:40.085098 zmc_m1[3722].DB2-/home/david/zoneminder/src/zm_rtsp.cpp/74 [Received RTSP response: RTSP/1.0 200 OK
Server: VLC/2.0.3
Content-Length: 0
Cseq: 1
Public: DESCRIBE,SETUP,TEARDOWN,PLAY,PAUSE,GET_PARAMETER

 (124 bytes)]
01/27/15 19:01:40.085215 zmc_m1[3722].DB2-/home/david/zoneminder/src/zm_rtsp.cpp/48 [Sending RTSP message: DESCRIBE rtsp://kam.gilmour.net:8554/cam.sdp RTSP/1.0
User-Agent: ZoneMinder/1.28.0
CSeq: 2

]
01/27/15 19:01:41.086964 zmc_m1[3722].DB2-/home/david/zoneminder/src/zm_rtsp.cpp/74 [Received RTSP response: RTSP/1.0 200 OK
Server: VLC/2.0.3
Date: Wed, 28 Jan 2015 03:01:45 GMT
Content-Type: application/sdp
Content-Base: rtsp://192.168.1.54:8554/cam.sdp
Content-Length: 355
Cache-Control: no-cache
Cseq: 2

v=0
o=- 15596756719339326206 15596756719339326206 IN IP4 kam
s=Unnamed
i=N/A
c=IN IP4 0.0.0.0
t=0 0
a=tool:vlc 2.0.3
a=recvonly
a=type:broadcast
a=charset:UTF-8
a=control:rtsp://192.168.1.54:8554/cam.sdp
m=video 0 RTP/AVP 96
b=RR:0
a=rtpmap:96 H264/90000
[...]

PR #687, opened 2015-01-31
process RTSP DESCRIBE response header

knnniggett:
potential fix for #685

### DIFF
--- a/src/zm_rtsp.cpp
+++ b/src/zm_rtsp.cpp
@@ -366,6 +366,22 @@ int RtspThread::run()
     size_t sdpStart = response.find( endOfHeaders );
     if( sdpStart == std::string::npos )
         return( -1 );
+
+    std::string DescHeader = response.substr( 0,sdpStart );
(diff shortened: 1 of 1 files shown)


connortechnology:
This looks fine to me.  

Why did you use size_t for i in your for loop?  unsigned int is the usual type...

knnniggett:
No good reason other than that is how the other for loops are set up in zm_rtsp, I was simply copying & pasting

sogood007:
FYI: size_t generally is the correct/better type for value return by size() .  It matters when running in 64b vs 32b.  size_t automatically adjusts. unsigned int will get compilation warning in 64b build and size() return 64b in 64b build.

connortechnology:
uh huh. Thing is, this is just a simple for loop counter.  It is not a pointer.  

I did some reading on the matter... seems it is common practice these days to use size_t.. I'm not convinced though as it is not intuitive.  However I just don't care.  

here is a good article: http://www.codeproject.com/Articles/60082/About-size_t-and-ptrdiff_t

PR #691, opened 2015-02-04
this fixes Digest Auth for the mjpeg stream on a TV-IP302PI

connortechnology:
Digest Auth worked for the RTSP stream, but not http.
This adds quotes around the cnonce value, which matches what curl and wget do and adds the algorithm line which also matches what curl and wget do.  With these two changes the auth is accepted.


kylejohnson:
Looks fabulous

knnniggett:
nice work.

PR #692, opened 2015-02-04
fix content-type parsing when there are options on it

connortechnology:
I have a camera a Trendnet TV-IP302PI that when it send the mjpeg stream it gives it a content-type of image/jpeg;size=

this is legal.  You can have options in the content-type.  So I've added some code to find the ; and drop the rest.


kylejohnson:
Look groovy

knnniggett:
looks good

PR #696, opened 2015-02-06
Improve delete event

connortechnology:
I noticed that deleting an event globbed the entire event structure under a monitor.  That made for some pretty bad performance on my systems with a lot of events.

This converts the deleteEvent function to handle an event hash or an eid and use the StartTime value to figure out which day the event happened and so we only glob that day which drastically reduces filesystem hits.

### DIFF
--- a/web/includes/actions.php
+++ b/web/includes/actions.php
@@ -574,9 +574,10 @@ if ( !empty($action) )
                             // well time out before completing, in which case zmaudit will still tidy up
                             if ( !ZM_OPT_FAST_DELETE )
                             {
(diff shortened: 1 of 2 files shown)


connortechnology:
Don't merge yet, it has a problem.

connortechnology:
should be ok now

knnniggett:
Looks good to me.
If I understand this correctly, this pr will only make a difference if if FAST_DELETE is off?

Obviously, if FAST_DELETE is on, we are letting the zmaudit daemon do the deleting.... but do we know if zmudit has similar code that could also be improved in the same manner?

connortechnology:
correct, I havn't looked at zmaudit yet.  I've just been auditing the web code trying to fix sluggishness.

PR #699, opened 2015-02-08
tmpfile.conf for systemd

clipo1979:
tmpfile to go with the service file for systemd intergration


knnniggett:
We are already doing this in our packaging scripts under /distros so I assume you want this for an open source build, right?  We can do that.

There are a couple things that need to be changed before we can merge this, however:
- Replace the hard coded web user account with the @WEB_USER@ and @WEB_GROUP@ macros
- Update autotools to do the same thing. <- just follow the pattern for how the zoneminder.service file is processed
- What is the "+" for in the config file?  I don't see that defined in the documentation: http://www.freedesktop.org/software/systemd/man/tmpfiles.d.html

clipo1979:
Knnnight, 

Thank you for the comments I should have completed the coding instead of being lazey.
This was originaly for a CentOS 7 build but was also needed for some other builds so I thought it would be good to complete the systemd config files within the misc directory.

I'm confused about the additional + I was about to say its a typo.

Hopefully should have updated the autotool conf files also just copied the zoneminder.service as you suggested.

knnniggett:
Perfect.  Thank you for doing the work. 
Nevermind the comment about the "+".  That was github notation.  Due to the way it showed on the screen, I misinterpreted that to be part of the config file when it was not.

PR #700, opened 2015-02-08
Centos 7 rpm packaging

knnniggett:
Initial upload of files needed for creation of a zoneminder rpm on CentOS 7.


connortechnology:
I don't know if anyone else is qualified to verify this PR.  So as far as I am concerned, you can merge it yourself.

knnniggett:
OK, thanks!

PR #702, opened 2015-02-10
small improvements:

connortechnology:
First is some more debugging lines useful for figuring out why automatica palette selection didn't work.
The other is a comment about how the code will segfault if there are more than 64 formats.  Currently there are not.
Lastly witch a variable from int to unsigned int.  There is no way for it to go negative and logically should always be positive, so by making it an unsigned in we don't have to cast it later.

### DIFF
--- a/src/zm_local_camera.cpp
+++ b/src/zm_local_camera.cpp
@@ -1122,6 +1122,7 @@ uint32_t LocalCamera::AutoSelectFormat(int p_colours) {
 	memset(&fmtinfo, 0, sizeof(fmtinfo));
 	fmtinfo.index = nIndex;
 	fmtinfo.type = V4L2_BUF_TYPE_VIDEO_CAPTURE;
+	// FIXME This will crash if there are more than 64 formats.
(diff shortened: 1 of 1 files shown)


kylejohnson:
@connortechnology Looking at this PR, I see what should be three separate commits.  I'm a bit anal when it comes to this stuff, because seeing 'small improvements' in a commit log really doesn't tell us much.  You know we appreciate the work, and I know it can be a pain, but please try to split up your commits like you've done in other PRs. :)

connortechnology:
I think you didn't click the ... to actually see the commit messages. 

I agree with what you are saying but I think you are being a little pedantic when the commits are a comment, two debug lines and one other one line change.

Which is also why I merged it myself to try to fix a problem a user was having. If there was any meat here at all I would have put more effort in.

kylejohnson:
You're right, I am being pedantic, but we can't out right call that a bad thing.  And you have merit as well - this was such a small change, that it almost doesn't matter.

All I'm saying is that smaller individual commits would be better, even if you do have a larger description for the single commit.

Anyway, it is all good.

PR #706, opened 2015-02-17
Update Version.txt to 1.28.1

SteveGilvarry:
FYI This file in the Github web is checked by zmupdate for latest version information, need to remember to update it with a release.


knnniggett:
The entire master branch revision has not been updated to 1.28.1 yet e.g. in cmakelist, confiure.ac, etc.  This is intentional.

I don't think we should make two commits, one on the release-1.28.1 branch and the other in master that do the the same thing.  The last time we released, we did something similar and ended up with merge conflicts, which got overlooked.

Instead I *think* we should cherry pick from 1.28.1 back to master, but I wanted to dicuss it with the team first before I did just in case there is not some better way.

knnniggett:
@connortechnology is working on an issue with the Ubuntu 1.28.1 packages.  I don't want to complicate the troubleshooting and will wait until that issue is resolve.  Following that, I will merge the changes from 1.28.1 back into master, which should get the master branch updated to 1.28.1 in all the right spots.

SteveGilvarry:
@knnniggett This is not in Master, it is gh-pages branch, which is just the github.io webpage. So will have no impact on release/master merging just controls what zmupdate sees as the latest version.

knnniggett:
Oh, fair enough.  I didn't notice that.

PR #707, opened 2015-02-17
Merging Video and feature-h264-videostorage update

SteveGilvarry:
**Branch still a work in progress so if you update let me know any issues, but be warned**
Some more work I did on using the video and jpeg control options at the camera level from video branch and applying them to feature-h264-videostorage. I believe the c++ side is looking OK. I think I need to implement some improvements in the capture function made in recent times into the captureandrecord function as they should mirror each other for JPEG grabbing.
Now working on the GUI, event.php will playback,using videojs cdn link. Seem some issues in FF using a VM with video control responsiveness, Safari into the VM content worked fine.


knnniggett:
We very much appreciate the amount of work you have put into this @SteveGilvarry

PR #708, opened 2015-02-17
remove redundant state property of the Monitor class.

connortechnology:
Maybe don't merge this without us doing some testing.

As near as I can tell the state property of the Monitor class is redundant.  It is only used in Analyse. 

So I was thinking about it, and the only reason I can think of for it to be there is to lessen the effects of concurrent changes to the state.  Think triggers, etc.  

I suspect that we should have a lock around the state.  Thoughts?


kylejohnson:
@web2wire any thoughts here?

knnniggett:
where are we at with this one?

PR #709, opened 2015-02-17
try harder to find arp.

connortechnology:
If type -p doesn't work, try which.  If that doesn't work, just look for /usr/sbin/arp.

Also don't bother with escapeshellcmd because these are not user inputs.

The reason for this instead of pr #682 is that this is a minimal fix that we could include in 1.28.2.

Once this is merged, I intend to take #682 and improve upon it to use either ip or arp. For now though, just want a quick fix for 1.28.2.

### DIFF
--- a/web/skins/classic/views/monitorprobe.php
+++ b/web/skins/classic/views/monitorprobe.php
@@ -292,7 +292,7 @@ unset($output);
 $arp_command = '';
 $result = exec( 'type -p arp', $output, $status );
 if ( $status ) {
-    Warning( "Unable to determine path for arp command, type -p arp returned '$status'" );
+    Warning( "Unable to determine path for arp command, type -p arp returned '$status' output is: " . implode( "\n", $output ) );
     unset($output);
     $result = exec( 'which arp', $output, $status );
     if ( $status ) {
--- a/web/skins/classic/views/monitorprobe.php
+++ b/web/skins/classic/views/monitorprobe.php
@@ -289,12 +289,25 @@ $macBases = array(
 unset($output);
 // Calling arp without the full path was reported to fail on some systems
 // Use the builtin unix command "type" to tell us where the command is
-$command = "type -p arp";
-$result = exec( escapeshellcmd($command), $output, $status );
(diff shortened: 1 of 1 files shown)


knnniggett:
My memory is coming back to be regarding this.
Recalling the error message reported by others, they are seeing:
`Unable to probe network cameras, status is 127`

I think this is an important clue that we are *not* seeing the message:
`Unable to determine path for arp command`

Looking at the php, that means the first time we use exec to try and find the arp tool actually succeeds. Or perhaps we can merely say that $status is zero.

Because of that, I think this PR would suffer from the same issue. The first exec command, calling "type -t arp",  returns success, and consequently the other methods to find arp will be skipped. We then try to exec the command, using arp, and run into "status 127".

I can still merge this if you want me to, but I thought I'd mention that I think the error is occurring further down, around L312.  

In pr #682 I added the is_executable command, which should give us the ability for better error handling, rather than just output "127".

connortechnology:
127 is the return code of type -p arp... when it fails.  We don't actually print/log the output of type -p arp, we just give up due to the error code. I'm pretty sure that type -p arp isn't succeeding.  I am sure of this because logging in as www-data and running it gives an error.  
This PR does in fact work on my system (ubuntu 14.10).  It falls back to which arp, which works and arp is run successfully.

connortechnology:
let me improve the error output/logging... don't merge yet.

connortechnology:
So here's the deal. If www-data's shell is /bin/sh, type doesn't support the -p parameters. So the first line of output is -p: not found. The second line is /usr/sbin/arp.  So in fact it is succeeding.
If www-data's shell is /bin/bash, then type supports -p and everything works.

I kinda think we should go without the -p, and just parse the output stripping off 'arp is: ' if it is there.

knnniggett:
There is some kind of Saucelabs VM tunnel that keeps timing out.  Not even sure exactly what that is.  I've restarted the Travis build.

knnniggett:
See #723 for an addition to this pull request.

PR #711, opened 2015-02-18
Fixes #710 Added libavformat version check around free context functions

### LINKED ISSUE #710: zm_rtsp Use of avformat_free_context not version checked
As per forum http://www.zoneminder.com/forums/viewtopic.php?f=32&t=22935

https://github.com/ZoneMinder/ZoneMinder/blame/master/src/zm_rtsp.cpp#L214

Using avformat_free_context here, but this is an new API function. https://github.com/FFmpeg/FFmpeg/blob/master/doc/APIchanges
2011-02-04 - f124b08 - lavf 52.96.0 - avformat_free_context()
Add avformat_free_context() in avformat.h.

Old version was av_free_format_context()
Need to version check for the correct version of ffmpeg or libav before picking a function.

I will look at submitting a PR for this in the next day or so

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/8679224-zm_rtsp-use-of-avformat_free_context-not-version-checked?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SteveGilvarry:
Just put in standard version check code for new ffmpeg API functions.

### DIFF
--- a/src/zm_rtsp.cpp
+++ b/src/zm_rtsp.cpp
@@ -211,7 +211,11 @@ RtspThread::~RtspThread()
 {
     if ( mFormatContext )
     {
+#if LIBAVFORMAT_VERSION_INT >= AV_VERSION_INT(52, 96, 0)
         avformat_free_context( mFormatContext );
+#else
+        av_free_format_context( mFormatContext );
+#endif
         mFormatContext = NULL;
     }
     if ( mSessDesc )


knnniggett:
nice catch @SteveGilvarry 
Fix for #710 <- to get github to link the two issues.

Looks good, but one question.  Did you verify whether or not (our nemesis) libav has this new function call?

knnniggett:
Can't seem to find a reference to this call in libav, but this might be due to user error.

SteveGilvarry:
Is from 2011, which is fork year, but I think it was pre-fork so same change is in both, hence no need for a micro version >100 check here.

https://github.com/libav/libav/blob/master/doc/APIchanges#L1055

PR #713, opened 2015-02-19
Check to make sure that skin and css are valid.

connortechnology:
This checks installed skins and associated css choices for validity and will default to classic.  This should fix the master CSS complaints when upgrading from 1.28.*


knnniggett:
Some kind of weird timeout happened with Travis.  Restarting the build to see if it was just a hiccup....

knnniggett:
Well, I wasn't sure.  I guess that means I am getting better at reading php.

PR #714, opened 2015-02-20
Update fr_fr.php

Jypy:


knnniggett:
@Jypy Thank you for taking the time to update the language file.

Jypy:
You are welcome :)

manupap1:
This update replaces some special characters by bad characters.
This is probably due to a bad file encoding type (it should be UTF-8).

Jypy:
Strange, I had issues with the "original" so I made those changes. Now I have again issues with the update you've made ?!

manupap1:
I have made no update to the file after your changes.

Jypy:
You're totally right! My mistake. I updated to 1.28.1 and I thought the file was changed again. But that was just the original file back :)
I'll change the encoding on mine and see if it fix on your side.

Jypy:
Looks like it's already UTF-8.

PR #718, opened 2015-02-21
Removed el6 from endif arguments

SteveGilvarry:
CMakeLists.txt was throwing a non critical error as If and EndIf arguments didn't match, just removed the matching argument that was removed in https://github.com/ZoneMinder/ZoneMinder/commit/e3351441c7602dc8ed9e11523e785dcf8cd6ca33


knnniggett:
My bad.  Looks like I missed something during a manual merge.

PR #719, opened 2015-02-21
Allow zm to build w/o ffmpeg

knnniggett:
Traditionally, we have maintained the ability to build zoneminder without ffmpeg.
On another note, I was surprised to see ffmpeg calls in the local device code.  Not that is a bad thing... ffmpeg has wiggled its way into most everything.


connortechnology:
yeah I'm not sure how much would work without it.  Might be important on small footprint devices like raspberry pi's and cameras. I will find out I guess.

Eventually I'm going to be putting zmc onto openwrt.. should be fun.  I have a dlink dcs933L running openwrt now....

PR #721, opened 2015-02-21
add onvif ptz control into update script

knnniggett:
First time attempting to create a re-runnable update script.
I need another set of eyes to look at this to make sure I am doing it right.


connortechnology:
Looks good to me, I tested it.

PR #725, opened 2015-02-22
freshen dB when needed

knnniggett:
Older versions of zoneminder called "zmupdate.pl -f" every time zoneminder was started.  However, that was inefficient.  More recent versions of zoneminder went to the other extreme and never called zmupdate.pl -f.  This leads to errors in the log on occasion that ask the user to manually run zmupdate.pl -f.

The logic to detect a Config mismatch already exists in zm_config.cpp.  This PR hooks into that logic and calls zmudpate.pl automatically.

Posting the PR simply for discussion.  
I'm not sure we should merge this as is.  Zoneminder probably needs to be restarted so that the updated config takes effect.

### DIFF
--- a/src/zm_config.cpp
+++ b/src/zm_config.cpp
@@ -104,6 +104,11 @@ void zmLoadConfig()
 	config.Assign();
 }
 
+int zmFreshenConfig()
+{
+	return( system( "zmupdate.pl -f" ) );
+}
+
 StaticConfig staticConfig;
 
 ConfigItem::ConfigItem( const char *p_name, const char *p_value, const char *const p_type )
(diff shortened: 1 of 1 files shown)


connortechnology:
I'm not sure what the right thing to do is.  I find the whole way we do the Config to be too fragile and annoying to add new entries.  However I havn't thought of a better way yet.

I do like that this change would auto-fix the problem.  However it might not.  I'm in a situation right now where zmupdate.pl is not fixing the problem.

connortechnology:
Oh... and in the multi-server case,  when upgrading one server, it updates the db, the old servers revert it... back and forth until all servers get updated... maybe I like a manual update.

knnniggett:
Going to close this one. Perhaps we can use this as a reference down the road. 
In the meantime, I am going to add a "zmupdate -f" command to the post installation scripts to the Zonemidner rpms in zmrepo.

PR #728, opened 2015-02-23
Fix event view

### LINKED ISSUE #717: Issue with event page for users with restricted monitors
If you restrict a user to 2+ monitors and give them view permissions, the event page on monitor 2-X is broken.  User is unable to click on the links and the top header, which includes the time, length, score, etc does not display properly.

Monitor 1 event page works as expected.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8770850-issue-with-event-page-for-users-with-restricted-monitors?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github).
</bountysource-plugin>

connortechnology:
Fixes #717 

The old dbEscape used to put quotes around the values so the preg_split was complex.  Now the monitors should just be comma separated and it only comes from DB so it SHOULD be clean.... so just explode it based on commas. 

No further db escaping is necessary due to our use of PDO.

### DIFF
--- a/web/skins/classic/views/event.php
+++ b/web/skins/classic/views/event.php
@@ -27,15 +27,13 @@ if ( !canView( 'Events' ) )
 $eid = validInt( $_REQUEST['eid'] );
 $fid = !empty($_REQUEST['fid'])?validInt($_REQUEST['fid']):1;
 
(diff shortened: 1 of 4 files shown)


connortechnology:
So... the others didn't actually have a problem.  You might call them a cleanup.  The thing is, $user{'MonitorIds'} should not have quotes, spaces or anything else in it.  We should ensure that when saving that value.  We SHOULD be able to assume that it is a comma-separated list of MonitorIDs for performance reasons.

PR #740, opened 2015-03-03
Add Development note for non-release builds

knnniggett:
This is not ready for merge.  I wrote this very quickly, and it will more than likely need additional love.
However, I wanted to show the idea I was working on and get feedback before I put much more time into this.

This pull request allows zoneminder to autodetect if it is a non-release build and will display a note on the web console accordingly. See the screenshot below.  

It does this via client side Git Hooks.  Server side git hooks would have been even better, but unfortunately github does not allow server-side hooks due to security reasons.

Side Note:
I abandoned the previously discussed idea of using the git commit hash in the development version number once I learned that server side hooks were not available.  We can discuss this if anyone wants to.

Pros:
-It is simple and straight forward
-This PR only displays a message on the console.  It does not require our current release versioning scheme to change.  

Cons:
-Requires strict naming of our release branches in the format "release-x.xx.x".  Maybe not a con.
-We all *have* to make sure and run "makehooks.sh" any time we create a new local repository on our local drive.  Actually, this isn't entirely true.  Just the person(s) who create the release branch will need to do this, but I'd rather we all run the script just in case.

Future:
I will likely expand the use of client-side git hooks to auto-bump the zoneminder version numbers (configure.ac, cmakelistst.txt, version, etc) for us when we create a new release branch.

screenshot-zm - console - mozilla firefox

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,12 @@ include (CheckPrototypeDefinition_fixed)
 include (CheckTypeSize)
 include (CheckStructHasMember)
 
+# Used to alert the user they are running a development snapshot
+set(ZM_DEVELOPMENT "")
+if(EXISTS "./utils/development.txt")
+	execute_process(COMMAND echo "DEVELOPMENT Snapshot built on "date +%D OUTPUT_VARIABLE ZM_DEVELOPMENT)
+endif()
+
 # Configuration options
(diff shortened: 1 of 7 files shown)


knnniggett:
I'm going to close this, as it did not receive approval from the rest of the team.  We can re-open it if we ever want to do something similar.

PR #744, opened 2015-03-11
Install CakePHP via Composer

devinellis:
Rather than checking in the whole CakePHP library, we should be installing it via PHP's package manager, [Composer](https://getcomposer.org/). This PR includes:
- Remove CakePHP library (900+ files)
- Update composer.json to require CakePHP
- Update some config within CakePHP to get it to load properly

This PR does __not__ include:
- Installation of Composer itself (see https://getcomposer.org/download/)
- Automatic installation of Composer packages (CakePHP, but the list could grow)
- Documentation updates

## Composer install instructions:
```sh
cd web/api
curl -sS https://getcomposer.org/installer | php
sudo php composer.phar install
```


kylejohnson:
I've thought about using Composer a few times, but it was too much for me to tackle, along with everything else.  So Composer sounds good.

That said, this PR would need to hand installation of composer - or at least address in documentation that Composer is now a requirement.  We'd also need to fit installation of composer packages into the make / cmake build process.

devinellis:
@kylejohnson I added some documentation to the API page. 

I've never used cmake and poking around at what I see is not helping much. I found a file that configures core.php and default.php. I agree that downloading composer and then installing the packages should be done automatically but I'm not familiar with ZM's build process, I installed it via a PPA I believe, and I hacked the API to get it to work. Is it possible to do these steps within cmake?

devinellis:
ping

kylejohnson:
Hi @devinellis - yes, this can all be done within cmake.  I haven't had the time to figure out the steps required, but I imagine it won't be too much work.

I'm having trouble seeing where composer actually pulls in the zoneminder API code.  commit 8e9e04f1f8d7a3fbefa0f646d6413612b76bd483 is the only changes that I see which might do this.  Could you point me in the right path?

Could you also list the steps required to install cakephp and zm api via composer?

devinellis:
I've added the composer install steps to the PR's description

onlyjob:
My concern is that in Debian we build packages in clean environment without internet access (offline). This means that source tarball must be comprehensive as it will not be possible to download components on build-time. I'd strongly suggest to avoid incorporating any downloadable stuff into build process... Besides downloading may pose a security risk...

davidsainty:
Me too.  Downloading at build time is a huge no-no.

I'm not sure if that is really the suggestion, but ZoneMinder is not a PHP package, it's a system that happens to use PHP for part of the system.  Coupling it too tightly to annoying PHP processes is going to be a hassle for package maintainers.

In Pkgsrc we have neither composer nor cakephp as a package, but we do have a number of other packages that seem to bundle cakephp files in their distribution.  Perhaps that's the way it should be?

knnniggett:
Just catching up on all the open pull requests.  Looks like this one still has some on going discussion so I'm going to label it "not ready for merge".  If the discussion reaches a resolution let me know and I can change the status.

onlyjob:
I think we should reject this distro unfriendly suggestion... Moreover it has potential to  produce unexpected results on upstream changes...

devinellis:
I assume your builds pull from Github at some point? Composer does the same. If it is that important, you might want to consider splitting ZM into two repositories, that way the PHP code can live on its own. Without investing in modernizing the PHP infrastructure, you will have a hard time finding people to contribute to it.

Running a directory analyzer shows that well over half of the entire project is the checked in CakePHP source code. Smells funny to me.

onlyjob:
May be git submodules could be the answer to this problem...

knnniggett:
Going to close this one out since we've chosen not to go this route.

PR #751, opened 2015-03-23
Perl vendor vs site

connortechnology:
Between 1.28.0 and 1.28.1 we changed the command passed to the perl lib build system, adding INSTALLDIRS=vendor.

This changes where the perl lib get install from something like /usr/share/perl5/ZoneMinder to /usr/share/perl/5.20.2 . It also changes what happens when a custom prefix is used.  

So this fixes it.  Needs testing on different dists I think.


knnniggett:
This change works for CentOS 6.  I still need to check centos 7 & fedora.

One thing I did notice is that, if no prefix is specified, Perl modules end up under /usr/share/perl... when building from source, rather than /usr/local/share/perl....
That doesn't sound right.

knnniggett:
Just noticed something on Fedora.
When I build from master and use a prefix of /usr/local, perl modules are placed here:
/usr/local/share/perl5/vendor_perl/ZoneMinder

which is correct.
When I apply this pull request and rebuild, they end up here:
/usr/local/lib/perl5/ZoneMinder

which is not right

connortechnology:
this PR is not correct, just a WIP

knnniggett:
I've been doing a bit of Perl research lately.  
Perl modules installed as part of a package should go into the "vendor" folders.  On Redhat those folders are:
```
[abauer@localhost ~]$ perl '-V:install.*' | grep "/usr/share/perl5/vendor_perl"
installvendorlib='/usr/share/perl5/vendor_perl';
```
And on Ubuntu those folders are:
```
abauer@techeng-wks01:~/git/ZoneMinder$ perl '-V:install.*' | grep installvendorlib
installvendorlib='/usr/share/perl5';
```

There seems to be a gray area (at least in my mind) when zoneminder is built directly from source as to whether they should go under "site" or "vendor".  "site" modules are those Perl modules installed by the system admin, while "vendor" modules are those modules written by a third party (usually packaged).  I can see how both of these could be considered as true when zoneminder is built directly from source.  

However, there seem to be some technical reasons for us to avoid "site" modules.  On Ubuntu, the site modules include the Perl version:
```
abauer@techeng-wks01:~/git/ZoneMinder$ perl '-V:install.*' | grep installsitelib
installsitelib='/usr/local/share/perl/5.20.1';
```

Additionally, Perl no longer supports uninstalling modules via make uninstall. If the user installs from srouce, and then changes to a package there is no automatic way for that user to get the old modules (which would be under "site") off their system as the new modules would be under "vendor".

TLDR:
I think we should always install to the OS'es Perl Vendor libraries, regardless of how ZoneMinder was built or installed.  I'm doing some related work in cmake and plan to make sure the end result is that all modules go into the vendor folders.  Once I am complete, I will move to autotools and verify the same thing.

Let me know what your thoughts are on this... arguments for or against.  
Should I implement the ability to override this?

Issue #753, opened 2015-03-25
Make Uninstall no longer removes Perl modules


knnniggett:
I just noticed the following when issuing a make uninstall from a from-source build:

```
Making uninstall in ZoneMinder
make[2]: Entering directory `/home/abauer/ZoneMinder/scripts/ZoneMinder'

Uninstall is unsafe and deprecated, the uninstallation was not performed.
We will show what would have been done.

unlink /usr/share/man/man3/ZoneMinder.3pm
unlink /usr/share/man/man3/ZoneMinder::Base.3pm
unlink /usr/share/man/man3/ZoneMinder::Config.3pm
unlink /usr/share/man/man3/ZoneMinder::ConfigAdmin.3pm
unlink /usr/share/man/man3/ZoneMinder::ConfigData.3pm
unlink /usr/share/man/man3/ZoneMinder::Control.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::3S.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::AxisV2.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::FI8608W_Y2k.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::FI8620_Y2k.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::FI8908W.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::FI9821W_Y2k.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::M8640.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::Ncs370.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::PanasonicIP.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::PelcoD.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::PelcoP.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::SkyIPCam7xx.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::TVIP862.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::Toshiba_IK_WB11A.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::Visca.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::Wanscam.3pm
unlink /usr/share/man/man3/ZoneMinder::Control::mjpgStreamer.3pm
[...]

PR #763, opened 2015-03-30
POD: zmupdate.pl converted to "pod2usage"

onlyjob:
Best practice for POD documentation is to (re-)use it for displaying usage
 info. Now `perldoc zmupdate.pl` shows expected information and we can even
 generate man page automatically with `pod2man zmupdate.pl`.

I'm looking forward to see more scripts converted like this...

### DIFF
--- a/scripts/zmupdate.pl.in
+++ b/scripts/zmupdate.pl.in
@@ -20,11 +20,33 @@
 # Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #
 # ==========================================================================
-#
-# This script just checks what the most recent release of ZoneMinder is
-# at the the moment. It will eventually be responsible for applying and
-# configuring upgrades etc, including on the fly upgrades.
-#
+
+=head1 NAME
+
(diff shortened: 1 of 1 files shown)


connortechnology:
Yeah that's pretty nice.

onlyjob:
Thanks. :)

knnniggett:
Huh, I had no idea this existed.  Since we've done this to zmupdate.pl, I presume we should make a similar change to all the other zm tools to keep things consistent?

onlyjob:
That would be great. Thanks.

knnniggett:
@onlyjob Since you seem to have a good working knowledge of Perl, would you have any idea what the "new" method is for uninstalling Perl models?  See #753 
It seems they have changed the way Perl modules should be uninstalled, and I am looking for a way to avoid spending an afternoon reading the Makemaker documentation.

Worst case I'll just do something like `while read file ; do rm "$file" ; done < /path/to/.packlist`
but that seems so hackish.

onlyjob:
@knnniggett, sorry mate, I don't know. I never use make (un-)install. I just let Debian package management to do the job for me.

PR #772, opened 2015-03-31
Don't trigger linked cameras on new events

balr0g:
Currently, linked monitors trigger every time a new event is created on the monitor linked to. So if I have a setup as follows:
mon1: mocord
mon2: nodect, linked to mon1

While mon1 has no alarms at all, mon2 triggers with an alarm (cause being Linked) every 10 minutes (or however long the event interval is). This causes unnecessary recording and events.

This patch attempts to fix it by not triggering an alarm in zm_monitor when the only thing that's happened is a new event. It still updates the value of last_event, though this does make me wonder just how last_event is used.

Testing currently in process.

### DIFF
--- a/src/zm_monitor.cpp
+++ b/src/zm_monitor.cpp
@@ -251,7 +251,6 @@ bool Monitor::MonitorLink::hasAlarmed()
     else if( shared_data->last_event != (unsigned int)last_event )
     {
         last_event = shared_data->last_event;
-        return( true );
     }
     return( false );
 }


balr0g:
Tested — this does prevent excessive recording. I have not yet been able to test what happens when an alarm is triggered.

balr0g:
This seems to have the expected effect but I'd like someone else to test it.

knnniggett:
No ignoring you.  Just been tied up in other things. 
I have never looked at zm_monitor.cpp indetail, but this is such a trivial change, I'll just go ahead and merge it into master.  It can be un-changed easily enough if there is some kind of unintended side-effect.

Issue #429, opened 2014-05-08
FreeBSD pbi update [$70]


petrklus:
Hi,

I was looking into getting ZM installed on my BSD machine but the latest version I could get was 1.25:

http://ftp.pcbsd.org/pub/pbioutgoing/amd64/9/multimedia/zoneminder/

Are there any plans on updating to latest 1.27?
##

NCommander:
I'm going to take a stab at updating the PBI to 1.28.1, I'll keep you apprised of my status.

josh4trunks:
@NCommander you mind hosting your endevors to update the freebsd port on github. I want to be able to follow and possibly contribute if I find time.

I worked on this a bit before. I had a port that compiled, but I think running zms or some of the other binaries had some issues.

NCommander:
I got horridly sidetracked on this and haven't had time to resume it. I
didn't get very far; I don't believe I even had the tree compiling let
alone test >.<;
Michael

connortechnology:
I had it compiling once... and recently took another look and found that it doesn't anymore.

Mostly the issues are just including the right include file.

I also had it compiling on kfreebsd for debian..

connortechnology:
I have fixed the basic autoconf/cmake build issues.

Sauraus:
@connortechnology I am about to give this a go all in my self interest of running ZoneMinder in a FreeBSD jail ;) where can I find your changes?

connortechnology:
They have all been merged

abishai:
It is done
https://www.freshports.org/multimedia/zoneminder/

josh4trunks:
As far as I understand PBIs are no longer used, though I could build one if really necessary, lol.
I haven't played with Docker at all, but that might be a way we could have a standalone (including dependencies) package of Zoneminder distributed.

Either way, this issue can be closed. Zoneminder v1.30 is straight forward to install on FreeBSD.
Not sure how the bounty works, or if it is still available, but @abishai did the bulk of the FreeBSD port work/debugging.

connortechnology:
I am happy to give @abishai my share of the bounty, I do want to say that it was not all him.  He has been the latest player and has put a lot of work in and has been instrumental in bringing the final bits into place. 

However, I would say that I have done at least 25% of the work, and I know @kylejohnson put a significant effort, and there have been others.  I just want to make sure that everyone gets thanked.  

Also, I might like to see the freebsd native local camera code get merged in....

[...]

PR #775, opened 2015-04-01
Freebsd fixes

connortechnology:
This PR allows ZM to build on FreeBSD 10.  Mainly the two issues were sendfile support for which FreeBSD and others have a different parameter list, and the syscall for getting the current pid.


connortechnology:
Anyone want to review these?  Or can I just merge them myself?

connortechnology:
This somewhat solves #429 . It's still up to the FreeBSD ports people to update to it, but someone can now checkout master+this branch and it will compile and work on freebsd.

josh4trunks:
@connortechnology Thanks for your changes!
Were you planning on updating the FreeBSD port as well? If not, I'll work on it, though I only have moderate FreeBSD porting experiencing so it may take me a while.

connortechnology:
I don't intend to put any further effort into updating FreeBSD ports.  I will happily merge any patches they send.

So go ahead.

josh4trunks:
OK, thanks I'll work on it as time permits.

My preliminary notes are here https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=190268#c7

There's an issue with scripts/ZoneMinder/CMakeLists.txt because FreeBSD needs to use gmake (GNU make) but this can be patched in the FreeBSD port.

Issue #558, opened 2014-11-06
V4L1 Support removal?


SteveGilvarry:
I was reviewing Debian package to identify patches that might be able to be integrated upstream and came across the one below around videodev.h. I guess here is as good a place as any for a discussion on maintaining V4L1 support as it has been dropped out of kernel. Just asking the question as it might be superfluous.

Debian Patch to swap it to libv4l1-videodev.h.

Use libv4l1-videodev.h from libv4l-dev (>= 0.8.3-1), as it is no 
longer shipped as linux/videodev.h in linux-libc-dev.
http://bugs.debian.org/619813
# Index: ZoneMinder-1.26.4/configure.ac

--- ZoneMinder-1.26.4.orig/configure.ac 2013-12-03 17:10:55.458414363 +1100
+++ ZoneMinder-1.26.4/configure.ac  2013-12-03 17:11:00.822290301 +1100
@@ -303,7 +303,7 @@
 AC_CHECK_HEADERS(ucontext.h,,,)
 AC_CHECK_HEADERS(sys/syscall.h,,,)
 AC_CHECK_HEADERS(pthread.h,,,)
-AC_CHECK_HEADERS(linux/videodev.h,AC_SUBST(ZM_HAS_V4L1,1),AC_SUBST(ZM_HAS_V4L1,0),)
+AC_CHECK_HEADERS(libv4l1-videodev.h,AC_SUBST(ZM_HAS_V4L1,1),AC_SUBST(ZM_HAS_V4L1,0),)
 AC_CHECK_HEADERS(linux/videodev2.h,AC_SUBST(ZM_HAS_V4L2,1),AC_SUBST(ZM_HAS_V4L2,0),)
 if test "$ZM_HAS_V4L1" == "1" || test "$ZM_HAS_V4L2" == "1"; then
 AC_SUBST(ZM_HAS_V4L,1)
# Index: ZoneMinder-1.26.4/src/zm_local_camera.h

--- ZoneMinder-1.26.4.orig/src/zm_local_camera.h    2013-12-03 17:10:48.122582847 +1100
+++ ZoneMinder-1.26.4/src/zm_local_camera.h 2013-12-03 17:13:31.874796744 +1100
@@ -26,9 +26,9 @@

 #if ZM_HAS_V4L

-#ifdef HAVE_LINUX_VIDEODEV_H
-#include <linux/videodev.h>
-#endif // HAVE_LINUX_VIDEODEV_H
[...]

PR #776, opened 2015-04-01
build: replace obsolete "linux/videodev.h" with "libv4l1-videodev.h":

onlyjob:
Use libv4l1-videodev.h from libv4l-dev (>= 0.8.3-1), as it is no
 longer shipped as linux/videodev.h in linux-libc-dev.

Note: we had this patch in Debian since 2011. I extended it to cover CMake.


SteveGilvarry:
Related #558

knnniggett:
Perhaps we should check for both. CentOS 6 still uses linux/videodev.h

onlyjob:
I guess I shouldn't be surprised how much CentOS is falling behind... :(
Is this old interface really worth the effort to support on old distros like CentOS-6 knowing that it won't last long? One upgrade and it will be gone...

Is "libv4l-dev" available on CentOS? If so we should be good with just my pull request...

knnniggett:
The RHEL/CentOS philosophy is different than Debian or Ubuntu.  That is by design and is a good thing.  They don't jump on the latest & greatest software for the sake of having the newest stuff.  Generally, the only thing you get as updates are (backported) bug fixes, which means libv4l1-videodev.h will never be available for CentOS 6 (yes, I checked).  The good news is that CentOS 7 uses libv4l1-videodev.h so the light is at the end of the tunnel, but we just aren't quite there yet.

Can you modify your pr to suit?  I can push a change against your pr if you want me to do it.

Since we tend to touch on the "what do we support" issue from time to time, I thought I'd document that I generally follow the "two major releases" rule.  My own goal is to provide support for RHEL/CentOS 6 & 7, Fedora 20 & 21, Ubuntu 12.04 & 14.04, etc.  I don't know how Debian releases their software, but I would expect to follow the same rule.  If anyone feels this warrants discussion, then please create a new issue and we can hammer out what our official policy should be.

onlyjob:
I've been using RHEL/CentOS and Debian professionally for many years and I could never understand why RHEL/CentOS is still around. I reckon that wouldn't happen without business sector being hooked on it. When Debian is available I don't see any reasons to deploy RHEL/CentOS at all. Also philosophy of Debian and Ubuntu is very different. IMHO Ubuntu break so much of what we do in Debian so I would definitely recommend CentOS over Ubuntu but only if Debian is not an option (which is never due to technical reasons).

As for "never to be available for CentOS 6" it is clear that Red Hat won't make it available but someone could package `libvl1` for community repositories. If it is not packaged yet then you might find interesting to know that we had `libv4l` in Debian since 2008 and in 2011 it was released with stable release "Squeeze" (now "oldstable"). Another interesting thing is that CentOS-5 was released with just ~2000 packages while Debian Squeeze had ~30000 (yes, nearly thirty thousand packages).

To learn more about Debian releases you can read the following document :

 * https://wiki.debian.org/DebianReleases

Also we have upstream guide which I would much recommend to read:

* https://wiki.debian.org/UpstreamGuide

As for my commit, please modify it as you wish. Personally I have no intention to support CentOS and I'm not in position to test in CentOS. I've tested `libv4l` adaptation in Debian.
I respect your commitment to support CentOS and I believe you are in much better position to make suitable changes. Thank you.

knnniggett:
see #781

onlyjob:
Awesome. :)

onlyjob:
Fixed by (merged) #781. Thanks.

PR #778, opened 2015-04-03
build: add PATH_MAX definitions (needed on GNU Hurd)

onlyjob:


connortechnology:
I'm not so sure I like this.

First of all, we use PATH_MAX all over the place.  Why does it only need to go into these two files?

Second, we should really be trying to not need to use PATH_MAX.

Third.. I would rather see this happen in one file which is generally included everywhere.  Something like zm_limits.h or maybe it could go into zm.h or something like that.

onlyjob:
Thanks for review, @connortechnology :)

>First of all, we use PATH_MAX all over the place. Why does it only need to go into these two files?

No idea. I merely added 'em where needed. These two are required on GNU/Hurd or compilation stops with error. Determined by experimentation on several build attempts.

> Second, we should really be trying to not need to use PATH_MAX.

Unfortunately if we want ZoneMinder on Hurd we will have to add `PATH_MAX` definitions  somewhere. IMHO it shouldn't be a big deal as it is safe. Besides probably well over 100 packages in Debian have similar workaround...

> Third.. I would rather see this happen in one file which is generally included everywhere. Something like zm_limits.h or maybe it could go into zm.h or something like that.

Agreed. Please feel free to modify as you like and I shall be happy to test. I'm illiterate in CPP so quality of my changes are not great (not to mention I'm not too comfortable making them)...

onlyjob:
Hi @connortechnology, 

I moved `PATH_MAX` definition to `zm.h` as advised. Thanks.

PR #780, opened 2015-04-03
Update de_de.php

seebaer1976:


connortechnology:
I see that this replaces some &thingy; with the actual character.  I guess this is fine, just not sure.

seebaer1976:
hi,
first of all I have translated everything roughly into German as well as utf-8. 
The detailed translation will follow soon.

knnniggett:
So it looks like commit ac8ace33f889366af5079cabbfcb8fb32cf7aafa conflicts with this pull request and we cannot auto-merge it.  @seebaer1976 can you take a look at that commit and see if you can resolve the conflicts?

Issue #506, opened 2014-08-19
Event export with ZM_USE_DEEP_STORAGE option set


DarthBrader:
When the ZM_USE_DEEP_STORAGE option is set, the event export function generates faulty javascript in the zmEventImagesMaster HTML file.  It does not use the deep path format, so the links to the actual event data are broken.  

The issue appears to be with the file  .../skins/classic/includes/export_functions.php.  Someone offered a solution on the ZM forums, but it effectively hard-codes deep paths.  See http://www.zoneminder.com/forums/viewtopic.php?f=29&t=18573.  

This issue has also been reported on Ubuntu on Launchpad.  https://bugs.launchpad.net/ubuntu/+source/zoneminder/+bug/1176128.  

I am running Ubuntu 14.04.1 LTS and ZoneMinder 1.27, which I compiled from source.

onlyjob:
This issue was reported in Debian as well: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=723706

I confirm it is still exist in v1.28.1.

Apparently this is because export is not using `mygetEventPath`...

knnniggett:
When this was reported initially, I thought this described an issue with the location on the filesystem where the export data was written.  Rather, the issue is with one of the generated html files *inside* the exported archive file. And it is just zmEventImagesMaster.  The other files in the exported archives all point to the correct location.  I see now. Merging

PR #782, opened 2015-04-04
skins/classic: fix HTML export with USE_DEEP_STORAGE (Closes: #506).

onlyjob:
Use mygetEventPath() to get valid path.

Signed-off-by: Dmitry Smirnov <onlyjob@member.fsf.org>


onlyjob:
Thanks!

Issue #769, opened 2015-03-31
CMake: incorrect installation of Perl modules [patch]


onlyjob:
When I build Debian package using CMake (_autoconf_ is OK because it uses `INSTALLDIRS=vendor`) Perl modules install into absolute location:

~~~~
Installing /mnt/tmpssd/src/zoneminder/tmp/ZoneMinder-1.28.1/debian/zoneminder/mnt/tmpssd/src/zoneminder/tmp/ZoneMinder-1.28.1/dbuild/scripts/ZoneMinder/output/lib/x86_64-linux-gnu/perl5/ZoneMinder.pm
~~~~

I believe this is happening due to gross disregard to `DESTDIR`. Currently `CMakeLists.txt` is not suitable for build environment because `make install` can not put files into package staging directory. In another words the problem is that `make install` do not put files into directory relative to `DESTDIR` because absolute paths are used when modules are pre-installed into temporary build directory.

To fix the problem I've managed to produce the following patch:

~~~~
--- a/scripts/ZoneMinder/CMakeLists.txt
+++ b/scripts/ZoneMinder/CMakeLists.txt
@@ -23,11 +23,20 @@
 	set(MAKEMAKER_NOECHO_COMMAND "NOECHO=\"1>/dev/null\"")
 endif(CMAKE_VERBOSE_MAKEFILE)
 
 # Add build target for the perl modules
-add_custom_target(zmperlmodules ALL perl Makefile.PL FIRST_MAKEFILE=MakefilePerl PREFIX="${CMAKE_CURRENT_BINARY_DIR}/output" LIB="${CMAKE_CURRENT_BINARY_DIR}/output/${ZM_PERL_SUBPREFIX}" INSTALLSITEMAN3DIR="${CMAKE_CURRENT_BINARY_DIR}/output/${CMAKE_INSTALL_MANDIR}/man3" ${MAKEMAKER_NOECHO_COMMAND} COMMAND make --makefile=MakefilePerl COMMAND make --makefile=MakefilePerl pure_install COMMENT "Building ZoneMinder perl modules")
+# PREFIX="${CMAKE_CURRENT_BINARY_DIR}/output" LIB="${CMAKE_CURRENT_BINARY_DIR}/output/${ZM_PERL_SUBPREFIX}" INSTALLSITEMAN3DIR="${CMAKE_CURRENT_BINARY_DIR}/output/${CMAKE_INSTALL_MANDIR}/man3"
[...]

Issue #783, opened 2015-04-04
Switching USE_DEEP_STORAGE deleted all events.


onlyjob:
On 1.28.1 using web interface I've changed _Options-->Paths-->USE_DEEP_STORAGE_ and restarted Zoneminder. As result all my events were deleted and all contents of `/var/cache/zoneminder/events` were wiped quite unexpectedly.

I would expect at least warning in USE_DEEP_STORAGE' description....

To avoid loss of sensitive data please at least inform user about consequences.

knnniggett:
It shouldn't delete events.  

I know I didn't lose any events the last time I ran it, but that was a long time ago.  You have to run "zmupdate.pl --migrate-events" to convert your filesystem, but I don't recall if that is supposed to happen automatically or not.  Have to look.  Back in the day I seem to recall running that command myself, from the command line, which produced output indicating what it was doing.

knnniggett:
Further inspection hints that this looks to be a documentation issue (or lack thereof).
I was not able to find anything that automates this process, so I will update the documentation accordingly.

Here is the proper way to convert events:
- Stop zoneminder
- If you have the available storage, back up your event filesystem & zm dB
- Enable Deep Storage under options
- Run "zmupdate.pl --migrate-events" from the command line
- Monitor the output for errors
- Start zoneminder after the conversion completes.

knnniggett:
Updated with these commits:
3e8c894a3eeac6ab3f93227fdfdb9228d18f9c0c
bde870e

onlyjob:
Thanks. Let's remove phrase "If you are upgrading from previous version of Zoneminder" please? I was not upgrading, just operating 1.28.1 (as you may recall I'm still new to Zoneminder). Events was deleted by `zmaudit`, right? May be mention that as well?

onlyjob:
There is more to this: `--migrate-events` is an undocumented option of `zmupdate.pl`.
Please update embedded POD documentation.

onlyjob:
I'm not sure why but even after updating with 3e8c894 URL `zm/index.php?view=optionhelp&option=ZM_USE_DEEP_STORAGE` still show old help even after restart of all services...

onlyjob:
I also feel that we need 

 * either to add confirmation on changing that option to the web interface;

 * or (better) remove this option from web interface and implement it in `zmupdate.pl` instead.

Latter seems to be much safer and easier because once changing database structures is required it make sense to make such change only together with database update.

[...]

PR #792, opened 2015-04-07
cmake - use perl INSTALLDIRS

knnniggett:
This PR takes a more traditional approach to installing zoneminder's Perl modules via INSTALLDIRS & DESTDIR paramters, rather then LIB & PREFIX, which are not recommended according to makemaker documentation.

This is essentially proof on concept stage.  I'm looking for feedback from non-Redhat distros.
The intent is to install zoneminder Perl's modules into the distro's Perl "vendor" folders, regardless of the installation method.  

The specific folder the modules go into will vary with each linux disto.  It can be viewed like so:
```
perl '-V:install.*' | grep installvendorlib
```
What needs to be verified is:
1) the build succeeds
2) the perl modules end up in the intended folder


onlyjob:
Almost perfect except that it installs templates `*.pm.in` along with generated modules.

onlyjob:
Another minor remark is about code style -- pretty much all code standards begin with avoiding long lines which is almost universally recognised as best practice. A little code wrapping would be helpful...

knnniggett:
Well, that's good news.
I thought I noticed the input files were getting copied over, but I'm not sure how many though. Are they all being copied over from every folder or just the input files under the /scripts/ZoneMinder?
I should be able to clean that up without too much problem.

I've read all the issues you've posted, including the one about code style.  Both time & learning curve prevent me from knocking them all out quickly.  For now, I'd prefer to stick with the existing style and then address the whole thing at once at a later time.

onlyjob:
Extra templates goes to `usr/share/perl5/ZoneMinder`, notably `Base.pm.in`, `Config.pm.in`, `ConfigData.pm.in` and `Memory.pm.in`.

Thank you. As far as I'm concerned there are no urgent issues (although I'm a bit worried about #783 as we probably still didn't do enough to prevent data loss..).

I just have a feeling that very-long-lines code formatting issue is actually slowing everybody down... Well maybe it is just me but I find this issue a bit irritating... What if I try to push some safe code-style-only pull requests? I hope you won't mind...

knnniggett:
.in files should no longer be copied into the staging folder with the latest commit.

This required changing from a cmake `execute_process` to a cmake `file COPY` command since the former did not support filtering via pattern.

Redhat's rpm build process does not run out-of-source so I can't test this commit directly.

knnniggett:
@onlyjob Would you mind testing this again with the latest commit?  The previously mentioned .in files should not leave the source folder.

onlyjob:
Sorry for delay. With a2d9c0c no `.pm.in` files are installed. All good, thank you.

knnniggett:
@barjac @PX03AFK 

Gentlemen,
This pull request represents a change in the way zoneminder's Perl modules are built using cmake.  The end result should not be any different for your packages than previous.  I thought I'd give you guys any opportunity to verify this is indeed the case before our next release.

Things to verify are:
1) The build completes w/o error
2) Perl modules are installed into your distro's perl vendor folders (should be the same folders you have been using).

Note that these cmake variables have been removed:
ZM_PERL_USE_PATH
ZM_PERL_SUBPREFIX

they have been replaced with these:
ZM_PERL_MM_PARMS
ZM_PERL_SEARCH_PATH

I don't expect you to need to use these new variables. Let me know if you do.

I will wait a few days to hear from you before I look to merge this into master.

barjac:
All OK here :)

mastertheknife:
Back then when i wrote that stuff, i did it because i didn't want MakeMaker to install those perl modules directly into the system, but rather let cmake do it, that way its nicely integrated and in full control of cmake, so things like "make uninstall" work properly.

Whatever directory you choose, do not forget to update perl's search path, just in case its not already in the include search path.

knnniggett:
@mastertheknife Nice to see you pop in.

This PR keeps the same process of using makemaker to install to a temporary output folder and then let cmake install into buildroot.  It just uses different makemaker parameters to achieve the same end result.  This was done to resolve #769 .  

By default, it will always install into the destination distro's perl vendor folders so I don't expect any issues with Perl's search path.  Preliminary testing on redhat and debian based distros looks good.

If the end user chooses, he/she can override this by supplying custom makemaker parameters along with a folder name to add to perl's search path.

Another, related perl issue, #753 
Perl no longer supports "make uninstall" of perl modules, and this is apparently by design. Naturally, that only affects those who build from source.  What I plan to do is update the documentation to instruct those who build from source to supply the UNINST=1 makemaker parameter. That will tell perl to search for and delete any perl modules with the same name that are already installed.

PX03AFK:
I'm way behind on doing anything with this.  I'll try to get back to it soon
and note the points made.

 

David 

 

  _____

knnniggett:
This is ready to merged into master, to receive additional testing.

PR #794, opened 2015-04-08
as discussed...

onlyjob:
some reformatting + Perl Best Practice level 5 + pod2usage


knnniggett:
Makes a mental note to use this as an example for the rest of the source

connortechnology:
There is some nice stuff in here.  I'm not totally on board with the line splitting, but I'm ok with it too.

onlyjob:
Thanks guys. :)

PR #796, opened 2015-04-10
one small fix for a log line where the 4th parameter wasn't included.  T...

connortechnology:
...he rest are very small optimisations.


knnniggett:
looks good.

PR #800, opened 2015-04-12
* use pthread_join instead of pthread_tryjoin_np

Sune1337:
Use pthread_join in ffmpeg_camera to make ZM compile on non-linux.


knnniggett:
thanks for following up with this.

PR #803, opened 2015-04-12
Update Mootools

knnniggett:
I have not tested this and plan to do that later, assuming Travis gives it a green light.

There is some confusion on my part as to whether we need "nc" and/or "yc" flavors of these libraries.  The mootools site does not appear to stick to the old naming convention so hopefully that will not cause any problems.


knnniggett:
hmmm.... initial tests don't look good for this pr.  PTZ controls don't work.

knnniggett:
Error when using mootools-core 1.5.1:
```
TypeError: Browser.Platform is undefined
TypeError: $(...) is null
```

Turns out Browser.Platform was depreciated.  It looks like the Jquery bundled with zoneminder was calling this function.  This is where I stopped and decided to just use the most recent version of mootools that would work w/o additional modifications.  Those versions are:
mootools-core-1.4.5-compat
mootools-more-1.5.1

In case you were wondering why the two versions do not match, upstream does not offer older versions of mootools-more.

knnniggett:
This PR is ready to be merged.

PR #804, opened 2015-04-13
Some UI improvements to feature-h264-videostorage

SteveGilvarry:
Some small tweaks to UI for videojs and thumbnails.


knnniggett:
This looks to be a combination of bug fixes and new features.
Any chance we could get a little background on each of these?  

While I believe cmake globs all files under the web folder, autotools definitely does not, so we will need to add an appropriate reference to the new file videojs.zoomrotate.js to the appropriate Makefile.am.

SteveGilvarry:
Yep I will fix the file reference for zoom rotate, might also look at bringing in local copy of videojs. Most fixes are from bhuddah. Only 2 actual changes in it.
1. Add the ability to detect and rotate video on rotated monitors. Video from camera is direct stream so orientation settings that normally work for jpeg saving don't work, instead on the client it is rotating the view.
2. Second was to fix snapshot function, couple is users mentioned it wasn't working, but I have not had time to check where it is used yet. Will double check that and update.

knnniggett:
Let me know what the status is of this when you have a moment.

PR #821, opened 2015-04-19
beautifying *.pm

onlyjob:


knnniggett:
Before merging, we need to test /ustils/zmeditconfigdata.sh against commit abc8d3f

knnniggett:
I'm curious what you used to find the random spaces that occasionally lurk at the end of a line of code.

Also, by design, we have been using tabs to indent our code, rather than spaces. 
@connortechnology @kylejohnson  - Comments?

onlyjob:
> Before merging, we need to test /ustils/zmeditconfigdata.sh against commit abc8d3f

Good point. I've just did that -- it works.

> I'm curious what you used to find the random spaces that occasionally lurk at the end of a line of code.

Nothing special -- Midnight Commander's built-in editor highlights them well enough. Also I often edit code in Geany where I can see spaces and tabs (with my current settings).

> Also, by design, we have been using tabs to indent our code, rather than spaces. 

Not quite... At least you may be using spaces but historically sources that I was editing was very inconsistent -- spaces and tabs were mixed together.
Spaces are better because code is always looks as written. Tabs can be displayed very differently depending on editor setting. I'm not going back to tabs and I'd recommend to leave them behind with no regrets. ;)

Thanks for feedback, @knnniggett.

connortechnology:
We had discussed, and all of the active devs at the time preferred  tabs.  

No one has cared enough to go through and change the formatting yet though.

I would also prefer to put the open brace on the same line...

connortechnology:
As to splitting long lines... while I agree with the concept, I think this goes too far.  For me, I am often more concerned with fitting more lines on screen and the days of 80 columns are over.  I think 120 columns is more realistic. 

But... I really don't care that much.

onlyjob:
> I would also prefer to put the open brace on the same line...

Me too. :)  Not always but most of the time. I just tried to preserve existing style. Also it would be tedious to change... Maybe we can change it slowly as we introduce other changes...

> As to splitting long lines... while I agree with the concept, I think this goes too far. For me, I am often more concerned with fitting more lines on screen and the days of 80 columns are over. I think 120 columns is more realistic. 

You may have already noticed that I tolerated some lines longer than 80.  But in general long lines even as long as 120 are bad practice impacting readability of code. My screen is 2560x1440 and even on such large monitor there are benefits of working with compact code (e.g. multiple windows, etc.). What goes too far (literally) is lines ~3000 characters long... Formatting code nicely is not just about using short lines but splitting parameters etc. -- all those tricks that make code more readable. Long lines make it ridiculously easy to miss condition in the end of line, especially when eyes should go too far to the right to read it, not  to mention horizontal scrolling issues...

New formatting is better, I hope we can all agree on that so let's merge and move on. :)

onlyjob:
Is anybody waiting for me to change anything here? I'd like to merge an move on please... ;)

knnniggett:
Just been busy.
I do this in my spare time as I have a steady day job I must give priority too.

onlyjob:
Thank you, Andrew. :)

knnniggett:
Guys, we have a problem with one of these commits.

While master builds, it currently produces an un-installable rpm:
```
error: Failed dependencies:
	perl(standard) is needed by zoneminder-1.28.1master-1.fc21.x86_64
```
The error goes away if I revert this pr in my local branch.

I'm doing this on Fedora 21, which ships with Perl 5.18.4, so I don't think this is a case of using an older version of perl.

knnniggett:
It was unfortunate, but due to the way the text was formatted in the file configdata.pm.in, rpmbuild incorrectly interpreted a line beginning with "use standard" as a call to load a Perl module called standard.

commit c7c062267378c7fe7ee640fd10bcb74251e784dc fixes this for this case.  

There are other cases in configdata.pm.in with plain text beginning with the word "use" but rpmbuild mysteriously did not have a problem with them.  I'm not sure if there is a good way to keep this from happening again other than to be careful where we break a line.

While troubleshooting, I also noticed that the qqq sub was replacing newlines with null, rather than a space.  Thus, the last word of the first line was butted up against the first word of the next line.  
commit c7a50a7cbca3cd7cef4dd4ffd0600b7ffef6d4d7 fixes that.

connortechnology:
Good job figuring that out.

Issue #688, opened 2015-02-01
Multiple issues..


TheHammer101:
Hi gents, first off since I am completely new to Git, and ZM. I gotta give you huge props for the amazing work you're doing!

That being said I would love it if someone could help me with a few issues as I am trying to set up a camera recording server for the company I work for with about 190 remote Panasonic IP cameras (mostly very low resolution, standard quality, .5 frame rate). I've got the cameras added and accessible with little to no problem.

I am running:
Ubuntu Server 14.04.01 with LAMP, OpenSSH, and shared memories to shmall = 1572864/shmmax = 6442450944
Xeon proc
12GB of RAM
Ubuntu sofware RAID0 boot disks (temporarily to get set up with before I move to RAID1)
a 3x1TB Seagate ZFS ZRAID1 array setup for the recordings mounted to /recording (will be expanding this size later)

I have everything set up and working on a small scale, but I have absolutely no idea how to fix the issues I'm seeing right now! If you would, read on, and let me know how I can help you help me. lol

Issue 1
When viewing recorded events of Panasonic BL-C???A cameras which are recording JPG streams of 320x240 with a frame rate of 0.5 frames per second the WebGUI playback controls are very sluggish to respond. Even with only 4 monitors added! No errors in the log file I can see.
---------------------------------------------------------------------------------------------
Issues 2
When viewing recorded events of Panasonic BL-C???A cameras which are recording JPG streams of 320x240 with a frame rate of 0.5 frames per second the fast forward button merely changes the number of playback speed, but does not increase the actual playback speed. I do not see any errors in the log for this.
---------------------------------------------------------------------------------------------
Issue 3
When viewing the "Montage" I can only see 6 cameras at one time (20 monitors are set to Monitor, Enabled), and the rest show white space similar to an image that's failed to load in a web page.
![miss](https://cloud.githubusercontent.com/assets/10791812/5990609/0a99a4ea-a97e-11e4-9fb7-8a9c854dbc53.jpg)
When clicking on the missing image it brings up the single camera live view, but the page doesn't load! Any other pages from ZM do not load as well! When I close the Montage page the single camera live view I brought up from the Montage page, or any other pages I've tried to open loads up with no problem. *shrugs* It throws the following errors (some I'm sure relating to two of the cameras actually being down at this point in time..):

`2015-01-31 19:15:37.401302	zms	4120	INF	Got signal 15 (Terminated), exiting	zm_signal.cpp	40
[...]

PR #825, opened 2015-04-22
Fix fast forward/reverse in event playback (#688)

rwg0:
Fixes subissue 2 of issue #688 

If the frame rate for an event is less than the maximum frame rate for playing events, then fast forward/reverse don't work properly.

Suppose a capture frame rate of 10fps, a frame skip of 9 (therefore 1fps saved when not in alarm mode). Maximum display fps set to 10 also. If the replay rate is 5x, then calls to updateFrameRate will set effective_fps to 5 and frame_skip to 1, since maxfps is 10.

However, actual playback of the event is controlled by delta_us in the runStream() loop. This is calculated from the time difference between frames and is only adjusted if the effective_fps is less than the base_fps of the stream (ie slowmo). This means that fast forward/reverse don't actually speed up until you get to the point that updateFrameRate sets frame_skip>1. 

This patch always scales delta_us by base_fps/effective_fps, which means that the frame loop will speed up when required to give proper FF/REW behaviour. A lower limit is applied to delta_us to ensure maxfps is not exceeded.

If maxfps can be zero, this line should be protected by 'if(maxfps>0)'.

Finally another idea, it would seem sensible to limit the delta_us maximum value too - for instance if there are several seconds between frames for some reason, it seems unlikely that anyone would want to sit through that period watching one frame waiting for the next to occur.

### DIFF
--- a/src/zm_event.cpp
+++ b/src/zm_event.cpp
@@ -1428,8 +1428,10 @@ void EventStream::runStream()
             if ( ((curr_frame_id-1)%frame_mod) == 0 )
             {
                 delta_us = (unsigned int)(frame_data->delta * 1000000);
-                if ( effective_fps < base_fps )
-                    delta_us = (unsigned int)((delta_us * base_fps)/effective_fps);
+                // if effective > base we should speed up frame delivery
+                delta_us = (unsigned int)((delta_us * base_fps)/effective_fps);
+                // but must not exceed maxfps
+                delta_us = max(delta_us, 1000000 / maxfps); 
                 send_frame = true;
             }
         }


knnniggett:
I created a build with this change and did not notice any side affects, but I've never used frame skip so I wasn't sure what to look for.

I did notice that the maxfps value being mentioned in this pr is not the same max fps field under the monitor config.  maxfps defaults to 10 so we should not have to worry about divide by zero.

rwg0:
The maxfps comes from the WEB_H_VIDEO_MAXFPS/WEB_M_VIDEO_MAXFPS etc settings if I have followed the code correctly. You would see effects if you set your monitor to say 1fps capture rate ( lower than the playback rate). With the old code at 5x speed you would see 1fps playback, but it would skip 4 out of 5 frames to give you 5x speed. This patch will give you 5fps playback in that situation, showing all frames.

PR #826, opened 2015-04-27
Update control.php

marcamon2013:
Bad path for control_functions.php

### DIFF
--- a/web/ajax/control.php
+++ b/web/ajax/control.php
@@ -1,5 +1,5 @@
 <?php
-require_once( 'includes/control_functions.php' );
+require_once( '../includes/control_functions.php' );
 
 // Monitor control actions, require a monitor id and control view permissions for that monitor
 if ( empty($_REQUEST['id']) )


connortechnology:
Okay.  This seems obvious... I'm good with merging it.

What I don't understand is why I've never noticed an error despite using the control features quite a bit.

knnniggett:
What is throwing me is that there are two control_functions.php in the source tree.  
There is one under web/includes and another under web/skins/classic/includes

I have not taken the time to figure out which one is the correct one to be called in this context.  It would be helpful if the OP verified the version of zoneminder he was running (I'm guessing he has built from master), what errors appeared in the log, and exactly under what conditions the error was triggered.

connortechnology:
well the one under skins is about how the controls are displayed.  The one not under skins is about how to build the control commands and so should be skin agnostic.

knnniggett:
Right, but that brings us back to what issue the OP had that caused him to report this problem.
Was the issue skin related or was it with the control commands. What was the context?
Maybe its obvious to others by just looking at the code.  I haven't had time to build master recently to try and duplicate the problem.

connortechnology:
From what I read, the require_once will use the path relative to index.php if it exists.

So this PR should be invalid from what I understand at this point.  You are right, we need to know why the OP felt the need to change this.

knnniggett:
I can confirm that this pr is invalid.
After building master and applying the change suggested in this pr, ptz control buttons no longer function.

PR #829, opened 2015-05-01
Separate css window sizes

connortechnology:
This moves the window popup sizes into a separate file where the filename is the same as the css skin chosen.  This allows the flat skin to use larger window sizes without affecting the classic css skin.


knnniggett:
Yeah, I noticed that too.  Good catch.

Issue #819, opened 2015-04-18
Empty string if missing index in lang files


manupap1:
No string is displayed in web pages if the requested index is missing in SLANG array.
I think the index should be displayed as a fallback method.
This simple function could be used in `web/includes/lang.php`:
```php
function translate( $name )
{
    global $SLANG;
    if ( array_key_exists( $name, $SLANG ) )
        return $SLANG[$name];
    else
        return $name;
}
```
This script could update all files in a row: 
```bash
#!/bin/bash
FPATH="web/ajax web/skins"
for folder in $FPATH
do
    find $folder -type f -name "*.php" -exec sed -i "s/\$SLANG\['\([a-zA-Z0-9\]*\)']/translate\('\1'\)/g" {} \;
done
```

knnniggett:
Please test #833

SteveGilvarry:
Closed via #833 and #845

PR #833, opened 2015-05-10
add translate function

knnniggett:
see #819 for description.

This should be tested first.


manupap1:
I tested it and I missed two replacements in `skins/classic/views/function.php` and `skins/classic/views/console.php` (because of concatenation in parameter).
Also files `skins/mobile/includes/control_functions.php`, `skins/classic/includes/control_functions.php`, `skins/classic/includes/export_functions.php` and `skins/classic/views/timeline.php` contain global declarations of `$SLANG` which are no longer required.
And file `skins/classic/views/options.php` contains an outdated comment with SLANG string, this comment is no longer required.

PR #834, opened 2015-05-11
Required Centos Perl Modules that were not present

rhiann0n:
Added the following as required by cmake & make:

perl-core
perl-ExtUtils-MakeMaker-Coverage
perl-ExtUtils-MakeMaker
perl-Getopt-Long-Descriptiveperl-DBD-MySQL
perl-LWP-Authen-Negotiate 
gcc-c++

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,11 +120,7 @@ Prioritize the repositories:
 
 Once your repos are in order, install the following:
 ```bash
-sudo yum install cmake bzip2-devel ffmpeg ffmpeg-devel gnutls-devel httpd libjpeg-turbo libjpeg-turbo-devel mysql-devel mysql-server pcre-devel \
-perl-Archive-Tar perl-Archive-Zip perl-Convert-BinHex perl-Date-Manip perl-DBD-MySQL perl-DBI perl-Device-SerialPort perl-Email-Date-Format perl-IO-stringy \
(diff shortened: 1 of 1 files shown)


knnniggett:
Did you know that zoneminder rpms are available from zmrepo?  You don't need to build from source.  

You did not state what version of CentOS you are using.  From the list I'd guess you are using CentOS 7.

The information in Readme.md is accurate for CentOS 6.  While I agree it should be updated, I do not completely agree with your list. 

For a list of the current dependencies look at the rpm specfile here:
https://github.com/ZoneMinder/ZoneMinder/blob/master/distros/redhat/zoneminder.el7.spec

You also removed the line continuations "/" which is not acceptable.

rhiann0n:
I'm using 6.5

knnniggett:
For CentOS 6, look here for the list of dependencies:
https://github.com/ZoneMinder/ZoneMinder/blob/master/distros/redhat/zoneminder.el6.spec

knnniggett:
I've updated the main Readme with a new set of installation instructions, which makes this particular PR no longer relevant.  However, I do need feedback on the new method, particularly how well it flows and how easy it is to understand.  Any questions or comments about the new instructions should be placed into a new issue.

PR #836, opened 2015-05-12
Update to FI982821W_Y2k

connortechnology:
Fix printMsg.  We don't use it as an object method. 

Update to parse a user:password out of ControlAddress which is what  a lot of other scripts do.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Control/FI9821W_Y2k.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Control/FI9821W_Y2k.pm
@@ -114,6 +114,10 @@ sub sendCmd
     my $result = undef;
 
 	my ($user, $password) = split /:/, $self->{Monitor}->{ControlDevice};
+	if ( ! $password ) {
+		$password = $user;
+		$user = 'admin';
+	}
 	$user = 'admin' if ! $user;
 	$password = 'pwd' if ! $password;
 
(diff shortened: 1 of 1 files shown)


knnniggett:
If they've just got a password in that field, then that will get passed into $user.  Not a show stopper, but I think we need to remember to put a statement in the release notes about this so anyone using this script knows to update their control field.

connortechnology:
This script didn't do that.  It just had hard coded values in it.  Other Foscam scripts do that.  It is on my todo list to update them as well.

knnniggett:
Negative.  The user name is hard coded as admin, but the password is currently retrieved from the ControlAddress field:
```perl
my $req = HTTP::Request->new( GET=>"http://".$self->{Monitor}->{ControlAddress}."/cgi-bin/CGIProxy.fcgi?usr%3Dadmin%26pwd%3D".$self->{Monitor}->{ControlDevice}."%26cmd%3D".$cmd."%26".$temps );
```

I'm also using this script myself.
I certainly agree with the idea of keeping all the ptz scripts consistent, but the way this pr is currently written will break existing users.

Another possibility would be to check how many values are returned by the split.  If only one value is returned then set $user=admin and $password={value returned by split}.
Would you be willing to do that?  Either that, or we make sure and remember to document this change in the release notes.

connortechnology:
AH, you are correct of course.  I was not looking hard enough.

I will amend the PR to handle the old way.

knnniggett:
thanks.

PR #837, opened 2015-05-12
added libav, libswscale, libv2l, etc to Dockerfile

jgarland79:
RTSP would not work without libav

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,9 @@ RUN apt-get update && apt-get install -y \
 	libwww-perl libdbd-mysql-perl libsys-mmap-perl yasm automake autoconf libjpeg-turbo8-dev \
 	libjpeg-turbo8 libtheora-dev libvorbis-dev libvpx-dev libx264-dev libmp4v2-dev ffmpeg mysql-client \
 	apache2 php5 php5-mysql apache2-mpm-prefork libapache2-mod-php5 php5-cli openssh-server \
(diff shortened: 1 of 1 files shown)


knnniggett:
What I think caused this is @kylejohnson wrote this back in the days of Ubuntu 12.04 when ffmpeg was still in the default repository.  ffmpeg has since been replaced by libav, which is a project that is just different enough from ffmpeg to cause us lots of grief.

Before this gets merged, we should ask @kylejohnson what direction he wants to head.  Should his Docker file use libav or ffmpeg?  Since ffmpeg is no longer in the default repository, extra work would be required to get it installed.

If we choose libav, then this PR should be modified to remove the ffmpeg package name from the Docker file, and add avconv (is that the name?) if it isn't already in the packages you've added.

Another thing to note is, if we choose libav, then this Docker file will only be able to be used in a Debian environment.

If ffmpeg is chosen, well then we'd have to find a repo that has has package for ubuntu.

knnniggett:
Alright, so Kyle and I had a discussion, and we are going to take the easier path and go with libav libraries, rather than ffmpeg.  

@jgarland79 what that means is that we need you to make the following changes to your pull request:
- replace ffmpeg package with libav-tools
- Add libavcodec-dev package
- Add libavfilter-dev package
- Add libavresample-dev package
- Add libpostproc-dev package

No idea why Debian chooses to split the library up into so many packages.  The latter three are not currently used, but I would much rather include all the libav libraries now so that we don't have to continue to manage the packages in the docker file.

Issue #756, opened 2015-03-30
VideoStream::OpenStream fails in write_header for mp4 and mov


whorfin:
I have a video source which is mpeg4 over rtsp
If I select "mov" or "mp4" via MPEG_LIVE_FORMAT, I get an error
?_write_header failed with error [whatever error EINVAL is on my OS] "invalid parameters"
and zms or nph-zms exits

Likewise, straight up running 
zmstreamer -o mp4
shows the same behavior [or -o mov]
gdb shows avformat_write_header() returning the badness

This is with version 1.28.1 on both Ubuntu and Solaris
[yes, Solaris - I've completed a port and would like to know where to submit the patches - just a pull request to here?]
Ubuntu 14 had .54 version of libavformat, I have .55 on OmniOS

Cheers

knnniggett:
ZoneMinder currently transcodes all incoming video streams into jpeg's. So it really doesn't matter if the video source from the camera is mjpeg, mpeg4, h264, etc ....the result is the same.... jpeg's on the disk.

When it comes time to stream video from the zoneminder server to a client browser, ZoneMinder reassembles those jpegs back into a stream. By default, that stream is mpjeg, but you can change WEB_H_STREAM_METHOD under each Options -> B/W tab to stream MPEG, which finally brings us to MPEG_LIVE_FORMAT.  

As described in the help text, the only know working format is asf.  The invalid parameter message you are receiving when attempting mov & mp4 is due to the version of ffmpeg (or libav) on your system.  I don't get that error. The video makes it to my browser, but it still won't play. 

It is unlikely mov will ever work, and we don't currently have an answer as to why mp4 doesn't work.  
However, others have been successful streaming mpeg to their browsers, and I suggest you search this forum, our user forum, and the wiki for clues that others have left.

While I'm not denying this is an area that can be improved, we do not have the resources to put this on our priority list.  Instead, I encourage you to look forward to our upcoming video method, which records incoming video straight to an mp4 container.  I'd expect streaming from the zoneminder server to the client browser to work then, or at the very least be worth fixing at that point.

PR #842, opened 2015-05-18
Port to OmniOS/Solaris

whorfin:
Before things drift too much further out of sync [I notice changes already since 1.28.1 involving
proliferation of SYS_gettid calls and rollback of nonportable pthread join], I wanted to get this out there.
I need to do further testing against this latest master, but this set of patches is running on 1.28.1
for some time.  [see issue #756 for announcement]

### DIFF
--- a/README-OmniOS.txt
+++ b/README-OmniOS.txt
@@ -1,0 +1,17 @@
+The following has worked [with newest patches] with buildtools and gcc48 from pkg,
+and ffmpeg2, LAMP and libjpeg-turbo from pkgin for OmniOS r151012 and r151014
+
+./bootstrap.sh
+# omnios default is 32bit, -m64 is unneeded if perl/libs versions match
+export CXXFLAGS=-m64
+export CFLAGS=-m64
+export PERL=/usr/perl5/5.16.1/bin/amd64/perl
+export CPPFLAGS="-I/opt/local/include/ffmpeg2 -I/opt/local/include/mysql
+-I/opt/local/include"
+# need the -R for runtime load library
+export LDFLAGS="-L/opt/local/lib/ffmpeg2 -L/opt/local/lib -R/opt/local/lib
+-R/opt/local/lib/ffmpeg2"
+export ZM_SSL_LIB=openssl
+# need -lsocket -lnsl for recv/send to work, and latest version needs -lsendfile
+./configure --with-webdir=/opt/local/share/httpd/htdocs/zm --with-extralibs="-lsocket -lnsl -lsendfile" --with-cgidir=/opt/local/libexec/cgi-bin --with-webuser=www --with-webgroup=www --with-webhost=MACHINENAME.local
+
--- a/configure.ac
+++ b/configure.ac
@@ -34,6 +34,11 @@ case $host_os in
 		HOST_OS='BSD'
 		AC_DEFINE(BSD,1,"This is a BSD system")
         ;;
+    *solaris*)
+        # Do something specific for Solaris
+		HOST_OS='solaris'
+		AC_DEFINE(SOLARIS,1,"We are running a Solaroid OS [tested on OmniOS]")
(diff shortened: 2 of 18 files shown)


knnniggett:
It is always interesting to see what kind of systems zoneminder has made its way onto.  

@whorfin Looks we need to work through some of these before this can be merged.  Some just are not right at all and would break ZoneMinder for the rest of us.  

Lastly, we have a long term goal to remove autotools, and eventually switch completely over to cmake.  That should be the next item on the agenda.

knnniggett:
I tell you what, let's do this in cmake. 
It looks like cmake is not checking the host_os like autotools is, so I need to add that. 
Just so I get it right, what exactly is the result of "uname -s" on your system?
In the meantime, can you modify this pull request, or generate a new one, that includes just the changes your configure.ac, the .cpp files, and .h files?  Once we get that far, then we can tackle to stuff with ZM_VERSION.

whorfin:
OK!
Push reversions and rollbacks as requested.

uname -s produces
SunOS
under OmniOS [and presumably most solarian devi
You're absolutely right about PATH, my issue with zmupdate was the usage of mysqldump, which
lives in /opt/local/bin with pkgin.  A quick grep shows a bazillion references to mysqldump in
various incorrect locations which won't work, like a hardcoded /usr/bin reference in zmeventdump.
So yes please, let's fix this more rigorously via path detection in cmake and proper referencing in
all the locations that call the various mysql.* utilities.

DanTheColoradan:
@whorfin :+1:  nice work.

knnniggett:
I'm in an off site training class the rest of this week so my zoneminder productivity is going to be low.  Go ahead and create an issue that tracks the remaining issues with Solaris.

whorfin:
Excellent, thanks!
My zoneminder productivity is also going to plummet as the millions of other things to tend to jump the line.
I will however create an obvious-looking issue with Solaris issues as I encounter them.

knnniggett:
@whorfin Now that #844 is merged, try building zoneminder with cmake.  Start with these steps:
```bash
cmake /
	-DZM_WEBDIR="/opt/local/share/httpd/htdocs/zm" /
	-DZM_CGIDIR="/opt/local/libexec/cgi-bin" /
	-DZM_WEB_USER="www" /
	-DZM_WEB_GROUP="www" /
.
make
```
Don't forget the dot "." and don't add any additional parameters unless it does not build.
You can see all the parameters one can send to cmake here:
https://github.com/ZoneMinder/ZoneMinder/blob/master/CMakeLists.txt

Each time you re-run cmake, delete the cache file from the previous attempt: CMakeCache.txt

My guess is you will run into issues.  If that happens, create a new github issue, and post your entire build output and the entire cmake command you used to start the process.

whorfin:
Rather than just complaining, I've attempted to create a patch set which resolves the [so far] issues
It is #851

knnniggett:
You should undo all the changes to .gitignore.  .gitignore is global, so we can't add references that exclude files simply on the grounds that it they are not used for *your* distro when they are needed for *other* distros.  It does not hurt anything to have those files lying around.

Did you make the changes to the cmake modules yourself, or did you get them from some other source?
The modules under cmake/Modules are canned, and they should be kept that way.  Troubleshooting custom changes to the modules would make things more difficult than it should be.  

We need to accomplish what you are trying to do within the CmakeList.txt files, or adding additional modules (or replacing them entirely is ok), as long are they come from a reputable source that has tested the modules.  

For example, does your system have access to a Solaris version of FindPolkit.cmake module?  We could check for Solaris in CMakeLists.txt and then call that module instead of the one already there .....Or, if we can verify the canned module is cross-platform then we could use that.

connortechnology:
Most of this looks fine.  I would prefer to find a replacement to get the thread id instead of just not doing it, but I'm ok with it for now.

PR #845, opened 2015-05-21
Additional SLANG changes 

knnniggett:
followup changes based on comments in #833

NOTE: I was not 100% certain about how I made the translate function calls, with all the string concatenation going on.  Could use someone more familiar with php to confirm I did that right.


connortechnology:
Looks ok to me

PR #847, opened 2015-05-21
faster shutdown

connortechnology:
alter shutdown behaviour to send a term signal to all processes and then go through again to send a KILL if necessary.  This eliminates unnecessary sleeping.

Please note that there is a lot of code here to improve the race condition behaviour, but I have not addressed the race condition.


connortechnology:
Still not fast enough. Have more work to do.

Issue #840, opened 2015-05-15
Adapt to changes in Ubuntu 15.04


seeebek:
There are changes in Ubuntu 15.04 which make the building and installing Zoneminder not trivial.
The two most significant:
1. Ffmpeg is again in the official repository
2. Initd was replaced by systemd

I suggest creating a new directory under distros named "ubuntu1504" since it is significantly different from 12.04. 
I already got Zoneminder (H264 branch) compiled by changing the dependencies to the new ffmpeg packages provided in 15.04. Now I will be working on replacing initd with systemd. Once I finished and tested my changes I will create a PR for you to review.
If you have any suggestions please let me know.

knnniggett:
Don't know how familiar you are with systemd, but many of us that work with Fedora & OpenSuse have been doing this for over a year.  We've been able to get ZoneMinder to work well with systemd.
Welcome to the party. :-)

In anticipation of this day, I worked with Bob to put together this guide here:
http://www.zoneminder.com/wiki/index.php/Ubuntu_Server_14.10_64-bit_server_with_Zoneminder_to_use_systemd

I'm sure you know, 14.10 shipped with systemd, but it was not "turned on".
For you, the relevant parts of that guide are the portions that show how to create the systemd service file and configuring the new tempfiles.d service.

seeebek:
Thanks. This is what I was intending to do. Just put it all together. I'm also new to systemd so I will "learn by doing".  I can't promise to do it very soon but I will try my best.

seeebek:
I made the changes for Ubuntu 15.04. It compiles and runs for me on a clean installation amd64. It would be nice if @connortechnology  could review it and see if I missed something. Especially the file postinst. There are two places I commented where I wasn't completelly sure.

seeebek:
@connortechnology 
I ran into problems with file permissions. caused by following entry in file postinst:
```sh
if [ -z "$2" ]; then
	chown www-data:www-data -R /var/cache/zoneminder
fi
```
$2 gives 1.28.1+1-trusty-SNAPSHOT2015030201 with the latest git master which prevents it from setting the right permissions. What is the design idea behind this?

There is also this:
```sh
if [ "$1" = "configure" ]; then
	if [ -z "$2" ]; then
		chown www-data:www-data /var/log/zm
		chown www-data:www-data /var/lib/zm/
		chown www-data:www-data -R /var/cache/zoneminder
	else
		chown www-data:www-data /var/log/zm
[...]

PR #850, opened 2015-05-24
ubuntu 15.04

seeebek:
Added changes for building and running on Ubuntu 15.04 which now uses systemd and has ffmpeg in main repository.


seeebek:
Not ready to merge. I still get "Can't symlink '1' to 'Monitor-1': Permission denied"

knnniggett:
That sounds like a file permission problem.  Make sure your packaging scripts are setting up the events folder with the owner and group set to www-data.

You might also want to make sure there aren't any leftover symlinks in this folder from prior build attempts.

seeebek:
Yep, that's the problem. Events folder and few others were set to root. Thanks for your help. 
I will try to get the script to set them to www-data.

seeebek:
I found the problem and I addressed it in #840 . I don't know the decisions made behind the problematic parts of the code and I will wait until @connortechnology has a look on it.

seeebek:
It was a false alarm. Apparently I broke something on the way. On a clean try it works. You are welcome to review it and merge.

knnniggett:
@connortechnology I'm hoping you can take a look at this and merge if you think it is good to go.

PR #857, opened 2015-05-29
Do not build ZM with OpenSSL by default when using cmake and force user to enable non GPL libraries

manupap1:
This PR to address a lintian error when building debian package with cmake.
OpenSSL library is not under GPL license and some terms of the OpenSSL license are incompatible with GPL.
Therefore linking ZoneMinder libraries or binaries with OpenSSL makes them unredistributable under current ZoneMinder license without adding an exception.


knnniggett:
I think Debian is splitting hairs here, especially since all we have to do is add an exception to our license to get around their self-imposed restrictions.

I'm not saying we shouldn't do anything about this, but I think this pr, the way it is written, is far too complicated.  We should look into something that creates just one new cmake variable, which would allow the user to manually turn on support for openssl, instead of gnutls.

We are depreciating autotools, so lets just focus on cmake.

manupap1:
PR updated as requested

knnniggett:
This PR isn't working correctly in an rpm environment.  It is not finding the md5 function provided by gnutls.  Not a total surprise considering Debian & Redhat package this library a bit differently.  I need to spend more time on this and figure it out.

knnniggett:
Redhat has intentionally pulled the openssl compatibility layer from gnutls packages.  I have not found anything that directly tells me if this was simply due to a symbol conflict with openssl or security vulnerabilities that they simply didn't want to deal with.

In any case, this might be reason to look at replacing the md5 function calls to something else entirely.

knnniggett:
Going to close this pr as it has been superseded by #1235

Issue #580, opened 2014-11-15
Implement version check for ffmpeg and libav


SteveGilvarry:
@knnniggett Reminded me of some information I found on handling ffmpeg and libav versions. Putting this in here as a reminder to me to come back and look at implementing it.

Looking for av_frame_alloc here (thanks @SteveGilvarry ) : https://github.com/FFmpeg/FFmpeg/blob/master/doc/APIchanges

I found the relevant text:

2013-12-11 - 29c83d2 / b9fb59d,409a143 / 9431356,44967ab / d7b3ee9 - lavc 55.45.101 / 55.28.1 - avcodec.h
av_frame_alloc(), av_frame_unref() and av_frame_free() now can and should be
used instead of avcodec_alloc_frame(), avcodec_get_frame_defaults() and
avcodec_free_frame() respectively. The latter three functions are deprecated.
This tells us that libav implemented av_frame_alloc() in version 55.45.101
while ffmpeg implemented the same function in version 55.28.1

(if the third octect in the version number is >100 then we know it applies to libav)

So this puts us in a bit of a pickle due to the differences in versions. Technically, the best "correct" way to do this is to implement what @SteveGilvarry found for us in the libvlc library:
https://github.com/videolan/vlc/blob/d36bc0a71a7a69afd085c8b2754ecfbc5876fd2b/modules/codec/avcodec/avcommon_compat.h

But that will require more time than either of us want to spend our weekend working on.

However, the following might work if we just check the larger version number:
# if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(55, 45, 101)

```
  av_frame_alloc()
```
# else

```
avcodec_alloc_frame()
```
# endif

It's not as efficient for ffmpeg since we won't be using the newer function call for all versions of ffmpeg that support it.

Yeah, I know. This sure is a lot of work for such a small change, and it is probably why we don't get many volunteers to help our with this kind of work. I learn a little but more each time we get a new pull request that deals with the c++ side of zoneminder.
##

connortechnology:
I think we should only support back to the last ubuntu lts release, which is 12.04.  Or debian wheezy.

knnniggett:
Relevant quote from @onlyjob in #762 

~~~~
Compilation on Debian Jessie revealed multiple deprecations:

[...]

PR #858, opened 2015-05-29
Fix detection of deprecated libav / ffmpeg functions

manupap1:
Implementation as per discussion on https://github.com/ZoneMinder/ZoneMinder/issues/580


connortechnology:
This looks pretty clean to me.  Thanks @manupap1!

PR #866, opened 2015-06-03
Don't die if db goes away during logging

connortechnology:
So I don't know why I was getting a mysql has gone away error (mysql wasn't logging anything).  

That is beside the point.  There is no reason to die because of it.  

Now this is a very minimal fix. It turns off db logging and continues.  However it should probably attempt to reconnect on every log line.


knnniggett:
I've run into the same issue of "mysql has gone away", but I've only noticed the issue during development, inside of a virtual machine and only when debug is turned on.  When I turn debug off, the issue goes away.

PR #867, opened 2015-06-03
analysis optimisations

connortechnology:
This change uses a member Image instead of constantly allocating a new Image.

This saves all the initialisation and the lines of debug.
Also in GetImage we potentially remove an entire copy if no scaling or timestamping is being done.

This works perfectly in testing for me but please look closely and think abut what I've done before merging.


knnniggett:
Let's get this into master and test the heck out of it.

PR #873, opened 2015-06-05
reduce the wait to 2/10ths instead of a whole second

connortechnology:
Found this out on the net, and I agree with it.  A second is a long time.


knnniggett:
Sounds good to me.  Did you see any reference as to why the original author choose 2/10's of a second rather than another value?  Was it just a guess or was it based on something more specific?

connortechnology:
I actually chose 2/10ths.  The original auth I think took it down to a 1000/th or something like that which is too much.

PR #875, opened 2015-06-08
Add version to the startup log line

connortechnology:
The purpose is so that when people paste their logs we can quickly identify the version of ZM in use instead of having to ask.


knnniggett:
good idea.

Issue #799, opened 2015-04-11
Thoughts about the ZM API (for mobile consumption)


pliablepixels:
I'm going to leave some more notes on how my (forthcoming) app can benefit more from the APIs. Not sure if this is the best way to do it - i.e. start an issue, so please advise if I need to present it any other way:

a) As I've already mentioned earlier, its critical the API ties into the same php auth ZM is using in opt_auth - an open API is unusable and insecure (I'm hoping you tell me I can fix it easily)

b) Mobile devices have memory restrictions, so http://zmdevapi/events.json can be a problem if the list is large. And this list can easily be very large. I'd recommend the following:
b.1) Add ability to get length of events without fetching all of them. 
b.2) Add ability to specify a "from" and "to" so I can chunk get events in my code to preserve memory
b.3) add ability to return events between a specified start date/time and stop date/time

c) There is no tie in with monitors and events. There should be. In addition to point b) above, I'd love to see events scoped inside monitors like so http://zmdevapi/monitors/1/events.json. All the operations of point b should work scoped inside a specific monitor too.

d) The monitor list should return a live view URL as something I can construct as an img-src - right now I am constructing it separately as img  ng-src="{{loginData.url}}/cgi-bin/nph-zms?mode=jpeg&amp;monitor={{monitor.ID}}&scale=100&maxfps=3&buffer=1000&user={{loginData.username}}&pass={{loginData.password}}&rand={{rand}}"   - there is no tie in to the API

e) The detailed events list should also return an img-src url to view the event

f) In detailed events, please separate out time from date

Notes on your server side:
a) As of today, the cmake install does not make the APIs active -- so if I am building an app that uses those APIs, pretty much no one can use it because even if they did do cmake, they need to add the API paths to apache configurations. I'd appreciate if you add that to your wiki. I added it via .htaccess - not sure if that is the right way (the APIs are unauthenticated - not sure if thats due to htaccess or not)

b) Consider adding Header set Access-Control-Allow-Origin "*" to your apache.conf to allow for desktop clients to be able to access your APIs

kylejohnson:
Hey there.  So -

> a) As I've already mentioned earlier, its critical the API ties into the same php auth ZM is using in opt_auth - an open API is unusable and insecure (I'm hoping you tell me I can fix it easily)

Let us be careful with our words here.  While the API does not have any authentication currently, it will.  That said, it will probably, almost certainly, **not** be the "same php auth" that the current frontend is using.

[...]

PR #877, opened 2015-06-08
799 api improvements

pliablepixels:
There is a need to expose a new API to get PTZ control definitions. The monitors API returns a controlID but the data mapped to the controlID is not returned. We need a mechanism to retrieve it to be able to construct a correct PTZ command to the camera (example, Move vs. MoveCon etc)
Refer to https://github.com/ZoneMinder/ZoneMinder/issues/799#issuecomment-105233112


knnniggett:
Looks good to me.

@kylejohnson Any comments before we merge?

Issue #881, opened 2015-06-10
Merging angular-ui API branch into master


pliablepixels:
1) We need to merge the angular-ui API branch into master to get the latest APIs. 
2) Following this we will also need to merge https://github.com/ZoneMinder/ZoneMinder/pull/877

I'll start working on 1 for now.

connortechnology:
Um... I just looked ... and there is an api under master...I'm wondering how much is actually there already... looking...

pliablepixels:
Correct - the master does have an API branch but is not as updated as the angular-ui api branch - many APIs are not present

pliablepixels:
I just realized I did not merge in the following code into bootstrap.php 
```
Configure::write('ZM_CONFIG',  '@ZM_CONFIG@');
Configure::write('ZM_VERSION', '@VERSION@');
```
This is needed for the APIs to access config and version info. I'll do a PR

knnniggett:
After you checkout the branch, you need to do a "git submodule init"  and then possibly folloed by "git submodule update" to populate the submodule folder.

@connortechnology made some fixes to bootstrap.php and the related CMakeListFile so that those macros would get expanded properly.  Perhaps we just need to merge his commits into master.

pliablepixels:
I think he has already merged it to master -- the substitution in bootstrap is working.

I'm thinking of adding a version for the API that will be a minor to the version of ZM like this:
```
Configure::write('ZM_API_VERSION', '@VERSION@.1');
```
If ZM version is 1.28.1
This will make ZM_API_VERSION as 1.28.1.1

and in /api/getVersion.json, I can return both ZM_VERSION and ZM_API_VERSION

This will help in us making updates to the API incrementally. What do you think?

pliablepixels:
I've reworked this and put in an API version in cmakelists --> since that is where ZM version is set. Have done a PR

pliablepixels:
@connortechnology have you given any thought on authentication? Do you plan to link ZM portal login to API access or are you looking at an API auth. separate from ZM?

knnniggett:
Closing, since the API is now in master.

PR #882, opened 2015-06-11
Merged Angular UI branch API to master

pliablepixels:
Merged angular-ui branch APIs to master as per @connortechnology request  https://github.com/ZoneMinder/ZoneMinder/issues/881. This merge also includes the additional API fixes I had committed to the angular-ui branch.

Note this includes crud code, so @knnniggett may want to change crud to a submodule - not quite sure how that works.


knnniggett:
Hmmm.... I'm not sure the Crud stuff looks right.  This PR appears to be putting the individual Crud project files directly into our project.

From yesterday's learning experience about git submodules, I don't think we should be able to see the individual files from the Crud project within our own project.  Instead we should see a link from the web ui, or an empty folder if looking at the project from your local disk.

connortechnology:
This code also contains cakephp... which from what I understand isn't strictly necessary either.

pliablepixels:
The api directory is basically what kyle has in the angular branch. That branch also has Cake files --> from what I understand, Cake and the Crud plugin is being used for the APIs. CakePhp provides the MVC framework for the APIs, so I am not sure if the APIs would work without cake

connortechnology:
yeah, but the point is that in a distribution like debian, these are provided already in the system. Packaging our own copy of the code is not ideal.  We do however have to target a certain version (cakephp 3 is very different from cakephp2).

connortechnology:
I was talking with @kylejohnson and he implied that there was stuff to be cherry  picked... but maybe we should just snarf the whole thing.

connortechnology:
uh... the other thing is that @kylejohnson didn't do any of the automake stuff for it either.  Maybe now is a good time to deprecate autotools.

pliablepixels:
Right, I actually copied in the Controller/Model/View directories only (and the Cake lib) and merged the rest. I don't think we can leave out any files in these directories as they are the essential API logic/model files. 

What I have checked in is basically a working version of the angular APIs merged into master, but what needs to be done is:

a) Modify the install process to correctly install these files in deployment
b) Make sure APIs are accessible (in my case, I had to modify .htaccess files in 2 places after doing a make install on the angular-ui branch). Kyle mentioned that may not be required but he never clarified
c) Remove the bundled CRUD and Cake libraries and install them as part of the install process, if needed

I'm not familiar with the ZM build process rules at all. I wonder if we can get Kyle to comment?

knnniggett:
@pliablepixels I'll look into the git submodule stuff and then push a change against your pr.

pliablepixels:
so does that mean I can checkout a version that has all the APIs ready to use?

SteveGilvarry:
If you checkout master it should now have same api directory.

knnniggett:
You will still have to modify your apache config.
Making the Apache part happen automatically will be part of each distros packaging scripts.

pliablepixels:
I just checked out master and noticed a few things after doing cmake :

a) The Crud plugin does not get installed
b) make install does not change permissions in the install dir for www
c) /www/api/.htaccess does not included ReWriteBase for /zm/api but /www/api/app/webroot/.htaccess does. --> I guess either both should have this change or none of them (without them the API is not accessible)
d) There seems to be a problem in the way the Crud plugin gets installed - basically it did not work for me  -- I had to completely delete the Crud folder inside app/plugin and re-install with:
```
cd www/api/app
curl -sS https://getcomposer.org/installer | php
./composer.phar require friendsofcake/crud:~3.0
```
e) The APIs don't tie into the ZM auth - I assume that will also be part of the post deployment?

SteveGilvarry:
Not so sure submodules are the best option. 
http://stackoverflow.com/questions/3796927/how-to-git-clone-including-submodules
With version 1.6.5 of Git and later, you can use:
```
git clone --recursive git://github.com/foo/bar.git
cd bar
```
For already cloned repos, or older Git versions, just use:
```
git clone git://github.com/foo/bar.git
cd bar
git submodule update --init --recursive
```

PR #885, opened 2015-06-14
Updated API documents

pliablepixels:
a) Deleted api.rst and created api.md (Markdown offers more formatting control)
b) Correct/updated examples

Would be great if I got direct write access to the API documentation because I am going to keep editing it as I also update the API code. IF that is a problem, no worries, I'll continue editing and PRing


SteveGilvarry:
Only problem I can see is I don't think Sphinx does markdown and that is what the rest of the documentation is setup for. http://zoneminder.readthedocs.org/en/latest/index.html

pliablepixels:
@SteveGilvarry it looks like it does - https://read-the-docs.readthedocs.org/en/latest/getting_started.html#in-markdown

While we are on that topic - any reason why its on readthedocs? Given ZM is in github, why not use the github wiki?

pliablepixels:
Hmm, looks like a lot of work has been done on readthedocs for ZM. I'll submit a new PR with the updated API page in sphinx format

SteveGilvarry:
@pliablepixels Just following the lead from @kylejohnson on using readthedocs, I believe the plan is to move zoneminder.com to github pages, wiki to github wiki and documentation to readthedocs. There is a difference between content in readthedocs and wiki. I can see readthedocs should be official project documentation, this goes through the same PR process as code changes, which keeps the content under the control of the project. The wiki more open for anyone to contribute at any point without review. I do have a github repo of the wiki that I was working on converting the content from the mediawiki on zoneminder.com, but keeps getting out of sync with people adding new content to wiki all the time. So I stopped working on it until we had the official word on moving to github pages for the website.
The API shouldn't really need instant updating of the documentation, remember we have releases, so as long as the documentation is up to date at release points it will be updated on readthedocs.
But looks like you are getting more time on a day to day basis than I do so if you want to pull all the wiki content from wiki to github wiki go for it.

PR #890, opened 2015-06-19
reverse the if statement to reduce indenting

connortechnology:
This alters a lot of lines, but it's really just a removal of indenting by reversing the if condition.  Please note that there is a secondary effect of not calling zmMemInvalidate for monitors that are Function==None, but we never attempt to connect to the shared mem for them, so that is fine.

I could have altered the SQL statement to just be Function!=None but maybe someday we will want to add a debug line for None Monitors, I don't know.


connortechnology:
This looks really confusing in the diff, but it's really simple actually.  Might help to look at the whole file before and after.

PR #891, opened 2015-06-19
Zmwatch cleanup2

connortechnology:
I don't know why these lines were commented out, but they are causing a serious problem for my systems, in that zmc is not being restarted despite being dead for days.


knnniggett:
I keep thinking that Stan had a reason to comment out those lines.  I wish he would have written what symptoms caused him to do that.  Going to merge this change and possibly find out....

connortechnology:
yeah.  I fixed the issue with zmdc, but thing is, if you stop a monitor (maybe manually) it will never get restarted unless zmwatch says so.  

I think re-enabling these lines will help us track down problems and lead to a more stable system.

Issue #893, opened 2015-06-19
updated 9831W control script


pliablepixels:
We need to update the foscam scripts to match the latest CGI document and modify the following:
a) take out user/password from within the script and move it to ControlDevice
b) switch off osd by default - it renames the device name and sometimes does not revert
c) change "reset" to "reboot"

PR #894, opened 2015-06-19
893 foscam 9831 w and other foscams

pliablepixels:
ref: https://github.com/ZoneMinder/ZoneMinder/issues/893

a) take out user/password from within the script and move it to ControlDevice
b) switch off osd by default - it renames the device name and sometimes does not revert
c) change "reset" to "reboot"


knnniggett:
Nice work.  There is one more thing to add.
In order for existing users to see the control script after an upgrade, you'll need to add the appropriate bits of sql to this file: https://github.com/ZoneMinder/ZoneMinder/blob/master/db/zm_update-1.28.99.sql

I've already added a control type called "ONVIF" inside that file.  You can use that as an example.

Issue #898, opened 2015-06-21
Add isRunning field to States


pliablepixels:
As of today, there is no way to know which run-state ZM is currently in. Each run-state comprises of a specific configuration of monitors. zmpkg.pl switches to a run state but does not store the active state.
This makes it impossible to write cron scripts where two or more states have the same monitor configurations (example, out-of-house and vacation --> two different situations with the same monitor configs)

The proposal is to 

1) add an isRunning tinyint field to States which gets set to 1 when zmpkg changes state
2) Display the run state in ZM console at the top

@knnniggett -- do I need to modify zmupdate.pl as well in addition to modifying zm_create.sql and zm_update-1.28.99.sql?

knnniggett:
No mods to zmupdate.pl necessary.
Just add the schema change to the newest zm_updatex.xx.sql script and zmupdatw.pl will find and apply it  automatically. The sale must be written such that it can be run more than once without causing an error. See the existing scripts under the dB folder for examples.

pliablepixels:
Interestingly all past examples of ALTER TABLE don't bother checking if it
was already altered. I guess it will throw an error anyway. It's a little
convoluted in MySQL to check for existence of a column, so I guess I'll
just follow past precedence and just issue ALTER TABLE. if it was already
altered trying it again will not  add a duplicate column .

knnniggett:
Sometimes you don't have to when using Alter Table, but when you are trying to add a column you do.

See this example that adds the ExtendAlarmFrames column to the Zones table:
https://github.com/ZoneMinder/ZoneMinder/blob/master/db/zm_update-1.27.99.0.sql#L112

pliablepixels:
Okay done -- I've done a column check. Note that even if you don't it won't let you add the column (duplicate column error) - this is what update scripts were doing pre 1.22 and previous to it. I've however done it the proper way as you've recommended

knnniggett:
That's correct the older versions of ZoneMinder, prior to ~1.26, didn't implement sql updates the way we do now.

We made a design change recently to implement SQL such that it doesn't cause zmupdate to crash if it is run multiple times (which is often required during development).  Since the particular circumstances that cause zmupdate to crash only appear during development, between releases, it is in your best interest to do this.  

Older versions of ZoneMinder were developed, slowly, via 1-2 people, so it was not such an issue back then.

PR #899, opened 2015-06-21
898 is running states

pliablepixels:
1) Updated zmpkg to update new IsActive column of states. If the user is switching to a custom state, IsActive for that state will be 1. If the user is not switching to a custom state (restart/stop/start) then IsActive will be 0 for all

2) Modified create DB script to add the new column

3) Modified updated DB script to update the new column

4) Modified console.php to reflect custom state new (convenience only). If the current state is just Running then nothing will show. This is only done in classic --> does not apply to mobile/xml

reference: https://github.com/ZoneMinder/ZoneMinder/issues/898

### DIFF
--- a/db/zm_create.sql.in
+++ b/db/zm_create.sql.in
@@ -382,12 +382,14 @@ CREATE TABLE `Monitors` (
 
 --
 -- Table structure for table `States`
+-- Added IsActive to track custom run states
 --
 
 DROP TABLE IF EXISTS `States`;
 CREATE TABLE `States` (
   `Name` varchar(64) NOT NULL default '',
   `Definition` text NOT NULL,
+  `IsActive` tinyint(3) unsigned NOT NULL default '0',
   PRIMARY KEY (`Name`)
 ) ENGINE=@ZM_MYSQL_ENGINE@;
 
--- a/db/zm_update-1.28.99.sql
+++ b/db/zm_update-1.28.99.sql
@@ -324,4 +324,20 @@ WHERE NOT EXISTS (
 --
 UPDATE `zm`.`Config` SET `Category`='hidden' WHERE `Name`='ZM_USE_DEEP_STORAGE';
 
+--- The States table will be updated to have a new column called IsActive
+--- used to keep track of which custom state is active (if any)
+SET @s = (SELECT IF(
+	(SELECT COUNT(*)
+	FROM INFORMATION_SCHEMA.COLUMNS
+	WHERE table_name = 'States'
+	AND table_schema = DATABASE()
+	AND column_name = 'IsActive'
+	) > 0,
+"SELECT 'Column IsActive  exists in States'",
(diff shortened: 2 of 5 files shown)


pliablepixels:
@knnniggett not sure why events API PR is also showing here. It was on a different branch and there is already a PR for that.

knnniggett:
Just thinking out loud here:
- Should add a row, called default, that becomes the active state if one if not already specified?
- I noticed that, existing behaviour prior to this pr, is to poll the State table each time zmpkg.pl is called.  I wonder if we can make this more efficient.  I'm not sure we need to do that.  Thinking we only need to check the state table if zmpkg.pl is called with the "start", "state", and possible the "restart" parameter.

pliablepixels:
@knnniggett the issue that I see is that we then have to exclude "default" as a possible run state to display or switch to. I think zmpkg and the php code basically assume each entry in the DB is a valid custom state and then also adds the standard "start" "stop" "restart" (www/skins/classic/views/state.php:54)

Given that state change is infrequent, I wonder if this optimization would help.

connortechnology:
I don't have a problem with how this is implemented.  I wonder if the table will accurately represent the state... or if it will get out of sync easily.  I guess we can address that later. 

As to default state... no I don't think we need an extra row for it...

connortechnology:
we may need to add Starting Up and Shutting down states....

pliablepixels:
@connortechnology what situations do you think may put this table out of sync?
I've handled the following situation:

a) When ZM starts, stops or restarts, zmpkg.pl resets all states to 0
b) When ZM switches to a custom state, zmpkg.pl resets all states to 0 then marks the currrent state as 1

In the event ZM crashes, and starts again, a) will apply

I could not find any part of the code that does not use zmpkg to change state or start/restart

knnniggett:
Hmmm... don't we want a selected custom state to survive a reboot/restart of the server/zoneminder?  Maybe I misunderstood, but I thought that was why were writing which state was active to the dB, rather than just storing it in a global variable.

connortechnology:
The other thing is multi-server.  Each server can be in it's own state. So this might not be the place to put this.  I need to finish up my multi-server patches so that we can store the state in the server table.

knnniggett:
I'll manually merge this over the weekend.

knnniggett:
Merged.  Note that I think we should look into adding the ability to retain the custom state it is in after a reboot.

pliablepixels:
@knnniggett, so let's talk about the reboot state retain part a little more:

a) state = "my_state", user does a "stop" and after a while "start" --> what should the state be - "running" or "my_state"
b) state = "my_state" user does a "restart" --> what should state be "running"  or "my_state"
c) state = "my_state", ZM crashes/machine reboots/etc --> I can understand state could be "my_state"

I am unclear when ZM should really preserve old state and when it should just go back to 'running'

The problem is zmkpg "start" is used to start ZM (restart or first start) --> how do we discern whether that means keep old state or not depending on what may have caused "zmpkg start"

PR #900, opened 2015-06-22
Updated to work with zoneminder v1.28.1

soana:
copied changes from FI9831W (declared so perl does not complain of globals)
Verified on my system - confirmed to work


connortechnology:
So... you just removed all the nice code I put in there  to handle whether there was a http at the beginning, or to handle the old auth cases...

And, you've removed the version from the Agent string, why?

soana:
I could not get it to work with the existing version. Please feel free to
reject my version.

knnniggett:
The 9821w is one of the primary cameras I do my testing on.  
The next time I build the master branch, I'll verify the 9821 control scripts still works.

PR #902, opened 2015-06-24
Fix el7 build

bill-mcgonigle:
In search of debug symbols, I wanted to be able to use the standard rpm workflow (yum-builddep and rpmbuild) on EL7 to build zoneminder (from the zmrepo SRPM) and these two issues prevented that.  With these solved, the SRPM builds out of the box.


knnniggett:
Nice catch.  I might change the jscalendar download failed message to a warning to make it more obvious.

Since we started using cmake, ffmpeg headers would be found under /usr/include/ffmpeg, but the C compiler still wouldn't see them, thus zoneminder would fail to build. Symlinking the folders under /usr/inlucde/ffmpeg to /usr/inlcude has been the workaround.

Looks like commit #ec1c3cf fixes that, I would presume.  Thank you. I had been beating my head against the wall for some time now, not knowing how to fix that.

bill-mcgonigle:
Thanks, @knnniggett - it would be great if zmrepo had the debug symbols and source in place such that `yumdownloader --source` and `debuginfo-install` worked automagically.  That would help ensure everybody on rpm builds had the same symbols for debugging, at least.

knnniggett:
@bill-mcgonigle 
I've created testing repositories on zmrepo with recent builds of the master branch.
Debug and SRPMS have also been uploaded.
http://zmrepo.zoneminder.com/

PR #909, opened 2015-06-26
fix montage view issue in mobile skin

knnniggett:
Montage view did not work for the mobile skin as reported in this forum post:
http://www.zoneminder.com/forums/viewtopic.php?f=32&t=23236&p=88222

### DIFF
--- a/web/skins/mobile/views/montage.php
+++ b/web/skins/mobile/views/montage.php
@@ -63,7 +63,7 @@ foreach( $monitors as $monitor )
     $scale = getDeviceScale( $monitor['Width'], $monitor['Height'], $imagesPerLine*1.1 );
     $imagePath = getStreamSrc( array( "mode=single", "monitor=".$monitor['Id'], "scale=".$scale ), '&amp;' );
 ?>
-        <a href="?view=watch&amp;mid=<?php echo $monitor['Id'] ?>"><img src="<?php echo viewImagePath( $imagePath ) ?>" alt="<?php echo $monitor['Name'] ?>"/></a>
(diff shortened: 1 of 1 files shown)


explanar:
I my experience the fix does not work as it should: randomly only 50% of the monitors are shown in montage view of mobile skin. (Ubuntu 14.04, ZM 1.28 stable PPA)

The alternative Fix in /includes/function.php: remove ZM_DIR_EVENTS from return( ZM_DIR_EVENTS.'/'.$path );
https://forums.zoneminder.com/viewtopic.php?f=32&t=23236#p88209

works better.

In both cases i don't see Zone Preview Pictures!

knnniggett:
@explanar We are depreciating the mobile skin of favor of using a mobile app, such as zmNinja.
As soon as we release 1.29.0, in the next couple of weeks, we will be deleting the mobile skin from the project. See the 1.29.0 release notes shown on our releases page.

connortechnology:
It might be nice to keep it around for really ancient mobile devices.. thinking WinCE devices.

knnniggett:
The removal of the mobile skin was one of a few legacy features slated for removal. It was brought up in #1115 and feedback was asked for at that time. Seeing there were no objections, we moved forward with the plan to remove it and included that in the 1.29.0 release notes.

Similar to the argument that was originally made to consolidate the flat & classic skins into a single folder structure, it doesn't make sense to have a second set of skins living outside this. That creates duplication of effort, requires us to patch multiple folders when skin changes are implemented, etc.

Moving forward, if someone wants to create a new mobile-friendly skin, based on the new way of creating skins through css changes, then that would certainly be a welcome compromise.

PR #910, opened 2015-06-26
Add Servers Table and add Id PRIMARY KEY to States

connortechnology:
I can't imagine how this will affect anything, other than to put in place the minimum infrastructure for the isRunning PR to be updated and merged.

### DIFF
--- a/db/zm_create.sql.in
+++ b/db/zm_create.sql.in
@@ -386,11 +386,23 @@ CREATE TABLE `Monitors` (
 
 DROP TABLE IF EXISTS `States`;
 CREATE TABLE `States` (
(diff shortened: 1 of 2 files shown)


knnniggett:
Travis failed when it tried to apply zm_create.sql
```
0.03s$ mysql -uzmuser -pzmpass < db/zm_create.sql

ERROR 1064 (42000) at line 388: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '

  `Name` varchar(64) NOT NULL default '',

  `Definition` text NOT NULL,

  PRIMA' at line 2
```

It is not obvious to me where the problem lies.

PR #916, opened 2015-06-30
Api auth

connortechnology:
Adds auth layer to api.
Currently just denies access if auth is turned on.
Login does not work becase cakephp doesn't not work with mysql's PASSWORD hashes.

### DIFF
--- a/web/api/app/Config/core.php
+++ b/web/api/app/Config/core.php
@@ -216,7 +216,8 @@
  *
  */
 	Configure::write('Session', array(
-		'defaults' => 'php'
+		'defaults' => 'php',
+		'cookie'=>'ZMSESSID'
 	));
 
 /**
--- a/web/api/app/Controller/AppController.php
+++ b/web/api/app/Controller/AppController.php
@@ -34,6 +34,17 @@ class AppController extends Controller {
 	use CrudControllerTrait;
 
 	public $components = [
+		'Session',
+        'Auth'  =>  [
+            'authenticate'  =>  [
+                'Form'  =>  [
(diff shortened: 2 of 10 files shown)


knnniggett:
I think our previous conversation in irc about php crypt applies to this pr.  
I ran the test that @SteveGilvarry suggested on my CentOS 6 platform 
http://www.the-art-of-web.com/php/blowfish-crypt/

CentOS 6 ships with php 5.3.3 and the good news is that the test described in the previous link was successful.  RYPT_BLOWFISH is enabled on CentOS 6.

SteveGilvarry:
An example of changing hash method that I liked. Not sure if we have a way implement the immediate hashing of old passwords in db on upgrade, but at least to change password hash on logon.
https://stackoverflow.com/questions/19955793/migrating-users-table-with-hashed-password-from-old-php-app-to-new-laravel-app

connortechnology:
I will read that now, but a thought that I had was that if we supported both old and new methods for login, then on successful login we could update the password entry to something newer.  I think it needs to be done.  video surveillance is where security is paramount.

connortechnology:
Ooh, I like the double hashing..

devinellis:
- Shouldn't the API be stateless? why are we adding a session cookie?
- Isn't it a json/xml API? Why are we adding html views?
- Doesnt Zoneminder already support some sort of Auth? Can't the API honor that?

pliablepixels:
we should close this PR

knnniggett:
@connortechnology should we close this per @pliablepixels request?

PR #918, opened 2015-07-02
Add ServerId to Monitors

connortechnology:
Next step, add a reference to the Servers table into the Monitors table.


knnniggett:
We need to update this to add to  db/zm_update-1.28.100.sql and then bump the version.

connortechnology:
yah, done

PR #919, opened 2015-07-03
add check for gettime in librt, needed for building on pi

connortechnology:
Shouldn't affect anybody else.


knnniggett:
Technically, this is only needed when running Raspbian.  This issue does not appear on Fedora for the pi. :-)

connortechnology:
I think it has to do with the glibc version.  Newer merged it into libc.  So raspbian jessie probably doesn't need this either.  I just thought I'd head off the endless bug reports of people trying to build on it.

PR #925, opened 2015-07-07
fix 1.28.99 db upgrade script

knnniggett:


knnniggett:
Works w/o any error messages on my Fedora demo system.

@connortechnology take a look at the sql and let me know if you agree this does what we need.

connortechnology:
that might work, but I think I've solved it in a more logical fashion.  We will see.

PR #926, opened 2015-07-07
Version to 1.28.99

connortechnology:
If we leave the version at 1.28.1, the db update scripts won't run.  

Technically, we should have been doing 1.28.1.1 and so on until we got to 1.28.1.99... but we didn't do that.

Isaac


SteveGilvarry:
We need a better way to manage db updates against, but we should rename this one 1.28.1.1 or we will prevent any further DB updates before the next release.

knnniggett:
I think this will be ok.  We unofficially used .99 last time around to make it easier to spot a development release. 

This is the same thing we did prior to last release.  You can look in the /db folder and see that the 1.27.99 file is still there.  x.xx.99 does not cause any problems, neither should x.xx.100, x.xx.101, etc.  As long as we keep bumping the last number up each time we merge a commit prior to release, we should be fine.

In the grand scheme of things, I do agree that we need a better way to manage db updates to master.  The last time we talked about it, the three of us each our own idea of how to do it, but I don't think any of our ideas didn't come with side effects of some sort.  

Too bad github doesn't allow server-side git hooks.  We could write something that just happens automatically anytime someone merges a new script to the /db folder.

connortechnology:
So the first thing that is key is that the zm version needs to bump with the db update scripts. 

I propose that for now we just move on to .100 and so on.  

In future I propose that we do 1.28.1.1 etc

newburns:
I support @connortechnology proposition....
(Even though I have little support and/or voting rights :) )

knnniggett:
Merging.  We need to remember to bump this again with any future commit that modifies the dB schema.

PR #930, opened 2015-07-13
don't include .cpp in man

connortechnology:
just doesn't include the .cpp in the man filename.


knnniggett:
Aww damn.  Sorry for missing that one.  Are there any remaining pod2man issues that you are aware of?

PR #936, opened 2015-07-15
add error handling on failure to open serial port

connortechnology:


knnniggett:
Serial ports?  I remember those.

PR #944, opened 2015-07-16
add php-gd to list of dependencies for debian and ubuntu builds

connortechnology:


knnniggett:
It looks like the package is called php5-gd in Debianland.

Issue #945, opened 2015-07-17
please update bundled CakePHP


onlyjob:
CakePHP released numerous updates (some with important security fixes).
Please consider upgrading bundled CakePHP to 2.7.0 (or at least to 2.6.9).

See https://github.com/cakephp/cakephp/releases

onlyjob:
Sorry, I've already reported it in #814...

Issue #814, opened 2015-04-16
Upgrade bundled CakePHP


onlyjob:
Recently CakePHP released new stable version _3.0.1_ while our bundled CakePHP is _2.5.0-RC1_. Let's upgrade please? See https://github.com/cakephp/cakephp/releases
##

kylejohnson:
Nope.  3.x is not backwards compatible.  We won't be upgrading for a loooooong time, at least until the new ui + api is finished.

On April 15, 2015 8:44:38 PM EDT, Dmitry Smirnov <notifications@github.com> wrote:
>Recently CakePHP released new stable version _3.0.1_ while our bundled
>CakePHP is _2.5.0-RC1_. Let's upgrade please? See
>https://github.com/cakephp/cakephp/releases
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/ZoneMinder/ZoneMinder/issues/814

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

kylejohnson:
That said, we may be able to upgrade to a later 2.x release.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

onlyjob:
Fair enough. Then perhaps 2.6.3 would be the good version to upgrade to and it is already in Debian. ;)

SteveGilvarry:
Don't want to say it but hoping this is the last version of CakePHP we have.

PR #949, opened 2015-07-17
Upgrade bundled CakePHP to version 2.7.0 (Closes: #945, #814).

onlyjob:


knnniggett:
I don't know if anyone else other than @kylejohnson is qualified to answer whether or not this version of CakePHP is compatible with our project.

SteveGilvarry:
Not ready for merge, need to pull this apart more as looks like it is overwriting zm code with standard framework files. https://github.com/ZoneMinder/ZoneMinder/pull/949/files#diff-f0f8b358fe760ba46e815b8553b6877eL72

onlyjob:
Two points of confidence:

 * "CakePHP release notes" promises API stability and full compatibility with previous 2.x releases.
 * Experimentation confirms the above statement (although I'd like someone else to try 2.7.0 as well). Zoneminder works fairly well with CakePHP 2.6.8; official Debian package already uses external/newer CakePHP for a while.

I've made an effort to avoid overwriting any Zoneminder code. Please point out if you spotted any problems. Specifically regarding dropping `CakePlugin::load('Crud');` I dropped it deliberately (to reduce delta to simplify future merges) because Crud directory is empty. Please let me know it that was a mistake. I could not find any traces of that plugin. As far as I can tell my local Zoneminder instance works perfectly with that change.

Besides customising 3rd parry library in source tree would be a serious error...

SteveGilvarry:
I agree on the 2.7 side, and not 100% on this but a lot of deletes in the app folder and that is where the code for the API is. I think a CakePHP update would be lib or vendor folders.
Crud is git submodule linked.

onlyjob:
A submodule, of course. :)
_Crud_ must be a recent addition -- I have checked latest release tarball and did not find it there. I restored `CakePlugin::load('Crud');` as well as `app/Controller`, `app/Model` and other customisations I've missed. Please advise if anything else should be preserved of if it would be better to leave `app` alone?
I couldn't identify anything else that is not part of the framework (I did 3-way compare between cake-2.5.0-RC1, repo and cake-2.7.0).
To make such work easier perhaps it would be better to mark customisations as such by leaving READMEs, comments and copyright headers...
`lib` is identical to original framework so there are no customisations there. Only `app` folder needs review.

connortechnology:
Another thing is that I don't think we should be including cakephp at all anyways. Especially in packages. cakephp comes already packaged, we just need to figure out how to use it.

Another point to make is that the classic UI does not use cakephp in any way.  Only the new api uses it.  Our use of it is so basic that I would be amazed if anything would be broken between 2.? versions.

SteveGilvarry:
It does seem to be a clunky process, app folder is a skeleton you build your own app in, as opposed to taking a copy. 3.0 seems to have split it out a little more.
A lot of the php stuff seems to want composer for install, cake 3.0 for example, but no idea how to integrate that to cmake. Plus not sure it fits in to Deb packaging process.

devinellis:
My old Composer PR https://github.com/ZoneMinder/ZoneMinder/pull/744 it's pretty trivial and should be easy to integrate into the current build process

onlyjob:
@devinellis your #744 may be trivial for _build_ but not for _release_ as it breaks comprehensive-ness of release tarball. You've never answered how shall we build packages once you remove bundled CakePHP from source tree.

devinellis:
I don't see much of a functional difference with git clone or submodules. I assume that is done during release--the copied source gets included in the release tar? Can't Composer packages be included in the same way? I don't have any experience delivering tarballs.

onlyjob:
Source tarballs downloaded from GitHub include no submodules. I'm not sure how source releases will be generated and where they will be published to if GitHub source releases won't suit package builders any more. What's the point of source tarballs if one can't build software from it?

knnniggett:
Obsoleted by #1306

Issue #950, opened 2015-07-17
Monitor configured with ffmpeg shows last part of URL instead of hostname in overview


schrorg:
Hi,

if a monitor is configured for ffmpeg, the main view does not show the correct part of the configured ffmpeg URL. For example, if I configure the url "http://hostname:port/stream", the main view should display "hostname" in the source tab, but instead displays "source" (the last part of the URL).

For libvlc sources, the file 'web/skins/classic/views/console.php' uses a parser, but not for ffmpeg sources. I will provide a pull request to correct this behaviour, shortly.

PR #951, opened 2015-07-17
Show correct part of URL (hostname) for ffmpeg sources in console

schrorg:


schrorg:
Closes #950

knnniggett:
Thanks for the help.

Issue #583, opened 2014-11-18
Wrong encoding for some lang files


manupap1:
Some lang files are encoded to non UTF8 encoding and this results in a bad translation for special characters. For example (french lang file fr_fr.php encoded to ISO-8859):
![zm_encoding_issue](https://cloud.githubusercontent.com/assets/7723405/5091552/cfde810e-6f49-11e4-8eb6-d99611e5b1bd.png)

All lang files should be encoded to UTF-8.
##

cpasqualini:
Workaround, at least in Debian official packages for my language:
cp /usr/share/zoneminder/www/lang/es_ar.php .
iconv -f iso8859-1 -t utf8 es_ar.php >/usr/share/zoneminder/www/lang/es_ar.php

best regards

cpasqualini:
Workaround, at least in Debian official packages for my language:
cp /usr/share/zoneminder/www/lang/es_ar.php .
iconv -f iso8859-1 -t utf8 es_ar.php >/usr/share/zoneminder/www/lang/es_ar.php

best regards

cpasqualini:
Hi all
Any progress on this issue?

knight-of-ni:
Fixed in #2131. Closing.

cpasqualini:
Thanks for merging this!!

PR #952, opened 2015-07-17
Change encoding of german language file to UTF-8

schrorg:
This addresses #583, but only for the german language.


knnniggett:
Looks right, but I've noticed a couple instances where we seemed to be getting conflicting pull requests from people with a couple of the other languages.  

@manupap1 I think you are more familiar with the language code than I.  Does this pr look ok?

manupap1:
There is no conflict.
This pr is ok to me.

PR #954, opened 2015-07-21
953 configs api return all values

pliablepixels:


pliablepixels:
Please only consider this change https://github.com/pliablepixels/ZoneMinder/commit/0235573cddee18fa2e2a975b3e78c6bbf19af42d -- don't know why the DB update came up. It was resolved earlier and merged

PR #961, opened 2015-07-23
960 montage review new feature

Linwood-F:
This adds a different sort of review display, not (explicitly) tied to the events view, but rather a montage of camera images with a timeline and a slider so you can scrub through them over time.  It is lightly tested on IE, Chrome and Firefox, and purposely uses newer HTML5 instead of the clustered DIV's in timeline so as to be more browser friendly; it also adjusts dynamically on window resize to be more friendly to tablets and phones (though it requires so much bandwidth and CPU that it might not be suitable to smaller devices at least on cellular vs wifi).  I have no idea how well this fits with the "ZoneMinder" philosophy so offer it primarily for comments and suggestions at this point, though as far as I know it should merge and run.


Linwood-F:
I have updated this with several new features, notably a live view and a fit-to-browser feature (on by default).  I think this has the potential to be a replacement for the montage view as well as the timeline, though it does not do MPEG replays and does not support all the various ways timeline can be called (I frankly cannot figure out what they all are). 

By switching to a pull scheme this avoids the issue with connection limits in browsers.  It is self-limiting in that it adjusts the rate at which it attempts to pull to try to balance response time so as not to bury the browser.

Finally while you can scale monitors arbitrarily (and let the browser scroll), it is aimed more at having a fixed browser size and scaling the monitors to fit.  It automatically scales monitors initially to have similar display area (e.g. a monitor with half the horizontal and half the vertical resolution as another will be scaled up by 4x relative to the larger monitor).  It provides the ability to also enlarge an individual monitor (or shrink it) and rearranges the screen accordingly, so you can enlarge a monitor of interest while reviewing history, or while watching live.

Because of all the sizing and resizing and rearranging it needs more testing.  I have tried in IE, Chrome and Firefox (current versions). It has minimal testing with mobile devices.  I would love if someone else could take a look at it, see if it merits continuing down this path.  Here are a few screen shots.

The first is the history view -- darker links on the timeline indicate scored frames.

History Review

This is the live view (the timeline is suppressed in this view, and the Live button changes to History to switch back: 

Live view

And here is an example after clicking the + in the corner of a monitor a few times to enlarge it:

Showing Zoom

pliablepixels:
@Linwood-F saw your call to get feedback on IRC. Happy to review.

IT looks like something that could be of great value to the web console, but I really could not get this to function at all - it looks like it is better than the first version I tried but I landed up into the following issues:

(Environment: OSX, Safari, Yosemite. Using montagereview.php inside my home LAN where ZM is also installed)

a) When I first went to https://myserver/zm/?view=montagereview, it took more than 20 seconds for the view itself to load. Is it possible to show some sort of loading screen ? i.e. transition to montage review immediately instead of waiting for any DOM elements to load and show a loader

b) I tried refreshing many times, but by the time an image loads ups (even a single frame) its well over a minute. In most cases it seems to be stuck at 'evaluting fps' forever. See screenshot - my slider is on what I think an event areas (red across multiple monitors) but I see no images

c) Not quite sure how the montage fit is written but note the size of the borders - wonder why one of my monitors is so tiny - I assume its my 640x480 and relative to the others (1280x960) it is half but not sure if it should by default show up as so small

d) Finally, the place holder image that says there is no event seems to be in serif style. I wonder if it would be better to make is sans (un-adorned, as that is more in line with the flat skin a lot of people seem to be using)

e) It looks like changing the slider results in a full page refresh - is it possible to only update the view of the montage and not the full page

d) I couldn't test much more as I really could not get it to react even after waiting at least a minute after each slide, which is odd --> I am on the same LAN

Linwood-F:
Re (a), mine loads in 4 seconds.  If you are running a recent copy (was it from today), it should load only the last hour of history, and default to live view (it still loads the hour of history in case you hit the history view).  The clear intent is that it not take long to load; that it is, is a problem, and I think I need to fix that rather than a loading screen.

Re (b) "evaluating fps" that does not make it wait, it just does not display anything there until it has about 15 complete cycles of load times (all images).  This is just a symptom of some loading problem as well I think.

(c) There's something wrong.  All my monitors are the same size, and I suspect I am doing something wrong (my guess is backwards) in the normalization step.  That code has been rewritten about 3 times since I first tried it, my guess is I flipped a divide to a multiply somewhere.  Let me set up a small monitor and test.  The intention is that small monitors are scaled up, and large ones scaled down to a happy medium so they take up the same area (it does this regardless of whether in "fit" mode or not).  So this looks definitely like a bug.

(d) Um... sure.  Though in retrospect now that I know how I think I'll just change it to be drawn on the canvas instead of an actual image to load.   So Arial?

e) Moving the slider should not result in a page refresh, ever.  That's why it loads and caches everything in javascript arrays.  The only thing that results in a page refresh (a real one) are the buttons that load a different time frame, or there is what may appear to be a refresh for either a switch from live to history or fit to non-fit as the browser repositions everything, but there is no actual refresh there.

The concern I have is the load times you are seeing.  I'm on a lightly powered ZM setup as well (or thought I was :)  ), in one case it is a HyperV VM running in my desktop while I use the desktop, and the desktop is about 3 years old itself.   One possibility is that you have a very large FRAMES and EVENTS table.  Can you run a couple MySql queries by chance?   See both how long they take and what they say?   The inner queries are the two that run when the page loads wide open (all data for all times), and on mine takes about 6 seconds to run, and shows about 12,000 which are how many events it loads (which are a lot), and 7,600 which are how many frames it reads through -- it consolidates as it reads so it is hard to know without looking in debug how many array elements it uses.

This at least will tell if it is mysql that is slow for the page -- but both mysql and the php web generation should be much faster with the default -- which is one hour's worth only of data.  So if that's slow as well... well, I am grasping a bit.

I can run this on a 5 year old android phone (Galaxy SII) and it comes in in about 8-9 seconds for the first screen.

Now all that said -- Safari is a browser I have not tried, it runs on IE, Chrome and Firefox for me (on windows).  Let me see if I can download a Safari for windows and give it a go -- no Mac's in the house here.

THANK YOU FOR TRYING IT!!!!

select count(*) 
from 
(
  select E.Id,E.Name,UNIX_TIMESTAMP(E.StartTime) as StartTimeSecs,UNIX_TIMESTAMP(max(DATE_ADD(E.StartTime, Interval Delta+0.5 Second))) as CalcEndTimeSecs, E.Length,max(F.FrameId) as Frames,E.MaxScore,E.Cause,E.Notes,E.Archived,E.MonitorId
  from Events as E
  inner join Monitors as M on (E.MonitorId = M.Id)
  inner join Frames F on F.EventId=E.Id
  where not isnull(E.Frames) and not isnull(StartTime) 
  group by E.Id,E.Name,E.StartTime,E.Length,E.Frames,E.MaxScore,E.Cause,E.Notes,E.Archived,E.MonitorId
) x
;
    
select count(*) 
from 
(
  select E.Id as eId, E.MonitorId, UNIX_TIMESTAMP(DATE_ADD(E.StartTime, Interval Delta Second)) as TimeStampSecs, max(F.Score) as Score
  from Events as E
  inner join Frames as F on (F.EventId = E.Id)
  where not isnull(StartTime) and F.Score>0
  group by E.Id, E.MonitorId, F.TimeStamp order by E.MonitorId, F.TimeStamp asc
) y

pliablepixels:
@Linwood-F , I just tried it on Win7/Firefox - exactly the same issues. Yes, I took your php and gif file as of 2 hrs ago.

pliablepixels:
Sorry, missed your request to run the SQL queries.

Query 1:
```
+----------+
| count(*) |
+----------+
|     8492 |
+----------+
1 row in set (18.62 sec)
```

Query 2:
```
+----------+
| count(*) |
+----------+
|    76478 |
+----------+
1 row in set (1.15 sec)
```

Linwood-F:
OK, I think I've made progress on this.

The biggest issue was load speed, and I think that was due to a query I have now optimized, plus made some other changes for code size and speed.  My loads for 2 weeks of data (continuous 7 camera 2 fps recording) is about 3 seconds.

The scaling issue for low res cameras was a reversed normalization factor, but that uncovered a few other issues related to scaling, now fixed.  Though it also uncovered an issue in the image scaling routines (another issue and PR open for those).  They should only show up if you have recent code; if so consider looking at #997.

I got rid of the graphic for "no data" and just wrote text in the window, sticking to simpler "No Data" and a sans serif font.  I could not find an issue with screen reload.

I'm still looking into the safari issues... at present best to test with IE, Chrome, or Firefox.

Linwood-F:
And now I think I've resolved the safari issues I can find.

Linwood-F:
Just to be clear while I think this works, I think it is premature to merge as I want to keep adding some features as I get feedback (and oh... feedback solicited). 

Right now I'm unhappy with the "fit" algorithm, but not sure how much better i can make it, so I may make it so one can manually resize and reposition monitor screens.  Except I have no idea how to do that yet.

knnniggett:
@Linwood-F It's been a while, but I like really like this. Did you plan to do any further development for this or are you waiting for us to provide more feedback?

Let me bring this up with the rest of the team and see what they think.

Linwood-F:
I've been using it continually myself, but it has some issues I intended to pursue further.  My real concern is that it seemed at odds with the whole video branch.  I don't think pulling historical frames from video is going to be fast enough to be useful.  Pulling live frames will likely work, so in live mode it's likely a workaround for the streaming limitations in browsers (albeit at a dynamically set but lower frame rate).   What would really be interesting is seeing if someone on the video branch could give it a go and see if it works at all.   That's on my to-do list but rewiring my attic and keeping up with basketball season (I photograph it) have somehow got in front of programming.

If it's thought to have a future I'm more than game to pick it up again and do some work on it.

connortechnology:
I vote to merge.  This doesn't affect anything else.

MadNachos:
Great addition, its working pretty well on my system outside of a few php related Apache errors.

Linwood-F:
Please feel free to provide some of the errors.  I plan to get a test system set up again and work on this some more in the coming weeks.

MadNachos:
At the moment the one I am seeing only comes up when clicking on the '1 hour' or '8 hour' boxes but it seems to work OK.

`[Sun Feb 14 12:52:18.062675 2016] [:error] [pid 21360] [client 192.168.11.2:37738] PHP Fatal error:  Call to undefined function imagejpg() in /var/www/zm/views/image.php on line 135, referer: http://192.168.11.6/zm/index.php?view=montagereview&fit=1&minTime=2016-02-14%2011:52:15&maxTime=2016-02-14%2012:52:14&current=2016-02-14%2012:22:13&displayinterval=1000&live=0&scale=1&speed=1`

Linwood-F:
I think that's an unrelated change made (sigh, by me) on a commit on Aug 9 to image.php to add scale as a optional feature.
` 
if($width==$oldWidth && $height==$oldHeight)  // See if we really need to scale
        {
            imagejpg($i);    >>>>>>  Mis-spelled add "e" to jpg
            imagedestroy($i);
        }
        else  // we do need to scale
        {
            $iScale = imagescale($i, $width, $height);
            imagejpeg($iScale);
            imagedestroy($i);
            imagedestroy($iScale);
        }
`

I am not set up right now to do a PR for this despite it being one character, house full of guests and have to spend some time getting a VM set back up.  

Can you edit that one line shown above and see if that fixes it appropriately, and if so I can either do a pull request later, or maybe someone else wants to.

Linwood-F:
Sorry, not sure why that is all run together, but hopefully you can see it.  It's the only place in image.php that the word **imagejpg** appears, as opposed to **imagejpeg**.

MadNachos:
Ahh..but of course, if I had paid attention to the highlighting in 'vi' I would have noticed that myself ;) 

Adding the 'e' took care of it. Thanks.

PR #965, opened 2015-07-26
Updates for feature-h264-videostorage

SteveGilvarry:
Mostly merging master, but looks like I had some other changes sitting around. If you want me to split them let me know. 
Some of my changes to debian distro folder got pulled in. These changes were using the debian packaging details from official package, merging some patches and pulling some out that were not essential.
Also a clean up of the videostore code as lots was not being used.


connortechnology:
I'm not wild about this other stuff getting sucked in, makes it hard to discuss the merits.

SteveGilvarry:
@connortechnology I don't disagree, didn't even realise I had done that until PR was done, the first part was all about pulling in Debian packaging, so the first two are patches in there. I will split them off into 2 PR's for master so that is avconv and apache logging.
A third for master at some point to bring official debian package into Upstream debian folder?
Then I can get back to the h264 specific ones. I hate cleaning up my mess in git.

PR #966, opened 2015-07-26
Auto generated changelog

SteveGilvarry:
FYI just played with https://github.com/skywinder/github-changelog-generator. Interested in opinions of something like this versus someone taking the time to maintain this.


knnniggett:
Good idea.  I think it beats what I have been doing, which was using git log to puke out some text, which I would then paste into the Github Release page.

Question: How/when would this be called?  Manually, whenever?  Or is there some way to auto-trigger this, like Travis?

SteveGilvarry:
It is a little bit chicken or egg, as the release needs to be tagged before it can be included in auto generated changelog, but the release needs an updated changelog. I haven't seen Travis push changes back, normally just status of build.

SteveGilvarry:
Looks like they already dealt with my chicken or egg, https://github.com/skywinder/github-changelog-generator/wiki/Advanced-change-log-generation-examples future-releases allows you to tag unreleased changes as the next version

PR #968, opened 2015-07-27
Apache.conf modifications

SteveGilvarry:
Most of these have come from official debian package. Remove trailing slash from ScriptAlias, add Require All Granted and comment out second log option. Comments welcome?

### DIFF
--- a/misc/apache.conf.in
+++ b/misc/apache.conf.in
@@ -14,17 +14,18 @@
         AllowOverride All
     </Directory>
 
-    ScriptAlias /cgi-bin/ "@CGI_PREFIX@"
+    ScriptAlias /cgi-bin "@CGI_PREFIX@"
     <Directory "@CGI_PREFIX@">
         Options +ExecCGI -MultiViews +SymLinksIfOwnerMatch
         AllowOverride All
+	Require All Granted
     </Directory>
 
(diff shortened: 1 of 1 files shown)


knnniggett:
Huh, didn't realize we had an apache config under /misc.  Always used the distro-specific ones under out of the /distro folders.

Couple of things:
1) The script, as it is written, will only work with Apache 2.4.  We should probably make it work with Apache 2.2 as well.  Here is an example:
https://github.com/ZoneMinder/ZoneMinder/blob/master/distros/redhat/zoneminder.el7.conf.in

2) This script has a default virtualhost directive in it.  However, most distros will already have a default virtuahost directive installed with the web server package.  Will having two <VirtualHost *:80> entries cause an overlap and/or create any other issues?

SteveGilvarry:
1. Lines 1 - 10 of that example reminds me to fix that sometime. Also it refers to a line below but I don't think that is there anymore, used to be a line that made the site not work post install due to that "insecurity" of images not being secure. But I will grab the rest of that example and integrate it.

2. These were mostly patches that the debian package was making and I figured we could merge upstream, but the distro one is better, I will merge them together and update this.

knnniggett:
See commit 800d7dd33ac2fedfaec612bf82b53212091cf26e as a follow-up to this pr

PR #974, opened 2015-07-28
Fix to display thumbnails when only storing snapshot.jpg

SteveGilvarry:
One of the new options in feature-h264-videostorage is a flag per monitor for saving jpegs. This checks that flag and adjusts thumbnail display based on mode.


knnniggett:
keep 'em coming.

PR #989, opened 2015-08-05
CSS[skins/classic]: highlight current row in tables.

onlyjob:
Minor CSS usability improvement.


onlyjob:
Wow, that was quick. :) Thank you Isaac for your awesome work. :)

Issue #991, opened 2015-08-06
Add option to make TimeStamp larger


schrorg:
I am working on a patch to double-size the timestamp in pictures (for better readability in scaled versions). I therefore created a "zm_bigfont.h" with a larger font and did a few modifications in zm_image.h and zm_image.cpp. Everything seems to work now, but I would like to make the size configurable for either every single monitor or the whole system.

Are there any concerns in making this an option? And would you prefer to have that a global option or an option per monitor?

schrorg:
Everything I've done until now is in https://github.com/schrorg/ZoneMinder/tree/large-timestamp

knnniggett:
We need you to take this a step further and generate a pull request.  
That is not a problem if you feel it is not quite ready.  Simply state that in the description.  
This makes it easier for us by highlighting exactly what you changed, and it makes it easier for you by avoiding future merges to your branch as new commits are added to the master branch.

pliablepixels:
@schrorg I looked at your changes. It looks like your code is basically replicating pixel rows of the font as ZM does not really use a font vector, it defines a bitmap for each character in an array. What is your thought on readability? While it does increase the size, do you find the resulting text is blocky and hard to read? IF so, would it make sense to include an option to overlay a vector font that is royalty free? (Any google font can be used as a local font file and is license compatible afaik)

schrorg:
@pliablepixels I created the big font data array by first duplicating every bit (pixel) horizontally and then duplicating each line. So basically, I made four pixels out of one. I did this with a few lines of perl. But the effect is a bit like if you would view a video in 2x scale:
![large-timestamp](https://cloud.githubusercontent.com/assets/7068299/9129639/91a67cc0-3cda-11e5-8ff1-fdeb9b3f9ba1.png)
It is absolutely okay for us but sure could be better.

I will see and talk to some colleagues about a vector font. It seems to me that the piece of code should be kept as fast as possible and I have no idea about how changes in that area would impact the image processing performance.

pliablepixels:
Thanks - it looks much better than the tiny display today. I'd recommend you do a PR as @knnniggett suggests with this enabled. You should also look at enabling this option via the web console otherwise people won't be able to use it.

schrorg:
@pliablepixels I did a pull request (#992) and I am working on an option for the web interface.

pliablepixels:
@schrorg - nice! thanks.

knnniggett:
#992 was merged. Closing #991 .

PR #992, opened 2015-08-07
Add option to make TimeStamp larger

schrorg:
This pull request addresses #991. It is basically working but not yet finished.


wambs:
Hi,

Can you also make the update where you can adjust the time stamp smaller as we.. Its too big on my output.

Kevin

Linwood-F:
Am new to how things are done -- where would a user make the selection of the larger size?

schrorg:
@Linwood-F Until now you have to adjust the constant ZM_TEXT_SIZE in zm_image.h and set it to either 1 (small/normal size) or 2 (double size). And then recompile.

@wambs, I don't understand what you mean with "smaller as we..". Do you want to have a size between the normal size and my large size?

Linwood-F:
@schrorg , Ah, got it.  Thank you.  I was hoping for something I was doing for an undiscovered but (non-building) user accessible technique that I had missed, as adding a new option through all the screens and SQL looks rather daunting and relatively undocumented.

schrorg:
I finished adding the font size option to the user interface. It can be found in the timestamp panel of every monitor.
Everything seems to run smoothly in my test environment, so I would consider this to be finished and ready for merge.

connortechnology:
For the most part, looks good to me.

Couple of things: 
#1, why does this include @pliablepixels API changes?
#2, you don't need to add lines to each language.  You just add it to en_gb and it will fall back to that. Only put actual translations into each translation file.
#3 in Image::Annotate you didn't make the int size a const...
#4 in zm_image.h you seem to have messed up the whitespace on that enum line...

schrorg:
Oops. Sorry for that. I fixed everything.

Lots of english text in all the language files made me assume, I would have to insert the string into every translation.

I'm not sure how exactly @pliablepixels API changes got into my branch. I guess I did a rebase in a wrong way or so. I rebased again (interactive) and leaved out @pliablepixels commits, so everything should be fine again.

schrorg:
I bumped the version to 1.28.102 and re-merged my branch, so that it is up-to-date with the base again.
Has anyone seen any issues with the larger timestamps so far?

knnniggett:
Why are you bumping the version to .102?
Prior to release, we only bump the version when a PR changes the dB schema.

Also, there already is a PR in the works, which is using .102.

schrorg:
Yeah, I was not sure how to deal with the version bump in master so I decided to go one version up with my changes.
I bumped to .103 now and added my SQL changes to zm_create.sql.in.

knnniggett:
Thank you.  I'm going to change this to "Not Ready for Merge" ... not that there is anything you still need to do, but because there is another PR out there that modify's the same files.  Merging the other PR first will be less consuing.

Issue #995, opened 2015-08-07
Events per page for APIs - too low


pliablepixels:
As of today, the APIs returns the configured # of items in ZM_WEB_EVENTS_PER_PAGE per API seek when they are paginated. This by default is 25 and is way too less for an API.

While this is acceptable for most of the APIs, the problem child is the Events API which runs into several thousand events. Returning only 25 at a time wastes too much of network resources.

I propose we change this limit to 100 for now, just for the events API. In future, we should create a nice item called ZM_API_ITEMS_PER_PAGE that controls pagination for all items.

PR #996, opened 2015-08-07
995 events count per api (bumps up # of events reported per API call)

pliablepixels:
Address issue https://github.com/ZoneMinder/ZoneMinder/issues/995


knnniggett:
sounds reasonable.  This pr also obsoletes #994

Issue #997, opened 2015-08-07
Image::Scale up has rounding issues which can cause corruption


Linwood-F:
This is a side effect of #932 and a related commit.

The Image::Scale routine has a home-grown scale up, and separate scale down routine in line. The scale up second has some rounding issues, as it steps through the old image to build the new, it tracks the position int he new image buffer.  Under certain scale factors combined with image widths, this tracking gets off, and causes the image to skew  (i.e. rows end up being too short, or too long, and so the next row starts shifted a pixel(s), all the way to the end).  

In the past, before #932 and associated fix, this generally caused the image to be larger, overrunning the buffer that was allocated based on the anticipated size, and returning the unscaled image (but logging an error).  Since mostly these were returned to an explicitly sized <img> tag, this was not noticable.

With the corrections in the above, it returns the scaled, but (occasionally) corrupted image, in some ways being worse than before the buffer overflow was fixed.

I have found a fix in the current scale up program.   It seems to require three kludges.  At the end of each row, beginning at current like 2633, and later at 2642, last_w_index and last_h_index are assigned from w_index and h_index respectively.  Each might exceed the row and height, so become:

last_w_index = w_index <= new_width ? w_index : new_width; 
last_h_index = h_index <= new_height ?  h_index : new_height;

And right after the latter, we need to reposition the running output pointer if it got skewed:

 // redo ps as it seems to be slipping at times                                                      <
pd = &scale_buffer[ last_h_index \* new_width \* colours];

Note that all of the above is something of a kludge to make up for a flawed algorithm.  There is likely a simple fix somewhere in the rounding in the current flow, but I did not see it.

Instead, I recommend adopting a more standardized algorithm that can do both up and down scaling.  I'm going to commit code shortly for a home-grown bilinear interpolation and some testing, once I clean that code up.

[...]

PR #998, opened 2015-08-07
Replace scale algorithm in Image::scale

Linwood-F:
There's a long discussion of where this came from and why in issue #997 .

I've suggested several alternatives there, and propose this as my preference since it is a cleaner and shorter algorithm, fairly standard and well know and reliable as algorithms go (obviously flaws in my coding are possible), and uses the same for up and down.

I think this works as shown, but it could use a very hard look and people's opinions.

PS. There is one slightly unrelated change, near the top -- in case things are scaling up or down but very close to 100% (as more dymanic programs like mobile apps and alternative views might), this gives a bit of a boost.  A shame to run through the whole scale for 98% for example, just let the browser do any fine tuning.

### DIFF
--- a/src/zm_image.cpp
+++ b/src/zm_image.cpp
@@ -2591,107 +2591,63 @@ void Image::Scale( unsigned int factor )
 		Error( "Bogus scale factor %d found", factor );
 		return;
 	}
-	if ( factor == ZM_SCALE_BASE )
+        // Don't scale if we are really close in size
+        if ( factor <= ZM_SCALE_BASE + 5 and factor >= ZM_SCALE_BASE - 5 )  // within 5%
 	{
 		return;
 	}
 
 	unsigned int new_width = (width*factor)/ZM_SCALE_BASE;
 	unsigned int new_height = (height*factor)/ZM_SCALE_BASE;
-	
 	size_t scale_buffer_size = (new_width+1) * (new_height+1) * colours;
-	
 	uint8_t* scale_buffer = AllocBuffer(scale_buffer_size);
-	
-	if ( factor > ZM_SCALE_BASE )
-	{
-		unsigned char *pd = scale_buffer;
-		unsigned int wc = width*colours;
-		unsigned int nwc = new_width*colours;
-		unsigned int h_count = ZM_SCALE_BASE/2;
-		unsigned int last_h_index = 0;
-		unsigned int last_w_index = 0;
-		unsigned int h_index;
-		for ( unsigned int y = 0; y < height; y++ )
-		{
-			unsigned char *ps = &buffer[y*wc];
-			unsigned int w_count = ZM_SCALE_BASE/2;
-			unsigned int w_index;
-			last_w_index = 0;
-			for ( unsigned int x = 0; x < width; x++ )
-			{
-				w_count += factor;
-				w_index = w_count/ZM_SCALE_BASE;
-				for (unsigned int f = last_w_index; f < w_index; f++ )
-				{
-					for ( unsigned int c = 0; c < colours; c++ )
-					{
-						*pd++ = *(ps+c);
-					}
-				}
-				ps += colours;
-				last_w_index = w_index;
-			}
-			h_count += factor;
-			h_index = h_count/ZM_SCALE_BASE;
-			for ( unsigned int f = last_h_index+1; f < h_index; f++ )
-			{
-				memcpy( pd, pd-nwc, nwc );
-				pd += nwc;
-			}
-			last_h_index = h_index;
-		}
-        new_width = last_w_index;
-        new_height = last_h_index;
-	}
-	else
-	{
-		unsigned char *pd = scale_buffer;
-		unsigned int wc = width*colours;
-		unsigned int xstart = factor/2;
-		unsigned int ystart = factor/2;
-		unsigned int h_count = ystart;
-		unsigned int last_h_index = 0;
-		unsigned int last_w_index = 0;
-		unsigned int h_index;
-		for ( unsigned int y = 0; y < (unsigned int)height; y++ )
-		{
-			h_count += factor;
-			h_index = h_count/ZM_SCALE_BASE;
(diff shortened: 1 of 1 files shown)


connortechnology:
I'm ok with this.

knnniggett:
I'm not sure I can agree that the slight benefit this PR brings to the image quality is worth using an algorithm that, according to #997 , is twice as slow as what we currently have.

Perhaps if the difference was more dramatic or the algorithm was not significantly slower.

Looking at the sample images, I still can't say I can clearly see a difference.  

Further discussion on this is welcome, especially from the other team members.

Linwood-F:
Let me add a couple of points.

First, the motivation for the change is that the old algorithm is incorrect.  It does not accurately track array bounds as it iterates, which is what has caused the errors in allocation in the past, and the skewing of images at certain magnifications.  So the motivation was to make it actually correct, and not allow run-amok array bounds.

Secondly, the slowdown is most noticeable in large enlargements. Both algorithms scale exponentially, so relatively small underlying differences are magnified as you go from 100% to 200% to 400%, etc.

It is not clear whether this is slower or not (or by how much) in scaling down.  My GUESS Is most scaling in ZoneMinder is down, not up.  Perhaps some of the longer term users can comment there.  My speed testing was done as an average over a complete run from something like 10% to 400% by 1% increments.  With exponential increase, this means the average is weighted almost entirely to the very high end.  It would be worth trying to arrange a more isolated test (not involving ZMS and web display) and testing it in smaller scales, say 25% to 150% and see individually how it does.  I will try to do that.

Finally, I think it is work explaining the algorithm difference a bit. In the prior zoneminder scale up algorithm, it is written as some kind of a push loop, i.e. it loops through the old image by bit, and decides where to push data from that bit into the new image's one or more bits. Since the ratio is almost never integral, it has to round to deal with fractional bits, and it is in this rounding that it is going wrong.  The proposed algorithm is a normal resampling algorithm, which pulls data for each bit in the new image, making it more understandable (for whatever that is worth) and also more simple to ensure it honors array boundaries.  And it provides the same mechanism for up or down (again, more about human understanding).

My GUESS by the way, is if you eventually want to move toward library routines you are going to find this kind of choice frequently -- hand written, purpose-written code may be faster than a packaged, proven algorithm.  Now THIS is not a packaged solution (though it is an implementation of a proven algorithm, but it could contain coding errors).  But this particular discussion is going to resurface when you do incorporate packaged algorithms, as I doubt seriously they will not bring with them more overhead.

If I can isolate and test I will post some numbers.

Linwood-F:
OK, I have actual times.  These are run on a HyperV VM so it is hard to say exactly how this compares to any given system, but for reference the VM will run 6 cameras at 2048 x 1536 and be about 40% busy.  These times were taken without zoneminder running.

The test was to take a 1024 x 768 JPG image and do 100 scales at various size, timing the loop, multiply by 10 to get milliseconds per scale (including one image copy overhead, which falls out at the 100% line). 

Scale (%)|New Time (ms)|Old Time (ms)|Increase
---------|-------------|-------------|--------
25|2.71|1.78|52%
50|6.26|1.54|306%
75|13.60|2.77|391%
100|0.39|0.38|<< no scale, overhead only
125|35.85|4.81|645%
150|52.12|5.59|832%
175|72.40|6.61|996%
200|94.64|7.36|1187%
225|117.20|8.61|1261%
250|143.35|10.03|1330%

As you can see, the overhead of ZMS and the web was hiding the actual increase, it is much larger than 3x.  On the other hand, the actual scale time is in small milliseconds when scaling down.  13ms (the 75%) supports 77 fps, but it is indeed almost 4x as much elapsed time (basically pure CPU) as the old program at just shy of 3ms.

The actual scale times go down with file size also.  My GUESS is that a 1024x768 image is never scaled up in real practice, but it is small VGA (etc.) images that may be scaled up.  I did the test again with a 640x480 image, and it scales like this: 25% = 1.7ms, 75% = 5.2ms, 150% = 20ms, 250%=56ms.   Comare the 56ms to the 143ms, and you can see it get better, but it still is long in comparison to the old algorithm.

I tend to be one that thinks you don't swap fast, incorrect code for slower correct code.  However, I identified over in another note a way I think you can prevent the bounds overrun in the old code (it does not make it right, it just makes it safer).  If you guys prefer that, it's easy to swap out.

We could also probably turn this into a nearest neighbor pull sample, though I doubt it makes a huge difference (they are the same order, just a bit less math for each bit). 

And someone better at optimizing C++ might want to look at the code for any obvious changes, like swapping some data types for faster math.

I still feel you are going to come back to this discussion again when someone actually integrates an image library to do these routine functions.  But I get that seeing 1330% is an eye catching number.

knnniggett:
@Linwood-F Haven't seen you on our IRC channel lately, so I'll post here. I'm not ignoring this issue, but I simply have had other priorities.  I agree there is likely something not quite right with the way zonemidner currently handles this.  Sometimes working towards something, only to find that is not the way we want to do it, is still knowledge gained. Your contributions are certainly appreciated.

knnniggett:
@Linwood-F I agree with the underlying thought that there is room for improvement in this area. However, I'm going to close this pr out for housekeeping sake, since increases in cpu requirements trump the benefits this pr would bring. Feel free create a new pr or new issue if you've got a new method.

Linwood-F:
Do you want to take the range protection fix instead that was mentioned in the first comment in issue #997

knnniggett:
I guess I misremembered. Thought resizing was necessary for the range protection which was causing the cpu spike. Yeah, go ahead and create a new pr if you can with the range protection fix.

Linwood-F:
No, the scaling was done to use a standard library routine instead of a home grown (and somewhat flawed) one.  And it did improve the image slightly, but it was slower.

I'll do this, it will take me a bit to get back to programming as have some commitments coming up.  The code is actually in that issue but setting up and testing it will take me a bit of time, but will.  I need to bring a test system up to date anyway and reconcile all the last few bazillion commits.

Issue #1000, opened 2015-08-09
Add optional scale/width/height to query in image.php


Linwood-F:
The image.php is used to return a single historical image.  It is currently an alternative to direct access via the /events... path, the latter of which is a security issue and seems doomed to going away.  The image.php is an alternative, though in the future I am a strong proponent of incorporating historical access to single frames in zms instead.  However that is a large task, and image.php is already used and simple, and more relevant is being incorporate in new programs.

This issue is to incorporate scaling as an option in this program.  There is a tradeoff in doing so, in that scaling on the server puts load there, but removes potentially very large bandwidth requirements if clients pull data with an explicit scale (as opposed to pulling the whole image and then scaling in the client).

pliablepixels:
This is a good feature to have. zmNinja uses view=image to apply authentication over event frames without the (current) overheads of zms. Till the time zms is fixed to a) return a single image for an event b) validated that it does not have any significant overhead compared to this method, I personally don't see any reason why this new addition should be a problem for what it does.
Hopefully, this will be accepted!

Linwood-F:
Thanks, I've pushed a change to permit either width or height or scale in combinations.

pliablepixels:
Seems to work fine for me. @Linwood-F you can always make (optional) changes to the algorithm in later pushes if you want, but its useful to have as it stands today. Would you want to request folks to merge it?

Linwood-F:
Yeah, as you saw offline I experimented with the other techniques and did not see improvements (and one of them crashes php consistently), so i think this is ready, will post a note in the pull request.

knnniggett:
PR was merged.  Can we close this issue out?

Linwood-F:
Sure, sorry, still learning the steps.

knnniggett:
No problem.  Thank you for all the work you've done for the project.

PR #1001, opened 2015-08-09
Add scale as optional feature to image.php

Linwood-F:
This relates to the issue described in #1000.

It adds the ability to invoke the image.php view from with any of: 

1) No scale - does exactly as before
2) scale parameter - scales as indicated
3) Both width and height pixel size - scale as indicated

For consistency with past programs both the default, and what happens if there are errors in the parameters, is to pass the image out unchanged.

Editorial comment: While this program is preferable to accessing images insecurely through the /events path, I think the long term direction should be to roll single frame historical access, similar to single frame live access, into zms.  But in the meantime this allows clients (such as new mobile programs) to make the choice as to whether have the server scale (slower server access but much lower bandwidth consumption) or the client scale (less server load, more load on client and network).

This requires php5-gd, but that was recently added as a dependency anyway.

### DIFF
--- a/web/views/image.php
+++ b/web/views/image.php
@@ -26,10 +26,10 @@
 //          Omitted or 100 = no scaling done, image passed through directly
 //          Scaling will increase response time slightly
 //
-//     width and height are optional, but if one is passed both should be passed (or they will be ignored)
-//          Pixel dimension to returned, scaling up or down
+//     width and height are each optional, ideally supply both, but if only one is supplied the other is calculated
+//          These are in pixels
 //
-//     If both scale and width & height are specified, scale is ignored
+//     If both scale and either width or height are specified, scale is ignored
 //
 
 if ( !canView( 'Events' ) )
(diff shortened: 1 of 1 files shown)


Linwood-F:
Don't merge this yet, I got an offline request to make height optional.  I had originally made it required because of a known issue in older php-gd versions, but I can simply calculate the height internally as well, and so make height optional.

Linwood-F:
OK, as requested, it now permits any of the three - scale, width or height.  It gives preference to width or height if specified and ignores scale.  If none are specified, or if they are not valid, or if the resulting scale is 100, it just passes the image out unscaled, so existing uses of this should see no change.

Linwood-F:
Hold for a bit -- got a big report I'm checking on.

Linwood-F:
I think this is ready now.  @pliablepixels and I chatted a bit about alternative algorithms but frankly all of them in GD did poorly in his scale-down example, so the default seems fine and I don't see any reason to expose the selection, especially as I think we should migrate toward ZMS anyway, this is just a stepping stone.

PR #1005, opened 2015-08-09
enable/disable RTSP Describe Header

knnniggett:
This logic is currently incomplete.

The goal of this PR is to allow one to enable/disable Processing the RTSP DESCRIBE response header.  Currently, this option is hardwired on, but should be turned off in cases where a camera is NAT'ed through a firewall for example.

So far, I have only focused on the C++ logic as that is where the biggest learning curve is for me.

Still to do:
- Add php bits to display the checkbox under Monitor -> Source tab for remote RTSP cameras
- Add associated help text
- Add the new column in the sql monitors table


knnniggett:
fixes #685

knnniggett:
I need to rewrite this.

knnniggett:
See #1028

PR #1008, opened 2015-08-11
Included logic to not enforce authentication in API layer if ZM auth is off

pliablepixels:
If ZM auth is off, API won't work. Changed this to make sure API disables auth if ZM disables auth


pliablepixels:
@connortechnology before you make the snapshot please consider merging this API auth change

Issue #946, opened 2015-07-17
minor Sphinx error


onlyjob:
During re-build of documentation Sphinx reported the following error:

    docs/api.rst:31: ERROR: Unexpected indentation.

Probably because it expects empty line (or double spaces followed by end-of-line on the previous line), as it seems from looking at the other paragraphs of the same file.
Affects "master".

pliablepixels:
Thanks -  I wasn't aware this needed to be tested with sphinx. I'll have a setup ready tomorrow and will make the changes.

pliablepixels:
this can be closed - was addressed (I can't close this as I am not the author)

PR #1014, opened 2015-08-14
Fix API indentation warning for api.rst

pliablepixels:
This addresses issue for https://github.com/ZoneMinder/ZoneMinder/issues/946
@onlyjob  can you please double check I did it right? 

I did a sphinx-build html and it gave no warnings
Also, while building I noticed we should have a _static directory as well, so I added it

I don't see any warnings now.


pliablepixels:
Oops I forgot to change branch. I'll redo it tomorrow.

PR #1015, opened 2015-08-14
946 minor api sphinx issue

pliablepixels:
Addresses https://github.com/ZoneMinder/ZoneMinder/issues/946
@onlyjob can you please check I did it correct? I validated with sphinx-build html

Also during build I noticed it also expects a _static directory (for future static resources) so I added that too


pliablepixels:
I am closing this. The same issue is also fixed as part of https://github.com/ZoneMinder/ZoneMinder/pull/1017

PR #1016, opened 2015-08-14
Fix issue with score values less than 0

### LINKED ISSUE #319: ERR: Can't insert frame: Out of range value for column 'Score' at row 1
I'm using ZoneMinder to capture an mjpeg stream from an ITS camera. My company has been using ZoneMinder for years with OpenSUSE 11.4 with great success, but I'm trying to move to OpenSUSE 13.1.

I just setup a brand new OpenSUSE 13.1 machine (physical machine, not VM) and installed ZM from the repo (version 1.26.4). It's running in conjunction with MariaDB 5.5.33.

When I tried to add my camera, I am receiving an error from zma "Can't insert frame: Out of range value for column 'Score' at row 1". This is from /src/zm_event.cpp line 568.

The schema in MariaDB shows the following for column Score in table Frames:

Field | Type | Null | Key | Default | Extra
------- | ------- | ----- | ------- | --------- | --------
Score | smallint(5) unsigned | NO | | 0 |

Could you offer any assistance here? Have I configured something improperly? I would love to get this running with OpenSUSE 13.1 since it was just picked up by Evergreen and we are moving our servers in that direction.

Thanks!

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1388156-err-can-t-insert-frame-out-of-range-value-for-column-score-at-row-1?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github).
</bountysource-plugin>

knnniggett:
fixes #319

### DIFF
--- a/src/zm_event.cpp
+++ b/src/zm_event.cpp
@@ -571,7 +571,7 @@ void Event::AddFrame( Image *image, struct timeval timestamp, int score, Image *
     if ( score < 0 )
         score = 0;
 
-    bool db_frame = (score>=0) || ((frames%config.bulk_frame_interval)==0) || !frames;
+    bool db_frame = (strcmp(frame_type,"Bulk") != 0) || ((frames%config.bulk_frame_interval)==0) || !frames;
     if ( db_frame )
     {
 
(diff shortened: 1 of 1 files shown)


connortechnology:
My only thought here is an efficicency one... doing string comparisons where maybe we could use an enum...

But I could do another pr later for that...

knnniggett:
Agreed.
My C++ knowledge is downright terrible.  I simply didn't know any better.
I look forward to learning how to make that more efficient.

PR #1018, opened 2015-08-15
New User Permission "Groups"

### LINKED ISSUE #541: Regular user cannot create or access groups
I'm under the impression that this feature (groups) is intended to be accessible to all web users, however testing in 1.28.0 it seems that the groups feature is only visible in the UI to admins with system edit rights.

Can this be changed so that any user can set his/her own groups for monitoring?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5499503-regular-user-cannot-create-or-access-groups?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github).
</bountysource-plugin>

knnniggett:
Fixes #541 
This PR adds a new permission called Groups, which gives the sys admin the ability to manage monitor group view/edit permissions without also giving the user permission to the rest of the System.

This turned out to be more involved than I had initially anticipated.  
@barjac owes me big.

Please test this and be on the look out for anything unexpected in the logs.


connortechnology:
The only issue I see is the use of short php tags in console.php

knnniggett:
Fixed. Good catch. I had copied/pasted that line from an old version of console.php.

PR #1026, opened 2015-08-20
Add new zone type - privacy zones

schrorg:
Today I needed the ability to hide some regions of a camera picture, so I added a new zone type 'Blacken' to mark such zones and black out the corresponding pixels early in the image capture process (just before adding the timestamp if TIMESTAMP_ON_CAPTURE is set).

'Blacken' zones can be created in the zones view, just like every other zone. The only thing that is missing until now is the functionality to refresh these zones in the monitor. For now, you have to restart the monitor by hand (i.e. open the monitor settings and 'save'). The reason is, that blacken zones are only calculated during monitor initialization by filling a bitmask.

I would be glad to have this feature pulled in whem I'm finished. Also, please let me know if there is a better name for 'Blacken'. I'm not a native english speaker but I didn't find a better word for it.


connortechnology:
Blackout might be better... but Blacken is fine by me.

SteveGilvarry:
English equivalent would be privacy zone or privacy mask, well that is what
my camera calls it.

knnniggett:
I'd like to match the semantics used by the rest of the industry when we can. 
Consequently, I vote for "privacy". 

We do need to remember to add this new zone type to our documentation.

Linwood-F:
I was just writing the same when Steve posted, I think I've seen "Privacy Mask" used more frequently, and many (some?  most?) newer cameras have them built in.  It's a nice reuse of zone definition to do this, though it raises the question whether it will be practical to do when (if?) zoneminder switches to h.264 video capture.

SteveGilvarry:
You could still do it with encoded h264, which takes jpegs to write them
out again as h264, which is the option for non h264 cameras that want to
store as h264. But h264 passthrough grabs the packet and writes that so
would not be possible. But still like your work Robin.

On Fri, Aug 21, 2015 at 12:41 AM, Linwood Ferguson <notifications@github.com

Linwood-F:
Yeah, that was just a comment, not meaning to imply it was not useful.  In fact a lot of options become moot with the passthru, like rotate.

schrorg:
Sure, if I had cameras with the ability to blacken areas, I would have done it there. But having it in ZoneMinder makes it much more usable and simple to define those zones. One of my cameras can mask zones, but it is difficult to set it up and it doesn't have the nice polygons :-)

I will rename 'Blacken' to 'Privacy'. That seems more suitable.

knnniggett:
@schrorg let us know when you are finished and we will merge this.

schrorg:
At the moment I am trying to change the privacy_bitmask during a running capture. This does not work until now, so I will need a few more hours.

Can anyone give me a hint on how to do that? The resulting behaviour should be like if I change a privacy zone (click 'save' on the web site), the privacy_bitmask should be regenerated to reflect the new or changed area. Just recalling Monitor::AddPrivacyBitmask() does not have any effect and I don't see why.

knnniggett:
Not sure if this is helpful or not....
When you hit save, is Monitor::Load being called?
Seems like that needs to be called with load_zones set to 1.

schrorg:
I got what I want by simply restarting the capture process if a Privacy zone got changed. This seems to be the savest solution by now. So if everything looks good, you can merge this PR now.

schrorg:
Well... please wait a few more minutes. I need to check the behaviour of this when adding or deleting Privacy zones! There seems to be a problem...

schrorg:
Okay. Everything is finished now.

PR #1032, opened 2015-08-22
update doc

tikismoke:
make it more readable on readthedoc


knnniggett:
Thank you. We can always use the help keeping our documentation up to date and well formatted.

Issue #1046, opened 2015-08-29
Optional google recaptcha


pliablepixels:
There was interest in adding a google recaptcha option to ZM which could be disabled/enabled via options. Default state is disabled. 

I'll be submitting a PR on this soon for review.

Note that the user will have to follow recaptcha instructions to configure a site and secret key for their ZM domain and enter those values in the options screen.

marcolino7:
Hi,
I updated my ZM installation from 1.28.102 to 1.28.106, and on login screen I got recaptcha widget with an error. I connot login on my installation.

![image](https://cloud.githubusercontent.com/assets/11460036/10041336/df8681f0-61e3-11e5-8a06-5c0621d18a16.png)

marcolino7:
Hi,
I run database upgrade script, and now work. Sorry for this.
Marco

PR #1048, opened 2015-08-29
Google recaptcha

pliablepixels:
Addresses issue https://github.com/ZoneMinder/ZoneMinder/issues/1046


knnniggett:
I didn't expect to see see Google's php being added to the source tree. 
Does this pr actually use the code under recaptcha/src or have you included the entire recaptcha folder simply to comply with the redistribution requirements?

Depending on the answer to that, perhaps we should consider adding this via git submodule:
https://github.com/google/recaptcha

pliablepixels:
It uses the code under recaptcha src directory to validate the server response.
I've removed the examples and test directories and some other root files that were not needed - hence did not think it necessary to git submodule it

pliablepixels:
Okay, the new version does the recaptcha check manually instead of using google's client library

knnniggett:
Yeah, good idea.  Doing this in a way that doesn't require any additional libraries is preferred.  That way we don't have to worry about stuff falling out of date with google.

In the future when we consider adding another library to the project, we should keep in mind how we manage our version control.  

It might sound trivial when the scope of what we are looking at is just one (small) library.  What's the big deal?  Just copy the php into the source tree, right? 

Only when you look at the big picture and see we've got cakephp, jscalendar, mootools, crud, cambozola, captcha, and likely others, does it become apparent that we should look for better ways to manage these libraries.  If the library in question is available via git submodule, then I'd argue, generally, we should go that route, regardless of how big or small that particular library is. This helps us down the road, long after we've all forgotten about how this new library works, and we find ourselves in a position where we need to update the version of the library.

pliablepixels:
One of the issues I am yet to handle is this:
If the user enables recaptcha and puts in invalid keys in both the public and secret textboxes, then the interface locks the person out as recaptcha fails every time. I'm handling the case when either is null --> we ignore recaptcha.

In this special case where the user puts in wrong keys, the only way out is to do a mysql query like so
```sudo mysql --u<dbuser> --p<password> zm -e "Update Config SET Value=0 WHERE Name='ZM_OPT_USE_GOOG_RECAPTCHA';"```

This case can likely be fixed, but I haven't gotten to it yet.

SteveGilvarry:
FYI You are going to hear from people on use of external links, I did when we used external library for videojs.

pliablepixels:
This is not using an external library, it is using an external service. I am hoping the saving grace here is that this particular reCaptcha is essentially a Google service - so when a person enables reCaptcha, they know they are enabling Google's reCaptcha and if its not enabled, the call to Google does not happen.

I am assuming video streaming being a core service, there were concern using a 3rd party JS player that was dialing home where it is not essential for that service (playing a video).

But in general,security and paranoia form a loving couple, so you may be right. It's disabled by default.

PR #1049, opened 2015-08-29
remove core.php, modify core.php.default

knnniggett:
core.php.default is currently a make input file.  
It gets processed by cmake and copied to php.core.
Consequently, all future cake config changes should be made to core.php.default and never core.php as that gets overwritten during the build process.  Core.php should never exist in the source tree.

Perhaps we should consider giving core.php.default a ".in" suffix to make it more intuitive that this is the case.


pliablepixels:
I'd support renaming it to ".in" as this is what ZM uses for template files. It seems ".default" is what PHP/Cake uses - given that these are static files not being pulled from other repos, renaming to .in should not adversely affect anything.

knnniggett:
@connortechnology can we merge this, please?

PR #1050, opened 2015-08-31
Add videostore to remote_rtsp, still need to check stream is h264

SteveGilvarry:
Still testing this, just putting this here for anyone else that might want to have a play with it and help test.


connortechnology:
Given that this would merge to the video branch, I think we can merge it.  To the h264 video branch that is..

SteveGilvarry:
Closing this out for some reason it was replicating commits, just starting fresh with the two required commits and still not recording so looking at that.

PR #1052, opened 2015-09-07
redhat rpm packaging modifications

knnniggett:
Update the README file, require HTTPS, other misc packaging mods.


knnniggett:
self-merging, since this only affects redhat rpm packaging

PR #1054, opened 2015-09-14
link to cambozola pacakge, rather than download during build

knnniggett:
applies to centos rpm specfiles


knnniggett:
self-merging, since this is strictly to do with rpm packaging.

Issue #1055, opened 2015-09-16
Data corruption with mmapped shm (shared_data size conflict)


connortechnology:
The vlc code exposes a problem where our mmap fd is actually 2, and hence library writes to stderr overwrite our shared memory.

See http://www.zoneminder.com/forums/viewtopic.php?f=32&t=23445

PR #1057, opened 2015-09-16
quick fix for #1055: make sure our mmap fd is > 2

connortechnology:


knnniggett:
nice fix. hadn't thought of that.

PR #1062, opened 2015-09-22
fixed security instructions for curl

pliablepixels:


knnniggett:
These two lines do not agree:

`curl -d "username=XXXX&password=YYYY&action=login&view=console" -c cookies.txt  http://yourzmip/zm/index.php`

and 

`Then for each of the examples below, add a ``-b cookies.txt``  to each of the commands below.`

PR #1063, opened 2015-09-22
added more details on how to use cookies with curl

pliablepixels:


knnniggett:
See commit f13dddb7190ee3e8cb192fd8398862a6630f5285

Issue #663, opened 2015-01-06
Modify Frames table to not have a composity key [$150]


kylejohnson:
Hey everyone.  Looking for feedback on modifying the ZM DB schema to NOT use a composite (multi-column) primary key on the Frames table.  The reason being a limitation with CakePHP - it does not support composite keys.  And as I am /heavily/ invested in CakePHP for the API, it makes more sense to modify the schema than it does to ditch the tool.

```
mysql> describe Frames;
+-----------+-------------------------------+------+-----+-------------------+-----------------------------+
| Field     | Type                          | Null | Key | Default           | Extra                       |
+-----------+-------------------------------+------+-----+-------------------+-----------------------------+
| EventId   | int(10) unsigned              | NO   | PRI | 0                 |                             |
| FrameId   | int(10) unsigned              | NO   | PRI | 0                 |                             |
| Type      | enum('Normal','Bulk','Alarm') | NO   | MUL | Normal            |                             |
| TimeStamp | timestamp                     | NO   | MUL | CURRENT_TIMESTAMP | on update CURRENT_TIMESTAMP |
| Delta     | decimal(8,2)                  | NO   |     | 0.00              |                             |
| Score     | smallint(5) unsigned          | NO   |     | 0                 |                             |
+-----------+-------------------------------+------+-----+-------------------+-----------------------------+
6 rows in set (0.00 sec)
```

What I propose is removing all of the current PRI and MUL key's, adding an 'Id' field and making that the PRI key, and then modifying the indexes appropriately (which I could use help with).

Thoughts?
I'd like to do this ASAP as it is blocking progress on the new frontend.

knnniggett:
I don't see a problem with this.  Would like to test with the legacy UI.

connortechnology:
also need to make sure we have indexes, but I'm fine with the idea

kylejohnson:
Sounds good - I will do some more testing with the legacy UI.

Could one of you help me with how to get the change in?  It has been a while since I messed with the DB upgrade stuff.

connortechnology:
Sure.  One thing, if we are going to do this, we go all the way and make it perfect. :)  So get rid of frame_id.... I'll start, make a branch for it all...  I can handle the legacy UI changes... you worry about the new UI changes...

kylejohnson:
oh boy.  My concerns were in c++ and perl code, which I'm not as involved in.

connortechnology:
[...]

PR #1074, opened 2015-09-29
fixes for event delete. 

pliablepixels:
Work in progress. Deleting an event is calling wrong commands for frame delete. I'm working on this and will keep pushing PRs till I solve.


pliablepixels:
verified, works. ready for merge.

kylejohnson:
Fairly certain this behaviour is caused by #663.  The real fix is to get #663 merged.

connortechnology:
Ok, so this point is that the frames need to be deleted first, right? 

Seems to me the Event->delete function should just delete the frames first... 

maybe I'm missing the bigger picture.

kylejohnson:
This has been fixed in #1080

Issue #1072, opened 2015-09-28
Events API delete fix


pliablepixels:
The Events API delete command does not delete the Frame entries. I will submit a PR to fix.

pliablepixels:
There are still problems with how the Events and frames are being debugged. Will need to submit a few more PRs while I figure this out

kylejohnson:
@pliablepixels #663 is causing this.

PR #1075, opened 2015-09-29
663 frames primary key

### LINKED ISSUE #663: Modify Frames table to not have a composity key [$150]

kylejohnson:
This should fix #663.  Maybe / related to #1072, #1073, #1074


knnniggett:
looks good to me

PR #1078, opened 2015-09-29
Use the 3.0 branch of crud, compatible with cakephp 2.x

kylejohnson:
The master branch of crud doesn't support cakephp 2.x.  The 3.0 branch does.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "web/api/app/Plugin/Crud"]
 	path = web/api/app/Plugin/Crud
 	url = https://github.com/FriendsOfCake/crud.git
+	branch = 3.0


kylejohnson:
@pliablepixels Please review.

pliablepixels:
@kylejohnson - looks good. Looks like master was updated after @knnniggett configured the gitmodule and ~3.0 is what we need today

knnniggett:
I forget what version of crud I last set the git submodule to, but I recall there was some compatibility issue with upgrading further.

pliablepixels:
The compatibility issue was pulling crud~4.0 which only works with cake 3.0 was it not?

knnniggett:
Actually, I'm not sure what this commit does exactly. 

Previously, I set the git submodule to commit c3976f1 of crud, which translates to version 3.0.10:
https://github.com/ZoneMinder/ZoneMinder/tree/master/web/api/app/Plugin

pliablepixels:
I think this PR is specifying the 3.0 branch right inside gitmodule which is probably easier to understand/maintain  going forward  but given you specified a particular commit inside the plugin directory I wonder why that did not work for @kylejohnson

PR #1079, opened 2015-09-30
Added control script for SunEyes SP-P1802SWPTZ

bofhdk:
(http://www.amazon.com/SunEyes-SP-P1802SWPTZ-Wireless-Camera-1080P/dp/B00MOTBXK6)


knnniggett:
Thank you for taking the time to contribute. However, you also need to add the necessary sql bits. Otherwise, it won't show up in the console. 

Like so:
https://github.com/ZoneMinder/ZoneMinder/blob/master/db/zm_create.sql.in#L587

bofhdk:
ahh! - sorry for the omission - it's on it's way through now :)

Thanks for the tip-off :)


/bobby

bofhdk:
Oh - MonitorPresets too .... coming through.

(one day I'll get the hang of this stuff (I hope) :) )

knnniggett:
Nice work.

bofhdk:
Thanks :)

PR #1084, opened 2015-10-01
Tz

connortechnology:
Setting the TZ ENV variable prevents every gettimeofday call from reading /etc/timezone which actually has a noticeable performance hit.

### DIFF
--- a/distros/debian/init.d
+++ b/distros/debian/init.d
@@ -23,6 +23,7 @@ command="$ZM_PATH_BIN/zmpkg.pl"
 
 start() {
 	echo -n "Starting $prog: "
+	export TZ=:/etc/localtime
 	mkdir -p $RUNDIR && chown www-data:www-data $RUNDIR
 	mkdir -p $TMPDIR && chown www-data:www-data $TMPDIR
 	$command start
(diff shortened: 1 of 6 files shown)


liucougar:
according to: http://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html

the default value for TZ would be ‘TZ=:/etc/localtime’.

may I know how do you test whether this has a performance hit or not?

connortechnology:
The problem is.. on my ubuntu and debian based servers, TZ is not set in the ENV.  

How I tested was watching cpu consumption on a raspberry pi.  I noticed this while stracing zmc, and then setting TZ and watching cpu load.  On a low end device like a raspberry pi it is noticeable.

liucougar:
it seems raspherry pi uses eglibc (not normal glibc), so it may have a different TZ default?

connortechnology:
This behaviour was seen on ubuntu as well.  

If you want to see the behaviour, run strace zmc -m 1 (or whatever monitor id).  seems to be worse on a http stream than an rtsp.  

Now set TZ and notice that it no longer is hitting /etc/localtime

TZ=:/etc/localtime strace zmc -m 1

liucougar:
just tried strace on ubuntu 15.04, against a RTSP camera, i am not seeing any call to gettimeofday in my strace log (I don't have http cameras to test against)

knnniggett:
I'm going to merge this as the discussion does not hint at any undesirable side effects.
Let's get this into master so others can test.

PR #1087, opened 2015-10-02
fix timeline view for SaveJPEGs monitors (without enabled VideoWriter)

liucougar:
in the videostorage branch, if a monitor is configured to have VideoWriter=0, and SaveJPEGs is non-zero, timeline view does not work

this fixes this problem.

in addition, it has some special handling for the case where SaveJPEGs=2 (only save analyse image) so that only analyse image will be loaded in the timeline imageSrc element

### DIFF
--- a/web/skins/classic/views/js/timeline.js
+++ b/web/skins/classic/views/js/timeline.js
@@ -22,7 +22,7 @@ function createEventHtml( event, frame )
     if ( event.Archived > 0 )
         eventHtml.addClass( 'archived' );
 
-    new Element( 'p' ).inject( eventHtml ).set( 'text', monitorNames[event.MonitorId] );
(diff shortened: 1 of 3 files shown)


knnniggett:
Thank you for taking the time to submit this pull request.
Since this is for the h264 video branch, we need to wait for @SteveGilvarry to respond as he is the author of that branch.

connortechnology:
Most of this I like.  At first I really liked the monitors = json_encode( $monitors ); but then  I realized that this can expose usernames and passwords set in the monitor but sticking it into the javascript.

So it will have to be sanitized a bit.

liucougar:
just pushed a change to just expose data needed in javascript. please let me know if there is anything else you think should be done, thanks

PR #1089, opened 2015-10-02
change types and move things around to remove compile warnings

connortechnology:


knnniggett:
That's an interesting type "off_t".  Seems that type is system dependant. 
For the sake of learning, what lead you to use that?

connortechnology:
off_t is POSIX.  It is what st_size in struct stat is defined as.

knnniggett:
fair enough.

connortechnology:
It is actually a signed type

PR #1090, opened 2015-10-02
Don't install zm.conf if it already exists

connortechnology:


knnniggett:
Not installing zm.conf at all would potentially prevent someone from receiving needed changes should we add new variables to zm.conf.  Instead of not installing it, can we install it as zm.conf.new?

connortechnology:
Seems reasonable.

knnniggett:
Thanks.

connortechnology:
So actually, renaming it to .new is annoyingly difficult.

Issue #1104, opened 2015-10-12
PHP Warnings


baffo32:
After a fresh install of zoneminder on my gentoo system, the console shows mostly PHP warnings.

![zm-console](https://cloud.githubusercontent.com/assets/240981/10419890/2bfb7e14-7053-11e5-9d6e-28270eeb8155.png)

knnniggett:
You need to do what it says and set date.timezone in your php.ini
This is a standard requirement for most distros.

baffo32:
Thanks, now all I have left is
> Warning: disk_total_space(): No such file or directory in /usr/share/zoneminder/www/includes/functions.php on line 1642
> Warning: disk_free_space(): No such file or directory in /usr/share/zoneminder/www/includes/functions.php on line 1643
> Warning: Division by zero in /usr/share/zoneminder/www/includes/functions.php on line 1643

This appears to be because ZM_DIR_EVENTS is unset.

baffo32:
Ah-ha, I'm missing the events folder

knnniggett:
Installation instructions for Gentoo are here:
https://wiki.gentoo.org/wiki/ZoneMinder

Please review it as the instructions for editing php.ini, along with many other required configuration steps, are there.

You appear to be attempting a from source installation, in which case you should also follow these instructions here: 
https://github.com/ZoneMinder/ZoneMinder/blob/master/INSTALL

This documents the use of zmlinkcontent.sh among other important steps.

ZoneMinder is not easy to build and install from source even for advanced users, which is why we recommend you install from package instead.  

If you need additional support then please move this conversation to the user forum or our IRC channel as our github forum is reserved for bug reports and feature requests only.
https://github.com/ZoneMinder/ZoneMinder/wiki/Github-Posting-Rules

PR #1105, opened 2015-10-12
Detect unset timezone

baffo32:
#1104

### DIFF
--- a/web/index.php
+++ b/web/index.php
@@ -46,6 +46,12 @@ if ( false )
     ob_end_clean();
 }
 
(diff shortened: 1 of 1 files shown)


knnniggett:
No, this should be Fatal, just as it currently is.  This forum post is a good example why:
http://www.zoneminder.com/forums/viewtopic.php?f=32&t=23511

baffo32:
It's not fatal, it issues a mound of warnings (top of #1104).  This PR changes it to one warning with instructions.

PR #1121, opened 2015-10-24
Use relative URL's instead of absolute

knnniggett:
I do not know of any negative side effects with using relative url's instead of absolute. 
Let's get this into master and test.

### DIFF
--- a/web/index.php
+++ b/web/index.php
@@ -58,7 +58,12 @@ else
     $protocol = 'http';
 }
 define( "ZM_BASE_PROTOCOL", $protocol );
-define( "ZM_BASE_URL", $protocol.'://'.$_SERVER['HTTP_HOST'] );
+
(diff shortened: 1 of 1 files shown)


SteveGilvarry:
No issues with my testing of this either, happy to merge? If so I will do it.
Only question I had is with only 3 places that use ZM_BASE_URL is this part one to ensure it has not effect before removing it altogether?

knnniggett:
Yeah, pretty much. I'm extremely hesitant to change things that have been working a particular way since the beginning of zoneminder. This is one of those cases. We have a really good reason to change it, but from past changes there have been those who use zoneminder in ways we don't anticipate whom we have affected. I can't think of what that would be in this case, but I still chose to play it safe.

If for some reason this merge causes an unforeseen problem, I made it real obvious for anyone to see how to change it back.  We can remove the comments in a future revision.

Thank you for merging.

Issue #1122, opened 2015-10-24
Unable to update db to 1.28.108


hans109h:
>zmupdate.pl
...
Database successfully upgraded to version 1.28.107.
Upgrading DB to 1.28.108 from 1.28.107
Warning: Using a password on the command line interface can be insecure.
ERROR 1064 (42000) at line 8: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 6
Output: 
Command 'mysql -hlocalhost -uXXX -pXXX zm < /usr/local/share/zoneminder/db/zm_update-1.28.108.sql' exited with status: 1

kylejohnson:
Thanks for reporting this.  Will look into it.

PR #1123, opened 2015-10-24
Stray semicolon causes SQL error

baffo32:
also normalized space around '='


kylejohnson:
Thanks for the catch.  This probably caused #1122

PR #1124, opened 2015-10-25
Install nph-zms with FILES so it is listed in install_manifest.txt

baffo32:
This allows it to be uninstalled with `make uninstall`


knnniggett:
You took out the DESTDIR variable, which will break package installs.

Note that adding this to the manifest won't affect it from being uninstalled via package anyway. This would only benefit uninstalling from source, which happens to be permanently broken. Perl, for example, no longer supports "make uninstalls" so all Perl modules are left on the system.

Issue #1128, opened 2015-10-28
support for ssmtp mail.


dreaddymck:
Often find myself updating /usr/bin/zmfilter.pl after each update.

            ### Send the Message
            # MIME::Lite->send( "smtp", $Config{ZM_EMAIL_HOST}, Timeout=>60 );
            # $mail->send();
            $mail->send('sendmail','/usr/sbin/ssmtp',$Config{ZM_EMAIL_ADDRESS});

Would it be possible to add support for ssmtp?

I may be able to do this. Not really clear on howto contribute. Is there an irc chatroom available that can help?

knnniggett:
Hey, this was merged. Closing....

Issue #1126, opened 2015-10-27
One solution for: Unable to generate video, check /usr/share/zoneminder/www/events/.../ffmpeg.log


dreaddy:
Running zm on ubuntu 14.xx

The following error was a common in my log

    zmvideo[10668]: INF [Creating video file Event-1028-r1-s1.avi for event 1028]
    zmvideo[10668]: ERR [Unable to generate video, check /usr/share/zoneminder/www/events/1/15/10/04/22/38/37/ffmpeg.log for details]

One quick solution:

    nano /usr/bin/zmvideo.pl

locate the following perl command in the code

    my $output = qx($command);

add a '\' in front of the $command

    my $output = qx(\$command);

save. 

Solution came via the following downvoted solution: http://stackoverflow.com/a/25561156/3347336

PR #1132, opened 2015-10-31
minor updates to address issues #1128 and #1126

dreaddy:


knnniggett:
When creating new ZM_ variables, you do that in this file here:
https://github.com/ZoneMinder/ZoneMinder/blob/master/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in

Creating a sql update script is not needed.
Once you do that, I don't think you need to check to if these variables are defined in your Perl. ZoneMinder simply won't run at all after an update until after the user runs "zmupdate -f" which will create the variables in the Config table.

For organization sake, please break this into two separate pull requests, each tied to the issue they refer to.

knnniggett:
If you get stuck, feel free to ask questions.

dreaddy:
I foolishly decided to build and install the latest version from github. 
'cause y'know: YOLO
My mission in life now is get zoneminder working again. :8ball:

PR #1133, opened 2015-11-01
fixed SaveJPEGs are not working

liucougar:
which is caused in errors introduced when merging with master


SteveGilvarry:
My bad, those 3 lines are a pita to merge.

Issue #1135, opened 2015-11-02
API modification: previous and next for event APIs


pliablepixels:
It is useful for the event API to return a "previous" and "next" event ID when invoking a specific event API (example, /zm/api/event/96.json should also return prev: 95 and next: 101 if these are the prev and next ids). Useful for gapless playback via APIs and forward back playback

PR #1136, opened 2015-11-02
added prev/next event ids

pliablepixels:
addresses #1135

### DIFF
--- a/web/api/app/Controller/EventsController.php
+++ b/web/api/app/Controller/EventsController.php
@@ -83,10 +83,14 @@ class EventsController extends AppController {
 		}
 		$options = array('conditions' => array('Event.' . $this->Event->primaryKey => $id));
 		$event = $this->Event->find('first', $options);
+		$this->Event->id = $id;
+                $event_neighbors = $this->Event->find('neighbors');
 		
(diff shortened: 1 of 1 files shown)


web/api/app/Controller/EventsController.php
```diff
@@ -83,10 +83,14 @@ public function view($id = null) {
...
 		$options = array('conditions' => array('Event.' . $this->Event->primaryKey => $id));
 		$event = $this->Event->find('first', $options);
+		$this->Event->id = $id;
+                $event_neighbors = $this->Event->find('neighbors');
 		
 		$path = $configs['ZM_DIR_EVENTS'].'/'.$this->Image->getEventPath($event).'/';
 
 		$event['Event']['BasePath'] = $path;
+		$event['Event']['Next'] = $event_neighbors['next']['Event']['Id'];
+                $event['Event']['Prev'] = $event_neighbors['prev']['Event']['Id'];
```
liucougar, line 94: would this throw exception if next/prev are not available (for the last/first event?)

pliablepixels, line 94: No, I tested it. returns a null (as in 'null' text)

connortechnology:
I'm ok with this.

PR #1137, opened 2015-11-03
Replace deprecated FFmpeg API

onlyjob:
Fixes FTBFS with ffmpeg-2.9; compatible with ffmpeg-2.8.

Author: Andreas Cadhalpun <Andreas.Cadhalpun@googlemail.com>
Bug-Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=803850

### DIFF
--- a/src/zm_ffmpeg_camera.cpp
+++ b/src/zm_ffmpeg_camera.cpp
@@ -62,13 +62,13 @@ FfmpegCamera::FfmpegCamera( int p_id, const std::string &p_path, const std::stri
 	/* Has to be located inside the constructor so other components such as zma will receive correct colours and subpixel order */
 	if(colours == ZM_COLOUR_RGB32) {
 		subpixelorder = ZM_SUBPIX_ORDER_RGBA;
-		imagePixFormat = PIX_FMT_RGBA;
+		imagePixFormat = AV_PIX_FMT_RGBA;
 	} else if(colours == ZM_COLOUR_RGB24) {
 		subpixelorder = ZM_SUBPIX_ORDER_RGB;
-		imagePixFormat = PIX_FMT_RGB24;
+		imagePixFormat = AV_PIX_FMT_RGB24;
 	} else if(colours == ZM_COLOUR_GRAY8) {
 		subpixelorder = ZM_SUBPIX_ORDER_NONE;
-		imagePixFormat = PIX_FMT_GRAY8;
+		imagePixFormat = AV_PIX_FMT_GRAY8;
 	} else {
 		Panic("Unexpected colours: %d",colours);
 	}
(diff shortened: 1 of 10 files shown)


SteveGilvarry:
This will bring up the discussion about old version support.
To facilitate the discussion
**av_frame_free**
libav10 https://github.com/libav/libav/blob/master/doc/APIchanges#L371
ffmpeg2.2 https://github.com/FFmpeg/FFmpeg/blob/master/doc/APIchanges#L640
**AVPixelFormat**
libav9 https://github.com/libav/libav/blob/master/doc/APIchanges#L574
ffmpeg1.0 https://github.com/FFmpeg/FFmpeg/blob/master/doc/APIchanges#L1094

I am all for moving to the latest version we can, but keeping in mind we would like users to be able to use new versions of Zoneminder on LTS distros. I think we do a version check on av_frame_free, to support older versions, but AVPixelFormat gets replaced.
Drawing the line at libav9/ffmpeg1.0 which is ubuntu 12.04. 
Wheezy is EOL Feb 2016 and it has libav8. 
Really need a list by distro.

src/zm_ffmpeg_camera.cpp
```diff
@@ -188,7 +188,7 @@ int FfmpegCamera::Capture( Image &image )
 #if HAVE_LIBSWSCALE
 		if(mConvertContext == NULL) {
 			if(config.cpu_extensions && sseversion >= 20) {
-				mConvertContext = sws_getContext( mCodecContext->width, mCodecContext->height, mCodecContext->pix_fmt, width, height, imagePixFormat, SWS_BICUBIC | SWS_CPU_CAPS_SSE2, NULL, NULL, NULL );
+				mConvertContext = sws_getContext( mCodecContext->width, mCodecContext->height, mCodecContext->pix_fmt, width, height, imagePixFormat, SWS_BICUBIC, NULL, NULL, NULL );
```
SteveGilvarry, line 191: Need to check out SWS_CPU_CAPS_SSE2, but there is 3 of these changes that seem to make the if and else statement identical.

onlyjob:
I'd much appreciate if someone could take this off my hands. I'm no expert and I merely forwarded patch that Andreas wrote. Also I have no time for testing or going deep into this matter...

onlyjob:
Also why do you need to care for old Ubuntu? Perhaps I've forgotten ZM support policy but Ubuntu leaches everything from Debian so IMHO there is no need to do anything special about Ubuntu...

SteveGilvarry:
@onlyjob I knew I would get a bite if I mentioned Ubuntu, only an example of old release that is still supported and could be running in production. I updated that to 14.04 in comment as the most recent LTS. No doubt CentOS will come up as well with 6 supported until 2020. And yes we know Debian is the best and only distro ;-)
ZM Support Policy isn't that clear/doesn't exist so this should prompt that discussion and we can establish a baseline for the next release.
I will wait for @connortechnology and others to chip in, but I am happy that we manage it from here. Thanks for forwarding upstream and we will certainly implement it in some form.

onlyjob:
> And yes we know Debian is the best and only distro ;-)

You know my obsession so well, Steve. ;-)

knnniggett:
I am certainly in favor of merging this in some form, just not in the form it is in now, not at this time.
Adding prepcrocessor conditionals to maintain the old code path would be much preferred at this point.

I rely on the RPMFusion repo for ffmpeg rpm's for Redhat based distros, and the version currently in the repo is 0.10 for CentOS 6. Yeah, 0.10. That's not a typo, and I agree. It is extremely old and should be updated.

I can try and backport a newer ffmepg package for CentOS 6, but if it turns into dependency hell, we will need to fall back to maintaining the old code path... at least for this next release. I'll need more time if it turns into that.

SteveGilvarry:
Off topic but found this that might help you. https://gist.github.com/gboudreau/f24aed76b4cc91bfb2c1

I will start having a look at the conditionals, at least I already did the work to identify versions. Well with this one I think I have them all.
https://github.com/FFmpeg/FFmpeg/blob/master/doc/APIchanges#L978

SteveGilvarry:
#1142 Replaces this PR will close when #1142 goes in.

PR #1138, opened 2015-11-03
#1128 - Added option to enable send email by ssmtp mail. Option for c…

dreaddy:
…ustom ssmtp path. Conditional checks for new options in filter

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
@@ -4055,7 +4055,9 @@ body = "ZM alarm detected - %EL% secs, %EF%/%EFA% frames, t%EST%/m%ESM%/a%ESA% s
 							{ name => "ZM_NEW_MAIL_MODULES", value => "yes" }
                        ],
         help        => qqq("
-            Please visit the following wiki page for more information on setting up ssmtp: http://www.zoneminder.com/wiki/index.php/How_to_get_ssmtp_working_with_Zoneminder.
+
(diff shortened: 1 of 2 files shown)


connortechnology:
This looks pretty good to me. 

I don't really like the quick goto wiki page for info on how to set it up.  I would prefer some brief instructions and then to go to for more information link.

The other thing is around 1172 where you do a which ssmtp and then if it is set in the config drop that and set it to what is in the config.  How about, if it is set in the config, use the config , else use which to find it.  Just for performance reasons.  Might be nice to log a warning telling them to set it in the config.

dreaddy:
"...I don't really like the quick goto wiki page for info on how to set it up. I would prefer some brief instructions and then to go to for more information link.."

help doesn't support formatting or markup. Adding commands and whatnot to the help text did not come across very clean. Adding examples for other linux flavors made the help section a bit difficult to read.

What do you suggest?

knnniggett:
Now that we have released 1.29.0, I am following up on our open pull requests. Generally I like the idea of this pr as long as we maintain backwards compatibility.

I tend to agree with @connortechnology that we want to avoid putting all the documentation on an external site, but on the other hand sometimes the best way to describe something is through a screenshot or some other method that just can't currently be done in our current help functionality.

@dreaddy What I suggest is that you try to include a basic description of how to set this up and then refer the end user to the readthedocs site. 

The files to modify are found under /docs. I think this is the right one:
https://github.com/ZoneMinder/ZoneMinder/blob/master/docs/userguide/options/options_email.rst

connortechnology:
I kinda think we should just merge this.

knnniggett:
Merged. The whitespace and formatting in this pr was way out of alignment and required moving things around to remain consistent with the rest of configdata.pm. We will find out soon enough if I missed a quote or a comma somewhere.

PR #1142, opened 2015-11-04
Update to #1137 for backward compatibility.

SteveGilvarry:
This PR replaces #1137, which assumed a complete swap to ffmpeg 2.8 or later. This updates that PR with checks for ffmpeg version for the relevant API changes, and maintains ability to use previous ffmpeg versions.
@knnniggett you want to check this against ffmpeg 0.10.15 in your rpmforge repo.
@onlyjob This will probably replace your PR I will close it once we have confirmed that.


knnniggett:
Tested. 
Works on CentOS 6, with ffmpeg 0.10 installed, and a Foscam demo camera.  If it works in this (old) environment, it should work in any other. Merging.

PR #1144, opened 2015-11-04
Support Vivotek ePTZ Camera Controls

schrorg:
I added a control module for our Vivotek cameras which support a software-based PTZ, called "ePTZ". A lot of Vivotek camera models support this simple protocol, so I believe it is useful for others.

This needs the version to be bumped because of the module being added to the database.

If it causes too much noise before the release of 1.29, you should hold this PR back until after the release.


kylejohnson:
Thank you for this.  I have a bunch of Vivotek cameras (they work great 
with ZM), but none which support PTZ, as such I can't test.

schrorg:
@kylejohnson Probably you can! We have some IB8369 and they even don't have real PTZ. But they have a thing called ePTZ. It's basically a Software-Zoom and if you have "zoomed" in, you can move around in the zoomed area (pan/tilt). It seems to me that this feature is generally included in the software of Vivotek cameras.

knnniggett:
The necessary SQL needs to be added to db_create.sql for new databases:
https://github.com/ZoneMinder/ZoneMinder/blob/master/db/zm_create.sql.in

schrorg:
Ah, I forgot! Fixed it now. Thank you for the hint.

knnniggett:
No problem. Thanks for doing the work. 

We currently have another open pr we plan on merging soon #1040 that is using .109.
I want to test and merge that first before we merge this pr.  You don't need to change anything in this pr to compensate for that. I'll manually merge this when the time comes.

connortechnology:
This is great, I am currently doing a 35+ camera install using vivotek cameras.

knnniggett:
Removed the zm_update script from this pr. We should start training our end users to use zmcamtoo.pl to pull in PTZ scripts missing from their existing database, rather than have to bump our version each time we add a control script.

schrorg:
Ah, I didn't know of that tool.

Thank you for merging!

PR #1147, opened 2015-11-05
Reorder RTSPDescribe to avoid -wreorder warnings

SteveGilvarry:
@knnniggett please check this out as this is your area, just moved the RTSPDescribe around a bit as I was getting build warnings and nobody likes them on a release.


knnniggett:
Yeah, I literally had my C++ book in my lap when I wrote that.  What were the error messages?  I'd really like to learn why the order matters in this particular case. Otherwise I'm likely to repeat the same mistake again.

SteveGilvarry:
I believe it is due to the Constructor, but in the interest of be open and honest I taught myself what it was based on the warning.
**Example of warning**
https://travis-ci.org/ZoneMinder/ZoneMinder/jobs/89247628#L3082
This seems like a fair explanation.
http://blog.kutny.net/2014/07/14/c-member-initializer-list-fallacy-a-k-a-wreorder/

connortechnology:
Yeah, that's a good explanation

knnniggett:
Wow, that's good to know. Going to bookmark that page.

Issue #1151, opened 2015-11-07
RTSP monitors don't work with IPv6


schrorg:
I am trying to build a IPv6-only system with about 100 cameras. That means I would like to have ZoneMinder connect to the cameras over IPv6. With ffmpeg and libvlc it would be no problem. I.e. #252 already solves this for ffmpeg.

But I would like to use the native RTSP code from ZoneMinder. And that seems to only support IPv4. The problem is, that the functions in `src/zm_comms.{h,cpp}` and some other files use `gethostbyname(3)`, which is not IPv6 compatible:

> 11/07/15 14:39:01.030891 zmc_m10[1652].DB1-zm_logger.cpp/234 [LogOpts: level=DB5/DB5, screen=OFF, database=INF, logfile=DB5->/var/log/zm/zm_debug.log.01652, syslog=WAR]
> 11/07/15 14:39:01.030951 zmc_m10[1652].DB1-zm_utils.cpp/262 [Detected a x86\x86-64 processor with SSSE3]
> 11/07/15 14:39:01.031433 zmc_m10[1652].DB1-zm_monitor.cpp/2726 [Got 1 monitors]
> 11/07/15 14:39:01.031449 zmc_m10[1652].DB1-zm_monitor.cpp/2759 [Got 1 for v4l_captures_per_frame]
> 11/07/15 14:39:01.031470 zmc_m10[1652].DB2-zm_camera.cpp/39 [New camera id: 10 width: 1280 height: 720 colours: 3 subpixelorder: 6 capture: 1]
> 11/07/15 14:39:01.033006 zmc_m10[1652].DB1-zm_thread.cpp/241 [Creating thread]
> 11/07/15 14:39:01.033031 zmc_m10[1652].DB2-zm_rtsp.cpp/198 [RTSP Local SSRC is 2112cd0c]
> 11/07/15 14:39:01.033040 zmc_m10[1652].DB1-zm_thread.cpp/281 [Starting thread]
> 11/07/15 14:39:01.033119 zmc_m10[1659].DB2-zm_thread.cpp/253 [Invoking thread]
> 11/07/15 14:39:01.033202 zmc_m10[1652].DB1-zm_thread.cpp/302 [Started thread 1659]
> 11/07/15 14:39:01.033232 zmc_m10[1652].DB2-zm_image.cpp/168 [Blend: Using SSE2 fast blend function]
[...]

PR #1152, opened 2015-11-07
Add support for IPv6 in RTSP code

schrorg:
Monitors with source type 'remote' can now be accessed over IPv6 when using the ZoneMinder RTSP code. This patch uses getaddrinfo(3) now instead of gethostbyname(3) - and changes a lot of networking stuff which **should be tested thoroughly**.

This PR addresses #1151. I would consider marking it **not ready for merge** until other some people tested and reviewed it.

### DIFF
--- a/src/zm_comms.cpp
+++ b/src/zm_comms.cpp
@@ -35,6 +35,8 @@
 #include <sys/ioctl.h>
 #include <sys/param.h>
 #include <netinet/tcp.h>
+#include <arpa/inet.h>   // for debug output
+#include <stdio.h>       // for snprintf
 
 #ifdef SOLARIS
 #include <sys/filio.h> // define FIONREAD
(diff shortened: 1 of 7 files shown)


cmcgriff:
Shouldn't [zm_sdp.cpp](https://github.com/ZoneMinder/ZoneMinder/blob/master/src/zm_sdp.cpp#L115) be updated to handle "IP6" as well.  Currently have an issue with my Hikvision camera where it responds with "IP6" and it throws an exception.

connortechnology:
This looks pretty good to me.  Yes, sdp.cpp should probably be updated as well.

schrorg:
Okay... there seems to be an issue with our cameras then. They are IPv6-only and work like a charm with this PR. Seems to need a deeper look into the connection. If I find the time, I will dive into it again on the weekend.

connortechnology:
@cmcgriff what model is your Hikvision camera?

cmcgriff:
DS-2CD2032F-I and DS-2CD2042WD-I

schrorg:
All of my test cameras return IP4 in SDP address type, so I could not really test what happens if a camera returns IP6. :-/
@cmcgriff, does the last push fix the problem for you?

cmcgriff:
Unfortunately I don't have a good setup to build/test with right now.

Looking at your change it appears to match the change that is suggested in the [Trendnet Wiki](https://wiki.zoneminder.com/Trendnet) for the TV-IP310PI using RTSP.

schrorg:
Cool, I didn't see this! I use a lot of Vivotek cameras and they all do IPv6 with no problem but say "IP4" in the SDP connection. That was the reason for the missing code change.

So we'll wait until someone tested it with a matching camera...

schrorg:
Although I am sure, this code works well, I would like to have it tested by some more people. So I would suggest to remove the label and merge the code into master.

knnniggett:
I'm onboard with this. My only concern is timing. We plan to create a 1.30 release candidate on Monday, and then release 1 week later. I'd suggest we merge this, following the creation of the 1.30 release candidate,

schrorg:
@knnniggett Okay, better merge after the 1.30 release. :-)

josh4trunks:
The change in `src/zm_sdp.cpp` also fixes "Remote" streaming for my Hikvision 2032 camera.
Otherwise I get the `Invalid SDP address type 'IP6' in connection info 'IN IP6 ::'` error

connortechnology:
Screw it, this code needs to be in here, and who knows when the next release will be.

Merged.

PR #1164, opened 2015-11-15
compat for php 5.4

abishai:
imagescale is a php 5.5 function, so compatibility layer is required to function under php 5.4 (ZM supported version)


pliablepixels:
I validated this code works well. Would appreciate if this is merged @knnniggett

Issue #1155, opened 2015-11-10
APIs need to honor user roles


pliablepixels:
As of today, if you are using APIs, as long as you have a valid login, you have full access. We need to tie this into the user permissions so the APIs don't return any more data than the role allows. (Issue originally reported here: https://forums.zoneminder.com/viewtopic.php?f=33&t=23653)
##

mnoorenberghe:
What's the policy on closing issues? I notice many issues that had fixed merged weren't closed. Are they intentionally left open until release and then sometimes forgotten about? I think in many cases it's because the committer didn't include "fixes" before the issue number and then nobody came back to close it.

SteveGilvarry:
Just time and resources. I actually said just 2 days ago to the other team members that I was going to start closing out a few to get numbers down. So if you want to help flagging them I will follow behind to close, as well as reviewing them myself. Hopefully no one gets offended by me closing their issue, nothing personal.  Some of these should be in the forums as opposed to actual bugs or development features.

PR #1165, opened 2015-11-17
User role based access for APIs

pliablepixels:
@kylejohnson please review interim progress (Please label Not Ready to Merge) #1155 

So far I've implemented user roles for the Monitors API. If this approach is ok, I'll continue down this path.

### DIFF
--- a/web/api/app/Controller/AppController.php
+++ b/web/api/app/Controller/AppController.php
@@ -63,9 +63,47 @@ class AppController extends Controller {
         	$zmOptAuth = $config['Config']['Value'];
         	if (!$this->Session->Read('user.Username') && ($zmOptAuth=='1'))
         	{       
-        		 throw new NotFoundException(__('Not Authenticated'));
+        		 throw new UnauthorizedException(__('Not Authenticated'));
         	 	return; 
         	}     
+		// PP #1155
+		else
+		{
+			$this->loadModel('User');
+			$loggedinUser = $this->Session->Read('user.Username');
+			$isEnabled = $this->Session->Read('user.Enabled');
+			// PP - this will likely never happen as if its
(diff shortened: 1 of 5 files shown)


web/api/app/Controller/AppController.php
```diff
@@ -63,9 +63,47 @@ public function beforeFilter() {
...
+			{
+				throw new UnauthorizedException(__('User is not enabled'));
+				return;
+			}
+
+			if ($zmOptAuth=='1')
+			{
+				$options = array ('conditions' => array ('User.Username' => $loggedinUser));
+				$userMonitors = $this->User->find('first', $options);
+				$this->Session->Write('allowedMonitors',$userMonitors['User']['MonitorIds']);
```
kylejohnson, line 88: I haven't read up on this in a while, but what do you think of `Session->Write` vs. `Configure::write`?  I am using the latter in other places.

kylejohnson:
Overall this looks pretty good.

Maybe I'm not reading the AppController class right - but how does a user actually log in to the API?

web/api/app/Controller/AppController.php
pliablepixels, line 88: @kylejohnson , based on my reading:

`Configure::write` is for application wide variables while `Session->Write` is for settings that get set only for that specific session. The latter is what we need for these variables as they will change depending on who has opened the session

pliablepixels:
@kylejohnson - the user does not log into the API. The user logs into ZM. This sets a set of variables - I am accessing those variables in the context of the same session to make sure the user has logged in.

pliablepixels:
I'm closing this as PR https://github.com/ZoneMinder/ZoneMinder/pull/1196 supersedes this and adds OPT_API to UI too

PR #1166, opened 2015-11-19
use webvvt to overlay timestamp on videos to videos in timeline and event

liucougar:
honoring Monitor.LabelFormat

also fixed bug which prevented seeking in timeline video preview

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -261,3 +261,38 @@ if ( focusWindow )
 }
 window.addEvent( 'domready', checkSize);
 
+function convertLabelFormat(LabelFormat, monitorName){
+	//convert label format from strftime to moment's format (modified from
+	//https://raw.githubusercontent.com/benjaminoakes/moment-strftime/master/lib/moment-strftime.js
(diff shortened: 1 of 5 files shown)


SteveGilvarry:
@liucougar Looks good I will try to check this on the weekend and merge before I do another master merge.

liucougar:
with this new timestamping, I noticed there is a bug in the creation of the recording videos (at least with the videos created by the `record` mode, each of which is about 10 mins long): the first 2-3 seconds of the video is not viewable in either video html5 element in browser nor vlc. in addition, there is 2-3 seconds extra at the end of the video in both html5 video and vlc. I think something is not quite right with the timing info in the recorded videos.

SteveGilvarry:
Keyframe issues I believe

liucougar:
i have 2 seconds key frames, so I guess that makes sense. Do you have any suggestions on how that could be fixed?

SteveGilvarry:
Fifo packet buffer, keep meaning to get to it, that and pre event buffer of
packets.

liucougar:
@SteveGilvarry any update on this please?

PR #1170, opened 2015-11-21
Simplification of README

pliablepixels:
Given that the README is part of the main ZM page, I think putting package build instructions confuses many users - they think this is what is needed to be done. I propose we make this README much shorter and if required move the build from source/create package stuff out to some other file


pliablepixels:
ill clean this up an PR again

knnniggett:
Agreed. It was getting a bit lengthy.

pliablepixels:
Sounds good. I submitted a different PR for this. 
https://github.com/ZoneMinder/ZoneMinder/pull/1171

PR #1171, opened 2015-11-21
README simplification

pliablepixels:
Given this is shown in the root zoneminder page, I'd like to propose we don't complicate by adding instructions for building packages etc - we could focus on installing from an existing package (which should serve most users) and link the rest into a different file.


knnniggett:
Was going to merge this when I had a thought.
Rather than move the content into new *.md files, would it be more appropriate to move this information into readthedocs?

pliablepixels:
Right, this change has two parts:

a) Deletion of "build from package" instructions from the main README.md I think that is still valid
b) Moving of those instructions to build-package.md - I think this part should not exist anymore, and instead simple add a link to README.md to point to read the docs build from package.

knnniggett:
You want to do that, or do you want me to push the change in b) to your pr?

pliablepixels:
Would you mind doing it? I've deleted that branch and can't quite figure out how to continue from where I left off

knnniggett:
Going to close this and create a new PR, which moves the building instructions to readthedocs.

PR #1173, opened 2015-11-22
Error on missing submodules

knnniggett:
Generate an error when the Crud submodule has not been initialized.
I left it plural, because of potential plans for additional submodules, pending peer approval.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,12 @@ set(zoneminder_VERSION "1.28.109")
 # make API version a minor of ZM version
 set(zoneminder_API_VERSION "${zoneminder_VERSION}.1")
 
+# Make sure the submodules are there
+if( NOT EXISTS "${CMAKE_SOURCE_DIR}/web/api/app/Plugin/Crud/.git" )
+message( SEND_ERROR "The git submodules are not available. Please run
+git submodule update --init --recursive")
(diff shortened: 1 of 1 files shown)


connortechnology:
Guys, this broke debian package building.

When building a package, the build scripts filter out .git and .svn and others.

knnniggett:
See #1178

SteveGilvarry:
I really think the submodule is going to cause more issues than it has solved, do we really need to be linking to them. Next big one is going to be on release the github source package download won't include it, so official packagers might miss it, already raised by onlyjob. Worth a discussion.

knnniggett:
That's exactly what this issue addresses. If the packagers miss it, then cmake will stop on an error.

SteveGilvarry:
Yes so they will need to roll their own source gz file rather than take from github and apply patches. Or we provide a release zip other than github one.

knnniggett:
That's one thing I don't know the answer to. What is the most common way package maintainers obtain the source tar ball? Do they get it from the releases page?

In rpm-land, one can just get the Source tag in the rpm specfile to point directly to our github repository. To get the sub-module installed, one needs to insert the appropriate submodule init command in the %prep portion of the packaging script. I don't have any idea if Debian can do this.

In any case, if there is a way to put a custom tarball, which includes the submodules, on the release page, then I've got no problem doing that.

PR #1184, opened 2015-12-09
Add code to detect the change in REALM from older to newer firmware

connortechnology:


kylejohnson:
Hey - being picky here as this is a small and innocent change, but
* If you're going to merge your own PR, please add a description of what drove the PR.
* You labeled it 'not ready for merge', and then merged it right after.
* Please provide a more descriptive commit message.  Looking at the commit, a better message might have been "Detect the change in REALM on TVIP862 cameras" .

PR #1185, opened 2015-12-09
Storageareas

connortechnology:


SteveGilvarry:
Milestone is a suggestion that we bring this in next version.

schrorg:
I am currently trying to hack around a problem with a multi-server setup and a slow connection between the servers. One of them holds the database and exports the filesystem via NFS to the other server over a slow connection (around 2 Mbit/s). Everything works fine, but this is too slow for recording events.

Would the Storageareas feature help in this situation? If yes, how could I help developing it?

connortechnology:
Storage areas would help, as you can use local storage in each server 
for recording.

What distro are you on?

schrorg:
Local storage would indeed solve my problem. I am on Ubuntu (the DB- NFS-exporting server) and Raspbian (~Debian 8).

connortechnology:
Well, you can use my storageareas ppa. ppa:iconnor/zoneminder-storageareas.

Be warned though, there is a ton of bleeding edge code in there, like video storage and some skin/console options reworking.

PR #1189, opened 2015-12-16
PTZ - Added autostop to Down command on FI8918W

marcolino7:
I added PTZ autostop to FI8918W


pliablepixels:
@knnniggett can we merge this? The current script keeps moving to the bottom -

knnniggett:
While we are evaluating the 1.29 release candidates, I had not been merging anything other than bug fixes. If what you say it true, then yeah I think that fits that category.

PR #1191, opened 2015-12-17
Filter object

connortechnology:
This works for me, but could probably use some cleanup/further work before merging.

The idea is to move the code from zmfilter.pl into a nice object, which we can use elsewhere, like zmvideo.pl.


connortechnology:
I already merged this to storageareas

PR #1192, opened 2015-12-17
Better zmvideo

connortechnology:
This code implements the ability to specify a saved filter, and to generate the videos for each event returned by that filter.
It also implements the ability to concatenate the videos into 1 big video.

It uses several new classes, and moves some code around.


SteveGilvarry:
@knnniggett We leaving this for post 1.29 or going in?

knnniggett:
After 1.29. We will never release if we keep adding stuff.

connortechnology:
I may have already merged it to storageareas, so let it live there.

PR #1194, opened 2015-12-19
Documentation

pliablepixels:
@SteveGilvarry  - can you check if this correctly resets master to rtd theme? It seems if we use "default" it uses rtd when hosted in readthedocs but in my local copy shows as classic which is confusing. So I :
a) remove my static import - wasn't needed
b) changed html style to rtd_them instead of default


SteveGilvarry:
@pliablepixels http://sgzoneminder.readthedocs.org/en/pliablepixels-documentation/installationguide/index.html
It seems to work but then I saw this saying it will fail the build if you try to use on RTD, it doesn't seem to so maybe their docs are out of date. http://docs.readthedocs.org/en/latest/theme.html#how-do-i-use-this-locally-and-on-read-the-docs
They really don't seem to have much on styling documented.

pliablepixels:
@SteveGilvarry  - thanks - in that case lets just keep it as is. I also figured out how to add custom styles without breaking rtd theme - so we can continue tweaking style elements in the future. For now, all the custom style does is add a black border around images - given there are many white on white screenshots

pliablepixels:
I'm done for now - if this latest PR also works for you @SteveGilvarry feel free to merge it

Issue #1195, opened 2015-12-19
api: more security


pliablepixels:
With the API being packaged with 1.29, I think we really should, if possible add an option to:
a) Disable the API via UI
b) With API enabled, do a minimum to honor user roles - don't return monitor data and event data for stuff a user is not allowed to see in ZM. Other elements (such as system view/edit) are less important for now.

I am going to submit this as a PR - if you want to include it in ZM 1.29 go ahead, or a milestone later. This supersedes #1165 - I'll bring in the changes from there to this one

PR #1196, opened 2015-12-19
Api more security

pliablepixels:
Addresses issue #1195

### DIFF
--- a/web/api/app/Controller/AppController.php
+++ b/web/api/app/Controller/AppController.php
@@ -77,7 +77,6 @@ class AppController extends Controller {
         		 throw new UnauthorizedException(__('Not Authenticated'));
         	 	return; 
         	}     
-		// #1155 fixes
 		else
 		{
 			$this->loadModel('User');
--- a/web/api/app/Controller/AppController.php
+++ b/web/api/app/Controller/AppController.php
@@ -58,14 +58,63 @@ class AppController extends Controller {
 	// Also checking to do this only if ZM_OPT_USE_AUTH is on
 	public function beforeFilter() {
 		$this->loadModel('Config');
+		
+        	$options = array('conditions' => array('Config.' . $this->Config->primaryKey => 'ZM_OPT_USE_API'));
+	 	$config = $this->Config->find('first', $options);
+        	$zmOptApi = $config['Config']['Value'];
+
+		if ($zmOptApi !='1')
+		{
+        		 throw new UnauthorizedException(__('API Disabled'));
+        	 	return; 
+		}
+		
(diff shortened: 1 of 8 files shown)


SteveGilvarry:
@pliablepixels Can I suggest that since github tracks all changes, blame, etc and we do not have any precedent of using comments to attribute a change that we remove the "//PP" comments?
Other opinions? @knnniggett @kylejohnson

pliablepixels:
@SteveGilvarry - yes, this was actually a way for me to track local changes before committing  - but its not needed - I'll strip and push again

SteveGilvarry:
Maybe the #1155 ones can go altogether also

pliablepixels:
Everything in #1155 is also in this (including bug fixes to it) - which is why I closed 1155

pliablepixels:
oh you mean PP? Yes, in the update I pushed above with a heart wrenching code from Lord Tennyson, PP is gone from there too. There will be other fragments of past PRs that will still have PP - I'll get around to removing them later, after I consult with my legacy planner on what this means to my legacy.

web/api/app/Controller/AppController.php
```diff
@@ -58,14 +58,63 @@ class AppController extends Controller {
...
         	$options = array('conditions' => array('Config.' . $this->Config->primaryKey => 'ZM_OPT_USE_AUTH'));
 	 	$config = $this->Config->find('first', $options);
         	$zmOptAuth = $config['Config']['Value'];
         	if (!$this->Session->Read('user.Username') && ($zmOptAuth=='1'))
         	{       
-        		 throw new NotFoundException(__('Not Authenticated'));
+        		 throw new UnauthorizedException(__('Not Authenticated'));
         	 	return; 
         	}     
+		// #1155 fixes
```
SteveGilvarry: @pliablepixels Sorry I meant these "// #1155 fixes" ones, just the comment as again we can always see what belongs to what in Github. Or unless you think something needs to say why it exists in which case maybe change the comment to something more meaningful.

pliablepixels: Oh ok  - yes, what I really should do is refer to issues in commit messages - like I do in zmN - its a good way to map code to issues later. But I don't think I can find any quote from tennyson that will be apt for this change

pliablepixels: Actually @SteveGilvarry - I just remembered, its an easy way for me to say "This change went in due to user role auth checking" - should I remove it?

SteveGilvarry: Something like //User role check if you think it would help someone reading the code understand it's purpose, but for me doesn't need to talk about a change. That is all covered by commits and PR in github.

pliablepixels:
done

PR #1201, opened 2015-12-22
Skip directories that have non-digits in them

connortechnology:


connortechnology:
I think this one is sorta important. Please merge it.

connortechnology:
Hmm, my commit message may not make it clear what the deal is here.  So the issue is that glob does not use regex's, it uses globbing.  so \d* gives us every file that starts with a digit.  So if your camera name starts with a digit, then a symlink will be put in that dir, and zmaudit will find that 1mycameraname camera does not exist in the db and will rm -r the symlink.  

I don't really think it's a big deal, because the rm should just delete the symlink which will get re-created... but it kinda gives me the willies.  Plus it adds extra processing time to zmaudit.

knnniggett:
Makes sense.

Issue #1204, opened 2015-12-25
All of the Perl code uses "#!/usr/bin/perl"


lembark:
This prevents using a modern perl on most systems (e.g., many RH systems are stuck at 5.8 compiled for binary compatability with 5.00503). Simple fix: use "#!/usr/bin/env perl", allowing the user to configure their environment for the preferred perl.

This arrangement also makes testing ZoneMinder with different versions of Perl a whole lot simpler since perlbrew (or a simple path update) can pick the version to use. It also simplifies installation into Docker since the full Perl on most distro's is a bit bloated.

knnniggett:
ZoneMinder works just fine on RedHat based systems and their binary clones. 

I do not agree with letting the system auto-pick a Perl interpreter using env for a system package like ZoneMinder.

One of the "features" of using env is all the process names get replaced with generic ones like "perl'. So instead of seeing zmpkg and zmdc in your process list, you see a bunch of instances of "perl" in your process list.

Indeed using env is documented as not allowed in the packaging documentations of distros like Debian in Redhat. 

For example:
https://fedoraproject.org/wiki/Script_Interpreters_%28draft%29

What you'll find if you look at other projects that use env to find their script interpretor, such as Ansible, is by the time the applications makes it into package form, the env command has been stripped out and the shebang has been put back into its more traditional form.

The env tool is handy for users who install and use Perl under their home directory, but ZoneMinder is a system tool, which is not intended to work in that kind of environment. 

ZoneMinder is intended to be built against a specific version of Perl found on the build host. If you want to change the version of Perl then rebuild ZoneMinder. This is by design, and is the same procedure one would have to make when changing the version of ffmpeg, liblv, etc.

connortechnology:
I could see us merging a PR that made the shebang line configurable so that a developer in your situation could build everything to use env, but our packaging scripts would use the standard /usr/bin/perl.

lembark:
On Sat, 26 Dec 2015 07:45:36 -0800
Isaac Connor <notifications@github.com> wrote:

> I could see us merging a PR that made the shebang line configurable 
> so that a developer in your situation could build everything to use 
> env, but our packaging scripts would use the standard 
> /usr/bin/perl.

Problem there is that /usr/bin/perl is useless on many distros,
through use of ancient or badly compiled perl executables. For
exmple, anyone useing perlbrew to get a current version for 
either use or testing cannot make use of ZoneMinder: they cannot
[...]

PR #1205, opened 2015-12-25
replace "#!/usr/bin/perl" with "#!/usr/bin/env perl"

lembark:
/usr/bin/perl on many systems is badly out of date and mis-compiled. Most RH systems are still at 5.8 compiled for 5.00503 compatability (i.e., about a decade out of date).

It is also difficult to re-build the distro's perl just to test new versions of Perl with packages (e.g., ZoneMinder). 

Simple fix: "#!/usr/bin/env perl" allows the local user to set up their environment to use an altenate (e.g., via perlbrew or specifying /opt/perl/<version>/bin/perl).


kylejohnson:
Thank you for submitting this PR.

While your changes look good, the build failed while trying to run two of the perl scripts.  I believe it is an issue with the build environment.

```shell
$ sudo zmpkg.pl start

/usr/bin/env: perl -wT: No such file or directory

The command "sudo zmpkg.pl start" exited with 127.

$ sudo zmfilter.pl -f purgewhenfull

/usr/bin/env: perl -wT: No such file or directory

The command "sudo zmfilter.pl -f purgewhenfull" exited with 127.

Done. Your build exited with 1.
```

knnniggett:
Won't merge. See #1204

PR #1209, opened 2015-12-30
Strict key checking

adamreiser:
Added a boolean option to require strict host key checking for SFTP uploads.


connortechnology:
I have no problem with merging this.

knnniggett:
I don't either, after the next release.

PR #1210, opened 2015-12-31
Restore missing docs to storageareas branch

SteveGilvarry:
@connortechnology Not sure why it went missing after another merge, but just cherry-picked it back in.


connortechnology:
It's something to do with the debian package build and me doing a git commit -a.  I really must be more careful, and also must audit those debian package build scripts.

Thanks so much!

Issue #1214, opened 2016-01-01
ZM_SERVER_ID uninitialized


knnniggett:
@connortechnology 
I'm not sure if this is by design or something that needs fixing.

On a standalone, non-multiserver, machine just upgraded to 1.29.0, when I start zoneminder, I get the following non-fatal error:
```
[abauer@localhost db]$ sudo service zoneminder start
Starting ZoneMinder: Use of uninitialized value $Config{"ZM_SERVER_ID"} in concatenation (.) or string at /usr/bin/zmpkg.pl line 229.
```
The logic that sets ZM_SERVER_ID is here:
https://github.com/ZoneMinder/ZoneMinder/blob/master/scripts/ZoneMinder/lib/ZoneMinder/Config.pm.in#L106

It first checks if ZM_SERVER_NAME is set and if it is not it checks ZM_SERVER_HOST.

However, in my case, neither ZM_SERVER_NAME nor ZM_SERVER_HOST are set.

So the question is, for a standalone machine running 1.29.0 and not using multiserver, should the logic in Config.pm consider this as a third case, or should it be required for the end user to always set ZM_SERVER_HOST in zm.conf?

connortechnology:
Yeah... sounds like it should set it, just to remove the warning.  Patch coming in a few.

This is the problem with this last-minute merging of multi-server I guess.  No one has been testing the upgrade case much.  Still, this is just a warning, everything still works. 

I have another patch to fix issues in the db logging.

knnniggett:
No sweat. This is why I made today's release just a release candidate.
With all the stuff we merged since our last release, I knew there would some things to work out.

knnniggett:
I'll do another round of testing tomorrow.

knnniggett:
Looks good.

PR #1215, opened 2016-01-02
Multi server

connortechnology:
Some fixes for database logging/filtering of logs based on Server.


connortechnology:
Hopefully fixes #1214

knnniggett:
Hmmm... something is still going on. The web log is no longer populating. I have not yet found an error message to describe the problem though. Still looking.

knnniggett:
@connortechnology 
I discovered an important clue:
If I add "localhost" as an entry under the Servers tab, the web log immediately springs to life.

connortechnology:
possibly.  I will look closer.

connortechnology:
I  really think we should default to only showing a subset of the log when you bring up the web log.  Showing the whole thing just kills firefox.   I'm thinking like the most recent 1000 items or something.

knnniggett:
Hmmm.... no I think I was mistaken. Changing log.Server to log.ServerId causes the Server column to show the ID, which in my case "0", rather than the Server name.

connortechnology:
I think 71 is correct.  The log has two server-related fields, ServerId which is the integer id #, and Server which is the text name.  the text name gets filled in in ajax/log.php.  

I am wondering if your issue was having a value in ServerId, but not having an entry in the Servers table.  We have no foreign keys in our schema, so that could happen.  Really need to improve our schema.  

Anyways, I have added a check which should make it not fail in this case.  Just testing before I commit and PR.

knnniggett:
Excellent. We are working now.

PR #1222, opened 2016-01-04
Multi server

### LINKED ISSUE #1219: monitorUrl incorrect in non-MultiServer Envinroment
@connortechnology 
Should be easy to fix.

monitorUrl, as defined in watch.js.php, looks like this when not in a multiserver environment:
 `var monitorUrl = 'https://';`

https://github.com/ZoneMinder/ZoneMinder/blob/master/web/skins/classic/views/js/watch.js.php#L51

That breaks a few of the widgets, like the state and the fps, in the Watch view.

So do we redefined $monitor->Server()->Url() to return empyset instead or just ignore it when it contains that particular value?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29551835-monitorurl-incorrect-in-non-multiserver-envinroment?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github).
</bountysource-plugin>

connortechnology:
Fixes #1219 

I decided that you were right knnniggett: The Server->Url should just return protocol/hostname, so in the multi-server case, just returning empty is preferred.

### DIFF
--- a/web/includes/Server.php
+++ b/web/includes/Server.php
@@ -34,7 +34,11 @@ class Server {
 	}
 
 	public function Url() {
-		return ZM_BASE_PROTOCOL . '://'. $this->Hostname();
+		if ( $this->Id() ) {
+			return ZM_BASE_PROTOCOL . '://'. $this->Hostname();
+		} else {
+			return ZM_BASE_PATH;
+		}
 	}
 	public function Hostname() {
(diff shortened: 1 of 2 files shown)


knnniggett:
Thanks. We can always change it again later if we want to rework how we do this.

PR #1223, opened 2016-01-04
Multi server

connortechnology:
This just removes the Server column for the non-multi-server case


connortechnology:
I did a seriously dumb thing in one of the previous PR's.  This needs merging.
@knnniggett

knnniggett:
That's ok. I won't tell many people. :-)

PR #1224, opened 2016-01-04
Updated Ubuntu installation docs to be consistent.

Simpler1:
Updated installation documentation for Ubuntu 15.x and 14.x and added a note about Linux Mint 17.x.

These two sets of instructions are almost identical and should probably be combined into one section (with the individual differences noted) so that any future changes just need to be done once to keep these consistent.


knnniggett:
Let's leave this open for now.
Superseded by the PR that will come out of #1231

knnniggett:
Closing. See #1231 and #1546

PR #1225, opened 2016-01-04
Added profiles for Wanscam-HW0025 and IPCC-7210W

Simpler1:
Added a profile for Wanscam HW0025 and a separate profile for IPCC 7210W.


knnniggett:
This is missing the bits of sql to make the control scripts show up in the web consoler:
https://github.com/ZoneMinder/ZoneMinder/blob/master/db/zm_create.sql.in#L591

Simpler1:
@knnniggett I added the two corresponding lines to the zm_create.sql.in.  I'm not sure if there is more that is needed.  If so, let me know.

Thanks...

knnniggett:
This is just an FYI.
These scripts will not show up in an existing database after an upgrade because the upgrade scripts were not part of this pr. 

The upgrade scripts are dependent on the version we want to call the next version of zoneminder, however, so we can create them just yet.  In any case, the end user can always use zmcamtool.pl to manually import new ptz control scripts into an existing database.

knnniggett:
This is just an FYI.
These scripts will not show up in an existing database after an upgrade because the upgrade scripts were not part of this pr. 

The upgrade scripts are dependent on the version we want to call the next version of zoneminder, however, so we can create them just yet.  In any case, the end user can always use zmcamtool.pl to manually import new ptz control scripts into an existing database.

connortechnology:
damn, I'll fix that

knnniggett:
Alternatively, we could start advertising zmcamtool.pl to do this after the fact. We could even call it in each of our post-install packaging scripts. Bumping the version number for each ptz control script is rather annoying.

connortechnology:
That's actually a good point.  Populating content in zm_create.sql isn't ideal.

knnniggett:
New SQL still needs to be added to `zm_create.sql` for each new ptz control script, but if we start using zmcamtool.pl, we won't need to create a `zm_update-x.xx.sql script` and bump the minor version each time we add one.

PR #1226, opened 2016-01-04
fixed install steps

pliablepixels:
https://forums.zoneminder.com/viewtopic.php?f=34&t=23841&p=91499#p91499

### DIFF
--- a/docs/installationguide/ubuntu.rst
+++ b/docs/installationguide/ubuntu.rst
@@ -397,12 +397,7 @@ After editing your /etc/apache2/conf-available/zoneminder.conf should look like:
 	sudo chown -R www-data:www-data /usr/local/share/zoneminder/
 
 
-**Step 12:** Make sure you can view Monitor View
-
-1. Open up ZM, configure your monitors and verify you can view Monitor feeds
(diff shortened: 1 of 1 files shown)


knnniggett:
Why does this set of instructions have the user create a custom apache config file when the one they should be using is here?
https://github.com/ZoneMinder/ZoneMinder/blob/master/misc/apache.conf.in

If that file is not what we want, then we should change it.

Simpler1:
@pliablepixels I made similar and additional changes in pull request #1224 earlier in the day.  

I guess great minds think alike.  :)

knnniggett:
@pliablepixels When you are back in town, we need to have a talk about these instructions as they don't step the user through creating a package at all, like the title says. 

We need to document a build procedure that builds zoneminder correctly during the build process, and does not require the end user to go back and hack config files after zoneminder has been built. Basically, these instructions should do what @connortechnology does when he creates a package for his ppa.

knnniggett:
Let's leave this open for now.
Superseded by the PR that will come out of #1231

pliablepixels:
I've updated the docs here https://github.com/pliablepixels/ZoneMinder/commit/0ef6d55e8ffad1f169ae661f06a1f46c1dce38ac so closing this

PR #1234, opened 2016-01-08
don't load fonts from googleapi

knnniggett:
Deletes the line that imports fonts from googleapi


knnniggett:
nevermind. @connortechnology beat me too it.

PR #1239, opened 2016-01-14
use trim on values when saving.

connortechnology:
Pretty simple, but may have unintended consequences. 

We have this getFormChanges function which we use to build an update sql for various objects.  This trims the values for basic scalar values.  

This is mostly important for cases like the path for a storage area.


connortechnology:
So I havn't had any troubles with this yet.

PR #1240, opened 2016-01-16
zmtrigger - process off+time delay condition

knnniggett:
The time delay effect was not working when the "off" action was specified because zmTriggerEventOff was being called immediately, rather than after the delay.

This PR fixes that. 
The code that handles the delay effect has been moved into its own subroutine.

### DIFF
--- a/scripts/zmtrigger.pl.in
+++ b/scripts/zmtrigger.pl.in
@@ -504,22 +504,11 @@ sub handleMessage
         Info( "Set monitor to $state\n" );
         if ( $delay )
         {
-            my $action_time = time()+$delay;
             my $action_text = $id."|".( ($state eq "enable")
                                         ? "disable"
                                         : "enable"
-                                      )
-            ;
-            my $action_array = $actions{$action_time};
-            if ( !$action_array )
-            {
-                $action_array = $actions{$action_time} = [];
-            }
-            push( @$action_array, { connection=>$connection,
-                                    message=>$action_text
-                                  }
-            );
-            Debug( "Added timed event '$action_text', expires at $action_time (+$delay secs)\n" );
+                                      );
+             handleDelay($delay, $connection, $action_text);
         }
     }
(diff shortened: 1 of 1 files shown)


knnniggett:
I slept on this and thought of a couple of discussion points related to this.

zmtrigger has undocumented enable/disable action functionality. It turns out that it has the same issue with $delay as the on/off action set. We should do something about this, but I'm not sure what exactly. 

Here are some possible solutions (not all are mutually exclusive):
1) Add the enable/disable functionality to the documentation 
2) Fix the $delay functionality so it works just like the on/off delay: enable+time - enables the monitor only for "time" seconds. disable+time - disable the monitor permanently after "time" seconds has elapsed.
3) Fix the $delay functionality such that: enable+time - enable the monitor permanently after "time" seconds has elapsed. disable+time - disable the monitor permanently after "time" seconds has elapsed.
4) Remove the the enable/disable functionality altogether, because I believe this has been superseded by the API.
5) Speaking of the API, how does that affect the future of zmtrigger as a whole? Note that zmtrigger opens a port on the server, which accepts unencrypted communication to any client without authentication (yikes).

I was getting ready to pound out items 1 and 2, but I guess my morning coffee kicked in and items 3-5 came to mind.

@connortechnology 
@pliablepixels 
@SteveGilvarry 

Looking for opinions here. I can move this into an issue if we think this requires a long term plan.

connortechnology:
I don't think we should remove it.  At least not right now.  api is too new, and I think this code will be useful for other situations.

I do think that zmtrigger should be manually turned on and well documented.

connortechnology:
btw, I'm also fine to merge this for 1.29.  It's a straight-forward bug fix.

knnniggett:
Let's just merge it as it is then. 
Creating an action of "disable+time" will disable the monitor immediately so technically we've still got a bug in that part, but I'm not sure if anyone is even using that feature. Not sure it is worth fixing right now.

PR #1243, opened 2016-01-19
use local user @ hostname.local for author

knnniggett:
When creating a local build, use the local user account name and the hostname at the author.


connortechnology:
You could remove all references to me.  I build as me, so your additions should pick up my name and address.

knnniggett:
@connortechnology 
Let me know what you think.

If `hostname -d` does not return a domain, then the email domain is created by concatenating ".local" to the hostname.

Getting the full name of the user account was surprisingly complex.

PR #1248, opened 2016-01-24
updated documentation for ubuntu

### LINKED ISSUE #1231: Update Ubuntu Installation Package Building Guide
This is a placeholder to update the Ubuntu Package Building Instructions to actually build a package:
https://github.com/pliablepixels/ZoneMinder/blob/8244496df1d3b0c8df30a6946930833665e8282f/docs/installationguide/ubuntu.rst#id3

We don't want to create any new procedures. Instead, we want our end users to perform the same steps the development team uses to build a package. For Ubuntu, that means we want to follow what @connortechnology currently does for the  packages in his ppa.

He uses the script found here to build a snapshot: `utils/do_debian_package.sh`
It is intended to be fully automated.

It uses the files found under `distros/ubuntu1204_cmake` .

@pliablepixels @Simpler1
Since I don't use Ubuntu, I ask for your participation

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29703886-update-ubuntu-installation-package-building-guide?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pliablepixels:
Resolves #1231

### DIFF
--- a/docs/installationguide/ubuntu.rst
+++ b/docs/installationguide/ubuntu.rst
@@ -1,5 +1,5 @@
-Ubuntu
-======
+Ubuntu Instruction
+===================
 
 .. contents::
 
--- a/docs/installationguide/ubuntu.rst
+++ b/docs/installationguide/ubuntu.rst
@@ -286,192 +286,108 @@ Harder Way: Build Package From Source
 -------------------------------------------
 (These instructions assume installation from source on a ubuntu 15.x+ system)
 
-**Step 1:** First make sure you have the needed tools
+**Step 1:** Grab the package installer script
+
+::
+
+	wget https://raw.githubusercontent.com/ZoneMinder/ZoneMinder/master/utils/do_debian_package.sh
+	chmod a+x do_debian_package.sh
+
+
+**Step 2:** Install some core dependencies
 
 ::
 
 	sudo apt-get update
-	sudo apt-get install cmake git
 
-**Step 2:** Next up make sure you have all the dependencies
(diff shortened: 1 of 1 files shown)


knnniggett:
@pliablepixels Was this the PR you were reminding me about?

Can we change step 2 to be:
```
sudo apt-get install devscripts equivs
sudo mk-build-deps -ir ./debian/control
```
This eliminates having to manage that God awful, ever-changing list of dependencies in multiple places. We just need to keep the control file up to date.

knnniggett:
Something else I just noticed. Instead of spawning a new apache config file in the rst documentation, let's refer to one that already exists, whether that be the generic one under /misc or an Ubuntu specific one under one of the folder under /distros.

pliablepixels:
@knnniggett yes this is the one.

I have some doubts:

a) There is no ``./debian/control`` at step 2, so the command you suggested fails 

```
root@ubuntu:/home/pp# sudo mk-build-deps -ir ./debian/control
W: Unable to locate package ./debian/control
mk-build-deps: Unable to find package name in `apt-cache showsrc ./debian/control'
```

b) I'm not sure which apache file you are referring to - the package instructions (build package from source) replaces the existing cmake instructions

knnniggett:
Right now I remember... we need to embed 
```
sudo apt-get install devscripts equivs
sudo mk-build-deps -ir ./debian/control
```

into the do_debian_package.sh, after it makes the debian symlink. Let me make a pr for that.

About the apache config, I misread the PR. You were removing the part I had a question about, not inserting it.

knnniggett:
I'll go with your advice and merge this now. When I make a PR with the do_debian_package.sh changes, I'll also include the doc changes.

PR #1249, opened 2016-01-24
Fix PelcoP checksum calculation

knnniggett:
Unlike the Pelco D protocol, the STX sync packet should be included in the checksum calculation as defined in the Pelco P protocol definition. This pr fixes that.

https://forums.zoneminder.com/viewtopic.php?f=32&t=23759

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Control/PelcoP.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Control/PelcoP.pm
@@ -133,8 +133,10 @@ sub sendCmd
 
     my $result = undef;
 
+    # Pelco P protocol checksum is created by XOR'ing the first seven bytes in the message
+    # including the first byte, the STX sync packet
     my $checksum = 0x00;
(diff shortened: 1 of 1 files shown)


knnniggett:
@kylejohnson @connortechnology @SteveGilvarry 

Can someone either comment on this or merge please?

connortechnology:
Sure would be nice to have a db of people with various camera models so that we could contact them for testing :)

I'm happy to merge it.  Worst case scenario is that it breaks someone's ptz... and then we fix it.

knnniggett:
Since most cameras of olden-day support both Pelco D and Pelco P protocol, my guess is owners of these old ptz cameras simply switched to the Pelco D protocol. This bug is pretty obvious. The Perl script was not calculating the checksum correctly according to the well defined Pelco P documentation.

But your right, we don't have any way to know for sure if the script, broken as it is, doesn't actually work for a camera which might have a broken implementation of Pelco P. Maybe the original author of the Pelco P perl script had a camera like that.

The only way I know to find out is to merge this. Anyone else got any thoughts?

PR #1254, opened 2016-01-29
1253 zms quit cmd

### LINKED ISSUE #1253: Provide an ability to quit zms for a specific connkey
ZMS is written to implement a long running TCP connection continuously streaming images. The only way to quit the connection is to "kill the browser window". While this is fine for a desktop browser, it makes it almost impossible to release zms resources in single page mobile apps (like zmNinja or any other app in the future). Any web application that uses ``img src`` to display images does not have control on the stream and cannot terminate the TCP object on its own which makes it exceeding hard to remove 'zms' processes even when nothing is being viewed.

There is an option to send a "Pause" and "Stop" control command today - that pauses the image streaming but the zms process still continues in memory and headers are still exchanged.

Proposal: We introduce a new CMD_QUIT to zm_event.cpp and zm_monitor.cpp
that basically does this:

```
case CMD_QUIT :
        {
            Debug( 1, "User initiated stream termination" );
            exit(-1);

        }
```
 
I haven't tested this - but I'm going to do that in the next day or two. @connortechnology @knnniggett  - do you have any issues with this approach? I'm currently going nuts trying to plug zms leaks and it shouldn't be so hard.

Its currently very hard to try and contain run away zms processes since web-apps don't have the ability to "close" the window 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30331464-provide-an-ability-to-quit-zms-for-a-specific-connkey?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pliablepixels:
Resolves #1253


kylejohnson:
Looks groovy to me.

Issue #1242, opened 2016-01-17
Dockerfile not working


vlkz:
```
E: Unable to locate package libavresample-dev
The command '/bin/sh -c apt-get update && apt-get install -y    libpolkit-gobject-1-dev build-essential libmysqlclient-dev libssl-dev libbz2-dev libpcre3-dev   libdbi-perl libarchive-zip-perl libdate-manip-perl libdevice-serialport-perl libmime-perl libpcre3  libwww-perl libdbd-mysql-perl libsys-mmap-perl yasm automake autoconf libjpeg-turbo8-dev    libjpeg-turbo8 libtheora-dev libvorbis-dev libvpx-dev libx264-dev libmp4v2-dev libav-tools mysql-client     apache2 php5 php5-mysql apache2-mpm-prefork libapache2-mod-php5 php5-cli openssh-server     mysql-server libvlc-dev libvlc5 libvlccore-dev libvlccore5 vlc-data libcurl4-openssl-dev    libavformat-dev libswscale-dev libavutil-dev libavcodec-dev libavfilter-dev     libavresample-dev libavdevice-dev libpostproc-dev libv4l-dev libtool libnetpbm10-dev    libmime-lite-perl dh-autoreconf dpatch' returned a non-zero code: 100
```
##

SteveGilvarry:
Closing as part of 2017 cleanup, not saying not valid, just no one is working on this and don't believe anyone is planning on doing so.

PR #1257, opened 2016-02-01
Fix docker file

salberin:
The current Dockerfile doesn't work. I upgraded the dockerfile to use Ubuntu Trusty Tahr and fixed some other problems in the proces. Fixes issue #1242


knnniggett:
Looks good, but can you change the date.timezone of "GMT" or "Greenwich"?

Will a /dev/shm of 2G cause issues if there isn't that much memory available on the host machine?

Perhaps we should lower it to something like 512MB, and then add an additional comment, warning the end user they will have to increase it if they are using more than a couple HD cameras.

salberin:
Will update the request... I'll add in a comment about memory.

knnniggett:
Thanks. Nice catch about cmake. I didn't even notice the dockerfile was still using autotools.

PR #1260, opened 2016-02-05
create raspbian file

kgooris:
this is a tuturial for installing zoneminder on raspberry pi

### DIFF
--- a/docs/installationguide/raspbian
+++ b/docs/installationguide/raspbian
@@ -1,0 +1,131 @@
+How to install ZoneMinder on Raspberry PI 1 or 2 with Raspbian 8.1 (Jessy) 64 bit
+==================================================================================
+
+A special Thanks to "Felixr" and "BBungle" from witch I took the initial FAQ.
+
+Prereq before starting:
+-----------------------
+ SD card created with raspbian Jessie 64bit
+ filesystem is extended.
+ the systems boots in console mode with autologin disabled (X is disabled)
+
+Log in as pi
+--------------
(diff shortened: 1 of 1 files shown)


docs/installationguide/raspbian
```diff
@@ -0,0 +1,131 @@
+How to install ZoneMinder on Raspberry PI 1 or 2 with Raspbian 8.1 (Jessy) 64 bit
```
mnoorenberghe, line 1: Drive-by nits:
s/Jessy/Jessie/
s/PI/Pi/

knnniggett:
We created the readthedocs site, because the legacy wiki at wiki.zoneminder.com has become an unmanageable mess. Consequently, we are going to prevent that from happening to our readthedocs site.

Simply copying and pasting content from wiki.zoneminder.com is not an acceptable way to get this information added into ZoneMinder's readthedocs site:
- all formatting needs to be changed to use resturcturedtext
- extra information, such as setting one's ip address and tweaking mysql, should be removed
- incorrect carry-over’s from old versions of zoneminder should be removed (editing small & shmax)
- This guide should live under the Debian guides, and generally, the guide should follow the same style and formatting as that guide: http://zoneminder.readthedocs.org/en/latest/installationguide/debian.html
- remove steps to install cambozola. This applies to all distros and should go in its own doc.
- You need to update the master index so that the guide shows up as a link in the table of contents

It you are willing, the guide basically needs to be rewritten and new pull request generated.

Issue #1263, opened 2016-02-08
jquery 1.7.2 is too old


onlyjob:
[web/skins/classic/js/jquery-1.7.2.js](../blob/master/web/skins/classic/js/jquery-1.7.2.js) is too old. Please update to 1.11.3 or above.

knnniggett:
Posting the following as a reminder to self that this might not be as easy as replacing the js file with a newer version: https://github.com/ZoneMinder/ZoneMinder/issues/785#issuecomment-95772526

onlyjob:
JQuery upgrade guide https://jquery.com/upgrade-guide/1.9/ may be useful...

connortechnology:
It probably is that easy.  We don't do much with it.

I'll pull it into my home system and see if anything breaks.

onlyjob:
No, @connortechnology, please no minification. How many times we have to go through this?
Minified files are effectively pre-compiled binaries, un-modifiable, unreadable and non-distributable in Debian (i.e. non-free). It is a bad practice to commit pre-built files to Vcs.
Moreover there are even more concerns, including security: https://wiki.debian.org/onlyjob/no-minification

Please always commit uncompressed files. If you believe in minification than you can compress on build time.

connortechnology:
Does debian not already package jquery?  Perhaps debian should use the debian package version as you do mootools?

onlyjob:
Yes we can use "libjs-jquery" (v1.11.3 at the moment) but that's not the point. It is the best to have no pre-minified files committed. Otherwise there will be maintenance overhead, repacking tarball to throw away non-free content etc. Really it is not worth it: there are serious cons and no pros...
Besides mootools-1.5.1 may be not safe to replace with mootools-1.4.5...

Issue #785, opened 2015-04-05
please update bundled jQuery


onlyjob:
Let's upgrade jquery from v1.4.2 to 1.7.2 please.
1.7.2 is the version that soon to-be-released Debian Jessie is shipping.
Official Debian packaging of Zoneminder uses system (not bundled) version of jquery.

I'm not sure if code changes will be needed but new version appears to be backwards compatible. Thanks.

knnniggett:
I think I want to let @kylejohnson respond to this.
I see on the jquery site that there are many versions to choose from.  If someone could give me a specific version number, I can create the pr.

onlyjob:
As far as I can tell jQuery 1.7.2 works just fine in Zoneminder. In fact I've never used Zoneminder with older bundled jQuery 1.4.2... I'm all for conservative upgrade, not to the latest version but to the well tested 1.7.2 for now.

knnniggett:
I noticed the new API Kyle is working on has some references to jquery in it.  I want make sure it doesn't affect that.

kylejohnson:
Any reason not to use the latest 1.11.2 version?  
The API should not be affected at all.

onlyjob:
It is good to know that jQuery maintains stable API.  
Zoneminder users in Debian rely on system jQuery (and therefore will not benefit from upgrade of a bundled jQuery). At this time Debian ship libjs-jquery/1.7.2 so naturally Zoneminder will be used with this particular version. I suggest upgrading to this particular version because it is the most tested one.

Please remember to create un-versioned symlink `jquery.js` --> `jquery-1.7.2.js` and update reference in `skins/classic/includes/export_functions.php` -- this will simplify future upgrades.

Also please never ever commit minified javascripts. They are same as pre-compiled binaries -- unreadable, unmodifiable, and non-distributable in Debian.

onlyjob:
See #809

onlyjob:
All good (merged), closing...

onlyjob:
@kylejohnson FYI [JQuery web site](https://jquery.com/download/) says:

> The jQuery 1.x line had major changes as of jQuery 1.9.0. We strongly recommend that you also use the jQuery Migrate plugin if you are upgrading from pre-1.9 versions of jQuery or need to use plugins that haven't yet been updated. Read the jQuery 1.9 Upgrade Guide and the jQuery 1.9 release blog post for more information.

I reckon we'll need more caution to upgrade to jQuery v1.9.0 and above...

PR #1270, opened 2016-02-08
Jquery1.12

connortechnology:
Works for me so far.  Just grabbed the minified jquery 1.12 and dropped it in. 

Solves #1263


knnniggett:
Heh, much nice looking than that last pr. :-)

There is symlink jsquery.js that should point to jquery-1.12.0.js

Looks like this commit introduces a minified version of jquery 1.12.0. 
@onlyjob warns against that, but I'm not sure why. https://github.com/ZoneMinder/ZoneMinder/issues/785#issuecomment-93141848

onlyjob:
There are problems: https://github.com/ZoneMinder/ZoneMinder/issues/1263#issuecomment-181432347   :-(  :-(

PR #1033, opened 2015-08-22
add warning and help text for maxfps fields

knnniggett:
fixes #754 

I don't find this solution visually elegant, and I'm open to suggestions if anyone has other ideas.

Screenshots are incoming....

NOTE that I did experiment with triggering a popup when the user clicked into the maxfps field, but that got annoying real quick.  I settled for a note, in red, appearing to the right of the box.

Also, what you see in the screenshots below does not appear for existing monitors with a type of "local' or "file" since the issue does not affect those types.

PR #1271, opened 2016-02-08
Updated alarmMaxFPS doc to note it won't work for remote streams

pliablepixels:


knnniggett:
> local camers or remote cameras in snapshot mode. 

small spelling error

> If you are using a remote stream like RTSP, MJPEG or others then the camera decides the stream fps and this parameter is ignored.

ZoneMinder does not ever ignore the contents of this field. It tries to use it in all cases, and it fails terribly when the remote device is streaming to zoneminder. 

See here for the help text that is now part of 1.29.0:
https://github.com/ZoneMinder/ZoneMinder/pull/1033

pliablepixels:
do you know where in code alarmMaxFPS is handled? I see code handling for maxFPS but I can't find any reference to where alarmMaxFPS is being referenced in any of the C code (except for a log printf)

knnniggett:
Updated via commit 3a804a0

PR #1277, opened 2016-02-09
check snprintf result before use

yumkam:
When snprintf would result in buffer overflow or other error happens, it returns value larger than specified buffer size or `-1`. If you directly add it to the pointer, bad things will happen.
Have not deeply verified, but some looks exploitable.

### DIFF
--- a/src/zm_event.cpp
+++ b/src/zm_event.cpp
@@ -100,7 +100,11 @@ Event::Event( Monitor *p_monitor, struct timeval p_start_time, const std::string
     if ( config.use_deep_storage )
     {
         char *path_ptr = path;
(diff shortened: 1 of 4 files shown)


connortechnology:
Thanks for doing this.. In general I'm happy with this except that for a lot of these cases the resulting behaviour is a little unclear.  In the early cases you just go Fatal, in the rest it is unclear what will happen.  It might help to at least document what will happen.  

Technically snprintf shouldn't ever return more than sizeof() but I'm cool with checking for it anwyays.

knnniggett:
I'm going to close this out. I think the idea is sound, but I'd rather wait until we encounter something more concrete, rather than speculative, that we really need to do something like this.

Issue #1274, opened 2016-02-09
1.29.0: pre-built file "zm_bigfont.h", no licensing information


onlyjob:
File `src/zm_bigfont.h` apparently generated from unknown source by unknown method and not re-built on build time. Licensing and copyright information is missing as well.
##

onlyjob:
@schrorg: Robin, could you please explain the license, origin of this file and how it is generated? This file should be distributed with sources. Thanks.

schrorg:
Indeed, one missing thing is the script I used to create the file. I'll make a pull request for that shortly.

And regarding the licence, the original file `src/zm_font.h` does not contain that information either, as you can see in the head of the file. So I would assume it would be GPLv2 as the rest of the code is.

onlyjob:
Fair enough, but likely this font file is derived from one of the web fonts that are most probably  distributed under different license (GPL-2+ is unusual license for fonts). We need to identify the original font to determine the license...

schrorg:
Okay, the best way would be to find "rthelen" and ask him or her. Does anybody know the author? A quick search discovered only http://en.pudn.com/downloads651/sourcecode/unix_linux/detail2648941_en.html and @rthelen.

POKKAHOH:
Excuse me, it may be better to change the method of generating a timestamps? Do not use ASKII table, but beter UTF-8 for support international characters?

schrorg:
@POKKAHOH The problem is, the addition of the timestamps to the images happen during the capture process and need to be extremely fast. One would have to use a library for that and then measure the impact of using it.

knnniggett:
fixed by #1278 so closing

PR #1278, opened 2016-02-10
Added script to generate src/zm_bigfont.h

schrorg:
This PR addresses some aspects of #1274.


onlyjob:
Sorry for being pedantic, but to avoid any doubts would it be possible to add licensing/copyright header to `mk_bigfont.pl` please? Thanks.

schrorg:
Ah. Sure! Will fix that.

PR #1279, opened 2016-02-10
Parallel filter

connortechnology:
Something to play with.

Adds a concurrent column to Filters, and then zmfilter.pl will launch a separate background process to run the filter if it is set.


connortechnology:
I give up. Use my fork if you want this feature.

PR #1280, opened 2016-02-10
1272 onvif copyright

altaroca:
This fixes copyright issues in folder onvif/ (issue 1272)


onlyjob:
Awesome. Thank you. I could only wish we had your contact email in README just in case.

PR #1285, opened 2016-02-13
Travis update to Trusty, add Clang build and up ffmpeg version

SteveGilvarry:
Converted to Trusty build environment on Travis, which meant removing libvlc5 and libvlccore5 packages, the new versions of these are dependencies of libvlc-dev and libvlccore-dev.

Updated ffmpeg to 2.8.6.

Added Clang as a compiler option, figured no harm in seeing what warnings it throws.


SteveGilvarry:
Updated to ffmpeg 3.0.
Any comments?

PR #1293, opened 2016-02-16
zm_monitor: fix overlap in memcpy buffers

kunkku:
behavior undefined by POSIX and causes a crash with FORTIFY_SOURCE


connortechnology:
While this is a good, catch... I'm not satisfied with the fix.  

msg_data is a char[256].  Would be a lot better if it was a struct status_msg, or do away with the struct status_msg and merge it with DataMsg.  We don't use DataMsg anywhere else anyways.

To anyone else trying to understand what the issue is: sizeof(status_msg.msg_data) is 256, sizeof(status_data) is something likely smaller than 256.  So we are copying more than we should.

PR #1296, opened 2016-02-16
security hardening: make static files non-writable by webuser

kunkku:


knnniggett:
Hi @kunkku 
We are no longer supporting autotools and have migrated to cmake instead.
All the autotools file will be removed shortly.

connortechnology:
Other than that, it's a great thing. Just make sure it is happening for cmake too.

knnniggett:
Obsoleted by #1308

PR #1298, opened 2016-02-17
Upgrade bundled jQuery (Closes: #1263)

onlyjob:


connortechnology:
Why jquery 1.11.3 when there is a 1.12 available?

onlyjob:
I knew you would ask that. :)
Just being super careful. I've already tried with 1.11.3 because this is the version Debian ship at the moment. Therefore 1.11.3 is much better tested. I'm not maintaining jQuery so I'm not sure what kind of concerns jQuery maintainers may have regarding newer version. However I know from experience that it is better/safer to exercise some caution about latest releases...
We can always upgrade again I suppose...

connortechnology:
Well, I don't think anyone cares, so here it is merged.

onlyjob:
Thank you. :)

PR #1300, opened 2016-02-19
Zm trigger improvements

connortechnology:
The use of croak means that the error message will only be seen if the user runs it from a command line.

So we log before croaking.

Ideally failing to open would not be fatal, and zmtrigger.pl could continue on it's other channels.


kylejohnson:
Looks good to me.

Issue #545, opened 2014-10-31
zmtrigger barf


pashdown:
1.27.99

zmtrigger eventually dies with this error:

Use of uninitialized value $nbytes in concatenation (.) or string at /usr/share/perl/5.18.2/ZoneMinder/Trigger/Channel/Handle.pm line 90.
Use of uninitialized value $nbytes in concatenation (.) or string at /usr/share/perl/5.18.2/ZoneMinder/Trigger/Channel/Handle.pm line 90.
variable is not a string at /usr/share/perl/5.18.2/ZoneMinder/Memory/Mapped.pm line 125.

CalebMuhia:
Hi, 

am also getting variable is not a string at /usr/share/perl5/ZoneMinder/Memory/Mapped.pm line 125.
i have zoneminder 1.28 build from source

zmtrigger.pl dies and does not work again till the next restart
would like to sort out the issue

Thanks

pashdown:
I had to resort to starting zmtrigger with a bash loop, so it gets restarted every time it dies.  And yes, I'm using 1.28.0.

connortechnology:
I've been tracking this on zoneminder.com too.  it is weird... something to do with the monitors being reloaded and the mmap variable getting corrupted. doesn't happen for me either, but it definitely is an ongoing problem.

CalebMuhia:
we are using video card for the capture. does it help when using ip cameras?

XenonofArcticus:
Currently happening for me with HD IP cameras.

cryo:
My speculation is this is due to a socket timeout from the IP camera (or a telnet connection to the zmtrigger port.  It still seems to want to use it until it is corrupted and crashes.  Implemented bash loop as mentioned to keep it going.

XenonofArcticus:
However, when it crashes, it does lose the event in progress, which is an undesired behavior so fixing this would be important.

knnniggett:
Putting this here, because I believe it will resolve this issue.

The following sample code from @pliablepixels is intended to fix a suspected problem with zmtrigger when a monitor is saved or modified. Zmtrigger does not update its memory handle and consequently crashes or fails when it attempts to read from the old, invalid handle.

```Perl
#!/usr/bin/perl
use ZoneMinder;
use DBI;

my @monitors = ();
my $dbh = zmDbConnect();

loadMonitors();

my $iter=1;

for (;;)
{
        foreach $mon (@monitors)
        {
               
[...]

PR #1302, opened 2016-02-21
Prevent zmtrigger from crashing when memory handles change

knnniggett:
Possible fix for #545 

When a monitor is created, deleted, or edited from the web console on a running system, the memory handles change. Zmtrigger is unaware of this change and proceeds to read the memory handle from the previous, now invalid, handle. This is my current theory to the cause of zmtigger crashing, or becoming unresponsive, over time.

This commit borrows on the idea from @pliablepixels . It attempts to detect when this condition occurs and remedy it by deleting the old and then recreating a new memory handle for the monitor in question.

Note that I have yet to experience this particular issue on my system, so I cannot verify this does indeed resolve the issue. Needs verification from someone who can duplicate this issue.

### DIFF
--- a/scripts/zmtrigger.pl.in
+++ b/scripts/zmtrigger.pl.in
@@ -135,14 +135,14 @@ use ZoneMinder::Trigger::Connection;
 my @connections;
 push( @connections,
       ZoneMinder::Trigger::Connection->new(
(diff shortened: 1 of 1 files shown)


knnniggett:
Wait a second. This isn't right. I undid some recent commits.

PR #1303, opened 2016-02-21
Prevent zmtrigger from crashing when memory handles change

knnniggett:
Let's try this again.
Possible fix for #545

When a monitor is created, deleted, or edited from the web console on a running system, the memory handles change. Zmtrigger is unaware of this change and proceeds to read the memory handle from the previous, now invalid, handle. This is my current theory to the cause of zmtigger crashing, or becoming unresponsive, over time.

This commit borrows on the idea from @pliablepixels . It attempts to detect when this condition occurs and remedy it by deleting the old and then recreating a new memory handle for the monitor in question.

Note that I have yet to experience this particular issue on my system, so I cannot verify this does indeed resolve it. Needs verification from someone who can duplicate this issue.

### DIFF
--- a/scripts/zmtrigger.pl.in
+++ b/scripts/zmtrigger.pl.in
@@ -219,7 +219,7 @@ while( 1 )
     {
         Debug( "Got input from $nfound connections\n" );
         foreach my $connection ( @in_select_connections ) {
-        {
+
             if ( vec( $rout, $connection->fileno(), 1 ) )
             {
                 Debug( "Got input from connection "
--- a/scripts/zmtrigger.pl.in
+++ b/scripts/zmtrigger.pl.in
@@ -197,6 +197,7 @@ my %spawned_connections;
 my %monitors;
 
 my $monitor_reload_time = 0;
+my $needsReload = 0;
 
 $! = undef;
 my $rin = '';
@@ -217,7 +218,7 @@ while( 1 )
     if ( $nfound > 0 )
     {
         Debug( "Got input from $nfound connections\n" );
-        foreach my $connection ( @in_select_connections )
+        foreach my $connection ( @in_select_connections ) {
         {
             if ( vec( $rout, $connection->fileno(), 1 ) )
             {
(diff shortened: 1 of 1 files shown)


connortechnology:
I can actually reproduce that here, so I will merge and test on my branch

connortechnology:
This PR does nothing to fix the problem. Sorry.

knnniggett:
No problem. What is the error message when the problem occurs?

connortechnology:
variable is not a string at /usr/share/perl5/ZoneMinder/Memory/Mapped.pm line 133.

knnniggett:
Interesting, it is failing when it is trying to detach the memory handle, rather then when it is trying to attach to the new handle.

knnniggett:
Perhaps we should check if $monitor->{MMap} is defined just before line 133.
If it isn't defined, just return.

knnniggett:
Well that won't work because that is what line 131 does. It checks first if $monitor->{MMap} is defined and non-zero

knnniggett:
@connortechnology Look at this:
https://rt.cpan.org/Public/Bug/Display.html?id=91483

Can you verify you are using a version of mmap that this fix in it?
https://github.com/toddr/Sys-Mmap/commit/a8f8b3ca4cbc08092c4881a81ffcde1d6fe6df8d

The version in the Ubuntu repo -s called 17-1, which might not be the same as what the mmap author's release called 17_1 on his github page.... usually the "-1" means package revision rather than software revision.

connortechnology:
I used to have to package sys-mmap-perl for old ubuntu distros.  I can certainly do that again, providing an up to date libsys-mmap-perl for ubuntu 14.04 and any others that need it.

connortechnology:
I have put a package up on my ppa.  Anyone subscribed to it who is on 14.04 should get the updated Sys::MMap and the variable is not a string problem should be resolved.

knnniggett:
From topics I have read on the forums, I suspect there is still another problem that has to do with expired memory handles. However, I want to wait to merge #1303 until we can get new data indicating this is still a problem.

connortechnology:
Yeah I can imagine that being a problem, so we will see.

knnniggett:
I'm loosing my marbles. I thought I closed this one out. Since no new memory handle issues have arisen, I'm going to close this out for housekeeping purposes. We can resurrect it later if something crops up.

PR #1305, opened 2016-02-24
add Netcat ONVIF compatible ptz control

knnniggett:
First implementation of the netcat, onvif compatible, ptz control script.

Reminder that, to get this script into an existing system, one should execute "zmcamtool.pl --import" from the command line.


connortechnology:
Looks good to me.  Reminds me though that a lot of code could be pulled into the parent object. I really need to go through all the PTZ scripts and pull out the common code.

Issue #1291, opened 2016-02-16
Upgrading cake-php that we bundle - APIs break with PHP7


pliablepixels:
ref: https://forums.zoneminder.com/viewtopic.php?f=34&t=24306
Looks like PHP7 is bundled with some systems and that's incompatible with CakePHP 2.5-RC1 which is what we bundle inside api/lib/Cake

@kylejohnson - I would need your help here - I tried the git cake upgrade tool but that threw a bunch of errors. So far I tried replacing 2.5 with 2.8 doing a simple
```
cd /usr/share/zoneminder/api/lib
mv Cake Cake_old
git clone -b 2.x git://github.com/cakephp/cakephp.git
ln -s cakephp/lib/Cake/ .
```
That works with PHP 5.4 but I don't have PHP7 to test. Specifically, I don't know if it will work if we don't also upgrade    the app dir via the tool.

onlyjob:
Related: #814

onlyjob:
https://secure.php.net/manual/en/migration70.php

By the way ZM interface works fine with PHP7 as far as I can tell. I did not test API though and I'm not sure how to do it. Do we have a test application to probe basic API functions?

knnniggett:
Resolved by #1306

Issue #1227, opened 2016-01-05
CakePHP version not compatible with PHP7


w41ru5:
My system upgraded to PHP7 I have ZoneMinder 1.29 running. I was wanting to try out the API layer, but I get an error with CakePHP:
`Error: Cannot use 'String' as class name as it is reserved`
`File: /srv/http/zoneminder/api/lib/Cake/Utility/String.php`
`Line: 25`

CakePHP (https://github.com/cakephp/cakephp/issues/7573) maked this as an issue they have resolved in versions 3+

onlyjob:
IMHO this is quite a problem as upcoming Debian release will be shipped without PHP-5 and only with PHP-7: https://lists.debian.org/debian-devel-announce/2016/01/msg00002.html

Related to #1291.

onlyjob:
The simple solution would be to upgrade CakePHP to 2.8 because "[CakePHP 2.8 is compatible with, and tested against PHP7][1]"

[1]: https://github.com/cakephp/cakephp/releases/tag/2.8.0

knnniggett:
Resolved by #1306

wansco:
This probably bears a new thread, but this is where I ended up, so I figured I'd at least post it here.

I had some API issues with a clean install on ubuntu 16.04 related to cakephp. The solution was to add the following lines to /etc/apache2/conf-enabled/zoneminder.conf in the directory www section

  AllowOverride All
  Require all granted

(Note: this isnt related to the original post, but it might be useful to someone)

PR #1306, opened 2016-02-25
Update to CakePHP 2.8.0 copy in lib folder

SteveGilvarry:
Path of least resistance for PHP7, was to move to CakePHP 2.8.0. Was just as simple as replacing the lib folder with the new CakePHP version.
Tested on Ubuntu 14.04 PHP 5.5 confirmed version and monitors json both working fine.
Upgraded to PHP 7, restarted apache, fixed new /etc/php/apache2/php.ini file for correct timezones.
Checked version and monitors json post upgrade and working fine and confirmed using /web/api that I was now running CakePHP 2.8.0.

@knnniggett want to test this on centos 6 to confirm it works with php 5.x on there?


SteveGilvarry:
Resolves #1291 and #1227

knnniggett:
What do you guys think of the possibility of making CakePHP a submodule?
It would certainly simplify the process of changing to a different version.
Of course, that means we will have nested submodules.

Because the API uses php traits, I can't use the API on CentOS 6.
I've gone ahead and packaged the current versions of cakephp and crud with the rpm, even though the API doesn't work. I guess I'm hopeful that someday we might be able to get some level of usage out of it.

Anyway, because the API does not work on CentOS 6, I'm not sure how I can test a newer version of CakePHP on this distro. Thoughts?

SteveGilvarry:
If you can make a submodule take just one folder of the other project it could work, but then again the next jump to 3 won't just be a swap.

Interesting that it doesn't work as CakePHP is PHP 5.3+ and looks like Centos 6 was 5.3.3. The only thing Cake does is the API so if it doesn't work now then no point trying to test it with an update

knnniggett:
It isn't Cake, but Crud, which uses php traits, which were implemented in php 5.4. The API needs both to work, which means we must follow the most stringent requirements.

PR #1308, opened 2016-02-26
Autotools delete

SteveGilvarry:
Pretty sure everyone was onboard with the move to cmake now. Just putting this out there so we no longer need to maintain these files.


knnniggett:
I'm onboard with this. My only concern is timing. We plan to create a 1.30 release candidate on Monday, and then release 1 week later. I'd suggest we merge this, following the creation of the 1.30 release candidate,

kylejohnson:
I am OK with getting this in for the 1.30 release.  The PR made it in in time, and is passing our tests.

Or am I missing something about our new process (did we decide on a merge window)?

knnniggett:
Just a preference. I'd feel more comfortable if this merge was in master a few weeks before we send it out into a release. This would also give us the opportunity to warn our user base one more/last time in the release notes about the pending removal.

PR #1311, opened 2016-02-27
typo in function name

pliablepixels:
will not render image in this case due to typo


Linwood-F:
Thank you for fixing my sloppy typing (and testing).

PR #1312, opened 2016-02-28
add redhat instructions to readthedocs

knnniggett:
This is the initial commit to migrate the installation instructions for redhat into readthedocs.

This is a work in progress.

TO-DO:
- formatting
- delete old redhat instructions from README.md
- also delete old debian/ubuntu instructions from README.md
- refer the reader to the appropriate readthedocs link


knnniggett:
I think I need another set of eyes to look at this.
Also, I'm using http://rst.ninjs.org/ as a wysiwyg editor, but it seems to format the text differently than what can be seen on our own redthedocs site. Any suggestion for a different restructured text edit (that can be found in package or online form) would be helpful.

knnniggett:
I am going to self merge this documentation-only pull request so I can see the result on readthedocs and create a follow-up pull request if there are any formatting issues.

PR #1314, opened 2016-03-01
add comment/warning for ZM_SERVER_HOST

knnniggett:
I've read at least three instances of forum users assigning a value to ZM_SERVER_HOST while at the same time not having any intention of using multi-server. This of course breaks their system. This is my attempt to fix stupid.

### DIFF
--- a/zm.conf.in
+++ b/zm.conf.in
@@ -46,5 +46,9 @@ ZM_DB_USER=@ZM_DB_USER@
 # ZoneMinder database password
 ZM_DB_PASS=@ZM_DB_PASS@
 
-# Host of this machine
+# Do NOT set ZM_SERVER_HOST if you are not using Multi-Server
+# You have been warned
+#
+# The name specified here must have a corresponding entry
+# in the Servers tab under Options
 ZM_SERVER_HOST=


connortechnology:
I kinda thought that the code would gracefully handle something being set here, but not in servers tables... I will check on that.

knnniggett:
Just for reference, here are the three relevant threads in the forums:
https://forums.zoneminder.com/viewtopic.php?f=32&t=24337#p93004
https://forums.zoneminder.com/viewtopic.php?f=34&t=24292
https://forums.zoneminder.com/viewtopic.php?f=34&t=24310

kylejohnson:
This was my experience as well.

connortechnology:
So... I don't think I need to make any changes.  It logs the error correctly and then behaves as if nothing was set.

knnniggett:
Checkout the previously referenced forums posts.
The condition occurs when the user sets ZM_SERVER_HOST, but does not populate the serves tab.
Streaming does not work in this situation. This makes sense why steaming doesn't work though. ZoneMinder sees a half-baked multi-server configuration and doesn't know what to do.

I can't think of anything better to do than to warn the user the configuration isn't right.

PR #1315, opened 2016-03-01
Thumbnail on console

connortechnology:
This is actually a conversion of the console from using tables, to using divs.  

It will eventually include showing a thumbnail on the console, and condensing the info into a multi-line format.


connortechnology:
How do we feel about removing the mark column, and requiring someone to open up the monitor and click Delete instead?

kylejohnson:
Yeah, way too many commits.  It looks like this PR has, among other things, storage areas, along with the console view update.  It has to be just the latter.

connortechnology:
The reason for all the commits is that it is based off my storageareas branch. When they get merged, this will become clearer.  I don't intend for this to be used by anyone anytime soon.  

So far I have only done the flat theme.

It should be trivial to make an empty name still clickable. I might be tempted to make the entire line clickable.

PR #1316, opened 2016-03-01
change multi-server error message text

knnniggett:
This is an attempt to clarify why this message appears. Thoughts?


connortechnology:
My problem is that one my test system, streaming work just fine in this situation.

PR #1317, opened 2016-03-01
fix setButtonStates.  The checkboxes are an array.

connortechnology:
I found that the edit and delete buttons didn't become enabled... this fixes that.  

checkboxes form an array, so setButtonStates must check in the array as well.

### DIFF
--- a/web/skins/classic/views/js/console.js
+++ b/web/skins/classic/views/js/console.js
@@ -11,18 +11,19 @@ function setButtonStates( element )
                 if ( checked++ > 1 )
                     break;
             }
-        }
-    } else if ( form.elements[i].length )
-    {
-		for( var j = 0, j_length = form.elements[i].length; j < j_length; j += 1 ) {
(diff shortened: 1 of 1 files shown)


knnniggett:
I'm pretty sure I've noticed this, but I can't remember where.

connortechnology:
When you check one of the checkboxes, does the line get highlighted?  Cuz it should. But it doesn't for me without this PR.

knnniggett:
Well I think so. See screenshot.

screenshot from 2016-03-03 17-13-56

connortechnology:
Yeah that looks right.  I don't really understand how it could be working though.

More investigation before merging.

knnniggett:
Does your web browser debugging console reveal any clues?

connortechnology:
I'll work on this tomorrow.

connortechnology:
I'm not able to reproduce the issue I was having that day.  So this might be unneccsary.

PR #1319, opened 2016-03-02
use Digest::SHA instead of Digest::SHA1.  API appears to be the same

connortechnology:
Works on ubuntu, to at least allow zmonvif-probe.pl to run.  

Please test.


knnniggett:
I need to test this on redhat, but right now I am in package dependency hell.

knnniggett:
Yes, this builds and installs in redhat. It changes the package dependency from perl-Digest-SHA1 to perl-Digest-SHA.

PR #1320, opened 2016-03-02
change TopicType to TopicNamespaceType

knnniggett:
Would like to wait until we hear from @altaroca before merging.

This was preventing ZoneMinder rpm's from installing onto the target system.

### DIFF
--- a/onvif/proxy/lib/ONVIF/Media/Types/TopicNamespaceType.pm
+++ b/onvif/proxy/lib/ONVIF/Media/Types/TopicNamespaceType.pm
@@ -63,7 +63,7 @@ sub __get_attr_class {
 }
 
 
-use base qw(ONVIF::Media::Types::TopicType);
+use base qw(ONVIF::Media::Types::TopicNamespaceType);
 # Variety: sequence
 use Class::Std::Fast::Storable constructor => 'none';
 use base qw(SOAP::WSDL::XSD::Typelib::ComplexType);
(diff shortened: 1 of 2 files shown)


altaroca:
Hi Andrew,
I have a TopicType.pm module in onvif/proxy/blib/lib/WSNotification/Types . 
These are all generated files. I think it is a bug in the generator. If media.wsdl includes wsnt-1.wsdl then the generated types in wsnt-1.wsdl should be in that package's namespace. However it seems TopicNamespaceType was generated in Media and in WSNotification. While TopicType is just in WSNotification.
Therefore I would rather change 
use base qw(ONVIF::Media::Types::TopicType)
to
use base qw(WSNotification::Types::TopicType);

knnniggett:
@altaroca Thanks for the feedback. However, the TopicType.pm module does not exist in our source code nor does a WSNotification folder exist. Are we missing something?

Is this related to the files that were recently removed due to licensing issues??

altaroca:
In my repository I have more ONVIF code such as motion detection on the camera. I was waiting for the first pull request to gain some more user acceptance and stability. Then I can publish the rest. 
I tried for a clean cut. Now it seems there is some dependency. I propose to publish all generated proxies to avoid more dependencies popping up. What do you think?

knnniggett:
Go ahead and generate a PR with everything else. I can rather quickly find out if that solves all our problems, or if any new problems are created. 

I guess I'll have to see how you chose to tie Onvif motion detection into the UI, but perhaps we could leave the additional UI changes out? That might allows us to bring in everything on the backend without the end user being affected. If that turns into a lot of extra work, then lets just bring in everything.

Note that I've got all the new, third party, Perl module dependencies uploaded to zmrepo.zoneminder.com so users of redhat distros do not need to install them via any other method (cpan).

knnniggett:
Obsoleted by #1338

Issue #1268, opened 2016-02-08
1.29.0: ffmpeg deprecations


onlyjob:
Spotted the following deprecations during compilation of ZoneMinder with ffmpeg-2.8.5:

```
/build/zoneminder-1.29.0/src/zm_mpeg.cpp:681:12: warning: 'AVCodecContext::coded_frame' is deprecated [-Wdeprecated-declarations] 
    if ( c->coded_frame->key_frame ) 
            ^ 
In file included from /build/zoneminder-1.29.0/src/zm_ffmpeg.h:110:0, 
                 from /build/zoneminder-1.29.0/src/zm_mpeg.h:23, 
                 from /build/zoneminder-1.29.0/src/zm_mpeg.cpp:25: 
/usr/include/x86_64-linux-gnu/libavcodec/avcodec.h:3024:42: note: declared here 
     attribute_deprecated AVFrame *coded_frame; 
                                          ^ 
/build/zoneminder-1.29.0/src/zm_mpeg.cpp:681:12: warning: 'AVCodecContext::coded_frame' is deprecated [-Wdeprecated-declarations] 
    if ( c->coded_frame->key_frame ) 
            ^ 
In file included from /build/zoneminder-1.29.0/src/zm_ffmpeg.h:110:0, 
                 from /build/zoneminder-1.29.0/src/zm_mpeg.h:23, 
                 from /build/zoneminder-1.29.0/src/zm_mpeg.cpp:25: 
/usr/include/x86_64-linux-gnu/libavcodec/avcodec.h:3024:42: note: declared here 
     attribute_deprecated AVFrame *coded_frame; 
                                          ^ 
/build/zoneminder-1.29.0/src/zm_mpeg.cpp:681:12: warning: 'AVCodecContext::coded_frame' is deprecated [-Wdeprecated-declarations] 
    if ( c->coded_frame->key_frame ) 
            ^ 
In file included from /build/zoneminder-1.29.0/src/zm_ffmpeg.h:110:0, 
                 from /build/zoneminder-1.29.0/src/zm_mpeg.h:23, 
                 from /build/zoneminder-1.29.0/src/zm_mpeg.cpp:25: 
/usr/include/x86_64-linux-gnu/libavcodec/avcodec.h:3024:42: note: declared here 
     attribute_deprecated AVFrame *coded_frame; 
                                          ^ 
```
##

PR #1321, opened 2016-03-03
FFMPEG 3.0 deprecation fixes

SteveGilvarry:
So need someone on some older versions of FFMPEG to try this. I set to not ready for merge until maybe @knnniggett and @connortechnology can review and remove the label.

I swapped out all the usages of avpicture with their av_image equivalents.

Only one change that I was not 100% on and that was the deprecation of coded_frame->key_frame. We use it to determine if the packet coming back from encoding is a Key frame, but the examples I found that stopped using it were not setting the Key Frame Flag on the packet, the only idea I have is it is being set at encoding as you would expect. I have just commented it out until we can test and discuss. This would address #1268.


SteveGilvarry:
Removing the tag as I believe after discussions that my concerns on the coded_frame deprecation are alleviated.

src/zm_ffmpeg.h
```diff
@@ -47,6 +47,12 @@ extern "C" {
 #else
 #include <libavcodec/opt.h>
 #endif
+    
+#if LIBAVUTIL_VERSION_CHECK(54, 6, 0, 6, 0)
+#include <libavutil/imgutils.h>
+#else
```
connortechnology: This is a bit messy.  I think instead of the else, since we are including avutil.h below here, we should do nothing.
I have also seen some comments on stack overflow about needing to include avutil.h before imgutils.h... not sure how accurate that is.

SteveGilvarry: Making me stretch to remember why it ended up like this.
1. ffmpeg/avutil.h <> libavutil/avutil.h. But actually it is already included above so it can go.
2. Didn't need imgutils.h until this version when avpicture_fill was deprecated in avcodec, so just the if will be fine.
imgutils.h includes avutil.h.
Lucky I need to fix spacing anyway.

Issue #1098, opened 2015-10-08
1.28.2 zmfilter.pl syntax error


mdetweiler:
root@ubuntu-video:~/bin# /usr/bin/zmfilter.pl 
Bareword found where operator expected at /usr/bin/zmfilter.pl line 828, near "Error"
	(Missing semicolon on previous line?)
syntax error at /usr/bin/zmfilter.pl line 828, near "Error"
Execution of /usr/bin/zmfilter.pl aborted due to compilation errors.
root@ubuntu-video:~/bin# vi /usr/bin/zmfilter.pl

root@ubuntu-video:~/bin# /usr/bin/zmfilter.pl 
Name "Net::SFTP::Foreign::debug" used only once: possible typo at /usr/bin/zmfilter.pl line 818.

knnniggett:
I was hoping @connortechnology would respond to this.
Zoneminder 1.28.2 only exists in his Ubuntu ppa. I don't have a copy of that to see what you've got. 
The copy of zmfilter.pl in master is clearly different than the one you are using:
https://github.com/ZoneMinder/ZoneMinder/blob/master/scripts/zmfilter.pl.in

You can ignore the message about Net::SFTP::Foreign::debug. It is just a warning and can be ignored. I don't think the author of Net::SFTP::Foreign wrote a debug module, so we simply need to disable the check for debug when using sftp as opposed to plain ftp. We are aware of it and it will get fixed eventually.

connortechnology:
I think the issue is a missing ; after the eval above that line.  When doing block evals I think you need a semi-colon after the end of the block;

So on line 827 change it to 

}; # end eval

mdetweiler:
Yes...  I had fixed it that way when I saw this issue on my system.

knnniggett:
Can someone with 1.28.2 compare their copy of zmfilter.pl to the copy in master and verify whether or not this problem hasn't already been fixed? 

Line 907 looks suspect:
https://github.com/ZoneMinder/ZoneMinder/blob/master/scripts/zmfilter.pl.in#L907

connortechnology:
The issue does not exist in master.  It exists only in 1.28.2.  Although I am going to look at merging the code into master.  The Net::SFTP stuff uses die and croak, so it can kill zmfilter.pl.  So we need to put it into an eval to prevent the failure of one filter from killing all filters.

knnniggett:
@mdetweiler If you are still around, can you test #1323 ?
[...]

PR #1323, opened 2016-03-04
fix missing Net::SFTP::Foreign::debug warning

knnniggett:
With this commit, Perl no longer complains about a missing debug module, which does not exist in Perl, but I don't currently have an environment set up to test SFTP transfers.

This modification follows the Perl documentation for this module and passes "-v" to the "more" key when Debug is turned on.


connortechnology:
I'd be fine to merge this.  Would be nice for someone to test it but I doubt that anyone will test it until we merge it.

knnniggett:
@connortechnology pssst.... merge me, merge me

knnniggett:
Oh right, I guess I should have stated this fixes #1098 . @mdetweiler posted in that issue he tested this pr with good results.

PR #1326, opened 2016-03-09
disable save button when nothing slected. fixes #1322

### LINKED ISSUE #1322: Add input checking / validation when creating a group of monitors.
Currently, it allows you to attempt to create a group, then press SAVE, but if you do not have any monitors selected it gives an SQL error. I think there should be some kind of validation that alerts the user that no monitors are selected, deactivates the SAVE button, or allow it to create an empty names group to add monitors to later.

`SQL-ERR 'SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'MonitorIds' cannot be null', statement was 'INSERT INTO Groups SET Name=?, MonitorIds=?'`

knnniggett:
Fixes #1322 on my machine. Please test.

### DIFF
--- a/web/skins/classic/views/group.php
+++ b/web/skins/classic/views/group.php
@@ -67,7 +67,7 @@ xhtmlHeaders( __FILE__, translate('Group')." - ".$newGroup['Name'] );
         if ( visibleMonitor( $monitor['Id'] ) )
         {
 ?>
-                  <option value="<?php echo $monitor['Id'] ?>"<?php if ( array_key_exists( $monitor['Id'], $monitorIds ) ) { ?> selected="selected"<?php } ?>><?php echo validHtmlStr($monitor['Name']) ?></option>
+                  <option value="<?php echo $monitor['Id'] ?>"<?php if ( array_key_exists( $monitor['Id'], $monitorIds ) ) { ?> selected="selected"<?php } ?> onclick="configureButtons( this );"><?php echo validHtmlStr($monitor['Name']) ?></option>
 <?php
         }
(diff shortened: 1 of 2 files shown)


connortechnology:
Um.. what is element.selected?  This is not standard DOM as far as I know.  Is it a jquery thing?

Seems to me it should be element.selectedIndex and we should be able to test for >= 0 (-1 means no selection).

knnniggett:
I'm just winging this php stuff.

I noticed that the selected property was being set for each monitor here:
https://github.com/ZoneMinder/ZoneMinder/blob/master/web/skins/classic/views/group.php#L70

I'm not entirely sure that "if" condition even needs to be in there.

connortechnology:
Oh... the option element has a selected property... but not the select element.
if ( element.selectedIndex >= 0 ) {

can you try that and see if it works?

knnniggett:
sure. give me a sec.

knnniggett:
Ah no, that does not work. The button never enables when I use `if ( element.selectedIndex >= 0 )` or even `if ( element.selectedIndex )`

What does work, is removing that `if` condition altogether.

knnniggett:
The behaviour we are looking for seems to be similar to this:
https://github.com/ZoneMinder/ZoneMinder/blob/master/web/skins/classic/views/js/monitorpreset.js#L10

Since that is not checking element.selectedIndex , I'll just go ahead and remove the the `if` condition.

connortechnology:
Yeah, after reading modern DOM docs... select.value should work just fine.

PR #1328, opened 2016-03-10
use make "-f" parameter instead of "--makefile"

knnniggett:
The version of `make` on FreeBSD 10.2 does not support the `--makefile` parameter. Fortunately, both Linux and FreeBSD support the `-f` parameter instead.

Symptoms include a build failure citing a missing "pure_install" target.


knnniggett:
woo hoo!

Issue #1288, opened 2016-02-14
ZM doesnt compile in FreeBSD


rajil:
I tried compiling version 1.29 in a FreeBSD 10.2 jail but it failed.

./configure --prefix=/usr/local/zoneminder --with-libarch=lib64 --with-mysql=/usr/local --with-mariadb=/usr/local --with-polkit=/usr/local
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... ./install-sh -c -d
checking for gawk... no
checking for mawk... no
checking for nawk... nawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking build system type... amd64-unknown-freebsd10.2
checking host system type... amd64-unknown-freebsd10.2
./configure: 3453: Syntax error: word unexpected (expecting ")")

cat config.log
```
This file contains any messages produced by compilers while
running configure, to aid debugging if configure makes a mistake.

It was created by zm configure 1.29.0, which was
generated by GNU Autoconf 2.69.  Invocation command line was

  $ ./configure --prefix=/usr/local/zoneminder --with-libarch=lib64 --with-mysql=/usr/local --with-mariadb=/usr/local --with-polkit=/usr/local

## --------- ##
## Platform. ##
## --------- ##

hostname = zm
uname -m = amd64
uname -r = 10.2-RELEASE-p12
uname -s = FreeBSD
uname -v = FreeBSD 10.2-RELEASE-p12 #4 r289129M: Sun Feb  7 08:12:08 CST 2016     rs4@nas:/usr/obj/usr/src/sys/MOREROUTINGTABLES 

/usr/bin/uname -p = amd64
/bin/uname -X     = unknown

/bin/arch              = unknown
/usr/bin/arch -k       = unknown
/usr/convex/getsysinfo = unknown
/usr/bin/hostinfo      = unknown
/bin/machine           = unknown
/usr/bin/oslevel       = unknown
/bin/universe          = unknown

PATH: /sbin
PATH: /bin
PATH: /usr/sbin
PATH: /usr/bin
PATH: /usr/games
PATH: /usr/local/sbin
PATH: /usr/local/bin
PATH: /home/zm/bin


## ----------- ##
## Core tests. ##
## ----------- ##

configure:2603: checking for a BSD-compatible install
configure:2671: result: /usr/bin/install -c
configure:2682: checking whether build environment is sane
[...]

PR #1330, opened 2016-03-11
add freeBSD cmake target distro

knnniggett:
Creates a set of standardized build options for freeBSD to simplify the cmake process and to achieve consistency between different builds.

Calling cmake now looks like this:
```cmake -DZM_TARGET_DISTRO="freeBSD" .```

The cmake values I have chosen in  b1d3be2 are based on how we build ZoneMinder for other distros. I recommend we avoid creating new target folder names, unique to freeBSD, unless there is some freeBSD documentation that indicates we should.

Where applicable, I have used a prefix of `/usr/local` since a set of packaging scripts has not been developed for freeBSD.

@kylejohnson Looking for feedback


kylejohnson:
Tagging #1288

PR #1333, opened 2016-03-11
Quick fix perl path

kylejohnson:
This addresses an issue noted in #1288


knnniggett:
This looks good for a short term fix.

This touches are a larger issue, which I have had in the back of my mind but failed to write down.

Distros that place executables into odd folders (odd by major Linux distro standards) aren't compatible with the default Perl paths hard coded here. Solaris is one example. 

Unless I can think of something better, I plan to create a make macro, which would have a default value of `/bin:/usr/bin:/usr/local/bin`. Most would not need to change this, but this does provide a way for the few affected to be able to change this at build time when they have to. I'm not sure there is a good way to auto-detect the right values so it would have to be changed manually.

There. I finally wrote it down.

PR #1334, opened 2016-03-12
Create CMS.php

amshaikh:
This utility will let the user view multiple monitors synchronously for a selected date. It helps tracing and reviewing any incident on a location.

### DIFF
--- a/web/CMS.php
+++ b/web/CMS.php
@@ -1,0 +1,498 @@
+<?php
+if (shell_exec('which gawk')=="")
+{
+?>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>zmCMS</title>
+  </head>
+  <body>
+    <h2> This utility requires 'gawk' to be installed on the server. </h2>
+  </body>
+</html>
+<?php	
+	die();
+}
+
+require_once( 'includes/config.php' );
+
+$qq = dbQuery("select * from Monitors");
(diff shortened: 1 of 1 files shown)


knnniggett:
Thank you for taking the time to create this pull request, but it does not currently meet our standards on a number of issues.

Rather than get into the details of what should change in the code just yet, first please provide us a better description of what this looks like. Screenshots would be most helpful here.

I'm particularly interested in knowing/seeing what this does that the following new feature doesn’t already do: https://github.com/ZoneMinder/ZoneMinder/pull/961

pliablepixels:
Gave it a try - nice job. A few notes:

a) There is a larger effort going on here https://github.com/ZoneMinder/ZoneMinder/pull/961 that does what you have done and some more - take a look and see if you want to contribute there

b) Its a better idea to put CMS inside the skins directory so its routed via index.php and security is applied

connortechnology:
I think we should insist on valid html5.

Also I'd like to see a lot of the table structure generated by 1 or 2 lines of code. 

Next parsing directory structure to get the images etc is bad. Once my storage areas code gets integrated there will be object methods to do a lot of the work.  So I see this code changing considerably.

knnniggett:
I'm closing this out since @amshaikh has not responded to our questions.
If that changes then we can reopen this.

Issue #1335, opened 2016-03-12
api security additions for StatesController


pliablepixels:
I will submit a PR that extends the existing api security to States controller so system permissions are honored. As of today, states controller does not implement security. Also, there is a typo in the monitor controller that I will fix.

PR #1336, opened 2016-03-12
api security enhancements issue #1335

pliablepixels:
address issue #1335

### DIFF
--- a/web/api/app/Controller/MonitorsController.php
+++ b/web/api/app/Controller/MonitorsController.php
@@ -97,7 +97,7 @@ public function beforeFilter() {
 
 			if ($this->Session->Read('systemPermission') != 'Edit')
 			{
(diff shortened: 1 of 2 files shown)


connortechnology:
Looks good to me

knnniggett:
Well, if this looks good to @connortechnology from a sunny beach in Florida, then it looks good to me too!

In other areas of ZoneMinder when we check permission, we simply set the view to "error" and then return. This code calls UnauthorizedException and then returns. Questions come to mind as to why we appear to be doing something different here. Is it because of cakephp? Should we update the rest of our php to use UnauthorizedException?  Anyway, I just want to know for the sake of knowing. Feel free to explain it offline if you feel that is more appropriate.

connortechnology:
It's a cakephp thing.

PR #1337, opened 2016-03-13
Couple quick console / login updates

kylejohnson:
The console view gets more eyeballs on it than any of the others, so I wanted to give it a little bit of quick love.


kylejohnson:
2016-03-12-184410_1333x626_scrot
2016-03-12-184356_1333x626_scrot

connortechnology:
That's way too many digits of precision in the load.

kylejohnson:
Come on, don't you like precision?  
It's got to be a freebsd-related issue.  I didn't modify the load function.

kylejohnson:
2016-03-15-124604_1902x1043_scrot

kylejohnson:
OK, going with this for the options view:
2016-03-15-132322_1902x1043_scrot

If I kept the tabs on top, and the screen isn't significantly wide, they'd collapse into two rows, breaking the 'tab' effect.

knnniggett:
The layout shown in the screenshots looks aweseom.
I love the sidebar. How hard would it be to be able to collapse or minimize the bar?

While I am absolutely all for giving ZoneMinder a facelift, the timing of this concerns me.

This pr makes a significant change to front end with no option of falling back to the old without rolling back to a previous version of ZoneMinder. While I don't necessarily disagree with this approach, I think it does justify a lengthy draft release period to allow others to test it. We would need to test these changes against all three css skins. 

In any case, at our next meeting, lets discuss revising our release strategy for next release. If we do agree to pushing the release date back to incorporate this change, I think it also makes sense to merge @SteveGilvarry 's changes to remove the widgets we said we were going to remove.

connortechnology:
Yeah I really like this.  

I agree we need to get our release out, and then we can merge this immediately.

connortechnology:
I don't know about the load/etc at the bottom instead of top.  When you have a lot of cameras it gets lost, and I kinda think that it is important information.

On the options, I'm not wild about the centered align.   It's ok in the left column, but for the available values I feel that they should be left-aligned.

connortechnology:
um... sorry they are left-aligned in your screen shots... but not in my resulting build.  Might not be your fault.

connortechnology:
for some reason I don't seem to be able the change the skin css anymore. It doesn't save. Also the # of Monitors on the nav bar has different style.  Since I can't change to classic css I can't test there.

connortechnology:
line 103 of options.php needs an if(window.opener){ around the window.opener.location.reload();}.  Since we are not in a popup anymore, the window.opener is null.

kylejohnson:
Good catches @connortechnology.  I've got some more styling changes for Options to push up, and then I say we merge.

kylejohnson:
2016-04-23-225000_1141x626_scrot

kylejohnson:
Currently working on unbreaking the monitor view.  Having issues with changing tabs and the validateForm function being called on each tab click.

knnniggett:
hmmm... not sure we should leave this one open. What do we want to do with this?

connortechnology:
This is a work in progress. I merged it into storage areas but it took a lot of work to make it useable.

kylejohnson:
Closing this out in favor of #1665

PR #1338, opened 2016-03-13
Added Proxies for WSNotification and ONVIF-Analytics

altaroca:
@knnniggett 
Sorry for the delay. PR for issue 1320. Let's see if this solves more problems than it introduces.


knnniggett:
Yes, this works. I was able to build and install rpm's for centos and fedora. The onvif probe is  returning a fault of some sort on the command line, but I can look into that later. I'll create a new issue for it, if I can't figure it out.

Issue #1346, opened 2016-03-19
ONVIF probing camera not detected


altaroca:
(forwarded on behalf of another user)
The ONVIF probing window opens and displays properly, but none of my cameras are discovered. Running 
`zmonvif-probe.pl -v probe` 
in terminal gives me the following output: 
`Probing for SOAP 1.1
Received message:
ARRAY(0x252ef50)
Error deserializing message:
<Fault xmlns="http://schemas.xmlsoap.org/soap/envelope/"><faultcode>SOAP-ENV:Server</faultcode><faultstring>Error deserializing message:
not well-formed (invalid token) at line 1, column 5, byte 5 at /usr/share/perl5/SOAP/WSDL/Expat/Base.pm line 79.
 at line 1 at /usr/share/perl5/SOAP/WSDL/Expat/Base.pm line 82.
.
Message was:
ARRAY(0x252ef50)</faultstring><faultactor>urn:localhost</faultactor></Fault>
Received message:
HASH(0x252f028)
Error deserializing message:
<Fault xmlns="http://schemas.xmlsoap.org/soap/envelope/"><faultcode>SOAP-ENV:Server</faultcode><faultstring>Error deserializing message:
not well-formed (invalid token) at line 1, column 4, byte 4 at /usr/share/perl5/SOAP/WSDL/Expat/Base.pm line 79.
 at line 1 at /usr/share/perl5/SOAP/WSDL/Expat/Base.pm line 82.
.
Message was:
HASH(0x252f028)</faultstring><faultactor>urn:localhost</faultactor></Fault>
Probing for SOAP 1.2
Received message:
<?xml version="1.0" encoding="UTF-8"?>
<SOAP-ENV:Fault xmlns:SOAP-ENV="http://www.w3.org/2003/05/soap-envelope" 
 xmlns:SOAP-ENC="http://schemas.xmlsoap.org/soap/encoding/" 
 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
 xmlns:xsd="http://www.w3.o rg/2001/XMLSchema" 
 xmlns:wsa="http://schemas.xmlsoap.org/ws/2004/08/addressing" 
 xmlns:wsdd="http://schemas.xmlsoap.org/ws/2005/04/discovery" 
 xmlns:chan="http://schemas.microsoft.com/ws/2005/02/duplex" 
 xmlns:wsa5= "http://www.w3.org/2005/08/addressing" 
[...]

PR #1347, opened 2016-03-19
ONVIF: Tenvis TH661 support

jburgess777:
These changes allow zmonvif-probe.pl to detect and determine the correct parameters for a Tenvis TH661 camera.

PR #1349, opened 2016-03-19
Improved Onvif discovery #1346

altaroca:
I generated proxies for  WSDiscovery-1.0 instead of  WSDiscovery-1.1 . This should improve onvif discovery in many cases. Because the onvif specs cite WSDiscovery-1.0 . 
Also improved reference handling in zmonvif-probe.pl as suggested by @knnniggett in #1346 .


onvif/scripts/zmonvif-probe.pl
```diff
@@ -95,7 +95,9 @@ sub deserialize_message
 
 sub interpret_messages
 {
-  my ($svc_discover, @responses, %services) = @_;
+  my ($svc_discover, $responses_ref, $services_ref) = @_;
+  my @responses = @$responses_ref;
+  my %services = %$services_ref;
```
jburgess777: Are you sure that this doesn't create a copy of the hash, preventing it from being used to suppress duplicate responses between the 1.1 and 1.2 discovery phases?

jburgess777:
I think there is an additional issue with the references because the two calls to interpret_messages() are different:
``` 
interpret_messages($svc_discover, \@responses, \%services); 
interpret_messages($svc_discover, @responses, \%services);
```
I took a slightly different approach to fixing these problems in #1347.

altaroca:
You have fixed some problems I did not see. So I just removed my commit. Your commit should cover it all.

knnniggett:
@altaroca After trying building and testing this, I just ran into a new Perl dependency on Class::Std. Surprisingly, rpmbuild did not automatically find it this time. 

Adding another dependency is not in itself a big deal, but we've already got a dependency on Class::Std::Fast, which is a similar but different module. If there is a good reason to depend on both of those modules?  I'm fine if there is. If there is not, can we easily change something so that we only need one of those modules?

jburgess777:
It looks like perl-Class-Std-Fast introduced the dependency on perl-Class-Std so it should probably be added to its Requires:

Can't locate Class/Std.pm in @INC (you may need to install the Class::Std module) (@INC contains: /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 .) at /usr/share/perl5/vendor_perl/Class/Std/Fast.pm line 19.

knnniggett:
The perl-class-std-fast package does not depend on perl-class-std:
```
[abauer@bauerhaus x86_64]$ rpm -qpR perl-Class-Std-Fast-0.0.8-1.el7.centos.noarch.rpm
perl(Carp)
perl(Storable)
perl(strict)
perl(version)
perl(warnings)
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
```
Rather than just knee-jerk add the dependency, I'm attempting to investigate why. This dependency creeped in today, and at the time I was wondering if @altaroca knows anything about it. If not, that's ok. 

Now that I've had another look, it seems like the problem is in the perl-class-std-fast rpm. That package came from repoforge. For some reason rpmbuild did not pick up the Class:Std dependency even though there is a "require Class::Std" clearly visible in the Perl module source.

altaroca:
The dependency on Class::Std::Fast::Storable has been there since my first PR. It's just the way SOAP::WSDL creates its proxies. 
For the module onvif/modules/lib/ONVIF/Client.pm I used Class::Std::Fast .
Both Class:Std::Fast and Class::Std::Fast::Storable require Class::Std .

knnniggett:
Thanks @altaroca . I thought  Class::Std::Fast had been a dependency for a while now, but only yesterday did zmonvif-probe.pl start complaining with the message that @jburgess777 posted. I've got a new  Class::Std::Fast rpm in zmrepo, which will now pull in Class::Std along with it. Hopefully we can put this one to rest.

PR #1352, opened 2016-03-20
Fix spelling in ffmpeg options help page

jburgess777:


connortechnology:
Note to self, look into eliminating all this duplication...

PR #1356, opened 2016-03-21
Minor improvements to documentation

violuke:


docs/installationguide/ubuntu.rst
```diff
@@ -198,7 +198,7 @@ As of Oct 3 2015, zm.conf is not readable by ZM. This is likely a bug and will g
 
 If you changed the  DB password **after** installing ZM, the APIs will not be able to connect to the DB.
 
-If you have, go to zoneminder/www/api/app/Config & Edit ``database.php``
+If you have, go to ``/usr/share/zoneminder/www/api/app/Config`` & Edit ``database.php``
```
kylejohnson, line 201: Not sure it makes sense to put an absolute path here, as ZM can be installed to many locations.  That said, this document does assume you're using the PPA, which installs there.  So nevermind.

docs/installationguide/ubuntu.rst
```diff
@@ -224,7 +224,7 @@ We also need to install php5-gd (as of 1.28.107, this is not installed)
 
 **Step 6**: Edit Timezone in PHP
 
-vi /etc/php5/apache2/php.ini
+``sudo nano /etc/php5/apache2/php.ini``
```
kylejohnson: Would rather this be changed back.  The `sudo` change makes sense, as `sudo` is used elsewhere in the doc.  I'd rather not have default editors changed, though.

kylejohnson:
Thank you.  Added a few small comments.  If you can fix those up, I'll merge.

violuke:
Thanks. I've changed the editor back. I only changed it in the first place as these instructions might be read by novices and typically people find nano more intuitive than vi, but probably wasn't best for me to make the docs inconsistent. Thanks.

PR #1360, opened 2016-03-24
Allow spaces in monitor names

schrorg:
I stumbled about not being able to enter spaces in monitor names, so I quickly investigated and did not find anything against doing it in the source code.

So if it really is no problem, please include this PR.

I am actually testing it with a single camera now...


connortechnology:
Yeah I can't think of any reason to not allow spaces.

knnniggett:
If we allow spaces, would it be a good idea to chomp off leading and trailing spaces?

kylejohnson:
> If we allow spaces, would it be a good idea to chomp off leading and trailing spaces?

I don't think it is worth doing.  If the user wants some extra padding, I don't see a reason to not allow it.  Besides, it would be extra code to maintain.

connortechnology:
Well I think when saving a Monitor all fields should be start and end trimmed. I have had issues with leading and trailing spaces. I forget what though.

knnniggett:
Yeah same here. There was a recent issue in the forum where the end user left a trailing space on one of the variables under the Paths tab (think it was PATH_ZMS). It caused the feature to not work and of course was difficult to isolate the cause of the problem.

schrorg:
Start and end trimming of all monitor values would be stuff for another PR, I think.

PR #1361, opened 2016-03-24
Fix coverity issues

connortechnology:
This fixes a few of them that are easy to handle.

The last one I think we should use boost to replace the algorithm.  More standard string functions, less custom crappy code.

Feel free to merge or wait for me to do that boost stuff.


knnniggett:
Looks good. Is coverity a word?

SteveGilvarry:
https://scan.coverity.com/projects/stevegilvarry-zoneminder?tab=overview I think I have to invite you to see this, pretty much c++ static code analysis. @kylejohnson We could set this up for the project directly as well as https://scrutinizer-ci.com/g/SteveGilvarry/ZoneMinder/ for PHP.

PR #1367, opened 2016-03-28
Record h264 video without analysis

mattpepin:
This pull request disables frame decoding if monitor type == RECORD and h264 passthrough recording is enabled. This brings down CPU load to 1-2% compared to 40-50% for Full HD streams on my processor. To do so:
- I changed CaptureAndRecord in zm_ffmpeg_camera
- I changed the protoype to "int recording" instead of "bool recording" in Camera::CaptureAndRecord and patched all subclasses
- If recording == 2 in CaptureAndRecord, it means to disable frame decoding
- Patched CheckSignal to always return true if frame decoding is disabled or else no events would be created

Not sure if it would be best to add an explicit option to disable frame decoding or if type==RECORD and h264 passthrough enabled is enough.

### DIFF
--- a/src/zm_camera.h
+++ b/src/zm_camera.h
@@ -80,7 +80,7 @@ public:
 	virtual int PreCapture()=0;
 	virtual int Capture( Image &image )=0;
 	virtual int PostCapture()=0;
-        virtual int CaptureAndRecord( Image &image, bool recording, char* event_directory)=0;
+        virtual int CaptureAndRecord( Image &image, int recording, char* event_directory)=0;
 };
 
 #endif // ZM_CAMERA_H
--- a/src/zm_curl_camera.cpp
+++ b/src/zm_curl_camera.cpp
@@ -311,7 +311,7 @@ int cURLCamera::PostCapture()
     return( 0 );
 }
 
-int cURLCamera::CaptureAndRecord( Image &image, bool recording, char* event_directory )
+int cURLCamera::CaptureAndRecord( Image &image, int recording, char* event_directory )
 {
     // Nothing to do here
     return( 0 );
--- a/src/zm_curl_camera.h
+++ b/src/zm_curl_camera.h
@@ -79,7 +79,7 @@ public:
 	int PreCapture();
 	int Capture( Image &image );
 	int PostCapture();
-	int CaptureAndRecord( Image &image, bool recording, char* event_directory);
+	int CaptureAndRecord( Image &image, int recording, char* event_directory);
 
 
 	size_t data_callback(void *buffer, size_t size, size_t nmemb, void *userdata);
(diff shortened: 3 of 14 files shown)


SteveGilvarry:
Looks good so far, will check further over next few days. Only comment is quite a few whitespace changes, makes it a little harder to identify the meaningful changes in the commit.

Maybe an extra option in the savejpeg output selection is needed, as Disable Decode. Only because I assume some people running record also want Live stream. I will have a think about some options.

mattpepin:
Yes sorry about that. I'll see if I resubmit zm_monitor without all
whitespace changes.

SteveGilvarry:
Don't worry about it this time I can work with it

connortechnology:
yeah.. minus the whitespace stuff this looks good, but is still a long way from where we need to be... which i s potentially stripped the audio out.  

I'm not totally happy with the recording going from bool to int.... logically it is confusing... but maybe it's the right thing.  I'll wait for others to weigh in.

josh4trunks:
I'm wondering, is this applicable to Nodect as well?
I noticed my 3MP camera does cause significant CPU load (~40% of a thread of an i3) in Nodect, but I'm not sure if this is from other processes like creating images from the stream.
Thanks

mattpepin:
Only monitor type=record bypasses frame analysis. That's why you still see CPU usage. I could easily change it to include nodect too.

josh4trunks:
I don't know the specifics of ZoneMinder well enough to know if that makes sense. But if it's possible without loosing the images frames zoneminder saves for events, that would be awesome.

SteveGilvarry:
Nodect would share an issue with modect in that the time it takes to start
recording and get a key frame you miss the actual event. I am working on a
buffer for the avpackets so we can buffer previous key frame. We will also
implement this PR in some form.

josh4trunks:
makes sense, thanks for the explanation.

sounds like some cool enhancements coming in future versions

techie66:
I was looking into trying to implement this, and then I see you already did it. Has this been worked on since March?

mattpepin:
I've been using it as is since March but I haven't updated it. Looks like it was never merged either.

techie66:
I see the issue with changing from bool to int, but I don't know if there is a better way, I'm going to try merging this with current and see how it goes. I also think a separate option to disable decoding is best, because it would prevent live streaming, I think.

connortechnology:
I think the way you really should go with this is to only decode when someone is watching the live stream.

connortechnology:
I am closing this because I want to delete the h264 branch which has been merged to master.
This idea will eventually be implemented but the approach here is not enough.  It would prevent live viewing of the stream.
My current work has progressed to removing the zma process in favour of a thread, which allows us to get rid of most of the shared memory allocation, and allows zms to actually do the live decoding.  
In any event, this PR has not been kept up to date.

PR #1376, opened 2016-03-31
Improve test multiserver restart cam.

laerciomotta:
Check if the server is correctly before restart cam

### DIFF
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -982,6 +982,7 @@ function daemonControl( $command, $daemon=false, $args=false )
 
 function zmcControl( $monitor, $mode=false )
 {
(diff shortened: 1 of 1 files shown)


connortechnology:
This is good, but we might want to also test for the non-multiserver case.  So 

if ( (!ZM_SERVER) or ( ZM_SERVER==$monitor[ServerId}) ) {

laerciomotta:
Nice, I think only in multiserver case.. Sorry!

connortechnology:
Also... I'd I've been wondering if we can use the api to cause the restart to happen on the other server

laerciomotta:
Yes, I think this too..
But I need understand better the code.

laerciomotta:
Change based in your tip.

PR #1377, opened 2016-03-31
Random fixes

connortechnology:
I'm not sure if this is a good way to go. Maybe you call can look over it.


connortechnology:
I am more convinced now that this is good.  Since the status is a class member, it stays around after thread exit.  So this is good.

PR #1384, opened 2016-04-04
tabs to spaces and use 2space indenting

connortechnology:
step 1 of following google code guidelines.  This is just the files in /src so far.


kylejohnson:
So, :+1: on putting this in its own PR!

Anyway, the google doc says this:

> Use only spaces, and indent 2 spaces at a time.
> 
> We use spaces for indentation. Do not use tabs in your code. You should set your editor to emit spaces when you hit the tab key.

That's just their standard, with no information on why.

While I do agree that our code is a mess of uneven tabs and spacing, why do you think we should go with spaces (over tabs)?

Would we then reject PRs down the line which use tabs over spaces?  (I'd say yes)

connortechnology:
Yeah we need to force people to follow the coding style.

Linwood-F:
Just in the FWIW department, tabs are not always interpreted the same in different IDE's, and often "move" after a copy/paste; spaces are uniform in how they are handled.

kylejohnson:
> Just in the FWIW department, tabs are not always interpreted the same in different IDE's, and often "move" after a copy/paste; spaces are uniform in how they are handled.

Now that's a good reason.

PR #1385, opened 2016-04-04
Remove the code looks for zm.conf in the current dir.

connortechnology:
I really don't like this code.  I think it is dangerous.


connortechnology:
We do however need a way to specify a different config file on the command line.

knnniggett:
Agreed. Will make a note to look into manually specifying a config file later.

PR #1386, opened 2016-04-04
Always show ptz

connortechnology:


knnniggett:
Testing this is still on my to-do list. Recall I have a camera which requires 256 presets (long story).
I don't envision this pr will work well with that, but let me try it and see what it looks like. Perhaps I can come up with an alternative.

connortechnology:
I have one that doesn't have presets, and the result is a little silly too.  We need some smart design in here.

SteveGilvarry:
No PTZ here so will need to wait for @knnniggett to get back from holidays.

knnniggett:
Since we tend to agree that we should implement a different solution, shall I close this?

connortechnology:
I don't see any need to close it. We need to do the work to come to a solution. Now that we are back in development more we can do so. I will take another look and think about it.

knnniggett:
I have used some surveillance software which, after hovering over a particular corner, will superimpose semi-transparant ptz controls onto the view. Thought that was cool.

connortechnology:
I like that. However I think right now clicking on the image zooms in.

knnniggett:
Right, but if the ptz controls are overlayed on top of the image, would that not cause zoneminder to call the onclick function associated with ptz control instead, assuming the "click" occurred over a ptz control widget?

knnniggett:
I'm going to close this simply for organizational purposes. I think we have all agreed that we want to do something different than what this pr does. Let's create a new pr when someone has time to revisit this in the future.

PR #1403, opened 2016-04-06
Fix grant

connortechnology:
I'm still doing some testing/further updates.

This PR adds code to the debian, ubuntu1204 and ubuntu1504 package dirs to do the following:
if the db is local, then check for the zm database and create it if it exists.  Also create the user with the required schema privileges.  Otherwise grant the required privileges to the user without specifying password (otherwise it will create a new user). 
Then run zmupdate.pl

I know debian has decided to go a different way with this stuff, but I don't think we have to do what debian says. I think this will make installs/updates much easier for the average user.

Thoughts?

### DIFF
--- a/distros/debian/postinst
+++ b/distros/debian/postinst
@@ -29,9 +29,9 @@ if [ "$1" = "configure" ]; then
                 if ! $(echo quit | mysql --defaults-file=/etc/mysql/debian.cnf zm > /dev/null 2> /dev/null) ; then
                     cat /usr/share/zoneminder/db/zm_create.sql | mysql --defaults-file=/etc/mysql/debian.cnf
                     # This creates the user.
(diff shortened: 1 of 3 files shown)


SteveGilvarry:
I think the question is what can we change upstream that Debian can use to
make the install easier for a user. I would actually like to ask Maintainer
for input?
If some of this stuff goes into an install process on first run rather than
the package, what is the scope of the install wizard?

connortechnology:
well @onlyjob is the maintainer.  He chose to remove most of this kind of stuff from the debian package.

onlyjob:
Some apps don't use upgrade scripts. For example _Zabbix_ just upgrade database schema silently while _CiviCRM_ advises to use dedicated web page to initiate schema upgrade manually once outdated schema is detected. Personally I'm OK with `zmupdate.pl`. IMHO there might be marginal benefit from incorporating `zmupdate.pl` into daemon or to PHP interface but is it really worth the effort?

PR #1406, opened 2016-04-08
replace the static zone image with a stream, and use SVG to draw the zones

connortechnology:
Just putting this here to help collaboration.

### DIFF
--- a/web/includes/config.php.in
+++ b/web/includes/config.php.in
@@ -104,6 +104,7 @@ define( "CMD_PREV", 12 );
 define( "CMD_NEXT", 13 );
 define( "CMD_SEEK", 14 );
 define( "CMD_VARPLAY", 15 );
+define( "CMD_QUIT", 17 );
 define( "CMD_QUERY", 99 );
 
 //
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -2486,4 +2486,22 @@ function validHtmlStr( $input )
     return( htmlspecialchars( $input, ENT_QUOTES ) );
 }
 
+function getStreamHTML( $monitor, $scale=100 ) {
+//FIXME, the width and height of the image need to be scaled.
+    if ( ZM_WEB_STREAM_METHOD == 'mpeg' && ZM_MPEG_LIVE_FORMAT ) {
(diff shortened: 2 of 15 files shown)


knnniggett:
Here are the spots where to use the new getStreamHTML:
https://github.com/ZoneMinder/ZoneMinder/blob/svg_zones/web/skins/classic/views/watch.php#L54
https://github.com/ZoneMinder/ZoneMinder/blob/svg_zones/web/skins/classic/views/zone.php#L125
https://github.com/ZoneMinder/ZoneMinder/blob/svg_zones/web/ajax/zone.php#L33

Notice that there are two zone.php files in tow different folders (threw me for a loop).

knnniggett:
Here is what I worked on today.

zone

I added two new features:
- I added the "State: xxx - x.xx1 fps" values under the image, just like the Watch view (a.k.a. live monitor)
- I added a "Pause" button to pause the video. The text on the button is dynamic and toggles from "Pause" to "Play" depending on current state of the stream.

I have not pushed any commits yet because I need to do some serious code cleanup. I did a lot of copy-and-paste from watch.js, and I need to go through and strip out unneeded code.

Short term Discussion item:
Currently one ends up with two duplicate streams: one stream in zones.php and another in zone.php. This is obviously inefficient. My thought on this is to revert zones.php back to a static image, but there might be other ways, such as eliminating zones.php altogether.

Long term feature requests:
- Add three bar graphs that represent the realtime pixel count for alarmed, filtered, and blobbed pxels
- Overlay the red-blob on the image

From my first crash course in learning how zoneminder php gets realtime data from zm's C binaries, we would need to add functionality to zms to send the data we need via socket and then modify the ajax code to retrieve this new data. While both of the mentioned features would be super cool, this looks to be more work than what I am willing to put in to the initial release of this feature.

connortechnology:
baby steps...

I think wecan merge zone and zones

knnniggett:
commit e5bf2d2 quits the stream in zones.php when the user either clicks a zone image or clicks the zone name in the bottom table. The stream continues normally in the new popup, zone.php. 
When the user saves his/her changes to the zone and the zone window closes, the stream starts back up automatically.

PR #1407, opened 2016-04-09
Better Error logging on disk_total_space failure

connortechnology:


knnniggett:
Should we add the same kind of check to disk_free_space as well, just a couple lines down?

knnniggett:
hmmmm.... looks like this commit is also in #1409 so perhaps we can close this pr out.

connortechnology:
yeah we should find and fix all cases where we dont handle failure

web/includes/functions.php
```diff
@@ -1668,6 +1668,10 @@ function getLoad()
 function getDiskPercent()
 {
     $total = disk_total_space(ZM_DIR_EVENTS);
+    if ( ! $total ) {
+        Error("disk_total_space returned false for " . ZM_DIR_EVENTS );
+        return 0;
```
knight-of-ni, line 1673: I'm thinking about the value we should return if disk_total_space fails:
- If we return 0, then the purge filter will not trigger and zoneminder will keep recording until writes fail due to disk full. I believe Zoneminder will simply stop running when this happens with a failed to write to disk message.
- if we return 100, then that will cause the purge filter to continually execute because it incorrectly thinks the disk is full. All events will eventually get deleted.
- if we return -1 to signal there was an error, then perhaps we can write additional code to handle that response and/or alert the user.

Thoughts?
Should we consider the third option?

connortechnology:
Well, I definitely wouldn't want events deleted due to a bug.  So we can't return 100. -1 is probably the best thing to do.

Filters actually run in the perl code, so this would only affect manual filter runs.

PR #1412, opened 2016-04-10
make cannot write to content dir an error, rather than fatal

knnniggett:
If under Options the end user enters an invalid path for either the events or images folders, the system is put into a catch 22. A fatal error is presented to the end user, but since it is fatal, the end user cannot proceed to fix the problem.  

I have changed the level of the message from Fatal to an Error. This should allow the end user to navigate to Options and fix the problem.


apetit518:
Where did you change the log level from Fatal to Error?

PR #1416, opened 2016-04-13
updated documentation to include runstates

pliablepixels:
Adds commentary on run states when describing the console as the runstate control is triggered from there. @SteveGilvarry ref https://forums.zoneminder.com/viewtopic.php?f=34&t=24547#p93917


SteveGilvarry:
@pliablepixels like having a personal assistant, make a comment and the next day it appears.

pliablepixels:
I grew up watching crocodile dundee.  I bet you own a rambo knife, so what option do I have but to comply?

PR #1422, opened 2016-04-16
suppress warnings of killall for FreeBSD

abishai:


connortechnology:
Under what circumstances does killall put out a warning?

connortechnology:
Oh... if none are matched.

connortechnology:
It will do this under ubuntu as well.

abishai:
If no childs found. I'm getting a lot of them on startup. This warnings go to zoneminder.pid (??!) on my system and breaking rc.d control script.

abishai:
Ubuntu shouldn't be affected - else { already has supression flag

PR #1428, opened 2016-04-20
Better zmvideo

connortechnology:
This PR includes my work on making nice abstracted Filter and Event objects. It adds the ability to specify a filter, thereby generating a video of all of the events matching that filter.


knnniggett:
ooooohhhhh.... a single video clip from multiple events. I likey.
Let's give this guy a raise!

PR #1429, opened 2016-04-20
add jscalendar as a submodule, modify rpm specfiles

knnniggett:
The ZoneMinder Filter view has an undocumented feature which supports integration of jscalendar:
https://github.com/ZoneMinder/ZoneMinder/blob/master/web/skins/classic/views/filter.php#L135

What this does is popup a small, graphical calendar when clicking in the date field:
screenshot from 2016-04-20 12-28-06

Previously, I have been adding this support into all the rpms I build. During buildtime, jscalendar would be downloaded and merged into the build root. I'd like to get away from that and instead use a git submodule. This also has the benefit of giving this feature to all ZoneMinder users, rather than just those who use my rpms.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 	path = web/api/app/Plugin/Crud
 	url = https://github.com/FriendsOfCake/crud.git
 	branch = 3.0
+[submodule "web/tools/jscalendar"]
+	path = web/tools/jscalendar
+	url = https://github.com/virasak/jscalendar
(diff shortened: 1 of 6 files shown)


connortechnology:
I'm happy with this, just want to test it with deb/ubuntu packages before merging.

I wonder what @onlyjob has to say about it though. It is GPL and doesn't seem to be available in debian so it should be ok.

I wonder though if long term we will drop this for a bootstrap based date picker...

onlyjob:
I think choice of this particular calendar is bad because it is unmaintained since 2008.

What about https://github.com/fullcalendar/fullcalendar or https://github.com/n9/bootstrap-datepicker ? The latter is in Debian as [libjs-twitter-bootstrap-datepicker](https://tracker.debian.org/pkg/libjs-twitter-bootstrap-datepicker).

knnniggett:
Due to time constraints and higher priorities, I have chosen to stick with jscalendar, integrated during the rpm build process.

PR #1433, opened 2016-04-23
(Docs) Added provisioner-friendly command for inserting sleep 10 

veloper:
Added *provisioner-friendly/scriptable* command for inserting sleep 10 into `/etc/init.d/zoneminder`

### DIFF
--- a/docs/installationguide/ubuntu.rst
+++ b/docs/installationguide/ubuntu.rst
@@ -18,7 +18,7 @@ if you don't have mysql already installed:
 
 ::
 
-	sudo apt-get install mysql-server 
+	sudo apt-get install mysql-server
 
 This will ask you to set up a master password for the DB (you are asked for the mysql root password when installing mysql server).
 
@@ -61,7 +61,7 @@ This is probably a bug with iconnor's PPA as of Oct 3, 2015 with package 1.28.10
 
 ::
 
-	sudo chown www-data:www-data /etc/zm/zm.conf 
+	sudo chown www-data:www-data /etc/zm/zm.conf
 
 We also need to install php5-gd (as of 1.28.107, this is not installed)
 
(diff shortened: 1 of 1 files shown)


connortechnology:
I used to care strongly about this, as I hated waiting 10 seconds to start/restart zoneminder.  Now I don't because the world has switch to systemd which has dependencies and so it can figure out whether or not mysql is running.

I would prefer to see a solution that checked for a mysql pid file or something instead of just waiting 10 seconds just in case.

veloper:
@connortechnology - I completely agree with your point. I too feel like there should be a more user friendly / hands-off solution for ensuring ZM is running; Upstart, Monit, Bluepill, God, and few other solutions come to mind. One could even take the extra step from there to create a Chef(-Solo) Recipe, or the equivalent in Ansible & Puppet.

When I was setting up ZM and following the instructions they seemed pretty scriptable. As I followed along with each step I started adding them to an idempotent bash script that I could use as a pseudo-provisioner. When I arrived at this last step I was met with manual instructions. This differed from the previous steps which consisted of scriptable shell commands.

Again, I'm 100% with you that there is a better solution out there than what is currently in the instructions. I also think you'd agree with me that the implementation, testing, and documentation of a solution like that falls clearly outside the scope of what this Pull Request aims to accomplish -- which is to enhance the _existing_ instructions.

Hopefully I've made my thought processes and intent around this PR is more apparent.

knnniggett:
For the record, this is what I'm doing on the one site I've got which uses @connortechnology 's ppa:

```bash
# ATB - wait for mysqld to start. Continue if it takes too long.
count=0
while [ ! -e /var/run/mysqld/mysqld.sock ] && [ $count -lt 60 ]
       do
               sleep 1
               count=$((count+1))
       done
```

knnniggett:
Obsoleted by #1549. Closing...

PR #1439, opened 2016-04-24
Update do_debian_package.sh to let a dev. make quick local mods

pliablepixels:
Add a read key right after the repo is cloned, so developers can make quick changes in another window and have them incorporated in the build process. If we don't do this, the only way is to do a do_debian, leave the repo at the end, make mods and do_debian again. This little hook lets me make local changes the first time in another terminal window.


pliablepixels:
closing - there are other ways to do this, including git cloning to zoneminder_release first and then running. Probably not worth it.

PR #1442, opened 2016-04-25
Fix rotate

connortechnology:


pliablepixels:
awesome - did a very quick check - seems to be working. Will do more tests!

pliablepixels:
tried both live and event footage -- with both rotated (ffmpeg and native RTSP) and non-rotated monitors. Works very well. I don't have events previously recorded before the change to see how they would work (its a new VM), but rest looks good.

knnniggett:
Just built this. Live view works for a remote & libvlc camera, normal orientation. Events previously recorded look good. Thumbnails in the event list and timeline look right.

However, what my system currently is not doing is recording events. If I click "force alarm", the monitor status stays "idle" and no alarm is recorded to disk. No errors appear in the log.

Can someone verify this on their system?

knnniggett:
OK, forced alarms are working now. Maybe I didn't wait long enough for the system to stabilize.

Previously, there always seemed to be an unusually long delay before the State and Frame Rate would appear below the monitor. With this pr, that information appears to show much quicker now.

knnniggett:
Everything seems to be working. 
I'm going to merge this into master so we can get others to test.

PR #1443, opened 2016-04-25
unbreak build under FreeBSD 32 bit

abishai:
time_t under i386 FreeBSD is int, so we can't compile ZM, as compiler see 2 equal functions here. On x64 and linux time_t is long. Changing time_t to long solves the matter.

### DIFF
--- a/src/zm_event.cpp
+++ b/src/zm_event.cpp
@@ -660,7 +660,7 @@ void Event::AddFrame( Image *image, struct timeval timestamp, int score, Image *
     */
 }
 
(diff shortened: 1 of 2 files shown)


connortechnology:
I don't like this.  How about we change the one that takes a frame id to an unsigned int, since that is logically what it is.

abishai:
Much more better solution than mine. Refactor the pull request or you commit tha change?

connortechnology:
I would prefer you to do it.  I'm out for a bit.

abishai:
Sure, I'll test it and redo.

PR #1444, opened 2016-04-26
handle disk_free_space error

connortechnology:


knnniggett:
If there were exactly zero bytes free on the target disk, this would trigger the error message, but what is the likeliness of exactly zero bytes free?

connortechnology:
that's a good point.  Perhaps we should not return 0, but just log the error....

kylejohnson:
My question would be, what made you need to make this change?  Were you hitting a condition where you were seeing a free disk space related error?

connortechnology:
we had a previous discussion about it. #1407

SteveGilvarry:
Resolved this?
> Perhaps we should not return 0, but just log the error....

Issue #1424, opened 2016-04-17
(FreeBSD) Incorrect socket name when binding for zms


abishai:
A lot of stalled 'soc' files left because clean-up code assumes they should be 'sock'
https://gist.github.com/abishai/2c11cda4db6f045ae1a4b0a9a7cbc6e4

abishai:
Problem is here https://github.com/ZoneMinder/ZoneMinder/blob/247cb3d0b2adb8941bd1cd400cf783bdfc71832d/src/zm_stream.cpp#L326 
If I add +1 to the string length, socket create with full name and deleted correctly when I stop streaming, so no more stalled 'soc' files.

I'm uncertain how to create platform agnostic solution as I don't understand the root of the problem.

PR #1449, opened 2016-04-28
fix 'soc' issue under FreeBSD

abishai:
Hack :(
On FreeBSD sock files created without ending 'k', so they are not cleared at all.
https://github.com/ZoneMinder/ZoneMinder/issues/1424
I don't know why it's happening and I've got a little help from BSD folks. Maybe you can suggest better solution?


connortechnology:
http://man7.org/linux/man-pages/man7/unix.7.html

The +1 should always be there.  I have a different PR that addresses this.

PR #1453, opened 2016-05-01
Ubuntu 1604 Packaging folder

SteveGilvarry:
Rename ubuntu1504, which is unsupported, to ubuntu 1604. Fix control file to current libav packages and php virtual packages. Update debian build script to use new folder for any non-trusty build. Should also still work for 1510.


connortechnology:
I'm not sure we need to rename it to 1604 if it still works for 1504.  Lots of people are going to still be running 1504 and may want to build zm on it.  Maybe it should be renamed 1504plus or something.

SteveGilvarry:
I some what agree with you, but according to Ubuntu it is 1504 who?
http://www.ubuntu.com/info/release-end-of-life
So I thought why bother with the standard releases once a LTS arrives we jump up.

PR #1455, opened 2016-05-02
rough in Server object support

connortechnology:
This roughs in Server object support in the API which is needed for zmNinja get the url to the recording server for live view.

I need to change the permissions requirements, will remove the not ready for merge when that is done.


kylejohnson:
Looks good so far

Issue #901, opened 2015-06-23
Allow use of UNIX File Sockets for MySQL Connection


josh4trunks:
Documentation for PDO is here.
http://php.net/manual/en/ref.pdo-mysql.connection.php

The $dbConn line needs to modified to allow inclusion of "unix_socket", and probably needs a bit of logic added to know when to include it.
ZoneMinder/web/includes/database.php

I'm not sure what else needs is crucial that needs modification for this to work. Most of the other instances of the database parameters being used seemed like distro specific stuff.

knnniggett:
Why?

It's working fine the way it is.  From the documentation you referenced:
```
 unix_socket

    The MySQL Unix socket (shouldn't be used with host or port).
```
We are specifying the host, so unix socket shouldn't be used:
https://github.com/ZoneMinder/ZoneMinder/blob/master/web/includes/database.php#L34

josh4trunks:
Im asking that the option be added, not that force admins to use either method. file sockets have higher performance because that aren't going through the network stack (being chunked, TCP reliability which is irrelevant on a local system)

It shouldn't be too hard to implement. I helped did this for owncloud and it when I have more time can try making a PR for zoneminder.

josh4trunks:
It should only take a line of logic to see if ZM_DB_SOCK ( or whatever ) is defined an to know what parameters to pass to PDO.

knnniggett:
Feel free to submit a pr.  Naturally, we want to make sure it doesn't break installations with a remote dB.

josh4trunks:
OK, will look into it sometime in the future.

knnniggett:
I'm closing this out due to inactivity.  The idea is sound.  If you choose to implement this, feel free to reopen this issue and generate a pull request.

josh4trunks:
I might try tackling this, this weekend.

It looks like zoneminder doesn't even take into account port numbers? so if someone tried connecting to a mysql running on a different port it wouldn't work?

josh4trunks:
@knnniggett I don't see a way to reopen this issue, can you do that?

Can someone please help me a little?
https://github.com/josh4trunks/ZoneMinder/tree/mysql_socket

I got the PHP done, but am having trouble with the Perl connection, at least I get an error when trying to start the zoneminder daemon.
It looks like `lib/ZoneMinder/Config.pm`, also has a db connection. Is there any possibilty to have this use `Database.pm` instead of having to have the database connection logic in 2 places for perl?
[...]

PR #1457, opened 2016-05-03
Adds logic to handle ports / unix socket

josh4trunks:
Logic is added to handle ports / unix file sockets for database connection.

For example the following now all work as expected
```
ZM_DB_HOST=localhost
ZM_DB_HOST=REMOTE_HOST:1234
ZM_DB_HOST=localhost:/path/to/non-default.sock
```

Fixes https://github.com/ZoneMinder/ZoneMinder/issues/901

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Config.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Config.pm.in
@@ -83,8 +83,25 @@ BEGIN
     close( $CONFIG );
 
     use DBI;
+    my $socket;
+    my ( $host, $portOrSocket ) = ( $Config{ZM_DB_HOST} =~ /^([^:]+)(?::(.+))?$/ );
+    if ( defined($portOrSocket) )
+    {
+        if ( $portOrSocket =~ /^[[:digit:]]+$/m )
+        {
+            $socket = ";host=".$host.";port=".$portOrSocket;
+        }
+        else
+        {
+            $socket = ";mysql_socket=".$portOrSocket;
+        }
+    }
+    else
+    {
+        $socket = ";$host=".$Config{ZM_DB_HOST};
+    }
     my $dbh = DBI->connect( "DBI:mysql:database=".$Config{ZM_DB_NAME}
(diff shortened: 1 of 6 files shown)


josh4trunks:
I'm not sure why Zoneminder has 3 different Perl files  and 2 different C++ files that connect to the database? If it's not necessary maybe we can consolidate the connections a bit?

josh4trunks:
@knnniggett @connortechnology

josh4trunks:
Please don't merge yet.
Looks like the API doesn't know how to handle the mysql connection. Anyone know where the connection is being initiated so I can fix it?

SteveGilvarry:
https://github.com/ZoneMinder/ZoneMinder/blob/master/web/api/app/Config/database.php.default

josh4trunks:
Thanks, I'll work on fixing this issue up when I get a chance

connortechnology:
I have a PR coming today that improves the api db config. you might want to wait on it

connortechnology:
You might want to base your changes off my pr #1458

josh4trunks:
sounds, good I'll wait. I gotta figure out the cake part anyway

josh4trunks:
@SteveGilvarry It looks like web/api/app/Config/database.php is installed during installation using the default values. If an admin is using any non-default database connect parameters (host/port/user/pass) are they supposed to edit this file as well as their zm.conf?

Maybe we should re-implement this so database.php just pulls the values from the zm.conf?

Here's an example giving parameters to a different framework, but looks similar and in this case the values are dynamic.
https://github.com/nZEDb/nZEDb/blob/0.x/app/config/bootstrap/connections.php

SteveGilvarry:
API is early stages of implementation so I assume no one has hit this yet,
but you would think a single place for all settings would make more sense.
I suspect database.php is cakephp out of the box.

connortechnology:
See #1470 which does exactly that.

josh4trunks:
sweet, DB_SOCKET is missing though

josh4trunks:
Replaced by https://github.com/ZoneMinder/ZoneMinder/pull/1492

PR #1458, opened 2016-05-03
Merge db connects

connortechnology:
This PR removes duplicated db connect code.  

Because the config loading code happened in a BEGIN statement, it got really complex about what had been required and what hadn't.  So made it a sub and added a BEGIN statement to Base, to ensure that things happen in the correct order automatically.


josh4trunks:
there's also 2 different connections in C++. not sure if it makes sense to consolidate those.

josh4trunks:
Is this ready for me to apply my changes for correct mysql port/socket parsing?
I noticed the failed tests, but I'm not sure if that means anything.

Thanks

PR #1462, opened 2016-05-05
Merge object model classes from storageareas

connortechnology:
This is meant to be an ongoing PR, where I make bits and pieces of my work on storageareas and video, etc available to be merged into master.  The idea to be to keep it in manageable chunks that don't affect anything else.

Please merge as you feel is appropriate. I don't want this to get too large.


connortechnology:
Step one, get the current versions of my crude DB object classes in.  This should not affect anything, as the ones that were in master have not changed much, and the others are not used at all in master.

PR #1463, opened 2016-05-06
Iconnor updated console

connortechnology:
deprecate all the sequence calculations.  They are overly complex.


knnniggett:
I'm not sure I understand the purpose of this pr, or perhaps I just don't see how this makes things any less complex. Typically I'd rather not change things that have been previously working, unless we have a really good reason. Not ruling out that there isn't something I just don't understand or overlooked in all the commits. Maybe we can discuss this one at our next Monday night meeting and/or get some other opinions on this one. @kylejohnson @SteveGilvarry

connortechnology:
This was a bunch of fixes for @kylejohnson 's old updated console.  I hadn't deleted it so that he could reference it, but he never did.  It can be deleted.

knnniggett:
Ah, okay. I was going crazy trying to make sense of this one.

connortechnology:
There is some good stuff in there but should be done separately.  I want to get rid of the up and down arrows everntually anyways.

PR #1473, opened 2016-05-09
when value to be stored is undefined or '', then set the field=NULL

connortechnology:
This is a solution to mysql5.7 strict mode.  

It is not ideal, as it may cause fields to be set to NULL that are NOT NULL by design.


connortechnology:
I'm putting this here for other people to test with.  I am running it in production, and will hopefully find any problems it introduces. So far I havn't found any.

connortechnology:
This apparently causes issues with linked monitors

PR #1475, opened 2016-05-12
introduce htmlselect as an alternative to buildselect

connortechnology:
I don't like buildselect. 

This introduces a simple function that takes a name, contents, selected value, etc.. and returns a string for the select.


SteveGilvarry:
@connortechnology Is it just me or is this introducing new function with nothing actually using it?

connortechnology:
That is exactly what it is. 

I have used this function extensively in storageareas.  I find it much more intuitive than buildselect.

I try to keep these PR's very small

SteveGilvarry:
@connortechnology update this and I will merge it since it has no impact until we start to use it, just make mention of it needing a test when you submit a PR that uses it.

connortechnology:
done

PR #1478, opened 2016-05-13
pass in the server url into the monitor object to use instead of the …

connortechnology:
…portal url

The ajax status queries, etc MUST go to the recording server.  So make that info available, and use it.


knnniggett:
Looks good.

PR #1486, opened 2016-05-16
Update image view

connortechnology:
This adds the ability to specify an image by fid= being the primary key of the Frames table, or eid= and fid= where eid is the event # and fid is the frameid in the event.  

The frame popup is updated to use this code.


knnniggett:
Appears to to work without negative consequences. Merging.

PR #1487, opened 2016-05-16
Initial port to OSX

pliablepixels:
At a point where cmake builds, make works - processes get created. Running them from command line works, but no packaging etc. has been modified. I don't plan to maintain this - just a starter for folks who are interested.


pliablepixels:
gaah lots of junk got uploaded. will fix.

src/zm_signal.cpp
```diff
@@ -66,13 +66,13 @@ RETSIGTYPE zm_die_handler(int signal)
 		ucontext_t *uc = (ucontext_t *) context;
 		cr2 = info->si_addr;
 #if defined(__x86_64__)
-	#ifdef __FreeBSD_kernel__
-		ip = (void *)(uc->uc_mcontext.mc_rip);
+	#if defined (__FreeBSD_kernel__) || defined (__MACH__)
+		//ip = (void *)(uc->uc_mcontext->mc_rip);
```
connortechnology, line 70: Um... this doesn't look right.  You added the MACH test, but remarked out the code... this would fail the FreeBSD case

src/zm_logger.cpp
```diff
@@ -16,7 +16,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 */ 
-
+#include "zm.h"
```
connortechnology, line 19: This shouldn't be neccessary, because you did it in zmlogger.h

pliablepixels:
yes, my fault. I'm going to close this anyway and do a new PR with only the required files. will address it there.

PR #1488, opened 2016-05-16
Hide rtspdescribe

connortechnology:
This hides the rtsp describe option when the selected method isn't rtsp.


knnniggett:
The monitor view no longer loads when I apply this charge. PHP syntax error is logged:
```[Tue May 24 07:42:27.947263 2016] [:error] [pid 22749] [client 127.0.0.1:42636] PHP Parse error:  syntax error, unexpected '}', expecting ',' or ';' in /usr/share/zoneminder/www/skins/classic/views/monitor.php on line 887, referer: https://localhost/zm/index.php```

connortechnology:
Le sigh.  Fixed I think.  I did test, and it worked for me.

knnniggett:
The page loads now, but the RTSP Describe checkbox stays hidden no matter what I select.

knnniggett:
The code that looks at the RemoteProtocol selection sets calls the show/hide method based on the user selection.

Further down, where the RTSPDescribe line is, you also set its style to display:none. Is this necessary? Isn't calling show/hide sufficient to show or hide the checkbox?

connortechnology:
fixed.

connortechnology:
let me test some more

connortechnology:
I really think it's ok now.

connortechnology:
The code down below sets the initial visible state.  It is necessary.

knnniggett:
Making progress, but it does not fit into the table properly. See screenshot.

screenshot from 2016-05-24 12-22-42

connortechnology:
I don't see how that could happen, and it's not happening for me.

Sigh.

connortechnology:
Okay, I'm serious, I've tried every browser, clean installs, etc.  It works perfectly for me and the resulting HTML validates.

Could it be a bungled merge for you?

knnniggett:
I'll try it again in a different environment at the office tomorrow.

knnniggett:
I built this again in a different environment, and I am still getting the same result. 

Without clicking any other buttons, here is what causes the issue every single time:
1) click add new monitor
2) select source type of remote
3) click source tab
4) observe rtspdescribe is not visible
5) select rtsp from the dropdown
6) observe rtspdescribe row is messed up as shown in the previous screenshot

I continued to experiment, and discovered the following:

7) click the general tab (or any other tab)
8) go back to the source tab
9) observe the rtspdescribe row appears correctly

So it appears that the issue occurs only the first time, when rtspdescribe is first made to appear. If one clicks to a different tab and then comes back, the issue is gone.

I have performed these tests using Firefox and Chromium on both the local zoneminder server and from a remote client.  No errors appear in the browsers debug console

connortechnology:
Okay, I can reproduce it too.  I was always testing by editing an existing monitor.  Not sure why that makes a difference.

The dom etc looks the same and correct, but I do notice that when it becomes visible, the display css rule becomes block, instead of maybe table-row.... I wonder if that is it.  Not sure why editing an existing monitor is different.  I will investigate some more.

connortechnology:
What a huge amount of wasted effort over such a simple little ui tweak.  Oh well.

I am now using setStyle('display','table-row') instead of show();

knnniggett:
Confirmed. Works.

PR #1495, opened 2016-05-19
added alarm apis

pliablepixels:
api doc updates


pliablepixels:
not sure why this failed build -  the changes are only for the docs and  doc build works. I can't see any cause for error in the travis logs. Any thoughts?

SteveGilvarry:
@pliablepixels master is borked by commit before, which was fixing master being borked.

PR #1498, opened 2016-05-22
Add support for MySQL Port / Unix Socket

### LINKED ISSUE #901: Allow use of UNIX File Sockets for MySQL Connection

josh4trunks:
Logic is added to handle specified port / unix file socket for database connections.
For example, the following now all work as expected.
```
ZM_DB_HOST=localhost
ZM_DB_HOST=REMOTE_HOST:3307
ZM_DB_HOST=localhost:/path/to/non-default.sock
```

Fixes #901

### DIFF
--- a/src/zm_db.cpp
+++ b/src/zm_db.cpp
@@ -37,12 +37,10 @@ void zmDbConnect()
   my_bool reconnect = 1;
   if ( mysql_options( &dbconn, MYSQL_OPT_RECONNECT, &reconnect ) )
     Fatal( "Can't set database auto reconnect option: %s", mysql_error( &dbconn ) );
-  std::string::size_type colonIndex = staticConfig.DB_HOST.find( ":/" );
-  if ( colonIndex != std::string::npos )
+  std::string::size_type colonIndex = staticConfig.DB_HOST.find( ":" );
+  if ( colonIndex == std::string::npos )
   {
-    std::string dbHost = staticConfig.DB_HOST.substr( 0, colonIndex );
-    std::string dbPort = staticConfig.DB_HOST.substr( colonIndex+1 );
-    if ( !mysql_real_connect( &dbconn, dbHost.c_str(), staticConfig.DB_USER.c_str(), staticConfig.DB_PASS.c_str(), 0, atoi(dbPort.c_str()), 0, 0 ) )
+    if ( !mysql_real_connect( &dbconn, staticConfig.DB_HOST.c_str(), staticConfig.DB_USER.c_str(), staticConfig.DB_PASS.c_str(), NULL, 0, NULL, 0 ) )
     {
       Error( "Can't connect to server: %s", mysql_error( &dbconn ) );
       exit( mysql_errno( &dbconn ) );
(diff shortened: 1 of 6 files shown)


josh4trunks:
@knnniggett @connortechnology 
Sorry to keep bringing this up, but if we could get this merged at some point I would appreciate it. Then I would not need to keep reapplying these fixes to get my production system working.

This fixes the currently incomplete logic below that does not work because it isn't implemented in all the places it needs to be.
https://github.com/ZoneMinder/ZoneMinder/blob/master/scripts/ZoneMinder/lib/ZoneMinder/Database.pm#L82
https://github.com/ZoneMinder/ZoneMinder/blob/master/scripts/ZoneMinder/lib/ZoneMinder/Logger.pm#L463

siigna:
👍  Would be awesome for this to get some traction.  I was [in the IRC channel](https://botbot.me/freenode/zoneminder/msg/64024998/) a while back asking about it.

josh4trunks:
It looks like none of the files this commit touches have changed in a way that will affect this PR.
Can someone please review and comment when time is available.

Thanks!

SteveGilvarry:
@josh4trunks this was literally 19mins after 1.30.0 release, so I take it you are keen to get it merged 😄. I did promise it would go in next so if @connortechnology or @knnniggett have no objections I will merge this weekend.

knnniggett:
Well, I was hoping to give the new release a few days out in the wild before we start merging the open pull requests. If something crops up, we won't have a bunch of recent merges in the way if we have to troubleshoot.

josh4trunks:
Yeah, I'd love to get this in so I don't need to keep mergeing in my changes for my production system =]

But, no need to rush this in before making sure we are good with 1.30.0

connortechnology:
Wouldn't fixes to 1.30 go on the 1.30 branch?  

I've been running with this code merged for a while.  It's pretty solid.

siigna:
> I've been running with this code merged for a while. It's pretty solid.

Ditto.  Been running this since 1.30.0-RC1 and haven't seen any issues.

knnniggett:
I did a quick built and test on Fedora 24 using a single camera and the normal database configuration. Everything worked. Since @connortechnology mentioned this was working for him, I will trust this does not (somehow) negatively affect multiserver. Not sure how it could. In any case, merging....

src/zm_db.cpp
```diff
@@ -37,23 +37,35 @@ void zmDbConnect()
   my_bool reconnect = 1;
   if ( mysql_options( &dbconn, MYSQL_OPT_RECONNECT, &reconnect ) )
     Fatal( "Can't set database auto reconnect option: %s", mysql_error( &dbconn ) );
-  std::string::size_type colonIndex = staticConfig.DB_HOST.find( ":/" );
-  if ( colonIndex != std::string::npos )
+  std::string::size_type colonIndex = staticConfig.DB_HOST.find( ":" );
```
connortechnology, line 40: I think there is a problem in here.  So, the new logic is, if we found a colon, then we look for a socket path?  That will never happen. The old code looked for either a colon or a slash, and then looked for a /. (By old code I mean some other patch that I merged into storageareas).

connortechnology:
Does this have any practical benefit?  Does it have a noticeable performance impact?

josh4trunks:
@connortechnology the line you are looking at finds the length of the hostname/IP part of the DB_HOST config value. The part which decides if it is a TCP or file(unix) connection is at line 53.

josh4trunks:
The practical benefit are you can now specify a socket path to connect to a mysql server. For example if your mysql server has a file socket at /non/default/path/mysql.sock you could enter `ZM_DB_HOST=localhost:/non/default/path/mysql.sock`
I also believe non-standard ports were not working before because of inconsistencies with this was implemented.

This makes the logic consistent amounts each file for a given language (1 PHP, 2 C++, and 3 perl). Though the way I implemented it is different for each language because I tried to use the most efficient method given the available functions for that language.

josh4trunks:
I assume there might be some small unnoticeable performance impacts because more checking/matching is done before connecting to mysql.

But I did my best to use keep things efficient by using character matching like `string[0] == /` instead of string matching

connortechnology:
Yes but it will only run line 53 if there was a colon.  the socket path doesn't have a colon so that code won't get run.

I think the real line we are talking about is 383.  Where it looks for the /.  That only gets run if a colon was found.

josh4trunks:
yeah, I assume values are in the format mentioned here.
https://github.com/ZoneMinder/ZoneMinder/pull/1498#issue-156160173

josh4trunks:
I've implemented this in a few different programs, and usually this is the format assumed. But if we want to also support '/path/to/socket' we could change the logic more. We'd need to also implement this in "web/api/app/Config/database.php.default" before passing to Cake. Not sure if that would be difficult but that logic also relieson the ":"

connortechnology:
Oh, I missed that.. I don't understand why we would specify a host if it is a local socket. Huh. I guess I've just never seen that before.  Anyone else care to weigh in?

josh4trunks:
You really don't need to, but that's how alot of programs I've worked with do it. owncloud/nextcloud, pydio, I think even wordpress.
you could just put ':/path/to/socket' and that should work.

src/zm_db.cpp
josh4trunks: See my reply here https://github.com/ZoneMinder/ZoneMinder/pull/1498#issuecomment-238396576.

I put it in the wrong spot =/

PR #1499, opened 2016-05-24
move multiserver message to display once during startup only

### LINKED ISSUE #1482: Do we need the multi-server "Not using multi server" logs?
Specifically,
https://github.com/ZoneMinder/ZoneMinder/blob/66a3fef61ebd10c45968a3706021d45d027e28be/src/zm_config.cpp#L138-L143

It seems when any operation results in some access to loadConfig, this log is generated. In R1.29 this was an INFO log, which we have demoted to Debug. We can keep it in Debug, but do note that for any operation, such as cgi-bin/zms view snapshot (or jpeg),any API that does a loadConfig (like alarmStatus) will keep generating this log. Folks using zmNinja with the next release of ZM (albeit in debug mode) will be rather overwhelmed as I am using alarm status APIs extensively.

@connortechnology thoughts?

knnniggett:
fixes #1482 

It should work, but I haven't tested this yet.

### DIFF
--- a/scripts/zmpkg.pl.in
+++ b/scripts/zmpkg.pl.in
@@ -226,7 +226,12 @@ if ( $command =~ /^(?:start|restart)$/ )
         zmMemTidy();
         runCommand( "zmdc.pl startup" );
 
-		Info( "Starting up services" . ( $Config{ZM_SERVER_ID} ? " for server $Config{ZM_SERVER_ID}\n" : "\n" ) );
+        if ( $Config{ZM_SERVER_ID} ) {
+            Info( "Multi-server configuration detected. Starting up services for server $Config{ZM_SERVER_ID}\n");
+        } else {
+            Info( "Single server configuration detected. Starting up services." );
+        }
+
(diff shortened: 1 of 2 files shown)


knnniggett:
single server mode tested and works as intended.

connortechnology:
I put these lines in for debugging purposes.  If people don't want to see them, they shouldn't have debugging on.

I object to them being removed.  I propose that we increase the debug level required to see them to 3 instead.

knnniggett:
The messages are being moved, not removed.

connortechnology:
They are moved to a perl script, which uses different code to parse the config.  I sometimes run zmc manually (when debugging).  I don't object to the lines in zmpkg.pl.  

Anyways, that is my opinion.

knnniggett:
Fair enough. We can do both.

knnniggett:
I have added debug level 3 message back into zm_config.cpp.

I have changed the wording slightly to make them appear more informational and less fatal (tried to minimize use of the words "no" and "not" as best I could). 

I'm tired of explaining to end users in the forums, that these messages do not necessarily mean there is a problem. At least three people in the forums have broken their system trying to "fix" these messages. ...probably can't fix stupid, but I'd like to try.

knnniggett:
whoops. missed bracket.

PR #1500, opened 2016-05-24
remove one unlink, add better error reporting to the rest

knnniggett:
On all the (redhat) systems I test on, the unlink statement I commented out with this pr fails every single time. Commenting it out appears to not have any negative affects. With the recent error reporting added to this statement, it was leaving an error in my logs every time zoneminder was started.

@connortechnology let me know if you agree with this. Alternatively, we could put the unlink statement back in, just without any error reporting.

Those of us who have worked on the project long enough know that "harmless" error messages can cause end users to blame all their problems on this message and respond with "OMG! Why don't you fix it now". Consequently, I'd like to nip this in the bud before it happens.

Note that, while I was researching this, I noticed that the zmtrigger and zmcontrol scripts did not clean up their sock files after zoneminder was started (zmcontrol was even left running). I'm writing this down so I remember to look into this sometime after our next release.

### DIFF
--- a/scripts/zmdc.pl.in
+++ b/scripts/zmdc.pl.in
@@ -262,8 +262,7 @@ sub run
     killAll( 1 );
 
     socket( SERVER, PF_UNIX, SOCK_STREAM, 0 ) or Fatal( "Can't open socket: $!" );
-#    Fails every time. Does not appear to be necessary.
-#    unlink( main::SOCK_FILE ) or Error( "Unable to unlink " . main::SOCK_FILE .". Error message was: $!" );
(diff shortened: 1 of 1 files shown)


connortechnology:
I think the unlink might be required... I added the error message at some point. Perhaps we need to just remove the error message.  Funny I've never seen that error message myself though.

knnniggett:
Since the error I'm getting is No Such File or Directory, I'll change it to only call the unlink if the file exists.

connortechnology:
I think that's a good idea.

knnniggett:
No errors on my end after the latest commit

PR #1501, opened 2016-05-24
Fix zone edit

connortechnology:
This fixes some problems with the work we did on streaming zone editing with svg.  

First off, the getImageSrc sets some global variables, that are needed in the js.  So we need to call it before outputting the headers.  So I've split out the functionality of a bunch of routines in includes/functions so that we can either just output the result, or return it as a string to be used later.

And the stuff I was doing with negative margins doesn't work so great, so redo that using absolute positioning which probably makes more sense anyways, and continues to work when you stretch the window wide enough to handle both the image and the svg.


knnniggett:
I'm testing this now.

Something else I justed noticed. We need to make the Zone link on the web console un-clickable (just like the monitor name) when zoneminder is not running.

PR #1506, opened 2016-05-29
Change encoding of web/lang/cs_cz.php from ISO-8859-2 to UTF-8

Disassembler0:
Addresses #583, for czech language.


connortechnology:
Uh... this seems like a good thing to my untrained eye

PR #1507, opened 2016-05-30
Remove deprecated register_globals from Apache/PHP configs

Disassembler0:
register_globals feature has been deprecated as of PHP 5.3.0 and removed as of PHP 5.4.0 - http://php.net/manual/en/security.globals.php

- Debian 7 comes with PHP 5.4.45
- Debian 8 comes with PHP 5.6.20
- Ubuntu 15.04 comes with PHP 5.6.4
- Ubuntu 16.04 comes with PHP 7.0.4

Moreover, if PHP-FPM is used (like in my case), the directive is not recognized at all.

### DIFF
--- a/distros/debian/apache.conf
+++ b/distros/debian/apache.conf
@@ -1,7 +1,6 @@
 Alias /zm /usr/share/zoneminder/www
 
 <Directory /usr/share/zoneminder/www>
-  php_flag register_globals off
   Options Indexes FollowSymLinks
   <IfModule mod_dir.c>
     DirectoryIndex index.php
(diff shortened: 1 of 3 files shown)


knnniggett:
We support Redhat distros as well as Debian based.
CentOS 6 ships with PHP 5.3.3 so this pr must maintain backwards compatibility with that.

Disassembler0:
There are more subdirectories under _distros_, so I sincerely hope that RedHat distros have configs in _redhat_ subdirectory, which wasn't modified by this commit. I have modified only _debian_, _ubuntu1504_ and _ubuntu1604_.
Alternative would be to wrap the line into `<IfModule mod_php5.c></IfModule>`

knnniggett:
I am on my way out the door and had not looked at your code yet. If this only modifies the code under debian distressed then obviously we are good.

PR #1511, opened 2016-06-02
Improve Docker features

TJC:
Hi,
I made some minor improvements to your Dockerfile and docker util scripts.

Cheers
Toby

PS. I note the Docker build doesn't actually seem to work -- apache's document root doesn't match the installed zoneminder location -- but my fixes here help a little anyway.

I also note that you shouldn't really bother with setting up an ssh daemon inside the docker container.
It's much better to just use "docker exec" to get a new shell inside a running container. I have left the ssh stuff however, for the time being.


michaelarnauts:
Looks fine! I hope ssh gets dropped anyway, it shouldn't be in the docker containers. `docker exec -it container bash` should be used for that.

TJC:
Exactly -- the ssh stuff feels like it's been left in from a much older era of Docker and should go now.

michaelarnauts:
Any reasons why this MR can't be merged in?

SteveGilvarry:
Master freeze in place until 1.30.0 is released, and even though this is standalone, I guess it waits, nothing stopping anyone pulling it down to separately in the meantime.

marcosscriven:
Would there be any chance of making the output of this the 'official' Docker image on Docker Hub? There's quite a few publishing: https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=zoneminder&starCount=0

The one from Quantum Object doesn't actually have the MySQL DB setup properly.

knnniggett:
No one on the development team is currently using docker.
You are welcome to become the maintainer if you are interested.

TJC:
@marcosscriven I reckon this one needs some more work before it's promoted as the best, official, image. As I noted in my original pull request, I was improving on the original work, but it's still not fully functional.

However I'd be happy to work with someone to try and achieve that. I use Docker very heavily in production in my day job, so I can help a lot there, but I'm only an amateur when it comes to ZoneMinder.

marcosscriven:
In fact, the Dockerfile here seems to be focussed more on building ZM, rather than being a convenient way to deploy. Using this as the official image would in any case be a bad idea, as there's lots of packages unnecessary for deployment.

I envisaged a much simpler Dockerfile, run whenever ZM is released, that required no more than mapping a certain directory to store ZMs DBs files. 

It could then be listed here: https://www.zoneminder.com/downloads/ as an easy, convenient, cross-platform deployment method.

This however would really require those who release ZM to agree with this strategy. It wouldn't require anything like the sort of maintenance as a Dockerfile used for building, just a version bump of the ZM binary it picks up.

michaelarnauts:
Maybe the guys from linux-server.io are interested?

-> https://forum.linuxserver.io/thread-256.html?highlight=zoneminder

knnniggett:
Docker is a great tool, but at the moment those of us on the development team are working on other priorities. This is one of those things I'd love to have time to play with, but I've always got something else I'm working on. We have and will continue to accept pull requests from strangers for our docker files. 

Months ago, at my local Linux club, I started on a Docker script to deploy a zoneminder rpm from zmrepo into a container, but I did not get very far. 

I looked at the link @michaelarnauts posted and it was interesting that SSL was mentioned. Note that all the zoneminder rpms in zmrepo are configured for SSL out of the box so it would be a matter of just deploying that.

A container solution would be preferable for any scenario where we have a third party app, which does not play nice with zoneminder. Such is the case with Asterisk, since it wants to run as its own user rather than the web account user. 

So yes, convenient way to deploy would be great, but that will require someone new to step up into the role of setting this up. We are usually in our irc channel on freenode if anyone wants to talk discuss it.

Issue #1485, opened 2016-05-16
Release 1.30.0


knnniggett:
This issue has been created as a placeholder to document what we need to address before our next release.

Due to some of the changes we've implemented, I plan to create a draft release before releasing. I estimate the draft release process to take approx. 1 month.

@ ZoneMinder Team
- Personally, I'm fine with calling our next release 1.29.1, instead of 1.30.0, but I'll go with popular opinion. Please voice your opinion on what we want to call our next release.
- We have a lot of open pull requests. Review and discuss which ones should be part of our next release

@kylejohnson 
- Communicate to the rest of the team the state of the telemetry server. We want to make sure it is running reliably before we release. If there is anything needed, such as hardware upgrades, to minimize risk of downtime then let's have that discussion before we release. 

@knnniggett
- ~~Review and test #1411~~
- ~~review and merge #1484~~
- ~~review and merge #1478~~
- ~~Address unlink error message #1500~~ 

@connortechnology
- ~~discuss and implement a solution to #1482 (see pr #1499)~~
- ~~#1484~~
- ~~#1501~~

@pliablepixels
- ~~#1486~~
- ~~#1478 to fix some multi-server problems.~~
- ~~#1470 makes configuring the api a lot simpler.~~

@SteveGilvarry 
- anything?

connortechnology:
I think 1.30 is the correct version  number.  It has onvif turned on, and telemetry.  These are new features.  1.29.1 would be a bug fix release.

We need #1486 in to make @pliablepixels happy.

Also #1478 to fix some multi-server problems.

#1470 makes configuring the api a lot simpler.

josh4trunks:
I'm hoping we get the FreeBSD port/pkg updated with the release of 1.30!

https://github.com/abishai/Zoneminder-port is working built off `ff3e9c8`, but I noticed some functions stopped working when I updated the port to `9377dca`. See https://github.com/abishai/Zoneminder-port/issues/4#issuecomment-218384770

I'll test again on the latest commit and post a bug report if it's still not working.

connortechnology:
We should probably fix #1464 too.

knnniggett:
Well, there is something still messed up with the master branch. I'm placing this hear for now until I've got time to troubleshoot this further:
```
Use of uninitialized value in concatenation (.) or string at /usr/share/perl5/vendor_perl/ZoneMinder/Logger.pm line 156.
[...]

PR #1515, opened 2016-06-03
Have cmake take the version from the version file

connortechnology:
Just putting this here.  Maybe no need to merge yet, I may find other things to change as well. Discuss.

This allows us to only have to change version file instead of both version and CMakeLists.txt.  Which is nice.


knnniggett:
Just to put what we talked about in irc here in writing, what our plans are:
- coordinate with @SteveGilvarry to create an empty, dead-end wiki page which contains the version number we want zmupdate on installed sites to see. Set permissions such that only we can change it.
- coordindate with @kylejohnson to get the url updates.zoneminder.com/version.txt to point to the newly created wiki page, instead of the version file in our github repo
- merge this pr

I believe this will resolve the issues we have surrounding the discussion in #1485

pliablepixels:
Moving this question here (see https://github.com/ZoneMinder/ZoneMinder/issues/1517)

Does this solve the case when building a _pre-release_ package using do_debian, the _filename_ of the package shows a different version?

knnniggett:
Yes! This solves exactly that problem.

pliablepixels:
But how? I just pulled that PR - it does not solve the problem! All that PR does is make sure the version in CMakeLists picks up the version from the version file. The core problem _is_ the version file - which does not say 1.30 for pre-releases (like now). So when I use do_debian to create a package, it creates _1.29*.deb while actually pulling from 1-30rc tag

knnniggett:
Because the overall solution involves more than just this pull request. 
Are you reading all the bullet points stated previously?

We have to get all the zoneminder installations out in the wild to look at a different location for their version number before we can change the version number store in the file "version" in our github repo. Once we do that (see the bullet points), then we will change the file "version" to contain 1.30.0 instead of 1.29.0.

So until all those steps happen, you can't use do_debian.sh without modifying how the script gets its version number, regardless of whether you merge this particular pr into your local branch or not.

pliablepixels:
@knnniggett - right, that is exactly what I was talking about and I had a suspicion we were talking past each other on some points.

My core question, if you could please clarify: In the future, after points a) and b) are addressed, lets say we have a stable v2.30 and you just released v2.31-rc1. Will you be updating https://github.com/ZoneMinder/ZoneMinder/blob/master/version to say 2.31 before 2.31 is released as a stable version? If so, all is good (in the future).

knnniggett:
No, the version number in that file does not change, nor has it ever changed previously, when going from rc1, rc2, ...rcN, and then finally to full release. This is basically a rehash of a much earlier discussion we had about how to differentiate versions of zoneminders between releases. That is up to the package maintainer to do handle that. 

Release candidates are provided as a convenience to the few advanced folks who might want to try it. We aren't pushing these to the masses and are not going to spend a lot of time on our end making sure end users can tell the difference between the different release candidates.

pliablepixels:
sounds good. Then all is fine

knnniggett:
@kylejohnson mentioned in slack that he stopepd update.zoneminder.com from updating the version with the contents of "version" in our github repo so I'm going to merge this. We will find out soon enough if this wasn't the case.

PR #1520, opened 2016-06-06
Set VIDEO_MAX_FRAME if it is not defined

knnniggett:
@abishai to test

@connortechnology to tell me if this is good place to make this change

From reading, it looks like VIDEO_MAX_FRAME used to be defined in v4l1 headers, but then got moved into v4l2 headers.... so any machine with v4l1 but not v4l2 will not build.

### DIFF
--- a/src/zm_local_camera.h
+++ b/src/zm_local_camera.h
@@ -105,6 +105,11 @@ protected:
   static V4L1Data     v4l1_data;
 #endif // ZM_HAS_V4L1
 
+// Required on systems with v4l1 but without v4l2 headers
+#ifndef VIDEO_MAX_FRAME
+#define VIDEO_MAX_FRAME               32
+#endif
+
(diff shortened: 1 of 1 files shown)


abishai:
It's to late to add it there. See https://github.com/knnniggett/ZoneMinder/blob/06923d44ba8ecd0ec02ab424ad998720e369e206/src/zm_local_camera.h#L94

And, as we need the same in cpp, better to put it there https://github.com/knnniggett/ZoneMinder/blob/06923d44ba8ecd0ec02ab424ad998720e369e206/src/zm_local_camera.h#L40

I was able to build ZM successfully after that.

knnniggett:
Right. I meant to put it right after including the v4l headers, but at the time I was in a hurry.

abishai:
I was able to build ZM with that. I think it can be merged.
Thank you!

PR #1521, opened 2016-06-07
update area when points move

connortechnology:


knnniggett:
I've got a less intrusive fix that simply adds a few lines to updateZoneImage

connortechnology:
So let's see it.

knnniggett:
working on it

knnniggett:
Tell you what, I'm juggling too many things right now to finish what I was working on so let me leave some notes here and perhaps we can merge the differences:

- What caused you to add updateArea() to updateActivePoint, delPoint, updateX, updateY? I didn't have to do that.

- In chat you mentioned you borrowed the code to determine the area of the polygon, which sums trapezoids, from our C code. That function is actually incomplete as it is written. It skips the last trapezoid in the series, determined by zone['Points'][n] and zone['Points'][0] where "n" is the last leg of the polygon.  This is how I calculated the area. It should look familiar.

```Java
    var polyArea = 0;
    for ( var i = 0; i < zone['Points'].length; i++ ) {
        var P1 = 0, P2 = 0;
        P1x = zone['Points'][i].x;
        P1y = zone['Points'][i].y;

        if ( i < (zone['Points'].length - 1) ) {
            P2x = zone['Points'][i+1].x;
            P2y = zone['Points'][i+1].y;
        } else {
            P2x = zone['Points'][0].x;
            P2y = zone['Points'][0].y;
        }

        polyArea += (P1y + P2y) * ( P2x - P1x ) / 2;
    }
    polyArea = Math.round(Math.abs(polyArea));

```

connortechnology:
I believe the functions are equivalent, although yours is slightly easier to read.  

delPoint needs an updateArea because the removal of a point can change the area.  We don't need it for add point because it is added on an existing line.  

The rest just looked like they should have them, but I think careful thought and review is in order.  I think most of this code is a mess.

knnniggett:
After applying this pr to my local branch and building, the area still appears as 100% no matter what size I make it.

connortechnology:
So I have put back a necessary call to updateArea.  So now it updates the area.  However while testing I found issues when editing the text fields for the points... so stay tuned.

connortechnology:
Okay, I wasn't able to reproduce the errors I was seeing last night.  However, I did it really annoying that we were still using onchange instead of oninput, so when entering in the x and y points nothing would update until the field lost focus.  So I changed that.  Then I got annoyed because everytime I hit enter it would submit the form.  So I disabled that.

SteveGilvarry:
@connortechnology text fields for points are not quite right, I had examples when I edited x the shape updated, then edited y and the point moved but shape didn't adjust. Firefox in Ubuntu 16.04.

connortechnology:
Hey you are right.... I can reproduce that.  Fix coming shortly

SteveGilvarry:
Seems to be working now. @knnniggett working for you now?
I was playing with an edge case, basically swap top left point with top right point and get two triangles, net result area = 0. No idea if such a zone would work now, I would assume not.

PR #1522, opened 2016-06-07
Move the session start down to where it is actually needed so that we…

connortechnology:
… don't hold it open while generating videos

Without this, the web ui stops completely while video generation goes on.


connortechnology:
Although there is no formal issue for it, a guy was complaining in irc about it.

SteveGilvarry:
Nominating as going in with 1.30 any objections?

PR #1524, opened 2016-06-07
if user=&pass= are in request, use them for auth in api

connortechnology:


SteveGilvarry:
@pliablepixels as primary user of the API for evil, is this change Ninja Approved?

pliablepixels:
@SteveGilvarry - how much time do I have to confirm? I am on travel this week and get get to it Monday - but if you need to make a release I can test over the weekend too.

SteveGilvarry:
@pliablepixels maybe I will setup auth on one of my machines and confirm that zmNinja keeps working, and merge if it does. But you will have time before a release.

iMarvinS:
I´ve just tried to use this authentication method but I don't seem to get it work properly. It gives me back an error which says that the user couldn't be found although username and password are the same as I enter them to login into ZM web interface. What´s the problem here?

PR #1527, opened 2016-06-11
revert recent changes to ConfigfromDB subs

knnniggett:
This fixes the warnings about the xxConfigFromDB functions being redefined.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Config.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Config.pm.in
@@ -45,12 +45,7 @@ use vars qw( %Config );
 our @EXPORT_CONFIG = qw( %Config ); # Get populated by BEGIN
 
 our %EXPORT_TAGS = (
-    functions => [ qw(
-        zmConfigLoad
-        loadConfigFromDB
-        saveConfigToDB
-    ) ],
-    constants => [ qw(
+    'constants' => [ qw(
         ZM_PID
     ) ]
 );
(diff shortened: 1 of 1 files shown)


connortechnology:
Unfortunately this moves backwards, not forwards.  So I'm not happy about that.  

See #1528

knnniggett:
The move backwards was intentional. Right now the priority is to get stable code so we can release. 

You've not given us a good reason why you made this change in the first place, other than you have a preference to have it in a different module. I don't see that as a particularly good reason to go changing code, which has worked fine for literally a decade, especially when we are trying to get a release out the door.

If you've got a good reason to do this, then let's do it after 1.30.0 release, and in a separate branch so we can safely test it, along with all the other changes we backed out recently.

connortechnology:
How about stop reverting bits and pieces of my code that were tested and were approved and were applied! 

Your PR removed the code, but didn't fix the calls which specify the Config module instead of the ConfigAdmin module.

Stop dicking around with my code.  I'm trying to get this release out too, but just reverting code without understanding the effect is not useful. Let's all try to understand the problem, and move forward.

I'm not approving your PR and I don't expect you to approve mine, so we are going to need some other people to come back from wherever they went to.

SteveGilvarry:
Both have valid points, and since I approved the PR's being reverted I guess I better get involved.

It was working so why change? It was part of a larger move to clean up perl modules. Am I correct that originally stemmed from the db connection being made in multiple places, since reverted? As it appears now, the benefit is one less module to build, and the issue was never caused by this change itself but a reversion of DB changes.

But I hear @knnniggett concerns about changing something that works now. So looking at that we have moved the functions without change to a new location and updated how they are called.
From my view it is fairly benign, I lean towards #1528, and only because it removes an extra module. I will have to build and test it.

connortechnology:
My position is just that I like to move forward rather than backwards.  Reverting has similar dangers.  I believe in understanding the problem, and fixing it correctly, rather than just putting it off to another day. 

I believe that my PR fixes the problem, and leaves us with better code.  So I am advocating that.  I was ok with the initial reversing, thinking it would simplify things, but it hasn't.  It has made things more complex.  I feel that we should have investigated, understood and fixed the problem instead.  Too late now.

knnniggett:
@connortechnology STOP FUCKING BREAKING CORE COMPONENTS OF ZONEMINDER AND I WON'T HAVE TO DICK WITH YOUR CODE! 

I come back from vacation and you managed break the master branch and no one had even known it. 
Then I've got to waste my time trying to figure out what happened and how to fix it.
We've spent the last month trying to undo what you did.

THAT RIGHT THERE GIVES ME A LICENSE TO DICK WITH YOUR CODE.

All your changes need to go into a feature branch, and then get merged into master at once, so the rest if us have a chance to figure out what the fuck you did when something else goes wrong.

connortechnology:
No.  It doesn't.  

Your treatment of me since you've gotten back has been bad.  I don't accept that it has anything to do with ZM.

I missed you so much while you were away.  It has been really unpleasant since you've been back though.

I look forward to our development moving forward. To that end, as I have said, I suggest we try to understand the problem, and fix it.  Instead of just reverting.  that is my position, I bow to the will of the democracy.

SteveGilvarry:
Jump on voice and hug it out gents, it is a simple difference of opinions. No ones changes are dicking with the code, they are just trying to solve problems, and happen to have a different approach to your own.
I merged the DB PR, so I broke master, it worked fine for me at the time. @knnniggett I am sorry that it has caused you so many issues, but it was solving an issue with DB connections being created in multiple places. 1.30 was not being even discussed, other than the intention to release more often. @connortechnology was totally agreeable to put that on the back burner so lets move on.

@knnniggett this PR doesn't fully revert yet, https://github.com/ZoneMinder/ZoneMinder/blob/master/scripts/zmupdate.pl.in#L343.
If you are not willing to look at #1528 as an option, then I guess the only solution will be this as a full revert.

SteveGilvarry:
FYI the related error from above when installing this PR against 1.29
Undefined subroutine &ZoneMinder::Config::saveConfigToDB called at /usr/bin/zmupdate.pl line 504.

SteveGilvarry:
#1528 solved it, I merged it.

PR #1529, opened 2016-06-12
connect needs a +1

### LINKED ISSUE #1526: 'soc' problem again - now with zmf
zmf bind code was forgotten - +1 should be added as with zms
https://github.com/ZoneMinder/ZoneMinder/blob/66a3fef61ebd10c45968a3706021d45d027e28be/src/zm_event.cpp#L275

connortechnology:
fix #1526

### DIFF
--- a/src/zm_event.cpp
+++ b/src/zm_event.cpp
@@ -272,7 +272,7 @@ bool Event::OpenFrameSocket( int monitor_id )
   strncpy( addr.sun_path, sock_path, sizeof(addr.sun_path) );
   addr.sun_family = AF_UNIX;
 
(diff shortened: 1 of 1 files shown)


SteveGilvarry:
Nominating for merging to 1.30, any objections?

connortechnology:
Yeah assuming I have managed to add a +1 correctly, I think this should go in.

abishai:
Maybe paste entire solution from zms ? It's checking max length, for example and warnings if it's too long.

SteveGilvarry:
@connortechnology Did you want to implement any of the checking @abishai is talking about, if not I will merge.

PR #1531, opened 2016-06-14
Adjust download file path to swap path to match the file generated.

SteveGilvarry:
I have seen a couple of people mention log export not working, decided to check and sure enough the generated file was changed to ZM_PATH_SWAP, but download path was still /temp.

### DIFF
--- a/web/ajax/log.php
+++ b/web/ajax/log.php
@@ -394,7 +394,7 @@ tr.log-dbg td {
         }
 
         $exportFile = "zm-log.$exportExt";
-        $exportPath = "temp/zm-log-$exportKey.$exportExt";
+        $exportPath = ZM_PATH_SWAP."/zm-log-$exportKey.$exportExt";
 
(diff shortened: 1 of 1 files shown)


knnniggett:
I don't think exporting any extra files to PATH_SWAP is a good idea. For some (most?) of us, PATH_SWAP points to some sort of RAM disk, which will naturally cause problems if the size of data being exported is larger than available ram.

Instead, I would recommend using ZM_TMPDIR, which is currently set at build time. At least on redhat based systems, this points to a folder on a real filesystem. We could expose this to the Options -> Paths if it made sense.

This applies to #1540 as well.

SteveGilvarry:
@knnniggett Yes I was only fixing the mismatch between generated file and web path, pretty much 18 months it was not working at all. If Web can access zm_tmpdir see no real issue where the file is put. Should we be deleting it also?

PR #1544, opened 2016-06-18
Feature-h264-videostorage - Fix missing } from merge that was preventing building.

SteveGilvarry:


connortechnology:
Thanks.. I thought I already fixed this, but maybe forgot to push it...

PR #1546, opened 2016-07-03
Updates to Ubuntu and Debian install guides

SteveGilvarry:
Can see the changes on my readthedocs account. http://sgzoneminder.readthedocs.io/en/docsupdate/index.html

### DIFF
--- a/docs/installationguide/debian.rst
+++ b/docs/installationguide/debian.rst
@@ -1,37 +1,154 @@
 Debian
 ======
 
-A fresh build based on master branch running Debian 7 (wheezy)\:
+.. contents::
+
+Easy Way: Debian Jessie
+-----------------------
+
+**Step 1**: Setup Sudo
+
+By default Debian does not come with sudo. Log in as root or use su command.
+N.B. The instructions below are for setting up sudo for your current account, you can
+do this as root if you prefer.
+
+::
+    
+    aptitude update
+    aptitude install sudo
+    usermod -a -G sudo <username>
+    exit
+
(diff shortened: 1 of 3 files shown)


knnniggett:
Step 12, in the Debian guide - Shouldn't the "service" command be replaced with the equivalnet "systemctl" commands in that step?

Not a really big concern, but I would ask we keep our documentation as consistent as we can in the way we do things. For example, it would be nice if we could all make reference to the same text editor. I don't really care which one we choose. Perhaps we should scan through our documentation and make all references use nano, instead of vi.

Does Debian not have apt and only aptitude? Just curious?

Also, I think a community member submitted a documentation pr some time ago. It looks like this pr will make  #1433 partially obsolete

SteveGilvarry:
Step 12 old habits, mind you service just calls systemctl and works, will change that.

Using nano for 2 reasons, firstly I based the Ubuntu one of bbunge's wiki, which he has tested and refined for the ID10T. Secondly vi is just being mean to new linux users. I didn't edit the 14.04 section that much so let me check that as I think it has a vi example still.

No idea why I chose aptitude in debian build?

That sleep command it the only change of substance and only for 14.04, and like I said above I didn't do much in there. I will take that part and put in 14.04 here.

SteveGilvarry:
@knnniggett How is it looking now?

knnniggett:
I've done a search through the project, and it seems the only place we mention the name of a text editor are in the debian and ubuntu guides so the consistency requirement has been satisfied.

PR #1548, opened 2016-07-03
fix jpeg buffer too small

### LINKED ISSUE #1547: libjpegturbo 1.5.0 causes streaming to break
Symptoms:
On a machine with the latest libjpeg-turbo 1.5.0, live streaming is highly intermittent/stutters and the zoneminder log files fill with "Buffer passed to JPEG library is too small"

First reported on Arch Linux here:
https://forums.zoneminder.com/viewtopic.php?p=95165#p95165

I discovered this problem also exists on Fedora 24.

Reproducible: Always
Steps to Reproduce: Install and run ZoneMinder on any machine with libjpeg-turbo 1.5.0.
Temporary Workaround: roll back to libjpeg-turbo 1.4.2

Research:
It seems this is a new "feature" of libjpeg-turbo:
https://github.com/libjpeg-turbo/libjpeg-turbo/issues/78

The commit description provides background on the change they made:
https://github.com/libjpeg-turbo/libjpeg-turbo/commit/68cf83db5647104d9d7ea289c399fe6a5c77da64

TLDR: We are apparently "doing it wrong" when we recycle the cinfo structure in cases when it has been called from other parts of ZoneMinder. I have a fix for this that is incoming.

Additional topic of discussion/possible scope creep:
Upon close inspection of the ZoneMinder source code, I noticed that we have duplicated some of the jpeg library source code, added a "zm_" prefix to the function calls, and placed it into zm_jpeg.cpp. This seems to be the root cause of the problem. Perhaps @mastertheknife could explain why this was done this way, if he is around, since he has done quite a bit of work in this part of the project.

knnniggett:
fixes #1547 

This PR attempts to solve the issue by splitting out jpg_dcinfo and jpg_ccinfo into four different structures.
jpg_dcinfo becomes readjpg_dcinfo and decodejpg_dcinfo
jpg_ccinfo becomes writejpg_ccinfo and encodejpg_ccinfo

By splitting these out in this manner, ZoneMinder no longer tries to use a cinfo structure, which was created in a different part of the ZoneMinder source code, which is what caused the libjpeg-turbo library to complain with ERREXITJERR_BUFFER_SIZE.

Note that a better solution could possibly be obtained by eliminating the duplicate jpeg library source code found in zm_jpeg.cpp. However, I do not have a clear enough understanding of why it is there in the first place to attempt it at this time.

### DIFF
--- a/src/zm_image.cpp
+++ b/src/zm_image.cpp
@@ -35,8 +35,10 @@ static short *g_u_table;
 static short *b_u_table;
 __attribute__((aligned(16))) static const uint8_t movemask[16] = {0,4,8,12,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF};
 
-jpeg_compress_struct *Image::jpg_ccinfo[101] = { 0 };
-jpeg_decompress_struct *Image::jpg_dcinfo = 0;
+jpeg_compress_struct *Image::writejpg_ccinfo[101] = { 0 };
+jpeg_compress_struct *Image::encodejpg_ccinfo[101] = { 0 };
+jpeg_decompress_struct *Image::readjpg_dcinfo = 0;
+jpeg_decompress_struct *Image::decodejpg_dcinfo = 0;
 struct zm_error_mgr Image::jpg_err;
 
 /* Pointer to blend function. */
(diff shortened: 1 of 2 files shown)


SteveGilvarry:
Confirmed issue was present in Debian Stretch running 1.29.0, updated to 1.30.0-rc1 still present, applied this PR and issue resolved.

PR #1549, opened 2016-07-04
Implement suggested wait for MySQL start

SteveGilvarry:
@knnniggett as you suggested in #1433. I removed anything in 14.04 install docs about inserting sleep. I need to do a test build myself, but putting up for comments. @connortechnology

### DIFF
--- a/distros/ubuntu1204/zoneminder.init
+++ b/distros/ubuntu1204/zoneminder.init
@@ -23,6 +23,13 @@ command="$ZM_PATH_BIN/zmpkg.pl"
 
 start() {
 	echo -n "Starting $prog: "
+        # ATB - wait for mysqld to start. Continue if it takes too long.
+            count=0
+            while [ ! -e /var/run/mysqld/mysqld.sock ] && [ $count -lt 60 ]
+                do
+                    sleep 1
+                    count=$((count+1))
+            done
 	export TZ=:/etc/localtime
 	mkdir -p "$RUNDIR" && chown www-data:www-data "$RUNDIR"
 	mkdir -p "$TMPDIR" && chown www-data:www-data "$TMPDIR"


connortechnology:
I like it

SteveGilvarry:
Ah just thought what about if SQL is not local? I know it will continue after 60 secs and still work. Just thinking about it...

connortechnology:
Zms.conf should be parsable by bash. Should be able to just source it and check contents of dB host

On 5 July 2016 00:30:45 CEST, Steve Gilvarry <notifications@github.com> wrote:
>Ah just thought what about if SQL is not local? I know it will continue
>after 60 secs and still work. Just thinking about it...
>
>
>---
>You are receiving this because you were mentioned.
>https://github.com/ZoneMinder/ZoneMinder/pull/1549#issuecomment-230360290

-- 

abishai:
The best way is parse zm.conf with commands like awk, for example
```awk -F "=" '/^ZM_DB_HOST/ {print $2}' {$CFG}```

knnniggett:
Well, one has to ask at what point do we cross the line of diminishing returns, for an init system that has been EOL'ed. 

We could check the dB host to verify it equals "localhost". If it is not then skip the check, and hence the delay. However, that won't work if the end user changes the localhost entry to the hostname of the local machine.... so now we need to check the hostname too. Does a minimal installation of Ubuntu 14.04 (server) ship with awk? I don't know the answer to that.

I'm fine with merging this as it is currently, based on the assumption that those who are able to configure zoneminder for a remote database should also be able to tweak their zoneminder init script. After all, that is what most users have already been doing.

I'll wait to merge this in case someone wants to tack on another commit to modify the check. If not, I'll merge it at it is tomorrow.

connortechnology:
Agreed about the question of whether it is worth it. However if someone is going to do the work I'm happy to include it.  I'm pretty sure basic Ubuntu has awk. I also think that a test for empty or localist is sufficient.

abishai:
I was writing the similar script for FreeBSD (it has init system and I doubt it will be deprecated in near future) and noticed that you do the same. awk should be available on any Unix-like system. I ended using awk and mysqltest tool to actually connect to mysql with zm's credentials. (mysqltest is a part of mysql-client, so ZM will pull it).

If Linux folks deprecated init system, maybe you are right. :)

abishai:
BTW, I'd like to ask a question - I've noticed that mysql shutdown influences ZM not in a good way. To say it simple, ZM dies. Do you think that ZM should recover itself (wait patiently for database connection, re-connect and restart it's processes)? If yes, we don't need to check mysql connection at all.

knnniggett:
@abishai Let's keep the discussion relevant to the topic of adding a delay to the init script. Please take general discussion and questions to our irc channel.

PR #1556, opened 2016-07-09
custom banner text on web console

knnniggett:
This is intended to go into the release after 1.30.0. I thought I'd create the PR now while I am thinking about it.

Each time I package a development snapshot or release candidate into an RPM, I patch the source code during the build to display a disclaimer, to make it obvious to the end user this is not an official release.

This PR allows me to get away from patching the source, and at the same time allow others to put their own custom text near the top of their web console. See screenshot below.

Currently I've statically assigned it to be cantered, H3 html text, because that suits my purposes. If anyone want to add the ability to control things like the font size, boldness, etc I am open to suggestions.

screenshot from 2016-07-09 13-17-17

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
@@ -1549,6 +1549,20 @@ our @options =
         category    => "web",
     },
     {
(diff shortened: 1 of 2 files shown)


knnniggett:
@SteveGilvarry @connortechnology 
Thoughts? Comments? Merge?

PR #1565, opened 2016-07-24
Enable local and travis ccache

SteveGilvarry:
On local build use ccache to speed up build if found, and use on travis. My testing shows once cached resulted in a halving time to build on travis from 14mins to 7mins. ffmpeg make down to 232 secs from 640 secs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ notifications:
 branches:
   except:
     - modern
+cache: ccache
 addons:
   sauce_connect:
     username: "zoneminder"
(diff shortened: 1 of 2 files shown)


SteveGilvarry:
I will run the travis build job again for the PR as first time will create the cache, same thing in master first run once this is merged won't improve, second run does. For PR's it looks for PR cache, if not found then master. I think that might even be it looks for the branch you are merging, and then master.

SteveGilvarry:
Normal build, and second run once cache was populated.

screen shot 2016-07-24 at 10 56 00 am
screen shot 2016-07-24 at 11 51 10 am

knnniggett:
Sounds harmless enough (famous last words)....

PR #1566, opened 2016-07-25
Control script for Keekoon cameras. 

gerdesj:
Tested with KK002.


knnniggett:
@gerdesj Thanks for doing the work. 

It looks like Travis failed the build because the ffmpeg site refused our attempt to git clone. I can't check it out from the command-line either so it looks like the ffmpeg site is borked.

In any case, we are in a code freeze as we test out our latest release candidate. Once 1.30.0 is released, we will merge this pr.

gerdesj:
Aah, ffmpeg continues to delight.  I'm still mentally scarred by the libav/ffmpeg thing.  My first ever public git PR goes down in flames.

SteveGilvarry:
Also need an sql update script for existing installs, I assume 1.30.1, but I do hate that sql updates are tied to software version.

knnniggett:
@SteveGilvarry We have stopped including sql update scripts for new ptz control scripts to avoid having to bump the ZoneMinder version each time someone contributes a new scripts. Instead, one can use zmcamtool.pl to import any missing scripts into an existing system. There is a previous conversation between myself and @connortechnology somewhere around here.

In order to make this happen automatically, one can execute zmcamtool.pl as part of the packaging scripts. I have already done that for redhat: https://github.com/ZoneMinder/ZoneMinder/blob/master/distros/fedora/zoneminder.f24.spec#L93

PR #1567, opened 2016-07-28
Use av_dict_parse_string to create AVDictionary of options

SteveGilvarry:
Use av_dict_parse_string to split ffmpeg options, and remove creating strings to split them in next step for rtsp_transport.


knnniggett:
Builds against ffmpeg 2.6.4, our new minimum supported version, on CentOS 6. Pushing the shiny green button.....

PR #1576, opened 2016-08-02
include polygons for the other zones.

connortechnology:
So this restores the old behaviour by adding svg polygons for the zones.  

I have concerns regarding colours. In 1.29 the editing zone was green.  

In 1.30, we use a css class to assign colour based on the zone type and we got red.  

The Alarm Colour RGB has no effect.  Maybe it should.


knnniggett:
@connortechnology Please see feedback from the user forum:
https://forums.zoneminder.com/viewtopic.php?p=95838#p95838

PR #1577, opened 2016-08-02
fix error in calling zmaControl

connortechnology:
zmaControl can take an id #, so need to move the check for local server test down.

I have kept this minimal, to go in 1.30.1, but I intend to really clean this up.

zmaControl should take a hash only, like zmcControl.

### DIFF
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -840,10 +840,10 @@ function zmcControl( $monitor, $mode=false ) {
 }
 
 function zmaControl( $monitor, $mode=false ) {
+  if ( !is_array( $monitor ) ) {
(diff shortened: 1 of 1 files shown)


knnniggett:
Just so we know, we are/were the symptoms or error message that this fix resolved?

connortechnology:
So, editing a zone, when saved calls zmaControl to restart the zma daemon.  An error will be logged because on line 843 when testing for if the monitor belongs to this server, we treat $monitor as a hash when it could be a scalar monitor id.

knnniggett:
Good to know in case someone reports it. We can tell them it has already been fixed.

PR #1581, opened 2016-08-03
quote the password

### LINKED ISSUE #1571: zmupdate.pl does not quote $dbPass for external command executions
For anything but the simplest passwords, this obviously will fail.

connortechnology:
fixes #1571

### DIFF
--- a/scripts/zmupdate.pl.in
+++ b/scripts/zmupdate.pl.in
@@ -460,7 +460,7 @@ if ( $version )
             $command .= " -u".$dbUser;
             if ( $dbPass )
             {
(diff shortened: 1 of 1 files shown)


jcasale:
Doesn't that miss the usage above at [line 422](https://github.com/ZoneMinder/ZoneMinder/blob/25d58f51923735764f6cb1a72314105a44fd2ee0/scripts/zmupdate.pl.in#L422) ?

connortechnology:
Yeah dang.

connortechnology:
Ok, So I just applied a commit that gets the other one.  Too simple to bother with a PR.

PR #1586, opened 2016-08-08
fix the logic problem where the if logic was reversed and we weren't …

connortechnology:
…looking for a slash anymore hence the socket test wouldn't get run

Comments ?  Am I right?

@josh4trunks


josh4trunks:
if you implement this here, you should also implement this in a few other places. 3 different perl files, 2 CPP, 2 PHP (including one that passes info to Cake)

josh4trunks:
this wouldn't work for cases where `ZM_DB_HOST=REMOTE_HOST:3307`

PR #1591, opened 2016-08-12
Timezone check

connortechnology:
Have noticed this causing problems a few times, after spending hours trying to figure out why auth isn't working.

Implements a warning so that hopefully no one spends hours on this again.


web/index.php
```diff
@@ -67,10 +67,15 @@
 // Use relative URL's instead
 define( "ZM_BASE_URL", "" );
 
+$system_timezone = preg_replace( "/\r|\n/", '', file_get_contents('/etc/timezone') );
```
knight-of-ni: The file /etc/timezone is Debian-specific. It does not exist on an rpm based distro

connortechnology:
Ok, can you give this a try on a redhat system?

knnniggett:
Whatever we end up doing, I think we should move this into functions.php.

The comments aren't quite right.
The third option, labeled for Mac's and older linux's, is actually for all newer linux's, debian and redhat alike, which use systemd. It is the other checks which are for older linux's. FreeBSD appears to use /var/db/zoneinfo so we would have to add that too.

However, I think we ought to look for a better way to do this. I need to take a look at how the time element is added to auth hash. Perhaps we can change how that is done so it does not matter if the php timezone differs from the local system timezone.

connortechnology:
Ok, I don't disagree.  I really think that it should go into the wizard that we plan to make. It should be a test that gets run when weird things are happening, instead of on every hit. 

I can clean up the comments (this code was cutnpasted from elsewhere) and stick it into a function in functions.php, not a problem.

SteveGilvarry:
OK so are we doing the mentioned change to put in functions.php now, or fixing the comments?

web/includes/functions.php
```diff
@@ -2129,4 +2129,37 @@ function getStreamHTML( $monitor, $scale=100 ) {
     }
 } // end function getStreamHTML
 
+function check_timezone() {
+  $system_timezone = '';
+  if (is_link('/etc/localtime')) {
```
kylejohnson, line 2134: On freebsd, `/etc/localtime` is a file.  Generally the file is copied from `/usr/share/zoneinfo/`, not symlinked.
Perhaps use `is_file()`, as that function appears to handle both files and symlinks?

web/includes/functions.php
```diff
@@ -2129,4 +2129,37 @@ function getStreamHTML( $monitor, $scale=100 ) {
...
+  }
+
+  $php_timezone = ini_get('date.timezone');
+
+  // Check time zone is set
+  if (! $php_timezone || !date_default_timezone_set(ini_get('date.timezone'))) {
+    date_default_timezone_set('UTC');
+    Fatal( "ZoneMinder is not installed properly: php's date.timezone is not set to a valid timezone" );
+  } else if ( $system_timezone != $php_timezone ) {
+    Warning( "System timezone is set to $system_timezone, but php's is set to $php_timezone. This can cause weird problems." );
```
kylejohnson: `This can cause weird problems.` doesn't sound specific enough to me.
Maybe `This can cause streaming problems when authentication is enabled.` ?

web/includes/functions.php
connortechnology: problem is that the contents of the file are not really parsable into a timezone.   The fact that it is a link is useful because we can read the timezone from the actual file name.

connortechnology:
I give up.

PR #1592, opened 2016-08-12
Fix analysis frame

connortechnology:
So getImageSrc expects a frame array, not an id, or an object.

So I made the frame object be upper case, to distinguish it from the frame array that we loaded from the db, and pass that instead of the frameid.


connortechnology:
So I had to add a bit more to this.  the view=image had no support for specifying capture or analyse image.
It will now.

knnniggett:
Previous behavior did not require one to click on the image to get the red blobs to appear.

connortechnology:
yeah I got the order of results in iAnalImage ? 'analyse' : 'capture' wrong.  fixed I think.

knnniggett:
works for me.

Issue #1601, opened 2016-08-27
Messages fail to send when using SSMTP


mdrush:
ZoneMinder has functions sendEmail and sendMessage in zmfilter.pl that are used when sending emails and messages. It appears that only the sendEmail function contains support for the SSMTP_MAIL option, probably due to the instructions at the bottom of [this page](https://wiki.zoneminder.com/How_to_get_ssmtp_working_with_Zoneminder) that only mention modifying sendEmail.

PR #1602, opened 2016-08-27
added missing ssmtp support in sendMessage

mdrush:
#1601


connortechnology:
Thank you for your submission, we really appreciate it!

First off, something going wacky with the indentation.  We decided to follow the google code style guidelines, so that means 2 space indents and no tabs.  

Second, if ssmtp wasn't installed, the which would fail.  I think it should maybe fall through to MIME::LIte->send.

PR #1607, opened 2016-09-01
Fix issue #1460

### LINKED ISSUE #1460: Remove stupid code in zm_image.cpp
around line 270 in zm_image.cpp in the initialize func, we algorithmically populate some colour tables.

This function gets called a lot.  There is no reason to generate these algorithmically.  We should just hard code them in zm_image.h and assign a pointer to them or something like that.

mattdurant:
Not sure if this is exactly what you had in mind for a permanent fix, but it does resolve #1460


connortechnology:
Yes, love it.  I'll do some testing.

kylejohnson:
Why not delete those lines instead of commenting them out?

connortechnology:
Well, I'd like them saved for future reference as to how the tables were generated.  They should probably be moved up to where the tables are.

connortechnology:
This code has made about a 1-2% reduced cpu load on my cpus per process.

PR #1609, opened 2016-09-01
implement suggested code to stop the monitor when Function is set to …

### LINKED ISSUE #1606: stopping zmc when setting monitor function to None trough APIs
Hello,

my problem is explained here :
https://forums.zoneminder.com/viewtopic.php?f=34&t=25205

in short, if i set my monitor function to None trought ZM web interface, zmc process is stopped and i can't see live stream (expected behavior).
but if i set it to None trough APIs, zmc process is stopped but then restarted, so i can see live stream perfectly (unexpected behavior).

to work around this i modified the edit method of MonitorsController class in this way:
```
[...]
                $this->set(array(
                        'message' => $message,
                        '_serialize' => array('message')
                ));

                $func = $this->Monitor->find('first', array(
                        'fields' => array('Function'),
                        'conditions' => array('Id' => $id)
                ))['Monitor']['Function'];

                if ($func == 'None') {
                    $cmd = 'stop';
                }
                else {
                    $cmd = 'restart';
                }

                // - restart this monitor after change
                $this->daemonControl($this->Monitor->id, $cmd);
        }

[...]
```

i did a lot of tests and that actually solved my problem, but i'd like to know your opinion about this.
in particular:

- is this problem known? someone else experimented it?
- is it right to stop zmc if monitor function is None? why using APIs it is currently restarted?

thank you.

connortechnology:
…None

implements the suggested code to fix #1606

### DIFF
--- a/web/api/app/Controller/MonitorsController.php
+++ b/web/api/app/Controller/MonitorsController.php
@@ -137,9 +137,18 @@ public function beforeFilter() {
 			'message' => $message,
 			'_serialize' => array('message')
 		));
-		// - restart this monitor after change
-		// We don't pass the request data as the monitor object because it may be a subset of the full monitor array
- 		$this->daemonControl( $this->Monitor->id, 'restart' );
+
+		// - restart or stop this monitor after change
+    $func = $this->Monitor->find('first', array(
+          'fields' => array('Function'),
+          'conditions' => array('Id' => $id)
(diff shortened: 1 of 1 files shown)


erelson:
@connortechnology I noticed from the diffs that there's a mix of tabs and spaces for indenting.

By anyways, thanks! I can confirm this fixes the issue!

PR #1614, opened 2016-09-03
1537 zone controller bug

### LINKED ISSUE #1537: Zones.json doesn't support pagination
If I curl http://server/zm/api/zones.json it appears to return a list of all available zones. Now that I've been adding and deleting zones, I noticed that only 20 zones are returned, whilst I have a few more. 

I don't see any pagination item, so there's no way to tell if there are more zones. Also, http://server/zm/api/zones.json?page=1 does nothing more that return the same 20 zones.

I am able to get the neccessary zone info by getting http://server/zm/api/zones/44.json, but seeing as deleting zones leaves holes in the zone numbering (zid), I have no way of knowing when to stop "getting zones" one by one. 

I know I'm using an undocumented API feature, but it would save me a lot of unneccessary calls if it would work as suspected ;).

Am I doing something horribly wrong, or is this a bug?

(using 1.29 cent os 7) (also posted in forum https://forums.zoneminder.com/viewtopic.php?f=34&t=24953)

pliablepixels:
Maybe Fixes #1537 by returning all zones. Doesn't add pagination. Not sure its really needed. 
 I tried with 22 zones and it seems to work.
@earlymeadow - give this a try?

### DIFF
--- a/web/api/app/Controller/ZonesController.php
+++ b/web/api/app/Controller/ZonesController.php
@@ -12,7 +12,7 @@ class ZonesController extends AppController {
  *      
  * @var array
  */     
-        public $components = array('RequestHandler');
+public $components = array('RequestHandler');
 
 public function beforeFilter() {
         parent::beforeFilter();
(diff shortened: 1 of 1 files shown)


earlymeadow:
I'll check it out this week and let you know.

earlymeadow:
The web/api/app/Controller/ZonesController.php in this commit works (with the 50 zones I have) very well, thank you so much. I'll be expanding my system to see how far I can go. ...../zm/api/zones.json now returns all defined zones. I too don't think pagination is neccessary. 

Also, the .....zones/forMonitor/X.json works too - only it's not documented ;) I tried to see if I could update the doc, but got a 404 when clikking the "Edit on GitHub" link.

Once again, thanks for the fix.

pliablepixels:
👍 sounds good, thanks for the report

PR #1617, opened 2016-09-09
Update German translation

coracis:
This is a result of apapting Zoneminder to my employee's needs. Further changes have been made to a branch German_adaption_beyond_simple_translation, which especially tries to make Zoneminder more understandable to non-developers (like renaming Modect and Mocord).

I added a new translation entry for Montage Review to en_gb.php but did not run updateLangs.php yet. I hope this does not break MR for languages other than English.

I have also removed the console link in MR, because users usually call MR from the console. Thus if they click on Console in MR they end up having _two_ consoles.


knnniggett:
I've never run updateLangs.php and I'm not sure what it is for. 

It hasn't been modified since we brought the project into github and predates the translate function call. In any case, if the translate function can't find a translation key then it defaults to the english translation, rather than respond with null or an error.  ....so I think we are good.

coracis:
FYI updateLangs.php adds entries that are missing in the translation files. I've just tried it. The script also adds beautiful "Added date" comments to the end of the line cluttering all the translation files. :) Thanks for the prompt merge btw.

PR #1625, opened 2016-09-14
Storageareas

bhaal2:
Modified web code so that filters now uses database table Id's instead of filterName's when editing filters, including saving changes.


bhaal2:
Fixed up order by, and modified delete to use Id's instead of Names

PR #1626, opened 2016-09-15
Storageareas

bhaal2:
Disk usage now displays for all storage areas.  Cannot personally verify this will work when there is no default, @connortechnology should be able to.


connortechnology:
Holy crap, you are burning through my todo list!  Good job!  I will peruse this a bit later after I get the kids off to school.

PR #1639, opened 2016-09-27
store the scale value for watch and event views in a cookie, differen…

connortechnology:
…tiated by monitorId.

Why monitorId?  Because I figure if you are using the scale function, it is likely because your camera is super high def and you want it to always scale.  But you might have older cameras that you don't scale.

### DIFF
--- a/web/skins/classic/views/js/event.js
+++ b/web/skins/classic/views/js/event.js
@@ -13,7 +13,7 @@ function changeScale()
     var newHeight = ( baseHeight * scale ) / SCALE_BASE;
 
     streamScale( scale );
-    Cookie.write( 'zmEventScale'+eventData.monitorId, scale, { duration: 10*365 } );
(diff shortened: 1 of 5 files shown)


connortechnology:
I have fixed the bracket problems.  This should be straightforward now.

web/skins/classic/views/js/event.js
```diff
@@ -13,6 +13,7 @@ function changeScale()
     var newHeight = ( baseHeight * scale ) / SCALE_BASE;
 
     streamScale( scale );
+    Cookie.write( 'zmEventScale'+eventData.monitorId, scale, { duration: 10*365 } );
```
kylejohnson: I think `monitorId` needs to be `MonitorId` per the below change in `eventData`
Cookie is being created as `zmEventScaleundefined`

connortechnology:
Huh... good catch.

PR #1641, opened 2016-09-27
Fix auth timeout

connortechnology:
This code stores a timestamp in the session for when the hash was last generated.  If an hour has gone by, ajax queries will generate and return a new auth hash.

Without this code, any stream will die after 2 hours when using auth hash.

I also used roudn instead of sprintf to do rounding.  sprintf does not use proper rounding rules and is a little heavy for the purpose.


connortechnology:
This is at least working for me now.

knnniggett:
This is on my to-do list. I just need to block out some time to review and test it.

knnniggett:
I'm not able to duplicate this issue exactly as you have described. I've got a monitor, which has been running fine for 2.5 hours now. 

As you can see I'm using hashed authentication:
https://localhost/cgi-bin-zm/nph-zms?mode=jpeg&scale=100&maxfps=15&buffer=1000&monitor=2&auth=6289c35eb27eb25cef1b048bd3c1283b&connkey=686549&rand=1475704611

ZoneMinder uses the hashed authentication just once to initiate the stream. Once it starts, then it does not retry that hash unless the stream fails for some reason.... so is that perhaps the conditions you are referring to? Not necessarily after 2 hours, but after 2 hours and a stream interruption of some sort?

connortechnology:
You might be right about it requiring stream interruption.  I wonder about the fps updates, etc though.  Will have to look closer.

connortechnology:
I give up

PR #1643, opened 2016-09-29
Add a TZ env setting for systemd.

connortechnology:
Unless there is a TZ env set, things like gettimeofday will stat /etc/localtime on every call.  We make A LOT of those calls.  So this actually makes a noticeable cpu and io load reduction.


knnniggett:
Yeah, this looks safe. /etc/localtime appears to be part of systemd itself, and does not appear to be specific to just one distro.

Issue #1649, opened 2016-10-10
Login form - username field autocapitalizes, autocorrects on mobile browsers


jvogt:
Opening this issue so I can submit a PR to update this

PR #1650, opened 2016-10-10
Disabled autocorrect, autocapitalize, spellcheck on username field

jvogt:
This supports a friendlier mobile experience when typing username on login form.

#1649


kylejohnson:
This resolves #1649

PR #1657, opened 2016-10-19
change fast_delete to off

### LINKED ISSUE #1597: PurgeWhenFull deletes too many events as zmaudit can't delete fast enough when OPT_FAST_DELETE enabled
I'm using Ubuntu 16.04.1LTS with ZoneMinder 1.30.0.

Consider the following scenario:
  1. `OPT_FAST_DELETE` is enabled
  1. `zm_audit.pl` is running daemonized as per default config
  1. Disk has only 5% free space left
  1. Default `PurgeWhenFull` filter kicks in, deleting the 100 oldest event
  1. `zmaudit.pl` doesn't delete the files immediately (this might be the bug)
  1. `PurgeWhenFull` runs again, deleting another 100 event until disk space eventually freed up

I myself lost around 600 events because of this.

As a workaround, I have disabled `OPT_FAST_DELETE`. I probably might also use cron to run `zm_audit.pl` instead of daemonizing it, but I think the root of the problem should be addressed instead.

If you need any data reproducing the issue please let me know!

knnniggett:
fixes #1597

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
@@ -1700,7 +1700,7 @@ our @options =
     },
     {
         name        => "ZM_OPT_FAST_DELETE",
-        default     => "yes",
+        default     => "no",
         description => "Delete only event database records for speed",
         help        => qqq("
             Normally an event created as the result of an alarm consists of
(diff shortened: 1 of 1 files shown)


connortechnology:
With fast delete off, perhaps we should change the audit frequency.  I like to run mine once per day or less. It is pretty io intensive, as it will hit every event.

knnniggett:
I'd rather we just make this one simple change now, let it ride, and then consider other options at a later time. I think the way we approach this is going to change once we start recording large video files to disk, rather than many small jpegs.

connortechnology:
Well, my other thought is that so far since 1.30.0 we have merged mostly bug fixes and are in a reasonable state to release 1.30.1.  This is a policy change, which I think should wait for 1.31.0.  Of course we can sort that out if/when we ever release 1.30.1 by just leaving it out.

knnniggett:
Yeah, waiting until 1.31.0 is fine with me.... As I mentioned earlier, I've already implemented this change in all zoneminder packages in zmrepo. It's been that way for at least a year. ...but I see you merged so this is moot.

PR #1665, opened 2016-11-01
Console, etc, restyle

kylejohnson:
This replaces #1337 due to that merge going so poorly.

I choose to leave out any javascript changes (no inclusion of bootstrap.js or jquery.js) due to conflicts with mootools - another time.  I also did not make any changes to the `monitor` view.

The montage views needed a little bit of love since the Monage dropdown in the header required jquery and bootstrap.js - as such I've restyled those pages to add a subheader.

Also a few changes to how we display the google recaptcha, if enabled:  The way it was coded before resulted in invalid HTML.

I need a few people to test this a bit, report any bugs (which should be minimal due to no JS changes), and I'll get them fixed.

2016-11-01-142833_1122x1020_scrot
2016-11-01-142823_1122x1020_scrot
2016-11-01-142808_1122x1020_scrot
2016-11-01-142753_1122x1020_scrot
2016-11-01-142746_1122x1020_scrot


connortechnology:
First glance comments:
I still disagree with putting the bandwidth and disk use at the bottom.  
You have reverted my show id on console code.
You appear to have removed MontageReview.  One thing to note when you put it back: It doesn't need the stream permission, it needs the canViewEvents permission.
Also, what's with the inline styles on the login?  You can create a css file that will only get included for that view....

kylejohnson:
> I still disagree with putting the bandwidth and disk use at the bottom.

Where do you think a better place would be?

> You have reverted my show id on console code.

Oops, I see that.  I'll get it re-added.

> You appear to have removed MontageReview. One thing to note when you put it back: It doesn't need the stream permission, it needs the canViewEvents permission.

Nope, it is still there.  See the first image that I uploaded above.

> Also, what's with the inline styles on the login? You can create a css file that will only get included for that view....

Honestly can't recall my reasoning for this.

kylejohnson:
@connortechnology I've re-added your Monitor ID on Console changes, and I've moved the login.php css to its own file.  
What are your thoughts about where to put the footer information?

knnniggett:
Yeah, I refer to see the load and disk use at the top too. Reminds me we should add /dev/shm usage in a separate pr.

connortechnology:
screenshot_2016-11-02_09-00-56

This is what I've got going in StorageAreas.  So basically the second line is status.  Now I'm running into problem because MontageReview is forcing the logged in part down to another line.  Also, On my cilent's machines there are so many storage areas that it is taking a second line to list them as well. So I was thinking of making that into a mouseover popdown with the detailed stats, and just have an overall average % when not moused over.  I was also thinking of colour coding it to warn of errors/full conditions.
Same with the load.  With multiple servers, it would be nice to see the load on all of them.

So anyways, in summary, they should go in line two of the header.

kylejohnson:
`groups` aren't working right.

kylejohnson:
Couple recent screenshots:

screen shot 2016-11-24 at 10 00 33 pm
screen shot 2016-11-24 at 10 00 37 pm
screen shot 2016-11-24 at 10 00 42 pm
screen shot 2016-11-24 at 10 00 46 pm

connortechnology:
You are doing google code style changes in this.  We cannot abide that.

kylejohnson:
I'm calling this ready for testing.  Please report any bugs.

knnniggett:
I really like the overall layout and I like how you have eliminated at least some of the popsup, opting to embed them into the main browser window.

I did notice a couple things, however.

This is a screenshot of the start/stop/restart popup. The contents are too big for the window.
screenshot from 2016-12-13 14-01-19

This is the log screen. The majority of the screen is filled with the selection boxes, which leaves little room for the events themselves. I guess more importantly, you see that no events are populating. 
screenshot from 2016-12-13 14-01-43

knnniggett:
I went ahead and built this pr on CentOS 7, which ships with php 5.4.16.
The log entries show up correctly, so perhaps we are looking at something related to php 7, since that is what the previous test was running on.

Can someone with a Debian/Ubuntu running php 7 test this hypothesis?

knnniggett:
I'm getting this error on both CentOS 7 and Fedora 25. version_alert appears to be part of this pr.
```
[Thu Dec 15 09:46:41.796993 2016] [:error] [pid 1021] [client 127.0.0.1:48642] PHP Notice:  Undefined variable: version_alert in /usr/share/zoneminder/www/skins/classic/views/header.php on line 19
[Thu Dec 15 09:46:45.743560 2016] [:error] [pid 1021] [client 127.0.0.1:48642] PHP Notice:  Undefined variable: version_alert in /usr/share/zoneminder/www/skins/classic/views/header.php on line 19, referer: https://localhost/zm/
```

kylejohnson:
Uh oh.  I'll look in to that.

knnniggett:
We have recently merged @connortechnology 's work into master, which includes much (all?) of the changes described here. Since this PR has become obsolete, I'm going to close this out.

PR #1666, opened 2016-11-03
tabs to spaces

connortechnology:
The ongoing work to fix the whitespace/google code style

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Base.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Base.pm.in
@@ -58,7 +58,7 @@ ZoneMinder::Base - Base perl module for ZoneMinder
 
 =head1 SYNOPSIS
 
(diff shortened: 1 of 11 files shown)


kylejohnson:
I think a criteria for this getting merged is that the number of changes (additions and deletions) stays equal; that'll help us to clearly see that this PR was only for what it should be.

connortechnology:
I don't see why the # of lines would change.  I think git is just being stupid.

connortechnology:
yeah something must have gone wrong. let me fix it.

connortechnology:
So git diff -w master shows no changes....

connortechnology:
Oh hey, # of changes is the same.... let's merge.

connortechnology:
I'm so confused now which branch is which, but I think I've got it cleaned up.  I don't want to do any more changes at this time because it is already too big to view here.

PR #1667, opened 2016-11-04
Change CREATE_ANALYSIS_IMAGES Help text

zestysoft:
The existing help text didn’t fully explain how blobs play a part in
this functionality.


zestysoft:
My thought process behind the edits  - besides trying to better explain what this feature does, I wanted to be as terse as possible since this is just a quick help the user guide should be the place for a more comprehensive discussion about this feature:

1. "By default during an alarm ZoneMinder records both the raw captured image and one that has been analysed and had areas where motion was detected outlined."

Motion detected only happens when blobs are used in active zones.  I felt that sentence was just too misleading as it made it sound like outlined motion was always available at a flip of this switch.  I also felt that the details about how this happens (the two images being created) doesn't belong in the quick help -- perhaps the user guide? A Comment somewhere in the code?  Seems like something only developers should be interested in anyway -- it should be enough to say that there will be a performance hit, since that's what most users would understand or care about.

2. "This can be very useful during zone configuration or in analysing why events occurred."

I figured this was self-evident.  Even if it wasn't, turning it on and then discovering it should make its usefulness immediately known... I mean, a red border around motion -- what else would that be useful for?

3. "However it also incurs some overhead and in a stable system may no longer be necessary."

I included the overhead part in my text, but got rid of the stable system part because it's an opinion.  I felt that it should be enough to say there's overhead -- if someone is later looking to optimize, they should then know this is something they can switch back off.  In my case, the overhead was negligible -- enough so that I plan on keeping in on forever.

4. "This parameter allows you to switch the generation of these images off."

I felt this was redundant from what I said in the first sentence: "If enabled" and also the sentence is a bit misleading -- This parameter allows you to switch the generation of these images on or off, not just off.  In fact enabling this feature turns these images on.

So, with my version:
"When enabled, for all alert frames within an event, all detected motion within zones that use blob as the 'Alarm Check Method' will be outlined in red."

This sentence covers 1, 2, and 4.

while:

" Note that this will incur some overhead."
covers sentence 3.

PR #1668, opened 2016-11-04
Add additional details to increasing mapped memory.

zestysoft:
Other distro’s do not use /run/shm as the memory mapped path, so added
information to cover this case.

Also filtered the output of df to focus our attention on the only mount
points that matter here.


connortechnology:
I'm not totally in love with the new text for CREATE_ANALYSIS_IMAGES. I like that the old gave more explanation.

zestysoft:
Ugh, those changes weren't supposed to be part of this pull request.  I've reverted them.  I'll detail the why behind the edits in the proper pull request.

PR #1670, opened 2016-11-04
Fix braces

connortechnology:
As per google style guide, and the personal preferences of most of our developers, put braces on the same line as the if/while/whatever


kylejohnson:
This looks good to me.  Is the PR ready @connortechnology ?

connortechnology:
I believe so.

PR #1675, opened 2016-11-06
Show available PATH_MAP percent on console

knnniggett:
Initial commit to show the end user how much swap space (usually /dev/shm these days) is in use.
Note that percent changes after a (auto)refresh such as when zoneminder is stopped/started. This makes it easier for the end user to see how much swap gets consumed the moment zoneminder is started.

I'm open to tweak this if anyone has any suggestions.

I would almost rather see free space in MB, but it was easier to show percent since the function already existed.

The semantic "Swap" is new, and thus not currently part of the existing language files. It will show in English until we get PR's from the various native speakers of other languages.

@kylejohnson Let me know how this affects your web UI pr

### DIFF
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -1429,15 +1429,15 @@ function getLoad() {
   return( $load[0] );
 }
 
-function getDiskPercent() {
-  $total = disk_total_space(ZM_DIR_EVENTS);
+function getDiskPercent($path = ZM_DIR_EVENTS) {
+  $total = disk_total_space($path);
   if ( ! $total ) {
-    Error("disk_total_space returned false for " . ZM_DIR_EVENTS );
+    Error("disk_total_space returned false for " . $path );
     return 0;
   }
-  $free = disk_free_space(ZM_DIR_EVENTS);
+  $free = disk_free_space($path);
   if ( ! $free ) {
-    Error("disk_free_space returned false for " . ZM_DIR_EVENTS );
+    Error("disk_free_space returned false for " . $path );
   }
   $space = round(($total - $free) / $total * 100);
   return( $space );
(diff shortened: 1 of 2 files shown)


knnniggett:
screenshot from 2016-11-06 11-45-09

kylejohnson:
@knnniggett I'll handle any conflicts within my branch, no worries.

However - I see you modifying the `getDiskPercent` regarding `$path`, but I don't see any changes for the console, for `PATH_SWAP`, or anything like that.  Did you forget to add a few files / commits?

knnniggett:
The changes to the console somehow ended up in commit 49aed79

knnniggett:
I can also see people confusing the word "Swap" to mean the Linux swap, but at the moment I can't think of a better word. We can't say "tmpfs" because the ZoneMinder Swap folder doesn't have to be in a tmpfs.

knnniggett:
959221b adds the change to console.php

knnniggett:
Don't know what I was thinking... ZM_PATH_MAP is the big memory eater. ZM_PATH_SWAP only uses some memory for buffer while streaming

connortechnology:
Well, it is a ramdisk, where we keep the mmap'd shared memory files.  

So how about Ramdisk? 

I think ramdisk is the closest thing that noobs will understand.  They may ask why it is using a ramdisk, but at least they will understand that it is ram that is involved.

knnniggett:
Technically it does not have to exist in the ramdisk, though. It can point to whatever folder the user chooses. It just happens to be the /dev/shm ramdisk in most cases. Maybe I'm overthinking this.

What if we called it "tmp"?

I noticed you've got something you are calling "tmp" in your sceenshot in pr #1665. Would this conflict with that? ...Or is this the same thing?

connortechnology:
The /tmp in my pic is because I actually /tmp setup as a storage area, so no, no conflicts.

How about we just list the actual path instead of giving it a name? So for most systems it would say /dev/shm

knnniggett:
Eureka moment! New commit is incoming....

knnniggett:
screenshot from 2016-11-08 12-54-07

knnniggett:
Notice we are using forward slashes as our deliminter between the different values.... might need to change that.

knnniggett:
Better!


screenshot from 2016-11-08 12-59-37

connortechnology:
Looks great! 

I think it will be confusing to some end users, but they can educate themselves.  

Or we could get crazy and add a popup help text explaining what the hell it is.

Can also colour code it.  I propose we leave all that for the todo list though.

kylejohnson:
Once you get in the `printWinningLottoNumber()` function this will be good to go.

PR #1676, opened 2016-11-06
prevent the end user from slecting an invalid configuration on the PT…

knnniggett:
…Z control configuraion "Move" tab

This pr presents an error message to the end user under the following conditions:
- "Can Move" is checked but none of the other movement types are checked (except Move Diagonally since that is optional)
- Any one or more of the others boxes are checked but "Can Move" is not checked.

This ensures that the underlying Perl script makes the right call to a subroutine name that actually exists.


knnniggett:
screenshot from 2016-11-06 12-48-22

connortechnology:
Note that this was based off your work on adding the shm swap line to the header.  Not a big deal, I imagine we will merge that soon.

knnniggett:
Son of a....
Let me fix that

knnniggett:
Actually, wait. If I fix this with another commit, it will make these PR's non-automergable. Maybe I should just leave it as is.

knnniggett:
f659cf8 reverses the change to console.php

PR #1681, opened 2016-11-13
unlink the mmap file when the monitor object is destroyed

knnniggett:
Currently, if a monitor’s status is disabled from the web console, ZoneMinder leaves the mmap file hanging around until ZoneMinder is stopped. This presents a potential problem for machines low on memory, since the mmap files usually reside in the ramdisk. This PR resolves that by unlinking the mmap file at the moment the monitor object is destroyed.

### DIFF
--- a/src/zm_monitor.cpp
+++ b/src/zm_monitor.cpp
@@ -674,6 +674,13 @@ Monitor::~Monitor()
     if ( munmap( mem_ptr, mem_size ) < 0 )
       Fatal( "Can't munmap: %s", strerror(errno) );
     close( map_fd );
+
+    char mmap_path[PATH_MAX] = "";
+    snprintf( mmap_path, sizeof(mmap_path), "%s/zm.mmap.%d", config.path_map, id );
+
+    if ( unlink( mmap_path ) < 0 ) {
+        Warning( "Can't unlink '%s': %s", mmap_path, strerror(errno) );
+    }
 #else // ZM_MEM_MAPPED
     struct shmid_ds shm_data;
     if ( shmctl( shm_id, IPC_STAT, &shm_data ) < 0 ) {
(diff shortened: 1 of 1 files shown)


connortechnology:
I could be wrong but I think you need to look at the mode in which the object was instantiated.  Otherwise you will have zma deleting the mmap file. Or zms.

knnniggett:
I was operating under the assumption that the backend, zma, zmc, etc would all share the same monitor object, and hence could exchange information between the various vm* processes w/o mapped memory. Wouldn't it be very memory inefficient if we spawned a monitor object, for the same monitor, multiple times?

In any case, this should be easy enough to test. Switching modes from Modect -> Monitor would cause this pr to fail.

connortechnology:
Well each process instantiates a Monitor Object.... and in the constructor for it, it mmaps the mmap file.... it's just that if zma dies for some reason and deletes the mmap file, then when zma starts back up, it will not be able to connect to the same mmap that zmc is using.  zmc will continue happily on, not caring that the mmap file was deleted.  

So we should only delete the mmap file if the monitor object was instantiated in capture mode.  I think. So only zmc will do the creating/deleting of the mmap file.

knnniggett:
I'll have it check if argv[0] = zmc similar to how zms checks how it was called.

Alternatively, we could do this in Perl and delete the mmap file when zmdc.pl is called to kill zmc, but I think it would be most efficient to do this in C.

connortechnology:
You can also look at the purpose variable.

knnniggett:
Oh, very nice. Someone already solved that problem.
if ( purpose == CAPTURE )

knnniggett:
Tested commit d3a7fca with one camera. Verified the mmap file was deleted if the monitor was disabled, and verified no errors occurred if the mode of the monitor was changes from Modect to Monitor. Thanks for the help @connortechnology

connortechnology:
That looks much better.  And now... to nitpick.. I'd be tempted to make the path part of the object so that it doesn't need to be computed again...

I'm good to merge though.

PR #1687, opened 2016-11-17
sort and remove duplicates using vims :%sort u command

connortechnology:


kylejohnson:
Looks much better.
Maybe at a comment at the top?
`# If you're adding to this file, please sort it with $command before committing`

PR #1692, opened 2016-11-22
Small fixes

connortechnology:
So these are old.  This is me cleaning out my old branches.

What we have here are a couple of conversions from "" to '' which is a micro-optimisation.
We have the addition of some ()'s which aren't strictly needed but really should be there.
We have a fix for the class on PanTilt which is missing a t
And the removal of superfluous global $SLANG; which we don't use anymore because we use a translate function.


knnniggett:
Travis did its random timeout again so I restarted the build

kylejohnson:
Looks OK to me.

PR #1693, opened 2016-11-22
add a zmMemInvalidate at the beginning of the while loop

connortechnology:
to catch unreleased mmap files.


knnniggett:
Ah, you beat me to it.

PR #1694, opened 2016-11-22
Disk space in events

connortechnology:
The gist of this is that it adds another column listing how much space each event is taking up. In the process I converted to using my Event object instead of just the array returned from the DB.

I have actually found the performance impact to be not too bad. Long term, this value should be stored in the db and updated whenever some change is made to the event.


knnniggett:
Shouldn't there be corresponding schema changes to zm_create.sql.in and create zm_update-1.30.1.sql? Or is it in there and I just don't see it?

connortechnology:
This doesn't touch the db.  It hits the fs.  changing the db is for future work.

I'm ok with this not getting merged right now.

knnniggett:
Hmmm.... how about making a toggle for this one in Options, so machines low on resources have the option to disable this?

connortechnology:
I certainly can.  

I do wonder how resource starved you would have to be for this to be a problem.  Honestly I expected it to be, but it simply isn't.  A little dir stat call seems to be completely negligible compared to everything else we are doing when listing events.  

The tradeoff is config complexity, etc.  Our config is already pretty daunting to a lot of end users.

knnniggett:
I think the general idea behind this PR is a great one. However, I don't agree we should introduce changes that will negatively effect performance, no matter how small, without a toggle. If we are making changes in other areas of ZoneMinder, on the grounds the new code improves speed and efficiency, then it is inconsistent if we, on the other hand, add other changes which slow it back down again.

I completely agree with the sentiment that making changes to ZoneMinder often seems harder than it should be. However, I've often admired how easy it is to add a new Config variable. That, to me anyway, has always seemed like a very nice exception.

knnniggett:
My apologies. I did not notice until now the latest commits pushed to this pr. Merging....

PR #1700, opened 2016-12-02
Update nl_NL translation

bajansen:
Translated some new untranslated strings and updated a lot of the older translations.

Edit: the Travis build fails but it does not appear to be related to my commits.


knnniggett:
Travis failed the check due to an unrelated issue with Travis and the mysql-server package. Merging anyway...

Issue #1704, opened 2016-12-11
Live scalable grid view (pull request) 


Goblenus:
Live scalable cameras grid view. Support fullscreen mode.
![17](https://cloud.githubusercontent.com/assets/16175474/21081793/1fe265a2-bfdf-11e6-9d66-59557a94a8c1.jpg)
![16](https://cloud.githubusercontent.com/assets/16175474/21081794/2725500e-bfdf-11e6-9769-b0f94643a8a3.jpg)
![15](https://cloud.githubusercontent.com/assets/16175474/21081797/3d0317bc-bfdf-11e6-892f-ad5d50f2ce43.jpg)
![14](https://cloud.githubusercontent.com/assets/16175474/21081798/3f08827c-bfdf-11e6-9372-a10e866e618f.jpg)
![13](https://cloud.githubusercontent.com/assets/16175474/21081799/44893f98-bfdf-11e6-9d82-db972882a9f0.jpg)
![12](https://cloud.githubusercontent.com/assets/16175474/21081801/47d548e0-bfdf-11e6-8fc4-989467b022b0.jpg)

Can use arrows to change grid view columns.

knnniggett:
This looks nice, but please move the php function calls out of gridview.php and into web/includes/functions.php

I personally don't use montage, but I'm wondering if, instead of creating a new view, we merge this capability into the standard montage. @connortechnology @kylejohnson

connortechnology:
Yes we already have too many views.

kylejohnson:
I've got an early version of this as part of the `console` view in my updated ui branch.  You can toggle between the table and the grid using the `Table | Grid` button in the page header.

I'm still figuring out what to display in the grid (e.g. do we want the event counts there, or just the simple picture?), but at this point I don't think it makes sense to have an entirely new view for this feature.

![2017-01-13-125250_1912x1052_scrot](https://cloud.githubusercontent.com/assets/114193/21943614/3ff44332-d98f-11e6-8f2a-750f24100d40.png)

Goblenus:
I don't need to see any events or smth like that at monitor page. The main task of this page to give me current state, nothing more. There are a lot of others pages, which can provide me that information.

Goblenus:
@knnniggett well, can you give me an idea what are you talking about...I mean, you tell put my code into web/includes/functions.php, and I'd like to know how do you expect to call it? Any input/output parameters? Which page will provide this view? I need basic understanding...

[...]

PR #1705, opened 2016-12-11
1704 gridview

Goblenus:
Live scalable cameras grid view. Support fullscreen mode.

https://github.com/ZoneMinder/ZoneMinder/issues/1704


knnniggett:
This looks nice, but please move the php function calls out of gridview.php and into web/includes/functions.php

I personally don't use montage, but I'm wondering if, instead of creating a new view, we merge this capability into the standard montage. @connortechnology @kylejohnson

PR #1706, opened 2016-12-13
Log sql when db insert fails

connortechnology:
So when a db insert fails, add the offending sql to the error report.
Second, just because one insert failed, doesn't mean we should turn off db logging, so get the db level before turning it off, and then restore it after logging the insert error.


connortechnology:
dang it... how that cmake stuff get in here?!

PR #1707, opened 2016-12-13
log failed sql when db insert fails. 

connortechnology:
Also, don't permanently turn off db logging just for one failed insert


knnniggett:
I do wonder if this might open us up to a situation where the logs are spammed with sql errors, but that might not be a bad thing... after all if zm can't write to the dB then we've got a big problem which the sys admin must address asap.

PR #1708, opened 2016-12-13
Message column to text

connortechnology:
Apparently at some point, we lost a db update script that changes the type of the Message column to TEXT.

Note: This bumps the version to 1.30.1


knnniggett:
Looks good. ZoneMinder bumped to version 1.30.1 now....

PR #1709, opened 2016-12-13
Fix 1703

connortechnology:
Put quotes around the link filename in the rm command so that it handles spaces.


knnniggett:
I'm not sure I've seen the use of ticks ` as a qq delimiter, but it should work.
Merging....

PR #1710, opened 2016-12-16
replace the old socket_sendto error message with something more useful

connortechnology:
So that people stop asking us how to fix it.

### DIFF
--- a/web/ajax/stream.php
+++ b/web/ajax/stream.php
@@ -52,7 +52,7 @@ while ( !file_exists($remSockFile) && $max_socket_tries-- ) { //sometimes we are
 }
 
 if ( !file_exists($remSockFile) ) {
-  ajaxError("Socket $ramSocketFile does not exist.  This file is created by zms, and since it does not exist, either zms did not run, or zms exited early.  Please check your zms logs and ensure that CGI is enabled in apache (sudo a2enmod CGI) and check that the PATH_ZMS is set correctly (typically /zm/cgi-bin/nph-zms).  Make sure that ZM is actually recording.  If you are trying to view a live stream and the capture process (zmc) is not running then zms will exit.");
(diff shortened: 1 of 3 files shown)


knnniggett:
Good idea trying to make this message more intuitive, but I think we should make the message more generic by removing the Debian specific portions. Leaving those in there will cause non-Debian users to dig a hole even deeper. The a2enmod tool is found only on Debian, along with the value for path_zms currently suggested in the error message.

Perhaps we could say something like: 
Consult the ZoneMinder documentation for your distro to verify cgi is enabled and PATH_ZMS is set correctly.

connortechnology:
The problem is that people won't read the documentation, then they will complain about how the docs are too confusing.

We will need to put a link in this message that takes you to the faq page, which needs an entry specifically for this.

knnniggett:
If we could do something that pointed the end user somewhere else for more details, I'd be in favor of that.
Will an http link work, embedded in an error message? Would it be clickable?

knnniggett:
While we are at it, I am some additional changes to the docs that are much needed. Will push them to your branch this morning...

knnniggett:
I sent you a pull request to your branch. You'll need to switch to your fork to see it if you did not get an email.

I did not realize how out of date our documentation was. I'm sure there are a lot of other updates that need to be made as well.

I'm not entirely certain if all my changes will format correctly, and I do not know how to preview these rst files before merging them.

web/ajax/stream.php
```diff
@@ -47,12 +47,16 @@
...
-while ( !file_exists($remSockFile) && $max_socket_tries-- ) //sometimes we are too fast for our own good, if it hasn't been setup yet give it a second.
-	usleep(200000);
+while ( !file_exists($remSockFile) && $max_socket_tries-- ) { //sometimes we are too fast for our own good, if it hasn't been setup yet give it a second.
+  usleep(200000);
+}
 
-if ( !@socket_sendto( $socket, $msg, strlen($msg), 0, $remSockFile ) )
-{
+if ( !file_exists($remSockFile) ) {
+  ajaxError("Socket $ramSocketFile does not exist.  This file is created by zms, and since it does not exist, either zms did not run, or zms exited early.  Please check your zms logs and ensure that CGI is enabled in apache and check that the PATH_ZMS is set correctly.  Make sure that ZM is actually recording.  If you are trying to view a live stream and the capture process (zmc) is not running then zms will exit. Please go to http://zoneminder.readthedocs.io/en/latest/faq.html#why-can-t-i-see-streamed-images-when-i-can-see-stills-in-the-zone-window-etc for more information.");
```
knight-of-ni, line 55: Whew, that's a long message. Out of curiosity, have you simulated this error condition to see how well this message shows up in the log window?

knnniggett:
Merging...

PR #1713, opened 2016-12-20
use zmMemDetach  instead of  zmMemInvalidate

knnniggett:
Current behavior of the master branch is (still) leaving file handles open on the mmap files.

To reproduce, simply open a monitor and then hit save. No changes to the monitor are necessary. If zmwatch has run, it will have an open file handle on the corresponding mmap file, and it will not let go if you wait for it.

I noticed the same behavior in zmtrigger. 

Experimentation revealed that replacing zmMemInvalidate with zmMemDetach corrects this. 

zmwatch no longer holds the file handle open, and zmtrigger holds the file open for a period of time up to MONITOR_RELOAD_INTERVAL (5 minutes). Maybe we should make that 60 seconds.

Since all the other cool scripts were doing it, I went ahead and made the same change to zmx10.pl, but I don't have a way to test that.

### DIFF
--- a/scripts/zmtrigger.pl.in
+++ b/scripts/zmtrigger.pl.in
@@ -418,7 +418,7 @@ while( 1 )
         foreach my $monitor ( values(%monitors) )
         {
             # Free up any used memory handle
-            zmMemInvalidate( $monitor );
+            zmMemDetach( $monitor );
         }
         loadMonitors();
     }
--- a/scripts/zmwatch.pl.in
+++ b/scripts/zmwatch.pl.in
@@ -89,8 +89,8 @@ while( 1 )
         my $restart = 0;
         # Prevent open handles building up if we have connect to shared memory
         # Many of our error checks below do a next without closing the mem handle.
-        # zmMemInvalidate will just return of nothing is open, so we can just do this here.
-        zmMemInvalidate( $monitor );
+        # zmMemDetach will just return of nothing is open, so we can just do this here.
+        zmMemDetach( $monitor );
         if ( zmMemVerify( $monitor )
              && zmMemRead( $monitor, "shared_data:valid" )
         )
@@ -141,6 +141,7 @@ while( 1 )
             {
                 $command = "zmdc.pl restart zmc -m $monitor->{Id}";
             }
+            zmMemDetach( $monitor ); # Close our file handle to the zmc process we are about to end
             runCommand( $command );
         }
         elsif ( $monitor->{Function} ne 'Monitor' )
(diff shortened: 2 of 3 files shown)


connortechnology:
testing

connortechnology:
So... is this actually working for you?  My zmwatch just crashes because zmMemDetach is not exported to it unlike zmMemInvalidate.  Also, zmMemInvaildate called zmMemDetach... so I'm not sure that I see how this could achieve the desired result.

Looking at the code... if for some reason the monitor object got corrupted and the key changed without detaching then we could lose track and not detach.... but I don't at this time see how that would happen.

connortechnology:
I wonder if we should open/close the db between zmwatch loops.  I wonder if there is a negative impact keeping that db connection open forever.

connortechnology:
So... we load the monitor data in a hash, called $monitor, we then pass this in to the Memory mapping functions, which store the MMAPAddr in it.  So it is absolutely crucial that we zmMemInvalidate in our inner loop because when we reload the monitors, that hash will be discarded. 

Also, our thought to call zmMemInvalidate at the top is completely bogus. We will be dealing with a fresh hash from DBI each time.

connortechnology:
Another thought... if we restart zmc.... we won't restart zma until the next pass.  But if we are restarting zmc... we really should restart zma as well.

knnniggett:
What is the exact error message you are receiving?

Interesting that it doesn't work for you, as it works w/o a hitch for me. It completely cleared up the issues I was having... been running with this change since I posted the pr.

I'm not sure I understand why zmMemInvalidate was written in the first place because it seems redundant. It calls zmMemDetach if $monitor->{MMapAddr} is defined.

However, zmMemDetach does the same thing and more:
- clears $monitor->{MMapAddr} if it exists
- clears $monitor->{MMapAddr} if it exists
- clears $monitor->{MMapHandle} if it exists

With zmMemInvalidate, if the first condition does not return true, then the other 2 are not checked and potentially cleared. ....so that implies, at least on my system, I'm seeing a case of a partially defined/corrupted mapped memory handle/key.

connortechnology:
zmMemInvalidate is a wrapper around whichever shared memory implementation is in use.  Memory.pm is the pubilc api, whereas Memory/Mapped provides the private implementation.  zmMemDetach is not exposed.  I don't understand how your zmwatch.pl isn't crashing with the same error as me, which is:

Undefined subroutine &main::zmMemDetach called at /usr/bin/zmwatch.pl line 91

Maybe it is, and that's why you don't have any open handles laying around... try running it from the command line, because that won't be logged anywhere.

knnniggett:
The issue described here still needs to be addressed. @connortechnology and I need to discuss the best approach to to resolve the issue and create a different pr.

PR #1714, opened 2016-12-20
remove the use of empty which on php < 5.5 only supports variables.

connortechnology:
When I switched to using object methods, this broke the use of empty() on php < 5.5.  

These monitor methods should return either a valid value or nothing, so we shouldn't need the empty.


knnniggett:
dang it, you beat me to to...

connortechnology:
Hey if I break it, I should fix it.... thanks for noticing the problem.

PR #1725, opened 2017-01-01
added RecordAudio to sql in README

davglass:
I pulled the latest PPA for this branch and installed it this morning. This field is missing from the README which caused all my camera's to fail to record. Just adding it here so others know :)

Thanks


connortechnology:
Thanks for contributing!

PR #1727, opened 2017-01-02
Add MonitorPreset for Qihan IP cameras via RTSP

StefanLindblom:
I've verified the SQL syntax + outcome using:
ZoneMinder 1.30.0
Linux Ubuntu 14.04.5 LTS, kernel 3.19.0-66
screen shot 2017-01-02 at 09 08 26


knnniggett:
Hi @StefanLindblom 
Glad to see someone is using the presets. Have you tried using the ffmpeg source type instead of Remote? We our discouraging end users from using Remote -> RTSP, unless they really have to, because we are no longer maintaining it.

StefanLindblom:
Hi @knnningett !
I actually just discovered presets after using ZM for quite some time...and thought that this brand would be nice to add since it seems to be gaining popularity.
But oh, I had no clue!
Maybe I should switch my own instance to use ffmpeg then... and re-visit this if it's working fine?
(Saying that, I do notice some drop outs today)

StefanLindblom:
Just tried ffmpeg with the same cameras. Works fine! Is there an article/blog post explaining why ffmpeg is now preferred? Would be interesting to know.

I'll try to amend my PR to use ffmpeg instead!
/Stefan

knnniggett:
There are currently three source types that all do the same thing: ffmpeg, libvlc, and Remote -> RTSP.
Start with ffmpeg, as it tends to work the best. It isn't perfect, but neither are the other two.

The mention of the Remote -> RTSP method being deprecated is shown in our latest documentation: 
http://zoneminder.readthedocs.io/en/latest/userguide/definemonitor.html#remote

It doesn't make a lot of sense for us to spend our time maintaining Remote -> RTSP when the ffmpeg & libvlc source are so much better. We will still accept pull requests from others against it if someone chooses to work on it.

knnniggett:
FYI for anyone who comes across this, once merged, this preset will become part of any _new_ database. To get this preset into an _existing_ database, the end user must use zmcamtool.pl.

knnniggett:
@StefanLindblom Just an FYI, that we are waiting for you to update this pr with a preset that uses ffmpeg. You can continue to push commits to your branch and they will automatically show up here.

connortechnology:
I'm kinda wondering.... should we have one for each?  I know we are moving towards deprecating remote... but we havn't actually yet.... and maybe in future might not.... I leave it to the submitter's discretion...

StefanLindblom:
Sorry for the delay here @knnniggett !

@connortechnology For now I just changed remote to ffmpeg.
I'm thinking that IF remote might be deprecated at some point, but ffmpeg will definitely stay - then it should be a (bit less) painful transition for users if they're already using ffmpeg.

I think I got the commit correct..

knnniggett:
@StefanLindblom Thanks for taking the time to learn Git and submitting this. It was a good first attempt for someone who may not have ever used git. If you want to submit another pr in the future, I can step you through doing it in such a way as to avoid merging all the extra stuff, but it's not a real big deal. Just affects readability.

PR #1729, opened 2017-01-02
Fix events list

connortechnology:
Currently the events list is broken.  I think I broke it with my showing disk space pr.  

This updates the Event Object to my latest version, which has an updated createListThumbnail funciton and fixes the events view.


knnniggett:
Ran out of time today to test this, but looks like someone in the forum ran into the same problem so I'll just merge it now and test later.

PR #1733, opened 2017-01-02
When use warnings is on and there is a null value in the monitor data…

connortechnology:
… it causes warnings.  So add skip if $_ is not defined.


knnniggett:
Hey, I'm just glad someone else besides me uses this tool.
@pliablepixels :-)

Issue #1720, opened 2016-12-27
Database table Users column MonitorIds not enough length


max252:
Hello,
I have about 90 cameras,
and when I add a user to a list of access to cameras (Options -> Users)
database truncates the data in column "MonitorIds"

I increased the size of the column from tinytext to text

alter table Users modify MonitorIds TEXT NOT NULL;

Now work

PR #1734, opened 2017-01-02
Fix 1720

connortechnology:
fix for #1720


knnniggett:
Shouldn't we also bump the version file to 1.30.2?

connortechnology:
crap, yes of course.

connortechnology:
the question becomes though... are we going to leave this for 1.30.2... or should we in fact merge this db update script to the one that does 1.30.1.. or.. .something else.

knnniggett:
Let's call the next release 1.30.2.
We know there are already some users in the wild running 1.30.1 master so in order to ensure they get the update in this pr, lets bump to 1.30.2.

PR #1738, opened 2017-01-06
Packpack

knnniggett:


knnniggett:
So looking through the travis details, we will need to make some additional changes.
It looks like each and every deployment is trying to rebuild an Ubuntu 14.04 from-source build, in addition to package build it is attempting to run. Will need to make some time to take a closer look at this.

knnniggett:
commit 6b7e3a1 modifies the rpm specfile to reference the zoneminder and crud source via https url. This should allow a build system to auto-retrieve the sources.

PR #1746, opened 2017-01-11
fix else behaviour by adding braces

connortechnology:
These are minor fixes for compiler warnings that turned up in Yakkety.


knnniggett:
I didn't realize that an if-else would even compile w/o braces.

Issue #1747, opened 2017-01-12
Docker image is broken. Returns a 403 accessing /


jbehrends:
When trying to run the [official docker image](https://hub.docker.com/r/kylejohnson/zoneminder/) as built with the Dockerfile in this project you receive a 403 error when trying to access the webserver:
![image](https://cloud.githubusercontent.com/assets/1915783/21873364/eca85b4a-d822-11e6-8a2d-a1e1ed20e032.png)

The Apache logs in the container show the following:
```
# cat /var/log/apache2/error.log
[Wed Jan 11 22:54:15.478184 2017] [mpm_prefork:notice] [pid 400] AH00163: Apache/2.4.7 (Ubuntu) PHP/5.5.9-1ubuntu4.20 configured -- resuming normal operations
[Wed Jan 11 22:54:15.478327 2017] [core:notice] [pid 400] AH00094: Command line: '/usr/sbin/apache2'
[Wed Jan 11 22:54:57.450305 2017] [authz_core:error] [pid 404] [client 172.17.0.1:43328] AH01630: client denied by server configuration: /usr/local/share/zoneminder/
[Wed Jan 11 22:54:57.662816 2017] [authz_core:error] [pid 404] [client 172.17.0.1:43328] AH01630: client denied by server configuration: /usr/local/share/zoneminder/favicon.ico, referer: http://localhost/
```

It appears this container MAY have been developed before apache 2.4 (which it now uses) and the apache site config needs to be updated.  Currently the documentroot is wrong, and permissions for "/" are missing:
Currently: **DocumentRoot /usr/local/share/zoneminder**
Should be: **DocumentRoot /usr/local/share/zoneminder/www**

I'll be submitting a PR shortly to fix this.

jbehrends:
Closing issue.  Was fixed as part of above PR that was merged.

PR #1749, opened 2017-01-12
Fixed apache documentroot, and fixed permissions for "/" in the project's Dockerfile

jbehrends:
This PR addresses #1747 by fixing the following two issues:

1. Documentroot was broken, so location was updated.
2. Apache was throwing a 403 because permissions were not set for "/".


knnniggett:
Thank you. We are relying on the community to manage the docker file so feel free to submit additional pr's against this in the future.

PR #1754, opened 2017-01-15
1716 doc img typo

### LINKED ISSUE #1716: Text mistake documentation screenshot
http://zoneminder.readthedocs.io/en/latest/userguide/components.html#system-overview
Screenshot text mistake: "Cature Video" --> Should be "Capture video, convert to frames"
ed-zm-system-overview

pliablepixels:
fixes #1716


connortechnology:
So not that I care or would prevent us from merging... but the xml was readable by some form of human... the new one... is not...

PR #1755, opened 2017-01-16
Update do debian release

connortechnology:
So this does a number of things. First and foremost it updates it to the version that i have been using.
Second, it implements a different way of specifying arguments.  Instead of positional, it uses -o=option style. It might be wise to wait on merging until I implement a --help option.  please note that the clean function still works as before. 
It also consults various ENV vars that are commonly used to populate the $AUTHOR and to control the use of signing keys.
Lastly it introduces uploading to the zmrepo debian ppa, which will of course only be useful to Isaac right now.
This might not be mergeable right now but is presented for discussion.


connortechnology:
Also just added an urgency flag...

SteveGilvarry:
Pretty sure we have some docs that refer to the old method so we should also find them and update to the new options. So people don't complain about it not working as per documentation.

knnniggett:
Added this to the 1.31.0 milestone. Please update the documentation to this pr.

knnniggett:
Someone want for fix the merge conflicts for this one?

PR #1756, opened 2017-01-16
Remove zmf

SteveGilvarry:
First go around at this and it builds, was some stuff in zm_event.cpp that I was a little dubious on removing, so check this and test it as I cannot at the moment, or just leave until I get back.


connortechnology:
Love it.

Issue #1530, opened 2016-06-13
(TRIVIAL) 1.30 is not compiling under FreeBSD11-ALPHA3


abishai:
FreeBSD11 is not yet released. however packages are built against it as well. 
We need a hint what abs() to use. Can we assume that *psrc is long ?

/wrkdirs/usr/ports/multimedia/zoneminder/work/ZoneMinder-1.30.0-rc1/src/zm_image.cpp:1624:24: error: call to 'abs' is ambiguous
        if ( (unsigned)abs((*psrc)-RGB_VAL(ref_colour,c)) >= RGB_VAL(threshold,c) )
                       ^~~
/usr/include/stdlib.h:83:6: note: candidate function
int      abs(int) __pure2;
         ^
/usr/include/c++/v1/stdlib.h:115:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long      abs(     long __x) _NOEXCEPT {return  labs(__x);}
                                           ^
/usr/include/c++/v1/stdlib.h:117:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long long abs(long long __x) _NOEXCEPT {return llabs(__x);}
                                           ^
/usr/include/c++/v1/math.h:646:1: note: candidate function
abs(float __lcpp_x) _NOEXCEPT {return fabsf(__lcpp_x);}
^
/usr/include/c++/v1/math.h:650:1: note: candidate function
abs(double __lcpp_x) _NOEXCEPT {return fabs(__lcpp_x);}
^
/usr/include/c++/v1/math.h:654:1: note: candidate function
abs(long double __lcpp_x) _NOEXCEPT {return fabsl(__lcpp_x);}

abishai:
Or use fabs(), it's less overloaded.

abishai:
Or labs() :)
I don't know what solution is better.

abishai:
As we already has workaround for SOLARIS here, maybe remove abs entirely if it's so confuse compiler?
        unsigned int diff = ((*psrc)-RGB_VAL(ref_colour,c)) > 0 ? (*psrc)-RGB_VAL(ref_colour,c) : RGB_VAL(ref_colour,c) - (*psrc);
        if (diff >= RGB_VAL(threshold,c))
        {
          count++;
        }

If you don't have better ideas, I can make a PR.

connortechnology:
[...]

Issue #1760, opened 2017-01-31
comparison between pointer and zero


abishai:
ZoneMinder can't be compiled with CLANG 4.0 It bumped severity of such comparison to error. 
src/zm_monitor.cpp contains comparison between (unsigned char*) mem_ptr and 0 in several places.
```
src/zm_monitor.cpp:197:18: error: ordered comparison between pointer and zero
('unsigned char *' and 'int')
    if ( mem_ptr > 0 )
	 ~~~~~~~ ^ ~
```

connortechnology:
The line should probably like this:
if ( mem_ptr > (void *)0 ) {

abishai:
Maybe, !=0 is OK too.

connortechnology:
No because failure to mmap results in -1.

abishai:
I added fix to my FreeBSD PR. Compiles on my test system with your fix.

mastertheknife:
What about comparing it to NULL?
NULL is typically defined as (void*)0, so mem_ptr > NULL should work.

arvidn:
for the record, ``mmap()`` returns ``MAP_FAILED`` on error (which may or may not be the same value as ``(void*)-1``). [posix spec.](http://pubs.opengroup.org/onlinepubs/009695399/functions/mmap.html#tag_03_379_04)

connortechnology:
It seems like we should follow posix.. defined constants are both readable... and defined.. hence... good.. Mmkay?

ali-babaei:
> The line should probably like this:
> if ( mem_ptr > (void *)0 ) {

on:
Telegram Android Client

I had following error:

jni/./gifvideo.cpp:151:28: error: ordered comparison between pointer and zero ('AVStream *' and 'int')
    if (info->video_stream <= 0) {
        ~~~~~~~~~~~~~~~~~~ ^  ~

Thanks @connortechnology, your solution worked for me too.

Just commented here for everybody facing this issue later!

PR #1761, opened 2017-01-31
[Upstream FreeBSD specific patches]

abishai:
- Remove usage of abs() . This one fixes https://github.com/ZoneMinder/ZoneMinder/issues/1530
- Switch to POSIX basename() FreeBSD project deprecates GNU functions if POSIX analogs available. Function was entirely removed in HEAD.
- cmake patch to remove polkit check for FreeBSD and provide more paths for mysql header lookup.
- fix llvm400 specific errors See https://github.com/ZoneMinder/ZoneMinder/issues/1760

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -386,13 +386,13 @@ find_library(MYSQLCLIENT_LIBRARIES mysqlclient PATH_SUFFIXES mysql)
 if(MYSQLCLIENT_LIBRARIES)
 	set(HAVE_LIBMYSQLCLIENT 1)
 	list(APPEND ZM_BIN_LIBS "${MYSQLCLIENT_LIBRARIES}")
-	find_path(MYSQLCLIENT_INCLUDE_DIR mysql/mysql.h)
+	find_path(MYSQLCLIENT_INCLUDE_DIR mysql.h /usr/local/include/mysql /usr/include/mysql)
 	if(MYSQLCLIENT_INCLUDE_DIR)
 		include_directories("${MYSQLCLIENT_INCLUDE_DIR}")
 		set(CMAKE_REQUIRED_INCLUDES "${MYSQLCLIENT_INCLUDE_DIR}")
 	endif(MYSQLCLIENT_INCLUDE_DIR)
 	mark_as_advanced(FORCE MYSQLCLIENT_LIBRARIES MYSQLCLIENT_INCLUDE_DIR)
-	check_include_file("mysql/mysql.h" HAVE_MYSQL_H)
+	check_include_file("mysql.h" HAVE_MYSQL_H)
 	if(NOT HAVE_MYSQL_H)
 		message(FATAL_ERROR 
                     "ZoneMinder requires MySQL headers - check that MySQL development packages are installed")
@@ -660,11 +660,13 @@ message(STATUS "Using web user: ${ZM_WEB_USER}")
 message(STATUS "Using web group: ${ZM_WEB_GROUP}")
 
 # Check for polkit
-find_package(Polkit)
-if(NOT POLKIT_FOUND)
-	message(FATAL_ERROR 
-            "Running ZoneMinder requires polkit. Building ZoneMinder requires the polkit development package.")
-endif(NOT POLKIT_FOUND)
+if(NOT BSD)
+    find_package(Polkit)
+    if(NOT POLKIT_FOUND)
+        message(FATAL_ERROR
+                "Running ZoneMinder requires polkit. Building ZoneMinder requires the polkit development package.")
+    endif(NOT POLKIT_FOUND)
+endif(NOT BSD)
 
 # Some variables that zm expects
 set(ZM_PID "${ZM_RUNDIR}/zm.pid")
--- a/src/zm_image.cpp
+++ b/src/zm_image.cpp
@@ -1658,11 +1658,9 @@ Image *Image::Highlight( unsigned int n_images, Image *images[], const Rgb thres
       {
         uint8_t *psrc = images[j]->buffer+c;
 
-#ifndef SOLARIS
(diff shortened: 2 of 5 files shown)


abishai:
Here is some patches I've made during porting ZM to FreeBSD. I included C and CMAKE patches to get feedback (I'm not programmer myself and have little experience in PR system). If they are OK, I'll provide more :)

knnniggett:
Thank you for submitting these patches. Some of these changes will need to be tweaked a bit.

We are making a final push to get the next release out the door. I'll provide further feedback following release.

src/zm_image.cpp
```diff
@@ -1658,11 +1658,9 @@ Image *Image::Highlight( unsigned int n_images, Image *images[], const Rgb thres
       {
         uint8_t *psrc = images[j]->buffer+c;
 
-#ifndef SOLARIS
-        if ( (unsigned)abs((*psrc)-RGB_VAL(ref_colour,c)) >= RGB_VAL(threshold,c) )
-#else
-        if ( (unsigned)std::abs((*psrc)-RGB_VAL(ref_colour,c)) >= RGB_VAL(threshold,c) )
-#endif
+	    unsigned int diff = ((*psrc)-RGB_VAL(ref_colour,c)) > 0 ? (*psrc)-RGB_VAL(ref_colour,c) : RGB_VAL(ref_colour,c) - (*psrc);
```
connortechnology, line 1661: I thought at first that this couldn't be correct... but it is.  You know what I would like to see though?  the result of RGB_VAL(ref_colour,c) being put into a variable and then used.  I know the compiler likely optimises it away anyways... but I think it would be more readable...

src/zm_logger.cpp
```diff
@@ -515,8 +515,9 @@ void Logger::logPrint( bool hex, const char * const filepath, const int line, co
     va_list     argPtr;
     struct timeval  timeVal;
 
-    const char * const file = basename(filepath);
-    
+    char *filecopy = strdup(filepath);
```
connortechnology, line 518: why are we copying filepath?

abishai: GNU and POSIX functions have different prototypes, one is accepting char*, another - const char*

abishai: https://www.freebsd.org/cgi/man.cgi?query=basename&apropos=0&sektion=3&manpath=FreeBSD+12-current&arch=default&format=html
https://www.freebsd.org/cgi/man.cgi?query=basename&sektion=3&apropos=0&manpath=FreeBSD+11.0-RELEASE+and+Ports

connortechnology: Okay... but you don't need to copy it.. just use a cast

abishai: Well, as I'm not programmer myself, I read manuals. It was said that such casting is a bad habit as const is a pointer to read only memory location and char* to read write. Recommendation was to use another pointer.

connortechnology: Ok, so apparently some implementations of basename will modify the input.  This is for multi-threaded support apparently.  So... I guess this change is actually correct.  But I don't want to be doing this on every logPrint.  I will see if I can refactor this to be more efficient.

knnniggett:
@abishai I sent a PR to your master branch to revise commit 0f23809
It is a better method for determining when to check for polkit.

Note the generating pr's from your own master branch can cause problems later on. It is best to always create a new branch first and do you work from that... e.g. `git checkout -b my_shiny_new_branch`

abishai:
@knnniggett Thank you! I'll test if it still builds.

OK, I'll try it. PR system is rather complicated to me. For example, my llvm patch was glued to this one. I tried to made it separated.

knnniggett:
> OK, I'll try it. PR system is rather complicated to me. For example, my llvm patch was glued to this one. I tried to made it separated.

Yes, that is because you are working out of your master branch.

abishai:
I'll remake clang patch. I shouldn't blindly change similar lines as only mmap returns -1.

src/zmf.cpp
```diff
@@ -331,7 +331,7 @@ int main( int argc, char *argv[] )
     Debug( 1, "Got image, writing to %s", path );
 
     FILE *fd = 0;
-    if ( (fd = fopen( path, "w" )) < 0 )
+    if ( (fd = fopen( path, "w" )) < (void *)0 )
```
connortechnology: I'm pretty sure this is incorrect... I will have to do some research to figure out what it should be.  I always thought that file descriptors were integers...

abishai: Yeah. I wrote that I'll remake the patch. Should be == NULL.

abishai:
Revised clang patch.

CMakeLists.txt
```diff
@@ -386,13 +394,13 @@ find_library(MYSQLCLIENT_LIBRARIES mysqlclient PATH_SUFFIXES mysql)
 if(MYSQLCLIENT_LIBRARIES)
 	set(HAVE_LIBMYSQLCLIENT 1)
 	list(APPEND ZM_BIN_LIBS "${MYSQLCLIENT_LIBRARIES}")
-	find_path(MYSQLCLIENT_INCLUDE_DIR mysql/mysql.h)
+	find_path(MYSQLCLIENT_INCLUDE_DIR mysql.h /usr/local/include/mysql /usr/include/mysql)
```
knight-of-ni, line 397: @abishai Can you try this instead?
`find_path(MYSQLCLIENT_INCLUDE_DIR mysql.h PATH_SUFFIXES mysql)`

knnniggett:
@abishai I changed find_path to use PATH_SUFFIXES before merging this. I believe it should work.
Thank you for submitting these. Now that 1.30.4 is released, we will merge any future pr's like this one much more quickly.  Remember though, first create a new branch in your fork and work out of that rather than master. :-)

PR #1764, opened 2017-02-04
sql injection and session fixation vulerability fixes

kylejohnson:
* SQL injection is addressed by sanitizing inputs and using real prepared statements (not emulating them).
* Session fixation is addressed by regenerating the session id after a successful log in.


knnniggett:
This was previously approved in the Pro fork. Merging...

PR #1765, opened 2017-02-05
Align Method description to what it is actually doing

SteveGilvarry:
Makes it hard for people to know what they are doing with the descriptions from Remote camera, so changed them to line up with FFmpeg option they are controlling. Does also impact libvlc, but these descriptions are closer also. Only fixing as it annoyed me.

### DIFF
--- a/web/lang/en_gb.php
+++ b/web/lang/en_gb.php
@@ -905,11 +905,11 @@ $OLANG = array(
 		          "\"loglevel=debug\" Set verbosity of FFmpeg (quiet, panic, fatal, error, warning, info, verbose, debug)"
 	),
         'OPTIONS_RTSPTrans' => array(
-		'Help' => "This sets the lower RTSP Transport Protocol for FFmpeg.~~ ".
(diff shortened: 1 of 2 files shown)


connortechnology:
I like this, but I think we should actually say that they might see corrupted images with UDP... I think those code makes TCP the default, right?  Which I agree with.   So maybe say that UDP is lighter weight but may cause image corruption.  Also... I think the word lower is unnecessary and will confuse noobs.

SteveGilvarry:
I actually want to find someone with UDP corruption to try fifo_size as a fix. But I will mention it can have issues against the UDP help text.

connortechnology:
Oh I have it... I use TCP with my cameras... but I can test.

connortechnology:
Looks good to me.

connortechnology:
Maybe we could even merge this for 1.30.2.

SteveGilvarry:
Nah it missed cutoff, unless everyone else wants to.

knnniggett:
Aw, what the heck.

Issue #1590, opened 2016-08-11
http cURL source with no content-length header fails


wilsonwaters:
Hi all,
I have a camera source which provides a jpeg snapshot through an http URL. Unfortunately it doesn't include the content-length http header with the data.
Zoneminder gives the following error when trying to read 

> Unable to get the size of the image

I had a quick look through zm_curl_camera.cpp and it seems after reading the data you're trying to find the size by calling `curl_easy_getinfo(c, CURLINFO_CONTENT_LENGTH_DOWNLOAD, &dSize);`. I suggest you should use `databuffer.size()` instead.

I don't know the codebase well enough so there's probably a good reason you do it this way, but it would be nice if we could have support for curl based sources which don't send the content-length header.

Thanks!

connortechnology:
I'm kindof curious what happens if you just use the remote http monitor type.

I'm not sure if we are going to maintain the curl type.  It was an interesting experiment, which ultimately did not solve the problem we were trying to solve.

That being said... databuffer.size may not be the entire image, as in, it may not all have arrived yet.  The logic was really not designed to handle this case, although obviously it should be.

What camera make/model is this?  Would you be able to expose it to me for testing?

connortechnology:
No worries, I have that app.  I'll do some testing when I find a moment.

mattdurant:
The problem is that without Content-Length, you have no choice but to read until the other end closes the socket.  In ReadData() of zm_remote_camera_http.cpp around line 193, it's assumed that if there is no data to read that it's because of an exception with the socket, no because the other end is finished sending the response in the case of a single image.

mattdurant:
I tested this before and after this PR, using the same app (I use it personally as well).  This fix resolved the issue for me.

wilsonwaters:
Nice. Thanks @mattdurant. I'll see if I can work out how to build Zoneminder and test it sometime soon.

Just had a quick look through the change. Would it be safer to test if the actual read returns 0, rather than what ioctl(FIONREAD) says is currently available on the socket?
Could there be a situation where the socket is open but the remote end doesn't immediately start writing data? Or if there is a pause half way though the transfer for some reason, wouldn't this cause the image to be truncated?

I'm not being very helpful sorry, I should build Zoneminder and test it myself but just wanted to mention it now.

mattdurant:
[...]

PR #1770, opened 2017-02-11
remove line that causes endless reading when doing single image mode

connortechnology:
See #1590 for discussion.


knnniggett:
regression. merging...

PR #1774, opened 2017-02-14
Remove SSH server from docker image

michaelarnauts:
Installing a ssh server to manage a docker container is bad. See https://jpetazzo.github.io/2014/06/23/docker-ssh-considered-evil/

You can enter a container by executing
```bash
$ docker exec -it <container> bash
```


markosv:
Totally agree!

SteveGilvarry:
This docker was originally setup as more a development version than production, it probably should be updated to xenial/php7 and use packages for installation. It is only maintained by community so feel free to make changes. I think I will merge this so we get closer to a production version.

michaelarnauts:
I think the same `Dockerfile` could be used for production and dev work. I think a good way is to build everything from source (as it is now) in the Dockerfile, and let dockerhub build the images for you. This way, it's really easy to use the latest dev version, but also stable releases.

It would probably also be best to move the mysql server out, and keep them connected with docker-compose.

PR #1780, opened 2017-02-15
use escapeshellarg on inputs to daemonControl and other functions

### LINKED ISSUE #1775: Command injection in device name
There is a post auth admin command injection in the device name of a monitor.
function zmcControl  in functions.php does not escape the device name and this later gets used by daemonControl(...)
escapeshellarg or escapeshellcmd on the device name would fix the issue.

connortechnology:
Fixes #1775 

Note the one in getDiskBlocks... you can edit your config to have ZM_DIR_EVENTS be a malicious command....


knnniggett:
looks good. Just to remind myself, escapeshellarg is a php internal function.

Issue #1781, opened 2017-02-15
Adjust score calculation algorithm


vajonam:
I am not convinced that the score is calculated correctly. When I specify the MAX_ALARMED, MAX_FILTER or MAX_BLOB size, those should be used to calculate the score rather the area of the whole polygon.

I might have a very large zones (less zones for faster), but smaller max (30-40% of the zone.area() )values which means the scoring should be higher, this is important as more higher scores are less likely to be false positives. 

https://github.com/ZoneMinder/ZoneMinder/blob/master/src/zm_zone.cpp#L273

https://github.com/ZoneMinder/ZoneMinder/blob/master/src/zm_zone.cpp#L368

https://github.com/ZoneMinder/ZoneMinder/blob/master/src/zm_zone.cpp#L667

I have PR ready for this if folks agree.

johncolby:
@connortechnology I agree that the score algorithm could be improved.

A more basic change would be letting scores have more precision one way or another (e.g. 1000* instead of 100*), so as to avoid the problem of many events being generically squashed to a score of 1 (see below, as also mentioned by @mastertheknife in #1782 ).

![screen shot 2018-04-20 at 1 30 17 pm](https://user-images.githubusercontent.com/473295/39072311-07867980-449f-11e8-89ff-3af4c9e7d467.png)

Then, @vajonam's suggestion could be accommodated by letting user have a display option for the Y-axis like "Display as..." "% of zone pixels" or "% of max threshold pixels" or "log scale" or whatever.

vajonam:
so what do I need to do to get this in? you new a new PR against current master/HEAD?

connortechnology:
Yes update it.  In the meantime I will make another attempt at understanding it.

vajonam:
So what I am proposing is that when we have any alarm and the max_alarm_pixes is not 0. i.e the user has specified that the max alarm aera in percent or pixels (gets converted to pixels anyway), then use that to calculate the score rather than to total size of the zone. 


```
if (max_alarm_pixels != 0)  // max size is specified use that to calculate the score
    score = (100*alarm_pixels)/max_alarm_pixels;
  else // max is not specified use total size of the zone to calculate the score
    score = (100*alarm_pixels)/polygon.Area();

```
[...]

PR #1782, opened 2017-02-15
fix scoring based on max value of zone, rather than area of entire zone

vajonam:
#1781 fixes for adjusting scoring algorithm based on largest allowable size


vajonam:
make sure we fall back to size of the zone polygon if max_* is 0

connortechnology:
My mind breaks down while looking at this code/logic.  It is going to take some serious looks by several people to agree on this.  

@mastertheknife ?

vajonam:
Agreed, but my point is that if you specify a max size for your alarmed area, you % or scre should be calculated against that max size, because if the alarmed area greater than the max size it will be ignored.

mastertheknife:
On my system, sometimes the score is really low, its 1 some events.
There were times it was less than 1, but a fix in the code (exists since 1.26 or so) makes sure the minimum is 1. The fix in the current code is like a band aid, perhaps we should really improve the scoring algorithm.
I haven't had time to understand the changes in this PR and its effect on the score. Its better to go safe and slow with this, its a significant change that will affect all users.

knnniggett:
Looks like this issue needs some follow up.
Assigning a milestone of 1.31.0 to this.

vajonam:
I am curious to see how many users actually use score for anything?

PR #1783, opened 2017-02-17
Add human filesize to the event filesize column 

marciopamplona:
Instead of showing the size in bytes without punctuation, shows the size in 2 decimal places with notation in human format (b, K, M, G).

### DIFF
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -2137,7 +2137,14 @@ function folder_size($dir) {
     foreach (glob(rtrim($dir, '/').'/*', GLOB_NOSORT) as $each) {
         $size += is_file($each) ? filesize($each) : folderSize($each);
     }
-    return $size;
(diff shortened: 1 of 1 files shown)


connortechnology:
Hey.. we are in feature freeze atm, which explains our silence.

On the other hand... I think I already implemented this in my branches... and I don't like folder_size returning human readable.  Because then I need another function that returns it in non-human readable format.  So folder_size should be a bytes value and whatever calls it should decide to make it human readable.  Also, I'm pretty sure there is already a human readable function somewhere in our source. I know it's a mess, but we have got to slim this code down.  I bet php also has a handy function for this.

connortechnology:
Thanks for your contribution though!  I really appreciate it.

knnniggett:
Closing this due to previous comments. 
If the previous comments can be addressed then feel free to open a new pull request.

Issue #1784, opened 2017-02-17
wrong key used for AlarmRefImageBlendPct


vajonam:
Incorrect key is used in monitor.php, it should be AlarmRefImageBlendPct not AlmRefImageBlendPct

kylejohnson:
Closed per #1785

PR #1785, opened 2017-02-17
fix usage of wrong key

vajonam:
fixes for #1784


vajonam:
Only broken for 2 years!

PR #1787, opened 2017-02-17
1786 dockerfile xenial

MarkosVakondios:
Image based on Ubuntu 16.04 Xenial and PHP 7


utils/docker/setup.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Start MySQL
-/usr/bin/mysqld_safe & 
+service mysql start
```
connortechnology: This is interesting.. not.. 

Should we not be using systemctl?  I know the backwards compatability stuff pretty much works... but not always..

MarkosVakondios:
Corrected both utils/docker/setup.sh & utils/docker/start.sh not to use upstart for MySQL starting, and hence no init system dependencies.

mnoorenberghe:
When I just tested this branch it gave a failure in "Step 6/15 : RUN cmake .": http://pastebin.com/cqzzcThG

Line 2512 looks suspicious. I don't know how to look at /ZoneMinder/CMakeFiles/CMakeOutput.log since the build didn't complete.

mnoorenberghe:
Hmm… I get the same error with `master` when I used that syntax so maybe something else is wrong with my setup [Docker Version 17.03.0-ce-mac2 (15654)]

SteveGilvarry:
Missing this must be done after cloning GitHub.

git submodule update --init --recursive

mnoorenberghe:
Yep, just realized that. See my update at https://github.com/ZoneMinder/ZoneMinder/pull/1787#issuecomment-287508749

knnniggett:
Hey, neat. Github supports resolving merge conflicts from the web ui now.

@mnoorenberghe I don't use Docker so please inspect the dockerfile to verify I manually merged the conflcits correctly: https://github.com/ZoneMinder/ZoneMinder/blob/master/Dockerfile#L45

mnoorenberghe:
LGTM but I don't have time to actually test it.

PR #1791, opened 2017-02-19
Additional controls for FI8908W cameras

dwalkes:
Mobile application ZM View uses relative move commands and presetGoto
commands which were not supported by the previous version of the script.

Add these commands based on documentation in
http://www.foscam.es/descarga/ipcam_cgi_sdk.pdf

Commands were tested on an FI8908W camera running version 11.37.2.65
firmware


connortechnology:
Just so @dwalkes doesn't feel ignored... we are currently in feature freeze in order to release 1.30.2   The moment we get that out, we will likely merge this.

PR #1794, opened 2017-02-21
use === operator in getDiskPercent function

knnniggett:
This change differentiates between a false condition and the value 0, which is valid.


connortechnology:
One problem, if $total is zero, then we have a div by zero error.  Need to check for it and handle it gracefully.

knnniggett:
Oops. Has been fixed. I also added more content to the error messages.

PR #1810, opened 2017-03-14
Minor zm_image SSSE3 improvements

mastertheknife:
Hello,
Its been few years since i last optimized things, its time for a 2017 refresh :)
I tried to squeeze some more performance, especially for the SSSE3 functions.

1) Decreased SSSE3 delta functions by 12 instructions per loop iteration.
This was possible thanks to SSSE3's horizontal instructions. Up to this point, almost all operations were vertical. Bytes had to shifted into the correct position, added and\or multiplied vertically.
The PMADDUBSW instructions multiplies bytes vertically and then adds the bytes horizontally.
Likewise, PHADDW adds words horizontally.
The code behaves exactly the same, but is now between 5-10% faster.
2) Applied change 1 also to the SSSE3 32bit->8bit convert function.
3) Added BGRA, ARGB and ABGR variants of SSSE3 32bit->8bit conversion.
4) Made the code actually use the SSSE3 32bit->8bit convert functions. The RGBA function existed, but nothing was using it.
5) Removed SSSE3 fourfield deinterlacing functions. They had equal or worse performance than the standard functions (when compiling with O2 or better). The functions were too complicated to be vectorized efficiently.
6) On 64bit systems, gcc\glibc uses optimized memcpy, which is equal or slightly faster than ours.

Thats all for now.
SSE2 is not worth optimizing, most people have processors that support SSSE3 or better by now.
If anything, i should focus on AVX2 which has larger registers (256bit vs 128bit) and allows for more pixels per register (8 32bit pixels vs 4 32bit pixels).


kylejohnson:
I haven't tested the performance, but the source code does compile and run here for me.  Have a single ffmpeg camera running with modect successfully.

knnniggett:
This is very nice. I do want to wait until after we release 1.30.2 to merge this.

So @mastertheknife is there any chance you would be willing to do the same kind of optimization for arm Neon instruction set? I'll buy you a Raspberry Pi 3 and ship it to you if you say yes.

connortechnology:
Id kinda like to see some tests that compare the outputs of each function to ensure they all work

On March 16, 2017 3:51:07 PM EDT, Andrew Bauer <notifications@github.com> wrote:
>This is very nice. I do want to wait until after we release 1.30.2 to
>merge this.
>
>So @mastertheknife is there any chance you would be willing to do the
>same kind of optimization for arm Neon instruction set? I'll buy you a
>Raspberry Pi 3 and ship it to you if you say yes.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/ZoneMinder/ZoneMinder/pull/1810#issuecomment-287171748

-- 

mastertheknife:
> I haven't tested the performance, but the source code does compile and run here for me. Have a single ffmpeg camera running with modect successfully.
Its unlikely to notice a performance increase, because its just 5-10% and only for this function. There two other heavy functions (blend and alarmed pixels detection) take the other 70% CPU usage of zma.

> So @mastertheknife is there any chance you would be willing to do the same kind of optimization for arm Neon instruction set? I'll buy you a Raspberry Pi 3 and ship it to you if you say yes.
I heard of ARM's Neon instruction set. Its very similar to Intel's SSE\AVX series, and has some advantages, e.g. instructions with more than 2 operands (Intel SSE series is limited to 2 operands). The ARM is already a very fast and efficient processor, the gains may not be great, but its worth a try anyway.
No need to buy anything, just give me SSH access (no need for root), i will write the code on my computer, scp to the arm, compile and test :)

> Id kinda like to see some tests that compare the outputs of each function to ensure they all work
Sure, no problem, i will upload and link to the code i use for benchmarking and testing the functions (for making sure they behave as expected).
For testing i have code that reads random numbers from /dev/urandom, feeds the same numbers to two functions: standard (expected result) and SIMD (testee function), and then compares the output. if it differs, the program stops. I normally let it run for hours.
Regarding benchmarking, this gives me an idea. Why not make it possible to benchmark slow functions (e.g. delta and blend for a start)? This will provide accurate evaluation of system performance when using ZoneMinder, because it runs the code it'll actually run when using ZM for motion detection, so its more accurate than generic benchmarks.

Also, i had some progress with delta in AVX2, but i have deleted the file by accident, need to start over.

connortechnology:
I will get you ssh access to my pi when I get back from vacation.  Right now it is sitting in my laptop bag disconnected...

knnniggett:
@mastertheknife I'm sending you a pm through the user forum

mastertheknife:
@knnniggett 
Thank you. A friend gave me SSH to his RasBerry2 yesterday. I already finished with the Delta function, but its just 22% faster. I guess this is because current ARM processors (until Cortex A15) process 128bit Neon instructions as two 64bit instructions.
Also, this is only compatible with AArch32. AArch64 is very different. It has different instructions and its registers don't overlap anymore. I will need access to a machine with AArch64 ARM running 64bit OS to develop and test that. For now the solution is to compile ZM as 32bit (AArch32) if running 64bit (AArch64) OS.
I am now working on the Blend function.

@connortechnology
I have uploaded the files i use for benchmarking and testing to here, including the Neon files:
http://51.15.134.207/~kfir/zm/
Typically need to compile like this:
gcc file.c -O2 -mssse3 (or -mfpu=neon for ARM).
Depending on the compiler, -std=c99 and -lrt might also be needed. Haven't spent much time into fixing this code to compile without C99

knnniggett:
Wow, that was fast. As far as I know, aarch64 is still experimental.
While some arm board manufacturers out there might have images available for download with custom, hacked aarch64 kernels, I'm not sure I've seen an official distro that has aarch64 out of beta yet. We probably have some time before that becomes a thing.

Whatever improvements you can get, likely are indeed worth while.
As I discovered at the end of my blog (see the table), a single 1080p camera @5 fps is just a bit too much for the raspberry pi 3 to handle. If we could lessen the cpu load, then a 1080p camera might be possible.
http://zoneminder.blogspot.com/p/raspberry-pi-3.html

connortechnology:
I think the pi3 actually is aarch64 just that Tapan doesn't use it yet...I could be wrong though. I think this will help the pi a lot.

mastertheknife:
Okay, if Aarch64 is not common yet, then we don't have to worry about it now :)
I finished the blend and recalculated the speedups for delta and blend.
Standard Fastblend (with -O2): 81 million colours/second, New: 148 million colours/second, 1.82x speedup
Standard Delta (with -O2): 25 million pixels/second, New: 32 million pixels/second, 1.28x speedup
Before i can push a branch, need to add runtime detection for Neon. its not easy as with x86, because the CPUID registers can only be read in privileged mode (kernel mode), so need to get it from the kernel (using hwcaps API)
Also, it doesn't compile without -mfpu=neon here, we'll need to add that somehow
Delta: Old:

mastertheknife:
I was curious why removing 12 instructions from the SSSE3 delta only increased performance by 5-10%
It appears that all instructions used, use 1 clock cycles, except for PHADDW and bit shifting. Bit shifting uses 2 clock cycles, but PHADDW is terrible. It uses 1-2 clock cycles on AMD processors, but anywhere from 3 (Haswell) to 7 (Core2, Atom) clock cycles on Intel processors. Its a relatively simple instruction, but Intel probably wanted to cut costs on less widely used instructions.

PR #1817, opened 2017-03-18
Add eslint configuration files extending Google's rules

mnoorenberghe:
According to https://github.com/ZoneMinder/ZoneMinder/wiki/Coding-Standards ZM follows Google's coding standards. In order to start to enforce this for JS I've added eslint configuration files extending their rules.

If this is something core devs are interested in I can also setup travis to enforce this and discuss/fix some of the other rules I've disabled for now. I used the `--fix` option to automatically fix some of the straightforward issues for now.

To test this you will need to run:
```
npm install --g eslint eslint-config-google
eslint .
```

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,4 @@
+web/api/lib
+web/skins/classic/js/jquery-1.11.3.js
+web/skins/classic/js/jquery.js
+web/tools/mootools
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,29 @@
+"use strict";
+
+module.exports = {
+  "env": {
+    "browser": true,
+  },
+  "extends": ["google"],
+  "rules": {
+    "brace-style": "off",
+    "camelcase": "off",
+    "comma-dangle": "off",
+    "key-spacing": "off",
+    "max-len": "off",
+    "new-cap": ["error", {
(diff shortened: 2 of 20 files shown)


connortechnology:
Thanks for this, I think we will definitely want this.  I am merging it to my branch, but I wouldn't expect anything else until after we get 1.30.2 out.

mnoorenberghe:
> I wouldn't expect anything else until after we get 1.30.2 out.

Could you elaborate on this? Do you mean that you're not accepting new patches on any branch until release? Eslint changes don't have to go into 1.30.2.

connortechnology:
Uh... like I said... I will merge to MY branch... other branches are up to those who maintain those branches.  

We may merge this immediately after 1.30.2 gets released... or we may not.. 

I'm just trying to give you some reasonable expectations, while also telling you that we appreciate your contribution. Right now, we are trying to focus on getting 1.30.2 out, and this will NOT be in it. It likely will be in 1.31.0, but that is not a promise.

knnniggett:
I agree with the premise, but I'd rather hold off on this until 1.31.0.

connortechnology:
Yeah so I merged it to storageareas... and fixed all the complaints there.  Not too bad actually.  In fact, I would prefer if it would be more strict.

knnniggett:
@connortechnology Is this pr ready to merge as-is or did you make further changes when testing in your storage areas branch?

connortechnology:
It is fine as-is

Issue #1275, opened 2016-02-09
incomplete copyright; license violation (removed copyright statement)


onlyjob:
[SPP1802SWPTZ.pm](../blob/master/scripts/ZoneMinder/lib/ZoneMinder/Control/SPP1802SWPTZ.pm) is based upon [FI8918W.pm](../blob/master/scripts/ZoneMinder/lib/ZoneMinder/Control/FI8918W.pm) therefore it is a derivative work. However original copyright statement(s) were removed/replaced with copyright "2015- Bobby Billingsley" which is a license violation. One shall not remove someone else's copyright statements and replace 'em with his own.

You can't do such things, @bofhdk. You should have preserve original copyright statements and add your own.
##

SteveGilvarry:
Fixed in #1819

PR #1819, opened 2017-03-18
Fix failure to attribute previous work

SteveGilvarry:
Just fixing this up as there is an issue #1275  logged from Debian for it.


knnniggett:
Seems harmless enough.

Issue #1340, opened 2016-03-14
DOC BUG: zmc has no option "-r" implemented


pmar5303:
Both manual page and "-h" help text of zmc declare the "-r" option for remote streaming. That option is not implemented at all in zmc, as is clear from the source code.

PR #1820, opened 2017-03-18
Remove zmc -r argument as it does not seem to be implemented.

SteveGilvarry:
#1340 bought this up and I believe it is correct -r is not implemented.

### DIFF
--- a/src/zmc.cpp
+++ b/src/zmc.cpp
@@ -27,7 +27,6 @@ zmc - The ZoneMinder Capture daemon
 
  zmc -d <device_path>
  zmc --device <device_path>
- zmc -r <proto> -H <host> -P <port> -p <path>
  zmc -f <file_path>
  zmc --file <file_path>
  zmc -m <monitor_id>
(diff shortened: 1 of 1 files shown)


connortechnology:
I kinda wonder if we shouldn't either implement it, or be even more aggressive and remove all the other options.  I kinda think the zmc should only take a monitor Id.

mastertheknife:
> I kinda wonder if we shouldn't either implement it, or be even more aggressive and remove all the other options. I kinda think the zmc should only take a monitor Id.

Please don't do that. Local cameras take -d parameter:
`[kfir@atom ~]$ ps ax | grep zmc
 1963 ?        S     80:23 /usr/local/bin/zmc -d /dev/video0
 1984 ?        S     81:18 /usr/local/bin/zmc -d /dev/video1
 2000 ?        S     80:20 /usr/local/bin/zmc -d /dev/video2`

I am not sure why, i believe its because its possible to use the same device by two(or more) monitors, but each one on a different channel.

connortechnology:
Okay, I'm fine with loading by -d... but all the host, password, etc...  what use?

connortechnology:
or better yet, replace them with a url parameter.

knnniggett:
Merging this for 1.30.3. Please create another pr if it is decided to make further changes.

PR #1822, opened 2017-03-19
Implement CSRF Mitigation

knnniggett:
Add csrf-magic library to zoneminder, and integrate into zoneminder web forms.
https://github.com/ezyang/csrf-magic

What this library does is generate a token, which is based off the users php session id. If the token submitted with the form does not match the token generated when index.php was first loaded, then the web server returns http 403 with the message "CSRF check failed. Your form session may have expired, or you may not have cookies enabled." You will know it, if you trigger the csrf prevention.

I have given this some testing, and it appears to work. However, ideally every form submit or save button in ZoneMinder should be tested to verify there are no false triggers.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
@@ -346,6 +346,26 @@ our @options = (
     category    => "system",
   },
   {
+    name        => "ZM_ENABLE_CSRF_MAGIC",
+    default     => "no",
+    description => "Enable csrf-magic library",
+    help        => qqq("
+      CSRF stands for Cross-Site Request Forgery which, under specific
+      circumstances, can allow an attacker to perform any task your
+      ZoneMinder user account has permission to perform. To accomplish
+      this, the attacker must write a very specific web page and get
+      you to navigate to it, while you are logged into the ZoneMinder
+      web console at the same time. Enabling ZM_ENABLE_CSRF_MAGIC will
+      help mitigate these kinds of attackes. Be warned this feature
+      is experimental and may cause problems, particularly with the API.
+      If you find a false positive and can document how to reproduce it,
+      then please report it. This feature defaults to OFF currently due to
+      its experimental nature.
+    "),
+    type        => $types{boolean},
+    category    => "system",
+  },
+  {
     name        => "ZM_OPT_USE_API",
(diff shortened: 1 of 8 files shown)


mnoorenberghe:
If this is used on the "login" action it will break many integrations such as [zmNinja](https://github.com/pliablepixels/zmNinja/blob/71b31f5e6d4a67e2af993a1dc2e0ca77154a83e2/www/js/app.js#L993) and [Home Assistant](https://github.com/home-assistant/home-assistant/blob/2b9fb73032cc3fa48302e4619dcb3323624de472/homeassistant/components/zoneminder.py#L75). Perhaps the "login" action can be exempt for now? A CSRF attack against the login form alone seems like a lesser issue.

Unrelated to the above I would suggest adding a way to opt-out of the CSRF protection unless we're willing to put out a quick point release for any things that it breaks.

SteveGilvarry:
@pliablepixels FYI what would work for zmninja if we were to implement a CSRF token?

I remember at some point the API not having a proper authentication method so we have to logon to web ui and use that session, what is the root cause of that? ZMS?

knnniggett:
@SteveGilvarry @connortechnology @pliablepixels @mnoorenberghe 

I've updated this pr to allow csrf-magic library to be turned off under Options.

Additionally, the csrf check is skipped during a login attempt. I *think* this does not create a way for an attacker to work around the csrf check.... or maybe I've been staring at this so long everything sounds good. Need another set of eyes.

This appears to work with zmNinja, but this will need a ton more resting. Would also like feedback from @mnoorenberghe regarding HA.

Note, that I still think we should change the way the API authenticates as described in #1827

connortechnology:
I will merge to storageareas and test.

connortechnology:
Looks good so far.  Everything appears to work, including zmNinja.

knnniggett:
@pliablepixels @mnoorenberghe 

Just wanted to give you guys a chance to provide feedback before we merge this.
We plan to merge this tomorrow if we do here from either of you.... so we can get 1.30.2 out the door.

pliablepixels:
Late to comment. But I've tested with Isaacs CSRF enabled server. No issues.

mnoorenberghe:
It seems to work fine for me too… I didn't test it's effectiveness but with the action=login exclusion I suggested this works fine for HASS so far. I've enabled it on my server so I'll report any issues.

PR #1826, opened 2017-03-20
Uri decode in zms

### LINKED ISSUE #1825: No URI decoding in zms.
zms doesn't do any uri-decoding on it's passed inputs.  This becomes a problem if you do something like use an email address for a username.  

In current master this isn't so much of a problem, as the ui doesn't encode the username, but it should.  In my branch I added proper encoding, and that broke zms.

connortechnology:
Fixes #1825.

So far I have only implemented a CURL based version.  We really should have a fallback function in case we build without CURL.


knnniggett:
Did you intend to include 4792d21 with this pr?

connortechnology:
Um.. no.

connortechnology:
As per my recent suggestion... I'm ok with this not being merged until I also add an automated test.

PR #1835, opened 2017-03-24
Setup api cache dirs

connortechnology:
So with the default cakephp api install, we don't auto-create the /usr/share/zoneminder/api/app/tmp link to /var/tmp.  I'm hoping this change fixes that.

The other problem is that cakephp will only create the directories under that when in debug mode.  We don't really want to default to debug mode, so let's create them manually during install. Since we are using /var/tmp which I don't think gets cleaned out on reboot, this should be enough.  Otherwise, we will have to make it part of the startup process/init script.


knnniggett:
For Debian packaging changes, I personally would not be opposed if you just pushed these changes strait to master.

PR #1836, opened 2017-03-29
fix inserting x10 record with missing comma

connortechnology:
See https://forums.zoneminder.com/viewtopic.php?f=36&t=25995

This fixes it.


knnniggett:
Wow, this has been like that for 3 years before it was noticed.

PR #1841, opened 2017-03-29
create cake cache folders

knnniggett:
This change is for the generic tempfiles.d folder under misc, which would typically be used when building from-source. We don't symlink the tmp folder to another location because the link target varies per distro.

### DIFF
--- a/misc/zoneminder-tmpfiles.conf.in
+++ b/misc/zoneminder-tmpfiles.conf.in
@@ -1,4 +1,9 @@
 D @ZM_RUNDIR@ 0755 @WEB_USER@ @WEB_GROUP@
 D @ZM_TMPDIR@ 0755 @WEB_USER@ @WEB_GROUP@
 D @ZM_SOCKDIR@ 0755 @WEB_USER@ @WEB_GROUP@
-
+D @ZM_WEBDIR@/api/app/tmp 0755 @WEB_USER@ @WEB_GROUP@
+D @ZM_WEBDIR@/api/app/tmp/logs 0755 @WEB_USER@ @WEB_GROUP@
+D @ZM_WEBDIR@/api/app/tmp/cache 0755 @WEB_USER@ @WEB_GROUP@
+D @ZM_WEBDIR@/api/app/tmp/cache/models 0755 @WEB_USER@ @WEB_GROUP@
+D @ZM_WEBDIR@/api/app/tmp/cache/persistent 0755 @WEB_USER@ @WEB_GROUP@
+D @ZM_WEBDIR@/api/app/tmp/cache/views 0755 @WEB_USER@ @WEB_GROUP@


connortechnology:
So in debian, and I think your redhat based dists, we made the tmp is the cakephp folders a link to @ZM_TMP_DIR@... but this PR looks like it creates a tmp dir in the cakephp folders. 

My concern with this is that it is deep in the fs and may fill up the drive and people won't know it's there... 
I like to put tmp files one of the main tmp dirs like /tmp or /var/tmp

Just my 2cents.

knnniggett:
Right, and I agree this isn't a perfect solution. That is exactly what I meant when I said the link target varies per distro.

Currently, the linking gets done in each distro's packaging scripts. This pr deals with the generic case, which means there is no packaging script to do the linking. Yes, we could create a cmake variable (ZM_CAKE_TMP ?) and then do the actual linking in cmake, but that would break all existing packaging scripts.... not sure we really need to go down that path for this case.

Anyway, this does not need to be merged soon.

mnoorenberghe:
Do we even need `ZM_CAKE_TMP` when we already have `ZM_TMPDIR`? Why not tell CakePHP to store its temp. files in subdirectories of `ZM_TMPDIR` instead of using a symlink? Wouldn't that be backwards compatible since distros already have to make `ZM_TMPDIR`? From looking at the cake docs it seems like it's a matter of adding `define('CACHE', '@ZM_TMPDIR@');` in `bootstrap.php.in`.

knnniggett:
Yes, that would make a much better solution. 

None of us, other than guy who wrote the API, know cake very well, so we simply did not know that feature existed. Now, what would be extra cool is if you could figure out how to make cake create it's own cache subfolders w/o debug being turned on (don't think its possible, but again I'm not that familiar with cake).

knnniggett:
@mnoorenberghe @connortechnology 
I'm looking into the possibility of changing the cake cache engine from flat files to Apc. With the cache now running in shared memory, we will no longer need to create Cake's cache folders.

I'd like your feedback and would also appreciate it if you both could test this.

What you need to do is install the php-apcu package on your system. On Fedora, the package is called php-pecl-apcu-bc, but Debian uses a different name. Once that is installed and the web server is restarted, make the following changes:
- Change the default cache engine in bootstrap.php from "File" to "Apc"
- Change the $engine in core.php from "File" to "Apc"

NOTE: Cake would not stop trying to write cache files to the cake tmp folder unless I made *both* of the changes above. The comments in core.php seem to imply I shouldn't have to do that, but that does not appear to be the case.

With these changes in place, delete any existing cake cache folders if they exist and temporarily turn cake debug to 2 in core.php. 

- Verify cake reports it is using the ApcEngine -> https://localhost/zm/api/
- Verify the API returnrs valid data -> https://localhost/zm/api/host/getVersion.json
- Verify cake has not re-created its cache folders on your filesystem

So far, I've only tested this on Apache web server.

connortechnology:
Works good for me.

knnniggett:
superseded by
#1852 
#1853
#1854

PR #1844, opened 2017-03-30
fix Monitors filtering SQL

### LINKED ISSUE #1812: MontageReview Restricted Monitors incorrect SQL
As logged here https://forums.zoneminder.com/viewtopic.php?p=99361#p99361
Essentially the following line `$monFilterSql = ' AND M.Id IN ('.$user['MonitorIds'].')';`
[See code](https://github.com/ZoneMinder/ZoneMinder/blob/master/web/skins/classic/views/montagereview.php#L148)
is being added to 3 SQL queries. Only one of the queries actually has the alias M in use, needs to either use correct filter for that table or Join to Monitors as M.

connortechnology:
fix #1812

### DIFF
--- a/web/skins/classic/views/montagereview.php
+++ b/web/skins/classic/views/montagereview.php
@@ -145,11 +145,9 @@ $frameSql = "
 
 if ( !empty($user['MonitorIds']) )
 {
-    $monFilterSql = ' AND M.Id IN ('.$user['MonitorIds'].')';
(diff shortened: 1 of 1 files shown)


Linwood-F:
Sorry about that and thank you.  I just didn't have a setup where I could reconcile git and fix it handy at present, off on other stuff.

Issue #1827, opened 2017-03-20
Add an API to authenticate to zms and the API without sending passwords with every API request


mnoorenberghe:
While working on the Home Assistant integration I realized that the API hasn't really been designed with security and ease-of-use in mind when it comes to authenticating to the API and authenticating to ZMS streams (when ZM auth is enabled).

Problems with the [current approach](https://zoneminder.readthedocs.io/en/stable/api.html#security):
1. API Consumers are required to forge a login form POST request to start a session. This gets in the way of CSRF protection (#1822) since the consumer would have to scrap the nonce.
2. It's really difficult, if not impossible, to calculate the auth hash for use in the zms URL. Without that, the zms URL needs the plaintext username and password query parameters which show up in server logs.

From the discussion with @pliablepixels on IRC and my experience with the HA integration, I think there should be an API to do the following:
Given a username and password in a POST, return the auth hash for the user.

That alone would fix (1) and (2) above for consumers who only want do get a ZMS stream but wouldn't help with consumers who want to use other API endpoints (assuming it didn't start a session). To protect the API endpoints against CSRF we can also support using the auth hash for API authentication.

I'll submit a patch this week (hopefully later today) to implement this. I still have to think through the various config options related to auth. to hopefully have this approach work for all combinations:
* OPT_USE_AUTH
* AUTH_TYPE
* AUTH_RELAY
* AUTH_HASH_IPS
* AUTH_HASH_LOGINS

(It would be nice to remove some of the options e.g. AUTH_HASH_LOGINS to simplify things).

mnoorenberghe:
WIP branch at https://github.com/ZoneMinder/ZoneMinder/compare/master...mnoorenberghe:api_login?expand=1

knnniggett:
Thank you for your help.

The two who have used the API the most, and are most familiar with the inner workings, are yourself and @pliablepixels . I'm not sure @kylejohnson has been involved with the API much since he wrote it. Consequently, I think the most sensible thing to do here is let you drive this ship.

I basically feel like a middleman here. I recently learned enough to implement the csrf-magic library, but I haven't used the API enough to have an answer as to how to implement it w/o breaking the API.

In hindsight, I think we all would like the API to not depend on the web frontend at all, but I think the scope of that is probably too big to implement at this time.

Eliminating redundant features sounds good, but of course we need to be careful how we handle that ....will need to make sure a clear migration path is documented.

knnniggett:
[...]

PR #1845, opened 2017-04-03
 Allow API authentication using the `auth` query parameter containing an auth. hash.

mnoorenberghe:
This implements #1827. Let's try to keep the discussion about the concept there and focus this pull request on the technical details of the implementation to avoid fragmenting the conversation.

The reason for moving the global `Debug` function was because it conflicted with a CakePHP method of the same name. I thought of a few different approaches to workaround the problem but I settled on moving it to a static method on Logger as it seemed cleanest.
> PHP Fatal error:  Cannot redeclare Debug() (previously declared in /usr/local/share/zoneminder/www/api/lib/Cake/basics.php:73) in /usr/local/share/zoneminder/www/includes/logger.php on line 508

### DIFF
--- a/web/ajax/stream.php
+++ b/web/ajax/stream.php
@@ -21,23 +21,23 @@ if ( !@socket_bind( $socket, $locSockFile ) )
 switch ( $_REQUEST['command'] )
 {
     case CMD_VARPLAY :
-        Debug( "Varplaying to ".$_REQUEST['rate'] );
+        Logger::Debug( "Varplaying to ".$_REQUEST['rate'] );
         $msg = pack( "lcn", MSG_CMD, $_REQUEST['command'], $_REQUEST['rate']+32768 );
         break;
     case CMD_ZOOMIN :
-        Debug( "Zooming to ".$_REQUEST['x'].",".$_REQUEST['y'] );
+        Logger::Debug( "Zooming to ".$_REQUEST['x'].",".$_REQUEST['y'] );
         $msg = pack( "lcnn", MSG_CMD, $_REQUEST['command'], $_REQUEST['x'], $_REQUEST['y'] );
         break;
     case CMD_PAN :
-        Debug( "Panning to ".$_REQUEST['x'].",".$_REQUEST['y'] );
+        Logger::Debug( "Panning to ".$_REQUEST['x'].",".$_REQUEST['y'] );
         $msg = pack( "lcnn", MSG_CMD, $_REQUEST['command'], $_REQUEST['x'], $_REQUEST['y'] );
         break;
     case CMD_SCALE :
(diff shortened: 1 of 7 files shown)


mnoorenberghe:
Gist with a shell script to test some combinations: https://gist.github.com/mnoorenberghe/f59156f0c2fe15467f4745f8204a3edd

web/ajax/stream.php
```diff
@@ -21,23 +21,23 @@
 switch ( $_REQUEST['command'] )
 {
     case CMD_VARPLAY :
-        Debug( "Varplaying to ".$_REQUEST['rate'] );
+        Logger::Debug( "Varplaying to ".$_REQUEST['rate'] );
```
connortechnology, line 24: Why put the Logger:: in front?  It's outside the scope of the PR.  I actually do like to fully specify function names, but this seems unnecessary.

mnoorenberghe: This is explained in the description of the pull request

web/index.php
```diff
@@ -175,8 +175,8 @@
 isset($request) || $request = NULL;
 isset($action) || $action = NULL;
 
-if ( ZM_ENABLE_CSRF_MAGIC && $action != 'login' ) {
-    Debug("Calling csrf_check with the following values: \$request = \"$request\", \$view = \"$view\", \$action = \"$action\"");
+if ( ZM_ENABLE_CSRF_MAGIC ) {
```
connortechnology, line 178: You took out the && $action != 'login' which is there for a reason.

mnoorenberghe: Right, I took it out since it was only a workaround for not having a proper way to authenticate to the API which this bug is resolving.

web/includes/functions.php
```diff
@@ -27,6 +27,9 @@ function ob_get_clean() {
   }
 }
 
+require_once( 'logger.php' );
+require_once( 'database.php' );
```
connortechnology, line 31: Why did you add these?  by the time anything includes functions.php it is generally understood that these will have been included already, in index.php.  Did you find that they were necessary?

connortechnology:
I am merging this to my storageareas branch to give it some testing.  

I get why we need the Logger::Debug change.. but I don't like it.  If we are going to make Debug into a static method, then I think Info and Warning and the rest should be too.  

I'm kinda wondering if we shouldn't do something like $Log->Debug() instead.  It's more what I am used to doing in my code.  Not saying it's better, but maybe it is.  Opinions?

web/index.php
knight-of-ni: While I do agree our goal should be to remove $action != login, I am concerned about the timing. In order to minimize the pain, perhaps we should first document the old auth method is deprecated, and then remove it in a future version.

PR #1847, opened 2017-04-06
Suppress debug messages on ZM service start/restart

ekwoodrich:
The tests that determine the format for superuser calls on the target environment  leak debug messages to STDOUT.  Starting the Zoneminder process will display an intimidating series of error messages that can safely be ignored by the user: 

e.g.
```
root@2bfdd23cc27a:~# service zoneminder start

Starting ZoneMinder: Can't exec "sudo": No such file or directory at /usr/share/perl5/ZoneMinder/General.pm line 1                                                   10.
Use of uninitialized value $output in scalar chomp at /usr/share/perl5/ZoneMinder/General.pm line 119.
Use of uninitialized value $output in concatenation (.) or string at /usr/share/perl5/ZoneMinder/General.pm line 1                                                   20.
success
```

This patch redirects STDERR from the qx(...) calls to the debug message., And if the output of qx(...) is undefined, replaces it with any error in $! resulting from the qx(...) call.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/General.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/General.pm
@@ -103,8 +103,10 @@ sub getCmdFormat {
   my $suffix = "";
   my $command = $prefix.$null_command.$suffix;
   Debug( "Testing \"$command\"\n" );
-  my $output = qx($command);
+  my $output = qx($command 2>&1);
   my $status = $? >> 8;
+  $output //= $!;
+  
   if ( !$status ) {
     Debug( "Test ok, using format \"$prefix<command>$suffix\"\n" );
     return( $prefix, $suffix );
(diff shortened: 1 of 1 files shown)


connortechnology:
In theory I like this.  I would like to be able to reproduce it first though, so I can be confident that I I understand all the issues.

Can you tell me more about your setup?  Obviously it lacks sudo for starters...

ekwoodrich:
I'm using the quantumobject/docker-zoneminder Docker image on a Ubuntu 16.04.02 host, Docker version 17.03.1. 

The Docker base images for Ubuntu-minimal don't include sudo, which doesn't usually break much because a Docker container uses root in a sandbox. I've also replicated this on 15.10 and 16.04 docker images [(this thread](https://github.com/tianon/docker-brew-ubuntu-core/issues/48) describes how sudo was removed in more recent Docker Ubuntu images).

If you have Docker installed, you can replicate by pulling a Zoneminder docker image:
`sudo docker run -d --shm-size=4096m -p 80 quantumobject/docker-zoneminder:latest`
Enter the container:
`docker exec -it [container_id] /bin/bash`,
And inside the container run
`service zoneminder restart`. 

The sudo error messages will be displayed, however the ZM install works perfectly otherwise. The errors reappear each time Zoneminder is restarted from the terminal within the container. 

For users who use a Docker packaged Zoneminder image, these messages are confusing and presumably any environment lacking sudo would have the same issue.

PR #1850, opened 2017-04-13
Support for hiding / showing advanced options in Options view

kylejohnson:
* Only a subset of the options have been set to advanced ('1').  Add more as needed.
* Toggled state does not survive tab changes.

I'll need someone to remind me the right way to submit DB changes in a PR.  I've included `db/zm_update-1.30.3.sql` for now.

### DIFF
--- a/web/skins/classic/views/options.php
+++ b/web/skins/classic/views/options.php
@@ -57,6 +57,7 @@ xhtmlHeaders( __FILE__, translate('Options') );
   <div id="page">
     <div id="header">
       <h2><?php echo translate('Options') ?></h2>
+      <input id="btnToggleAdvanced" type="button" onclick="toggleAdvanced();" value="Show Advanced"></input>
     </div>
     <div id="content">
       <ul class="tabList">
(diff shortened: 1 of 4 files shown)


connortechnology:
So I'm cool with this... but I don't think it should go in 1.30.3. I think master should strike out for 1.31 territory, and we can work on 1.30.3 as a branch off of the 1.30.2 release branch.

So you should make your db update file be 1.31.0

knnniggett:
@kylejohnson please update this from zm_update-1.30.3.sql to zm_update-1.31.0.sql

Before we can merge this, we should figure out how to manage changes like this with our new build system. Let's discuss in our Slack channel rather than in this issue.

web/skins/classic/views/options.php
```diff
@@ -57,6 +57,7 @@
   <div id="page">
     <div id="header">
       <h2><?php echo translate('Options') ?></h2>
+      <input id="btnToggleAdvanced" type="button" onclick="toggleAdvanced();" value="Show Advanced"></input>
```
connortechnology, line 60: invalid html

knight-of-ni:
I certainly agree with the idea of this pr, but the code has fallen out of date.
Let's implement this in a new PR.

PR #1851, opened 2017-04-13
invert timestamp_on_capture condition

knnniggett:
Currently the timestamp is not visible during event playback (only) if TIMESTAMP_ON_CAPTURE is off under options. This issue was traced back to when Exif metadata feature was merged.

This pr inverts the condition which checks for the timestamp flag.

I've built and tested this on my machine with one camera, and this does appear to fix the problem. I did not test to see if the Exif metadata was affected, but that appears unlikely.
Additional tests are welcome.


connortechnology:
Looks like you got some extra files in this... otherwise I'd merge imediately

knnniggett:
Right. Was doing too many things at the same time. Fixed.

PR #1853, opened 2017-04-14
change cake cache engine from File -> Apc

knnniggett:
From discussion in #1841 
This PR changes the cake cache engine from File -> Apc.
The Apc cache engine uses shared memory, which gets us out of having to manage cake's cache folders ourselves.

This needs to be tested to verify the additional memory requirements, no new problems are created, etc.

@onlyjob 
@barjac 
@connortechnology 

If this PR is merged, it will require a new package dependency on php-apcu. Since the name of the package varies by distro, I have attempted to track down a few of them:
Fedora: php-pecl-apcu-bc
CentOS: php-pecl-apcu
Ubuntu (Debian?): php-apcu

### DIFF
--- a/web/api/app/Config/bootstrap.php.in
+++ b/web/api/app/Config/bootstrap.php.in
@@ -23,7 +23,7 @@
  */
 
 // Setup a 'default' cache configuration for use in the application.
-Cache::config('default', array('engine' => 'File'));
+Cache::config('default', array('engine' => 'Apc'));
 
 /**
  * The settings below can be used to set additional paths to models, views and controllers.
--- a/web/api/app/Config/core.php.default
+++ b/web/api/app/Config/core.php.default
@@ -352,7 +352,7 @@
  *       Please check the comments in bootstrap.php for more info on the cache engines available
  *       and their settings.
  */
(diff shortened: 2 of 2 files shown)


josh4trunks:
I noticed my API stopped working with Zoneminder 1.30.4, PHP 7.1, and APCu. That is because APCu for PHP 7+ dropped support for the old APC calls. APCu-bc is an option for some operating systems, but we don't currently have a port of that for FreeBSD.

Browsing the CakePHP repo it looks like this has been addressed in more recent versions. Can we expect Zoneminder to update CakePHP at some point?

Thanks!

knnniggett:
@josh4trunks According to the user in this thread, pecl-apcu4 works on freebsd:
https://forums.zoneminder.com/viewtopic.php?f=36&t=26131&p=100361&hilit=freebsd+apcu#p100361

josh4trunks:
sweet, I'll try that for now

josh4trunks:
Hmm, I do not see how he was able to install that package while having PHP7.0 or 7.1 installed.
When I try to install it, pkg tries to downgrade me to PHP5.6

PR #1856, opened 2017-04-16
use dos2unix to remove BOM

### LINKED ISSUE #1855: BOM in some files
I received the following email regarding this issue:

Hi
I was digging for a quite some of time to know why I can't see images using zoneminder-1.30.13~20170413095351-xenial. There were always extra 3B in image and browser refuses to show it. Reason for that was in polish translation file(pl_pl.php) that contains Byte Order Mark(ef bb bf). After removing it everything is working fine.

Now I'm asking you to remove BOM from files in git repository simply by using dos2unix on those files, here is the list:
zoneminder.git$ find . -printf '%p ' -exec hd -n 3 UTF8 {} \; 2>/dev/null |grep "ef bb"
./web/api/lib/Cake/Test/Fixture/sample.xml 00000000  ef bb bf                                          |...|
./web/lang/pl_pl.php 00000000  ef bb bf                                          |...|
./web/skins/classic/includes/export_functions.php 00000000 ef bb bf                                          |...|

-- 
fri.K 

Basically there is some non-utf8 code at the beginning of these three files.  

diff --git a/web/lang/pl_pl.php b/web/lang/pl_pl.php
index 4edb6b118..325c2e1b9 100644
--- a/web/lang/pl_pl.php
+++ b/web/lang/pl_pl.php
@@ -1,4 +1,4 @@
-<U+FEFF><?php
+<?php
 //
 // ZoneMinder web Polish language file, $Date$, $Revision$
 // Copyright (C) 2001-2008 Philip Coombes

connortechnology:
Fixes #1855

### DIFF
--- a/web/api/lib/Cake/Test/Fixture/sample.xml
+++ b/web/api/lib/Cake/Test/Fixture/sample.xml
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0"?>
+<?xml version="1.0"?>
 <tags>
 	<tag id="1">
 		<name>defect</name>
--- a/web/lang/pl_pl.php
+++ b/web/lang/pl_pl.php
@@ -1,4 +1,4 @@
-﻿<?php
+<?php
 //
 // ZoneMinder web Polish language file, $Date$, $Revision$
 // Copyright (C) 2001-2008 Philip Coombes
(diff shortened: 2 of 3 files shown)


knnniggett:
I don't understand why the byte order mask of U+FEFF causes a problem, but the Wiki says it is optional.

Reminds me that, I'm not sure we need the cake files found under the cake Test folder, but that is another discussion for another day.

PR #1857, opened 2017-04-18
Storageareas

SteveGilvarry:
Putting this out there for first look at differences


SteveGilvarry:
To do

- [ ] Revert Readme changes
- [ ] Consolidate SQL changes into correct version file for upgrade from 1.30.1 to 1.31.x

PR #1858, opened 2017-04-18
fixes escaping

### LINKED ISSUE #1849: Editing a Monitor doesn't restart zmc/zma
In an effort to protect our calls to exec, we(I) botched the calls to zmdc such that it no longer correctly parses the command line and doesn't restart zma/zmc.  

I will have a PR soon to fix this, but I want to go further, so it will take some time.  We need to check zmdc.pl output, and I think we should check for a pid change.  Makes restarting zma/zmc heavier but more robust.

Also, we need an automated testing infrastructure.  I intend to start roughing that in.

This affects 1.30.2.

connortechnology:
escapeshellarg adds quotes, which is bad.  Use escapeshellcmd on the whole string instead.

fix #1849

### DIFF
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -832,12 +832,13 @@ function packageControl( $command ) {
 function daemonControl( $command, $daemon=false, $args=false ) {
   $string = ZM_PATH_BIN."/zmdc.pl $command";
   if ( $daemon ) {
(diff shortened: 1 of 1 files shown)


knnniggett:
I put a $15 bounty on #1849
go claim it, and don't spend it all in one place.

PR #1859, opened 2017-04-21
add a line to pick up the host to connect to for mysql

connortechnology:
The default database.php include doesn't have a line to pick up the DB_HOST setting. This works fine for a local db server.  However if a remote db server is used, api config won't work.  This should have no ill effects, and is well testing in my storageareas branch.

### DIFF
--- a/web/api/app/Config/database.php.default
+++ b/web/api/app/Config/database.php.default
@@ -67,6 +67,7 @@ class DATABASE_CONFIG {
 	public $default = array(
 		'datasource' => 'Database/Mysql',
 		'persistent' => false,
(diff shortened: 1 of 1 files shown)


josh4trunks:
@connortechnology 
This change is not needed.
If you see near the bottom of this file ZM_DB_HOST is broken up into 'host'/'port'/'unix_socket', and added to $default.

Did you find a scenario where the current code is not working?

connortechnology:
yeah must be a holdover from before we merged that stuff.

josh4trunks:
sounds good =]

PR #1470, opened 2016-05-07
Api improvements

connortechnology:
This simplifies the configuration of the API by pulling the db connection info from zm.conf.

PR #1860, opened 2017-04-21
Update documentation

josh4trunks:
ZM_DB_HOST should be documented somewhere. Tell me if there is a better spot.
Also, `web/api/app/Config/database.php` is handled automatically thanks to https://github.com/ZoneMinder/ZoneMinder/pull/1470


josh4trunks:
=]

PR #1861, opened 2017-04-25
Use ctype_digit

josh4trunks:
I believe the Ctype extension is now a dependency of ZoneMinder.
If so, this use case could be done slightly more efficiently (faster/less overhead) with ctype_digit() than is_numeric().

If the Ctype extension is not a required dependency please close this Pull Request.
Thanks


connortechnology:
technically is_numeric will return true for floats as well, so ctype_digit is actually preferred for correctness sake.

josh4trunks:
awesome =]

PR #1863, opened 2017-04-26
don't overwrite zm.conf

connortechnology:
This is another attempt to get CMAKE to not overwrite an existing zm.conf if you do make install.


knnniggett:
How about we save the file as zm.conf.new instead if the file already exists?
That way the end user has a way to know if we slip in a new variable into that config file.

PR #1864, opened 2017-04-26
Icon

connortechnology:
This PR is all about adding an icon/menu entry for ZM.  Hopefully that will help people who are left wondering how to access their newly installed zoneminder.  Another thing it may allow us to do is switch to using a non-standard port for ZM, which would simplify apache configuration.


CMakeLists.txt
```diff
@@ -749,3 +749,5 @@ if(CCACHE_FOUND)
         set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
         set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
 endif(CCACHE_FOUND)
+
+INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/misc/zoneminder.desktop DESTINATION share/applications)
```
knight-of-ni, line 753: Instead of hard coding the destination share/applications, can we use ${CMAKE_INSTALL_DATAROOTDIR}/applications instead? That's how we've done it elsewhere.

knnniggett:
This failed because it could not find zoneminder.desktop
```
CMake Error at cmake_install.cmake:52 (FILE):

  file INSTALL cannot find

  "/build/zoneminder-1.30.2+97/dbuild/misc/zoneminder.desktop".
```

This could be because the file was not copied into the buildroot during an out-of-source build. Files that are not ".in" need a copy command to copy the file into the cmake_binary_dir

Try doing something like this:
https://github.com/ZoneMinder/ZoneMinder/blob/master/scripts/ZoneMinder/CMakeLists.txt#L4

PR #1867, opened 2017-04-29
Add Socket Support to zmupdate

josh4trunks:
I will report back when I have a chance to test this as working.
Feel free to test and report back in the meantime though =]

### DIFF
--- a/scripts/zmupdate.pl.in
+++ b/scripts/zmupdate.pl.in
@@ -412,12 +412,12 @@ if ( $version )
         {
             my ( $host, $portOrSocket ) = ( $Config{ZM_DB_HOST} =~ /^([^:]+)(?::(.+))?$/ );
             my $command = "mysqldump";
-	    if ( defined($portOrSocket) )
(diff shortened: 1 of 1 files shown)


connortechnology:
Looks good except that we really need to start enforcing the use of google code style rules.

josh4trunks:
I'm not familiar with them. What should I reading?

Is there anything I should look at cleaning up for this PR, or are referring to this script in general?

josh4trunks:
Ok, I tested this as working for a few different cases including without a port/socket specified, with a custom port specified, and custom socket specified.

knnniggett:
For the sake of learning, what is the motivator behind this? Why do we want to do this?

josh4trunks:
@knnniggett 
Without the proposed fix, this script will not handle a **ZM_DB_HOST** like `localhost:/path/to/mysql.sock` correctly. It would pass the socket path as a port argument.

josh4trunks:
=]

PR #1872, opened 2017-05-05
fix uridecode(password) to uridecode(value)

connortechnology:
This fixes the auth failure by user&pass


knnniggett:
I verified this fixes the issue where url's in the following format did not work:
https://localhost/cgi-bin-zm/zms?mode=single&scale=100&monitor=4&user=username&pass=password

PR #1893, opened 2017-05-24
New skin. Bork

xztraz:
Hi. Got bored of the original skins and mashed something together without thinking about uploading it. But it got quite nice so here i am.
I might have made a bit of a mess using v1.30.2 files as a template. so some work probably remains.

Here are som screens: 

image

image

image

image

image

image


xztraz:
Hm build failed. 

ah yeah. based on flat so a diff against the current version should be done.

xztraz:
These are the only files iv'e changed/added from flat (the rest could just be overwritten with latest versions from flat but i suck at git.. so please help)

css/bork/skin.css
css/bork/views/event.css
css/bork/views/montage_3wide50enlarge.css
css/bork/views/montage.css
css/bork/views/watch.css
js/bork.js

xztraz:
diffed my changed files against v1.30.4. Should be ok. so just overvrite the other files with latest versions from flat and it should work.  sorry for messy upload

connortechnology:
Hey thanks for the contribution!

I probably won't run this, because I don't like dark themes.

This isn't that far off the dark version of the flat theme.  Perhaps it could take the place of the dark theme.  

I don't think we should merge this right now... as we are really unstable.  Also, I plan to alter the classic skin so that themes that are just colour/spacing changes are minimized...

So I hope you will stick around and play an important role in improving our interfaces.

xztraz:
Hi!

well. i like dark themes.. :p 

it's not that much changed from flat but nicer to the eyes if looking at dark cameras a long time. and some depth effects with some easy shadows.

Also adjusted all poppus i could since they are a bit under sized in flat.

Could you use say flat as a base and only lift in the changed css:es from another skin? only changing colors doesn't give the 3d effect.

xztraz:
Listen to that guy. 😀

Zoneminder is great at surveilance. stable and we'll coded but I've always been missing a bit more updated gui. This is a little step atleast. Hope it can be implemented for others to enjoy.

Something I'm missing is graphical buttons for play stop and such. 

Do you have statistics on with skins are used most?

connortechnology:
No we don't .  

This PR won't be merged until it is updated to master.  I expect master to not be developed further (by me anyways) so it is less of a moving target.

PR #1894, opened 2017-05-25
Dragndrop monitor sorting

connortechnology:
The pulls out the code from SA that replaces the up/down arrows with a dragndrop function.


scripts/ZoneMinder/lib/ZoneMinder/Filter.pm
```diff
@@ -240,6 +195,9 @@ sub Sql {
                 # This gets used later, I forget for what
                 $$self{Server} = new ZoneMinder::Server( $temp_value );
               }
+            } elsif ( $term->{attr} eq 'StorageId' ) {
```
SteveGilvarry, line 198: @connortechnology this supposed to be here?

connortechnology:
Oh no.. there is tons of stuff in here that shouldn't be there.

PR #1896, opened 2017-05-26
Dragndrop monitor sorting

connortechnology:


connortechnology:
screenshot_2017-05-26_16-08-58

This is what it looks like.  So you click and drag on the up/down arrows. The images are part of the glyphicon set that comes with bootstrap.  

The code uses jQuery, so we are including that as well. Shouldn't conflict with mootools.

web/skins/classic/includes/functions.php
```diff
@@ -78,6 +76,20 @@ function xhtmlHeaders( $file, $title )
   <script type="text/javascript" src="tools/mootools/mootools-core.js"></script>
   <script type="text/javascript" src="tools/mootools/mootools-more.js"></script>
   <script type="text/javascript" src="js/mootools.ext.js"></script>
+  <script type="text/javascript" src="skins/<?php echo $skin; ?>/js/jquery-1.11.3.js"></script>
```
knight-of-ni: Instead of referencing jquery-1.11.3.js here, can we instead reference the symlink jquery.js which, today, points to jquery-1.11.3? Should make future upgrades of jquery easier.

connortechnology: agreed.

PR #1897, opened 2017-05-29
Enchancing Dockerfile and docker execution

SeraphicRav:
Following best practices from Docker
Added some advices to build and run docker image
Solved timezone problem and allow customization
Expose MySQL data and ZoneMinder logs as a volume (I still have to find where events and images are saved to do the same with them)


SeraphicRav:
Made a mistake on when to launch the setup script

Issue #1900, opened 2017-06-02
Danish translation lagging behind


jedelbo:
dk_dk.php is out of sync with e.g en_gb.php, so a lot of terms are not translated. Some of the current translations are also questionable.

SteveGilvarry:
All translations are community managed so happy to help you if you have suggested changes.

knnniggett:
#1901 fixed this so closing

PR #1901, opened 2017-06-02
Update danish translation file

jedelbo:
Solves #1900


SteveGilvarry:
Thank you for taking the time to fix.

PR #1908, opened 2017-06-06
Migrate values from Options -> Paths to zm.conf

### LINKED ISSUE #790: DB is used to store essential config paths
Normally variables like `ZM_PATH_SOCKS` should not be stored in the database because they may need to be used by service startup scripts (init.d, systemd).
It would be highly undesirable to write service startup scripts with functionality to fetch configuration from database.
From packaging prospective it is very inconvenient that location of sock files or `PATH_SWAP` can't be changed by re-deploying binary package built with customised `ZM_SOCKDIR` because those locations can be changed only by modifying database. 

Usually path to socks directory is saved to configuration file under `/etc`.
It is unnecessary (and perhaps undesirable) to allow changing such essential variables through web interface. Also there is a considerable security risk when unprivileged user (not root) can change essential service startup parameters.

Besides it is very counter-intuitive that variables like `ZM_SOCKDIR` given on build time are used merely to generate database initialisation script.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10545158-db-is-used-to-store-essential-config-paths?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F188329&utm_medium=issues&utm_source=github).
</bountysource-plugin>

knnniggett:
fixes #790

The values on the paths tab are not portable. When the end user backs up the database from one system and restores it to another, the Options -> Paths values must be reviewed for accuracy on the new system. Much time can be spent troubleshooting if the end user does not know to check this. This pr does away with that issue by migrating the values out of the database and into zm.conf.

This pr also simplifies the next pr I plan to make, which will remove zoneminder's reliance on symlinks.

This needs testing with debian packages built against this pr to verify I made the appropriate changes to the packaging scripts.

@connortechnology please advise how this will impact storage areas.


distros/ubuntu1604/rules
```diff
@@ -25,7 +25,10 @@ override_dh_auto_configure:
             -DZM_SOCKDIR="/var/run/zm"                 \
             -DZM_TMPDIR="/tmp/zm"                      \
             -DZM_CGIDIR="/usr/lib/zoneminder/cgi-bin"  \
-            -DZM_CONTENTDIR="/var/cache/zoneminder"
+            -DZM_CONTENTDIR="/var/cache/zoneminder"    \
```
connortechnology, line 28: I don't require anything to be done about this, but I can't find anywhere where ZM_CONTENTDIR is used.

connortechnology:
This looks pretty good to me. My only concern is going to be people who have changed their DIR_EVENTS setting.  

I'll see if I can figure/PR some package update scripts to take care of detecting that and adding a file into conf.d

connortechnology:
This update could be catastrophic if anything goes wrong.  I am considering disabling zmaudit automatically during upgrade. Thoughts?

I think we need a perl script zmconfig.pl that reads/reports/edits the db config.

knnniggett:
That's correct. If the end user has changed the value of DIR_EVENTS to something else then data loss could occur if they try to start zoneminder w/o first adding a new config file under conf.d. Our documentation (help text & the dedicated drive wiki) both warn against directly changing DIR_EVENTS, however. I do plan to add a note about this in the 1.31.1 release notes. Perhaps we could also output a warning in zmupdate.pl.

Note the a zmupdate.pl -f will wipe the values from the database, which on some systems happens during package installation.... so I don't currently know of a fool proof way to retrieve custom values before they get deleted.

distros/ubuntu1604/rules
knight-of-ni: I might have gotten ahead of myself on this one.... when I modify zoneminder and our packaging scripts to not do symlinks, I expect to need this to create the contentdir folder. If not, I'll remove it.

knnniggett:
Actually, maybe that is the key to making this foolproof....

During a "zmupdate -f" first check `DIR_EVENTS != events`. If that is true then autocreate a config file under conf.d that sets DIR_EVENTS to the end users custom value

knnniggett:
Just pushed a commit which retrieves the current values of the variables on Options -> Paths and writes them to a config file under conf.d just before the dB is freshened and the variables are deleted from the database.

This needs a little more tweaking as it is triggering too often under certain conditions.

knnniggett:
I believe 827d729 resolves the issue where the new subroutine writes out config variables too often. Will test this tomorrow.

PR #1911, opened 2017-06-09
Fix 1498

connortechnology:
#1498 

There is a comment in there regarding this, but no response.

@josh4trunks 

Can you review/comment?


connortechnology:
One might ask: why is the db connect logic in zm_logger.cpp?

josh4trunks:
@connortechnology , did you mean this comment https://github.com/ZoneMinder/ZoneMinder/pull/1498#r73949773?
I addressed it here https://github.com/ZoneMinder/ZoneMinder/pull/1498#issuecomment-238396576.

I don't know why this file has the logic instead of using it from another file, someone else structured it this way. I just made sure it works properly.

PR #1912, opened 2017-06-09
minimize Travis log output

knnniggett:
I want to watch the results of this before merging.

It should pipe the output of packpack to a file, thus avoid the Travis 4MB log output limitation


knnniggett:
Self merging this to get into master. We need to wait a day for Travis to spin up a cron task on master before we will know this fixes the log limitation problem w/o creating any new problems.

PR #1919, opened 2017-06-13
Fix FTBS on F27

knnniggett:
move include <sys/uio.h> outside defined(BSD) block

Tested not to negatively affect:
el6
el7
f25
f26
Ubuntu 17.04


connortechnology:
ubuntu 1604 is good too.

PR #1925, opened 2017-06-18
Fix spelling mistake in link

doctormo:


knnniggett:
I still couldn't immediately see that missing "i" even after github highlighted the word.

PR #1929, opened 2017-06-21
Fix event viewing

connortechnology:
split streamPause and streamPlay into two functions.  Because the response also calls streamPause and streamPlay, it loops forever.  Now clicking sends the command and the response updates the buttons.

Change logic a bit to always spit out a frame immediately when zooming changes.


knnniggett:
@connortechnology Strange, I didn't see this PR earlier. Maybe I should have merged this one instead of #1930. It looks like we have merge conflicts now. Can you fix?

connortechnology:
Yeah I'll fix it up.

PR #1933, opened 2017-06-23
I believe level should be effectiveLevel here

connortechnology:
effectiveLevel exists to provide a simple check whether we should be logging anything.  So it should be the lowest value of any of the output specific levels.  

The fix seems obvious, but maybe too obvious.  I don't see how this could have been bungled.


knnniggett:
Why did you merge this yourself?
I was attempting to test this pr this morning but was hit with the "not null" issue with the Options column.
When that was fixed, I returned to this pr only to find you merged it already.

PR #1937, opened 2017-07-02
reorganize zm config files

knnniggett:
The goal of this pr is to avoid the condition where the target system package manager defaults to not overwrite config files.

This is done by putting zm.conf back to the state it was in prior to 1.31.0 and creating a new config file under conf.d called 01-system-paths.conf containing the variable newly migrated out of the dB.

This pr also includes rpm specfile changes to accommodate the new files. Changes to the debian packaging scripts were not required.


connortechnology:
I tested this and the files put in conf.d had @ZM_BLAH@ replacements instead of the values. Have we missed a step?

knnniggett:
Oh right, I need to make this a .in file and tell cmake to process it.

knnniggett:
should work now

PR #1938, opened 2017-07-03
Flush output buffer in view_video.php

### LINKED ISSUE #1914: Issue playing back high resolution h264 events
I'm running a pull of the master branch from 6/9/17 (1.31.0). I seem to be able to playback events from my lower resolution sub streams (640x360), but can't playback events that are triggered for the higher resolution stream (2048x1536).

I'm using Firefox and get the error **`The video could not be loaded, either because the server or network failed or because the format is not supported`**

When using IE it just spins trying to load forever.

I can see the events are being generated, as I can SCP to the server and pull down the MP4's which play fine locally. Not sure where to start looking to resolve this, whether it's an apache issue or if it's some configuration within zoneminder.

AndrolGenhald:
Fixes #1914

### DIFF
--- a/web/views/view_video.php
+++ b/web/views/view_video.php
@@ -98,6 +98,7 @@ while( $length && ( ! feof( $fh ) ) && ( connection_status() == 0 ) ) {
   $amount = min( 1024*16, $length );
 
   print fread( $fh, $amount );
+  ob_flush();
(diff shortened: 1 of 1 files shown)


connortechnology:
#1914 is no longer an issue. This is not required.

PR #1940, opened 2017-07-05
Montagereview rework

### LINKED ISSUE #1611: Issues with image rotation in montage review view
This is related to "Issues with image rotation #1431".  When a camera is rotated 90 deg in the source, the montage review view shows the image shows rotated, but uses a non-rotated resolution.  So a 1920x1080 camera that is rotated 90deg should show as a 1080x1920 image, but instead it is showing as a squashed rotated image at 1920x1080.

Issue 1431 fixed this in all other views.

connortechnology:
This may not be complete, but it works for me.

Mostly this is a code cleanup, in an effort to make myself understand how it works.
It has the benefit of splitting most of the js into it's own file which can get cached.
It also uses the monitor Object, which automatically does the rotation, so this fixes #1611 
Also it is a lot faster.

Further work might be to use an mjpeg stream instead of single images


knnniggett:
Yeah, I had wanted to split out the js, but it was not high on my priority list.

connortechnology:
WHoo.. I better do some testing and build some packages I guess.

PR #1941, opened 2017-07-05
add some comments and one change that may fix effectiveLevel use

connortechnology:
I'm not sure this has the fix, but I've gone through commented on some things.  Would appreciate some additional brains commenting while I test.


knnniggett:
Since the logger in master is already broken, I'll merge this now and then test it later.

PR #1942, opened 2017-07-06
Strip non 7-bit ascii chars from log message

### LINKED ISSUE #1363: Empty System Log screen
When I open the System Log, it's not refreshing. In Chrome's Developer toolbar, I can see a request to index.php with the following parameters:

view:request
request:log
task:query
limit:100
minTime:0

And this is the content I'm getting back:

https://gist.github.com/czettnersandor/332d962d0e1328421ea9

Zoneminder is installed to an Ubuntu 14.04.1 server (32 bit) using the official documenattion:
http://zoneminder.readthedocs.org/en/stable/installationguide/ubuntu.html#easy-way-install-zoneminder-from-a-package-ubuntu-14-x

There was log, but then it broke after a few days of usage.

connortechnology:
Fix #1363

Some elements of our code can insert non-7-bit ascii chars into the logs.  When encoding/decoding these to/from json to view in the web ui the json decode dies.  

So we strip anything that is non 7-bit ascii out.

### DIFF
--- a/web/ajax/log.php
+++ b/web/ajax/log.php
@@ -95,6 +95,7 @@ switch ( $_REQUEST['task'] )
         foreach ( dbFetchAll( $sql, NULL, $values ) as $log ) {
             $log['DateTime'] = preg_replace( '/^\d+/', strftime( "%Y-%m-%d %H:%M:%S", intval($log['TimeKey']) ), $log['TimeKey'] );
             $log['Server'] = ( $log['ServerId'] and isset($servers_by_Id[$log['ServerId']]) ) ? $servers_by_Id[$log['ServerId']]->Name() : '';
+            $log['Message'] = preg_replace('/[\x00-\x1F\x7F-\xFF]/', '', $log['Message'] );
             $logs[] = $log;
         }
         $options = array();


fhenryco:
I have the same problem (empty log screen) and i just did the correction to log.php file  however the logs screen remains empty after restarting zm. May be this is because some files having garbage characters  should be purged so i set log_database_limit to 1 hour and waiting for the log screen to be up again... is this the only thing i should do?  i'm not able to find on my system where are those files with garbage characters , for sure not in /var/log/zm , so what is the name of those db log files to help me find them and can i access them and delete their garbage characters by hand ?

connortechnology:
Db logs are stored in the database.  You need to use a mysql client like this:
mysql -u zmuser -p zm

then enter

select * from Logs;

etc.

fhenryco:
OK thank you. Fortunately after waiting a couple of hours the bad characters have been purged from the database and the Logs screen is up again. I say fortunately because i could enter the db with 
mysql  -p zm  (did not define a zm user after zm install) but then i have no experience with operations inside a db.

PR #1945, opened 2017-07-13
make code more robust by handling an undefined get_VideoEncoderConfig

connortechnology:
https://forums.zoneminder.com/viewtopic.php?f=36&t=26333&sid=1400af332e8f28407b974e19855afb58


connortechnology:
Fixed and tested.

Issue #1946, opened 2017-07-13
Add support for Maginon Supra IPC cameras


MartinGoodwell:
I created a script for that cam and I'd love to contribute it.

connortechnology:
Ok.  You can either send it to me at isaac@zoneminder.com, or fork our repo, add your script, and make a pull request....
If you aren't familiar with github development, just sending it to me might be the easiest.

MartinGoodwell:
Awesome. Created a pull request

connortechnology:
Welcome to ZoneMinder development!  Thanks for your contribution!

MartinGoodwell:
Thanks a lot!

PR #1947, opened 2017-07-13
Introducing Maginon Supra IPC camera Controls and script

MartinGoodwell:
Hope I did that right, according to http://zoneminder.readthedocs.io/en/stable/contributing.html


connortechnology:
fix #1946

PR #1948, opened 2017-07-14
Fix empty monitor source column

connortechnology:
https://forums.zoneminder.com/viewtopic.php?f=37&t=26314

I also just find this code easier to read.


knnniggett:
oh yeah, that's much better.

PR #1954, opened 2017-07-25
Fix install location for config files when building to alternate directory.

ralimi:
With the previous code, we ended up with a directory structure like the following:

$ find /etc/zm/conf.d/
/etc/zm/conf.d/
/etc/zm/conf.d/01-system-paths.conf
/etc/zm/conf.d/conf.d
/etc/zm/conf.d/conf.d/README
/etc/zm/conf.d/conf.d/02-multiserver.conf


ralimi:
Oops. Bug in this. I forgot the README file. I will close this and submit a different request.

Issue #1964, opened 2017-08-13
Support database communication over SSL


ralimi:
My setup has a few Raspberry PIs that I'm trying to keep rather lean and simple, and I'm hosting the database in a cloud provider.  With this setup, I'd like to have mysql connections SSL encrypted and also authenticated with a client certificate.

I've got this working and will send a pull request with the accompanying code.

knnniggett:
closed via #1965

PR #1955, opened 2017-07-25
Fix location for configs when building to alternate directory.

ralimi:
With the previous code, we ended up with a directory structure like the following:

$ find /etc/zm/conf.d/
/etc/zm/conf.d/
/etc/zm/conf.d/01-system-paths.conf
/etc/zm/conf.d/conf.d
/etc/zm/conf.d/conf.d/README
/etc/zm/conf.d/conf.d/02-multiserver.conf

PR #1965, opened 2017-08-13
Support SSL for mysql connections

ralimi:
This addresses #1964.

In theory, this should support cases with and without client certificates specified.  I have only validated it with client certificates, though.

I'm clearly still learning the ropes of how to manage git branches.  This unfortunately includes some of the changes in #1955.  I'm guessing those would be silently ignored if that pull request is accepted sooner?  Let me know if this is a problem and I'll go fix things up on my end.


CMakeLists.txt
```diff
@@ -801,7 +801,8 @@ endif(ZM_PERL_SEARCH_PATH)
 
 # If this is an out-of-source build, copy the files we need to the binary directory
 if(NOT (CMAKE_BINARY_DIR STREQUAL CMAKE_SOURCE_DIR))
-	file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/conf.d" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/conf.d" PATTERN "*.in" EXCLUDE)
+	file(GLOB ZM_CONFD_FILES "${CMAKE_CURRENT_SOURCE_DIR}/conf.d/*")
+	file(COPY ${ZM_CONFD_FILES} DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/conf.d" PATTERN "*.in" EXCLUDE)
```
knight-of-ni: This change is (already) part of #1955. Having this same change in pr #1965 will cause a merge conflict in whichever pr we attempt to merge second. You should remove this change from #1965. UPDATE: Right, I see your comment now that you noticed this too. The way to handle this is to just push a new commit to this pr which puts line 804 back to the way it was.

knnniggett:
We know from past experience that our debian userbase tends to accept the default of "no" when it prompts to overwrite an existing zm.conf with the new one, which it will in this case. That means they won't get the changes in this pr.

Not sure if this is going how big of an issue this will really be, but I'm tempted to create a new config file "conf.d/03-database.conf.in" which contains all of the ZM_DB vars, old and new. Suppose we could always do this later.

@connortechnology thoughts?

connortechnology:
Answering N won't break existing systems though, so I'm good with it.

knnniggett:
@connortechnology @ralimi 
This PR has caused default instances of zoneminder (with default sql config) to fail out of the box.

The zm web portal shows nothing but:
Unable to connect to ZM db.SQLSTATE[HY000] [2006] MySQL server has gone away

And the http log shows:
```
[Tue Aug 15 17:28:32.353923 2017] [php7:error] [pid 841] [client 127.0.0.1:48974] PHP Fatal error:  Uncaught Error: Call to a member function query() on null in /usr/share/zoneminder/www/includes/config.php:160\n
Stack trace:\n
#0 /usr/share/zoneminder/www/includes/config.php(134): loadConfig()\n
#1 /usr/share/zoneminder/www/index.php(46): require_once('/usr/share/zone...')\n
#2 {main}\n
  thrown in /usr/share/zoneminder/www/includes/config.php on line 160
```
This goes away if I roll back database.php to the prior commit.
That's as far as I a have been able to troubleshoot at the moment.

connortechnology:
I just pushed a fix that I am using

knnniggett:
Cool, that fixed it. Thanks

ralimi:
Thank you!  Apologies for the bug with non-SSL connections.

PR #1971, opened 2017-08-20
Added user-based quota support

keynet:
This branch adds simple user-based quota support so that the admin doesn't need to maintain a separate disk partition for ZM events to use filters effectively, or change disk partition sizes or use other methods of disk management as needs change.
The % disk used displayed on the GUI, and used by all the filters, respects a soft user quota if configured for the user running Zoneminder, (e.g. wwwrun in my case).  If no quota is configured, ZM maintains the original behaviour. The quota can be configured by a standard command line tool, or for example Webmin, and is picked up immediately.
User quotas are used on the basis that a) project quotas are only to be readable by root, only work with certain file systems like XFS and require additional configuration amongst other issues which add complexity, and b) Usually a non-root, non-regular user runs ZM, so setting a quota for that user for the partition storing events should rarely create unintended restrictions.  I've been using this code on a live system for a few months now.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Filter.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Filter.pm
@@ -321,21 +321,43 @@ sub Sql {
 } # end sub Sql
 
 sub getDiskPercent {
-  my $command = "df " . ($_[0] ? $_[0] : '.');
-  my $df = qx( $command );
+#RF Add quota support for user running filter - as functions.php
   my $space = -1;
-  if ( $df =~ /\s(\d+)%/ms ) {
-    $space = $1;
+  #get mountpoint
+  my $command = "stat -c %m " . ($_[0] ? $_[0] : '.');
+  my $df = qx( $command );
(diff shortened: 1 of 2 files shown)


connortechnology:
Just a comment about this... I don't use quota, so I don't have the quota tools installed.  So if we merged this, there would be errors in the logs, etc.
I kinda feel like it should be wrapped in a config entry to turn on quota support.  I feel like we already have too many config options.  I wonder if we can auto-detect quota usage.

keynet:
Agree don't really want more config options, which is why I wrote it the way I did. On my system (Opensuse Leap 42.2) uninstalling the quota package produces no log errors, even with debug switched on - but I may not have all the right debug options to see it of course.

If I'm right, if the quota package isn't installed, or the filesystem doesn't support quotas, or no quotas configured, ZM falls back silently to expected original behaviour.  So this is kind-of auto-detect?

We should perhaps add a dependency on quota tools in the RPM builds? Pretty standard package, only ~1MB installed size, so no burden on stripped down systems.

PR #1972, opened 2017-08-20
allow changing of the timestamp char code delimeter

### LINKED ISSUE #1865: Timestamp Label Format (%N, %f, %Q) / behaviour of strftime() on FreeBSD
When using ZoneMinder on FreeBSD (at least on FreeBSD 11.0), all my cameras are named "N" in the image timestamps. This is because strftime() on FreeBSD removes the `%` sign of unknown conversion specifications instead of not touching it.
On Linux, where strftime() does not touch the unknown `%N`, `%f`, and `%Q` in the LabelFormat string, camera names are printed correctly into the images.
As a workaround, putting `%%N` in the LabelFormat string of a camera makes the substitiutions work on both Linux and FreeBSD for my cameras, because strftime() than simply substitutes `%%` with `%` and so leaves `%N` in the string.

I would like to fix this but are not sure what solution is the "best". There are several possible solutions:

- add a second `%` sign to the default LabelFormat string for the camera name (make it `%%N - %d/%m/%y %H:%M:%S`) or
- use a different character for substitutions that are manually done after strftime() (i.e. `!N - %d/%m/%y %H:%M:%S`) or
- change the logic in `src/zm_monitor.cpp` to first replace ZoneMinder substitutions and than use strftime() on the LabelFormat string (this would be in `Monitor::TimestampImage()` beginning around line 3151 and could be dangerous because of user input).

Unfortunately the behaviour of strftime() in situations where it finds unknown conversion specifications is undefined, so I would prefer to change the default LabelFormat string to double the `%`. So every existing installation would work as before and new cameras on updated or new installations would have the new default LabelFormat string. Also, the documentation should be updated then.

What do you think?

Thanks for reading this longish text...
Robin

knnniggett:
fixes #1865 

This PR allows one to change the timestamp code character, under Options -> Config, from the default "%" to anything the user wants. Note that it only affects ZoneMinder's builtin codes. This has no affect on the standard strftime codes.

As described in #1865, this is one way to address the issue of BSD operating systems handling the % character differently than their Linux counterpart.

I'm willing to hear alternative methods to fix this.

@schrorg  @abishai 
Please test this and provide feedback as I don't currently have any FreeBSD systems running zoneminder.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
@@ -701,6 +701,25 @@ our @options = (
     category    => 'config',
   },
   {
+    name        => 'ZM_TIMESTAMP_CODE_CHAR',
+    default     => '%',
+    description => 'Character to used to identify timestamp codes',
+    help        => q`
+      There are a few codes one can use to tell ZoneMinder to insert
+      data into the timestamp of each image. Traditionally, the
+      percent (%) character has been used to identify these codes since
+      the current character codes do not conflict with the strftime
+      codes, which can also be used in the timestamp. While this works
+      well for Linux, this does not work well for BSD operating systems.
+      Changing the default character to something else, such as an
+      exclamation point (!), resolves the issue. Note this only affects
+      the timestamp codes built into ZoneMinder. It has no effect on
+      the family of strftime codes one can use.
+      `,
+    type        => $types{string},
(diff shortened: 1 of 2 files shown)


schrorg:
I will test this and report back, but it can take until the end of the week. Lots of other work to do here...

knnniggett:
Not a problem @schrorg 
I don't have a timeline for this.

connortechnology:
I have no problem merging this, it is a simple fix.

PR #1981, opened 2017-09-09
initial commit to change color of web console systemstats

knnniggett:
This checks the three system stats on the web console: system load, disk space, and path_map (usually /dev/shm).

It checks each parameter against various thresholds and changes the color of the parameter from Normal -> Orange (yellow looked really bad) -> Red -> Blinking Red.

Initial testing was done with the classic skin. I used inline css to change the colors. This needs testing against the flat skin. If this does not look right with the flat skin then I will integrate the css into the skin files. Feedback is requested.

In addition to this, I am also considering using zmwatch or zmaudit to log periodic warnings to the zoneminder log should one of the parameters enter the red zone.

### DIFF
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -1517,20 +1517,20 @@ function systemStats() {
     if ( $normalized_load <= 0.75 ) {
         $htmlLoad=$load;
     } elseif ( $normalized_load <= 0.9 ) {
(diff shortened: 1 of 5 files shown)


connortechnology:
I think we should use a class instead of hard coding the colours.  So assign a class of warning, error, critical and then set the colours in the .css files.

knnniggett:
@connortechnology updated to make use of css classes

PR #1997, opened 2017-10-09
Added control script for Trendnet TV-IP672WI IP camera

michaelbarkdoll:
Added control script for Trendnet TV-IP672WI IP camera


michaelbarkdoll:
Note: Previously released TV-IP672PI script doesn't work with my TV-IP672WI.  This uploaded version works with additional error checking and parsing capabilities for the username/password fields from various Control Device/Control Address entries.

knnniggett:
Thank you for this, but you need to add the necessary bits of sql to db/zm_create.sql.in so this shows up in the web console.

Here is an example:
https://github.com/ZoneMinder/ZoneMinder/commit/4be668f838b59614d204c963ffec5953bdda1786#diff-748ea98014ecf152d14a6001cf4b673c

michaelbarkdoll:
Added control script and SQL insertion for Trendnet TV-IP672WI IP camera

connortechnology:
I'm going to look this over and see if we can't merge with scripts for other TrendNet cameras.  Maybe we can make 1 that does all their models or something.

michaelbarkdoll:
I have another Trendnet camera with a similar model number that uses digest camera controls (much different than this script).  I'd suggest making it per model number to avoid inaccuracies and a script that is overly complex to diagnosis for other camera models...

michaelbarkdoll:
Any updates on a better script for this purpose?

connortechnology:
Allow me to explain why we didn't do anything with this.  We already have a mess of scripts for different cameras with only subtle differences between them. We need to move towards better scripts that handle more cameras.  Also in this and the parent scripts, there are problems that are not being fixed.

I instead took a close look at improving the situation. I took the existing TVIP862 script and improved it.  I believe I have fixed most of the glaring issues and solved the auth issues.  I will rename it to Trendnet.pm as I believe it will handle most TrendNet cameras.  

I expect to have it available by Monday.  

Thanks for your efforts.

PR #2002, opened 2017-10-29
fix generated video download

### LINKED ISSUE #2001: 1.30.4 Download generated video exhausts memory and fails.
After generating a video file under the video link in event view, clicking the download link will 404 if the file is too big due to exhausting phps allowed memory.  Someone in irc reported this and it looks like it only affects 1.30.4, not current master.  

It's due to output buffering trying to load the entire file into memory.  A solution is to add 
`    while (ob_get_level() > 0) {
     ob_end_clean();
    }
`
to /usr/share/zoneminder/www/skins/classic/views/video.php at line 104, just before readfile is called.  This is not the only solution since this bug doesn't exist in 1.31.

digital-gnome:
Fixes out of memory php error when downloading large generated video files.  Fixes #2001

### DIFF
--- a/web/skins/classic/views/video.php
+++ b/web/skins/classic/views/video.php
@@ -101,6 +101,9 @@ if ( isset($_REQUEST['downloadIndex']) )
     header( "Content-Transfer-Encoding: binary" );
     header( "Content-Type: application/force-download" );
     header( "Content-Length: ".filesize($videoFiles[$downloadIndex]) ); 
+    while ( ob_get_level() > 0 ) {
+     ob_end_clean();
+    }
     readfile( $videoFiles[$downloadIndex] );
     exit;
 }


connortechnology:
I'm not sure that this is correct (It might be, I'm just not sure).

I don't see how discarding the output buffer would help.  Theoretically it should be empty at this point, other than the headers we just output.

Typically you to do something other than readfile... something   like

while( $length && ( ! feof( $fh ) ) && ( connection_status() == 0 ) ) {
  $amount = min( 1024*16, $length );

  print fread( $fh, $amount );
  $length -= $amount;
   flush();
}

Can you comment on why this fixes the problem?

connortechnology:
I think it's the csrf stuff.  Problem is eventually there will be a file large enough to run out of ram when using readfile. Need to process it in chunks.

digital-gnome:
Php docs say
>readfile() will not present any memory issues, even when sending large files, on its own. If you encounter an out of memory error ensure that output buffering is off with ob_get_level().

I tested it with a file twice the size of the ram allocated to my vm.  Im pretty sure it's just the nested output buffers.

connortechnology:
I think the reason output buffering is on is the csrf stuff.  I'm not sure if turning off the buffering defeats the csrf protection.  Seems like it should be ok, as the tokens will have been checked before getting to this code.

connortechnology:
So I'm not going to merge this as-is, because it has to go to the release-1.30.5 branch which I will be creating soon.

digital-gnome:
Back at a real keyboard.  Output buffering is on in master but it works fine.  From what I can find about this it's nested buffering that breaks readfile.  1.30.4 has two levels of buffering as shown by ob_get_level().  Both of them appear to be empty but not closed at the point where readfile is called.  It works on master because it's only a single level of buffering.   Turning off the buffering just before the download should still be ok because it's the end of the line and as you said everything should have been checked before this last action.  

Looking at my bug report I forgot about the nested buffering part and that's key.  Sorry about that.  I thought I'd typed that in somewhere.

Issue #2014, opened 2017-12-06
Sorting monitors doesn't work


kimmeld:
Sorting monitors in the console does not work.  Drag-and-drop changes the order in the table, however this is not persisted in the database.  A JavaScript error is thrown.

kimmeld:
Fixed.

PR #2015, opened 2017-12-06
Fix monitor sorting

kimmeld:
Fixed issue #2014 

The AJAX request to update the monitor sorting goes to /index.php, which doesn't exist. This modifies it to go to index.php (which results in going to /zm/index.php).


connortechnology:
Nice find!

PR #2017, opened 2017-12-07
Fix Montage CSS

### LINKED ISSUE #2016: CSS Weirdness in Montage view
When viewing the montage, monitors that are in Alert status get a bunch of extra padding compared to Idle or Alarm status.  This makes the monitors jump around a little as they transition through the states.

kimmeld:
Fixes #2016 

For some reason, the alert style used in the montage view picks up a padding attribute from elsewhere.  I've added "padding: 0;" to the alert style to prevent this and to the alarm and idle styles for consistencies.


digital-gnome:
Is it still doing this after  #2013 was merged?  I thought I'd fixed this.

kimmeld:
I'm betting that zmrepo didn't have that when I last updated as your fix is pretty new.  Your fix is more comprehensive than mine, so I'm going to close this PR.

PR #2018, opened 2017-12-07
add a robots.txt to the docroot to tell bots to not scrape our ZM install

connortechnology:
Won't do anything for sites configured just using the /zm alias, but for those of who dedicate a virtualhost to ZM, it will stop our logs from being filled up with errors from bots crawling our ZM install.


joshuaherman:
It won't do anything to bots crawling IP addresses.

PR #2021, opened 2017-12-09
call zmcamtool during postinst

knnniggett:
@connortechnology 
Setting this as a pr so you see this.

This uses zmcamtool to non-destructively import new ptz configs during an upgrade. It will not overwrite.

Been doing this for some time with the redhat packaging scripts:
https://github.com/ZoneMinder/ZoneMinder/blob/master/distros/redhat/zoneminder.spec#L198


connortechnology:
nice!

PR #2022, opened 2017-12-11
alter the type of the Channel field to match the update that occurred…

connortechnology:
… in 1.23.3

### DIFF
--- a/db/zm_create.sql.in
+++ b/db/zm_create.sql.in
@@ -291,7 +291,7 @@ CREATE TABLE `MonitorPresets` (
   `Name` varchar(64) NOT NULL default '',
   `Type` enum('Local','Remote','File','Ffmpeg','Libvlc','cURL') NOT NULL default 'Local',
   `Device` tinytext,
-  `Channel` tinytext,
+  `Channel` tinyint(3) unsigned default NULL,
   `Format` int(10) unsigned default NULL,
   `Protocol` varchar(16) default NULL,
   `Method` varchar(16) default NULL,


knnniggett:
Wow. Just wow. 10 year old bug?

knnniggett:
@connortechnology I just found out why that field had a value of tinytext, rather than tinyint.

Creating a new database currently fails with:
`ERROR 1366 (HY000) at line 611: Incorrect integer value: 'simple' for column 'Channel' at row 1`

And line 611 is this:
https://github.com/ZoneMinder/zoneminder/blob/master/db/zm_create.sql.in#L611

Our database create script has the value "simple" for many of our presets. I'm not sure why yet so I am unsure if we should change the default value for our presets are put the file type back to the way it was.

knnniggett:
OK, the fields are misaligned somehow. The "simple" text should be going into the method column, not the channel column. I suspect the monitorpreset table is missing some of the more recent columns added to the monitors table... investigating...

connortechnology:
I don't see how 'Simple' is a valid value for Channel.  Will have to look closer.

knnniggett:
Still looking, but I found the commit that altered the channel column to tinytext:
https://github.com/ZoneMinder/zoneminder/commit/67595c1c842581efa8ffb02ab45339737785ae38

It fixed the same error message I am seeing at the moment, but I am coming to the conclusion that real problem lies in some of the presets themselves. Take this one for example:
`INSERT INTO MonitorPresets VALUES (NULL,'Axis IP, 320x240, mpjpeg','Remote','http','simple',NULL,NULL,NULL,'<ip-address>',80,'/axis-cgi/mjpg/video.cgi?resolution=320x240',NULL,320,240,3,NULL,0,NULL,NULL,NULL,100,100);
`
The "simple" text, which we know should go into the method column, is the 5th field in that sql statement, but if you open the MonitorPresets table in an existing zm db, the method column is the 8th field.... guess which field is the 5th field.... channel.

connortechnology:
I'm tempted to convert them to the Field=Value format just for readability and future-proofing.  I know that in my branch zmcamtool fails for similar reasons. It's been on my todo list.

knnniggett:
proposed a fix in #2022

PR #2024, opened 2017-12-14
Making a functional Dockerfile

rileyschuit:
**Issue:**  Running sample command at the end of the Dockerfile didn't work because the team doesn't have an automated build.  Building the image locally also didn't work because it had no mysql service to connect to.  This should allow that now.

**Summary:**  The Dockerfile allows for a enclosed mysql server or to point to a separate mysql server.  If the env var is not defined, it will build up a local service and let people run out of a single container.  Examples are given at the end file.

Keep up the good work guys.  Love the project and I hope this helps it.  This is my first real contribution to a project, so please be gentle!  And hoping to delete [my automated build](https://hub.docker.com/r/rileyschuit/zoneminder/) when this is accepted. :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,7 @@
-# ZoneMinder, you need the GIT repository code and submodules (git submodule update --init --recursive)
-
 FROM ubuntu:xenial
-MAINTAINER Markos Vakondios <mvakondios@gmail.com>
+MAINTAINER Markos Vakondios <mvakondios@gmail.com> Riley Schuit <riley.schuit@gmail.com>
 
-# Resynchronize the package index files 
+# Resynchronize the package index files
 RUN apt-get update \
 	&& DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
     apache2 \
@@ -11,6 +9,7 @@ RUN apt-get update \
     cmake \
     dh-autoreconf \
     dpatch \
+		git \
     libapache2-mod-php \
     libarchive-zip-perl \
     libavcodec-dev \
@@ -73,6 +72,8 @@ ADD web /ZoneMinder/web/
 ADD cmakecacheimport.sh CMakeLists.txt version zm.conf.in zmconfgen.pl.in zmlinkcontent.sh.in zoneminder-config.cmake /ZoneMinder/
 ADD conf.d /ZoneMinder/conf.d
 
+RUN git clone --depth 1 -b 3.0 https://github.com/FriendsOfCake/crud.git /ZoneMinder/web/api/app/Plugin/Crud
+
 # Change into the ZoneMinder directory
 WORKDIR /ZoneMinder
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -73,6 +72,8 @@ ADD web /ZoneMinder/web/
 ADD cmakecacheimport.sh CMakeLists.txt version zm.conf.in zmconfgen.pl.in zmlinkcontent.sh.in zoneminder-config.cmake /ZoneMinder/
 ADD conf.d /ZoneMinder/conf.d
 
+RUN git clone --depth 1 -b 3.0 https://github.com/FriendsOfCake/crud.git /ZoneMinder/web/api/app/Plugin/Crud
```
knight-of-ni, line 75: Rather than git clone crud separately, do this instead:
```git submodule update --init --recursive```

This ensures you end up with the version of crud the project is intended to be used with (currently 3.0.10) and will pull down any other submodules we might add in the future.

Comment (rileyschuit) [reply Dockerfile:75]: This doesn't look like an option when you are building the image.  Always results in an error.  I know this is noted as a requirement if you build the package but in an automated build for docker hub, it's not an option.

An alternative is to make travis ci build and push out to docker hub.  

It's up to you guys, if you leave it, it can be an automated build on docker hub.  Or we can make travis ci push it out.

Comment (knight-of-ni) [reply Dockerfile:75]: I'd rather not hardcode this. If `git clone` works here then `git submodule update` should work just as well. Try calling this command a couple lines down, after WORKDIR is set to ZoneMinder. Alternatively, you can specify the path directly in the command: `git submodule update --init --recursive /ZoneMinder`. If you get an error, then let us know what it is.

File: Dockerfile [hunk 2]
```diff
@@ -106,6 +107,46 @@ VOLUME /var/lib/zoneminder/images /var/lib/zoneminder/events /var/lib/mysql /var
...
+#    -v /disk/zoneminder/images:/var/lib/zoneminder/images \
+#    -v /disk/zoneminder/mysql:/var/lib/mysql \
+#    -v /disk/zoneminder/logs:/var/log/zm \
+#    --name zoneminder \
+#    zoneminder/zoneminder
+
+# Run example if you have a seperate db:
+# docker run -d -t -p 1080:80 \
+#     -e PHP_TIMEZONE='America/Los_Angeles' \
+#     -e ZM_DB_NAME='zmuser' \
```
CodyWoolaver, line 144: Don't mind me, but this should be `ZM_DB_USER`

Thanks for putting this together.

Dockerfile
rileyschuit: This is what I see when I attempt to add the .git directory and the .gitmodules file:
```
Step 14/27 : ADD .git /ZoneMinder/.git
 ---> 06223b78c962
Step 15/27 : ADD .gitmodules /ZoneMinder/.gitmodules
 ---> 184287c58e62
Step 16/27 : RUN git submodule update --init --recursive /ZoneMinder.
 ---> Running in e6073273071e
fatal: Not a git repository (or any of the parent directories): .git
The command '/bin/sh -c git submodule update --init --recursive /ZoneMinder.' returned a non-zero code: 128
```

Hopefully just a simple thing I'm doing wrong.

Dockerfile
rileyschuit: Thanks!  I'll make sure to get that in there.  It also points out that my env var for timezones (in the example) is incorrect as well.

CodyWoolaver: Actually looking at this... and how those variables are used... Doesnt this only work when building the dockerfile? These variables get used when building around line 115... Does the CMD command get rerun so we are able to update the /etc/zm.conf? It seems that we would need to use some kind of other startup script that would modify that file on each startup based on env variables... Unless I have just completely overlooked something.

Dockerfile
knight-of-ni: Sorry for the delay. I did some looking into and it seems you can git checkout the project and get the submodules all with one command:
`git clone --recursive https://github.com/ZoneMinder/ZoneMinder`

Seems to work for me. I'll test it in docker...

rileyschuit: Nice, testing now.  Looks to work well.  I'll get it in there.

knight-of-ni: Yeah, alright. It's building on my end. 

I ended up removing the block of ADD's, which copied the local code into the container. Git clone does that for us now.

Further down, you may want to change:
`ADD utils/docker/start.sh /tmp/start.sh`
to
`RUN cp -v utils/docker/start.sh /tmp/start.sh`

I think that is the right syntax. Not sure why we are putting start.sh into /tmp. Maybe that's for speed since that folder is likely in ram.

Then near the end, we no longer need this:
`ADD utils/docker /ZoneMinder/utils/docker/`

This folder will be there already thanks to the git clone.

This of course makes the whole container portable... you now only need the just Dockerfile and not the rest of the zm repository.

knight-of-ni: Here is a gist showing the whole thing after changes:
https://gist.github.com/knnniggett/6a83fbf30c82235bdc63de59214a8ddf

knnniggett:
I've merged the changes discussed here as part of the initial commit to our new dockerfiles repo:
https://github.com/ZoneMinder/zmdockerfiles

@rileyschuit I've extended the same offer to Markos. We would be willing to give you commit access to the new repo if you are willing to maintain the dockerfile. Let me know.

PR #2025, opened 2017-12-18
Explain where the frontend is found

Pete-C2:


knnniggett:
Unfortunately, it is not that simple.
Not all zoneminder installations can be viewed from the url `https://localhost/zm`.

Something like this would be more appropriate:
```
Many, but not all, ZoneMinder installations can be viewed from the url https://localhost/zm. Refer to the documentation included with your particular Linux distro to learn what the proper url should be.
```

knnniggett:
I'm going to close this out. Feel free to submit a new pr using the verbiage mentioned in my last post.

PR #2037, opened 2018-01-18
Update packpack.rst

kb5won:
added path to git clone command.  (I'm following install guide for first time and that step didn't work without the path)


knnniggett:
Good catch. That documentation is bleeding edge new, and it is good hear someone is attempting to use it.

PR #2039, opened 2018-01-20
add step that fixes API functionality

kb5won:
There was a critical step in https://wiki.zoneminder.com/Ubuntu_Server_16.04_64-bit_with_Zoneminder_1.29.0_the_easy_way that was not in this document.

### DIFF
--- a/docs/installationguide/ubuntu.rst
+++ b/docs/installationguide/ubuntu.rst
@@ -123,6 +123,26 @@ Set /etc/zm/zm.conf to root:www-data 740 and www-data access to content
 	a2enmod cgi
         a2enmod rewrite
 
+**Step 7B.** Fix permissions so API access works
+
+Edit zoneminder.conf
+
+nano /etc/apache2/conf-available/zoneminder.conf
+Add a - before Indexes and a + before FollowSymLinks
+
+<Directory /usr/share/zoneminder/www>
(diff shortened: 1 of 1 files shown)


knnniggett:
The documentation in our readthedocs site is correct as it applies to the latest stable release, ZoneMinder 1.30.4. 1.29.0 is obsolete, and the only easily available installation documentation for that is the version is the wiki link you referenced. 

It is certainly possible one of our older packages had the issue you described. It does sound familiar, but the apache config file we use for our current build is shown here:
https://github.com/ZoneMinder/zoneminder/blob/master/distros/ubuntu1604/conf/apache2/zoneminder.conf

As you can see, it already has the change in it which you described.

kb5won:
sorry about that!  I guess I made the mistake of reading the part in step 2 "oneMinder 1.29.0 is now part of the current standard Ubuntu repository" as advice that would be OK to install from the Ubuntu repository so I did that.  Might help others to put a note on that line that warns 1.29 is obsolete.

knnniggett:
Oh I see. ZoneMinder is community based and unfortunately that can lead to inconsistencies like this.
I'll update the reference to zm 1.29 on our readthedocs page.

knnniggett:
see commit 8dc6910

PR #2041, opened 2018-02-06
Added control script for Reolink cameras

flashoftheblades:
Added control script that works for Reolink RLC-423, RLC-420, and RLC-411 cameras. The script uses the Netcat ONVIF script as its base and adds in a bunch of stuff to handle the ONVIF authentication piece. It may work with other Reolink cameras and other ONVIF cameras that require authentication, but the ones listed are the ones I was able to test it against.


knnniggett:
Thank you for taking the time to add these.

PR #2051, opened 2018-03-06
Packaging cleanup for Ubuntu 18.04

MTres19:
I made rather a lot of changes here so I'm not yet trying to change `do_debian_package.sh`. I have a few questions still:
 - Is it OK that I removed SysV/LSB init support? I was noticing that despite not being part of the `dh --with` command in `debian/rules`, it still got included in the package and was running update-rc.d even on a systemd system.
 - Are my changes to `CMakeLists.txt` correct?
 - Should the `override_dh_systemd_enable` and `override_dh_systemd_install` targets be removed since Debian Policy recommends services to be enabled by default on install?

Hope this is helpful,
Matthew


SteveGilvarry:
@connortechnology new distro folder or merge into 1604 to cater for the differences?

knnniggett:
I certainly appreciate all the work that has been done here, but I'm not sure we need an additional set of readme instructions. We've already got 2, and I think that is one too many. Perhaps a better thing to do is link the wiki since that is currently the go-to place for debian distros.

The following applies to all our debian packaging scripts.
ZoneMinder no longer requires the events & images symlinks under the webroot and the scripting that sets these up should be removed.

connortechnology:
I'm not sure any of this is necessary. ZM builds fine on 18.04 with the current contents of 1604.
The CMakeLists.txt changes are not correct, why did you feel the need to make them?
The SysV stuff should be good to have.  I'd really like that cleaned up, but that stuff is also applicable back to 1604.

knnniggett:
Yeah, I too am leaning towards reusing the 1604 folder rather than create a new 1804 folder. It's just less maintenance if we can minimize the number of these.

We do, however, need to update the php5 dependencies to include the php7 counterparts whichever path we decide to take.

MTres19:
@knnniggett I didn't write the readme; I only copied it from the ubuntu1604 folder. I'll try to remove those symlinks later. Perhaps the php5 deps could just be removed since there are generic virtual packages providing the latest versions?

@connortechnology Have you tried recently? The libx264 dependency change is required at least (probably libx254-152 wasn't in the repos when you last updated your launchpad packages). I admit that I do not know much about CMake, but [this StackOverflow](https://stackoverflow.com/questions/18808733/how-to-use-check-include-files-macro-in-cmake) question seems to indicate that it should only be used if find_file succeeds. I made the changes because it caused CMake to throw a fatal error which bubbled up to debuild. The reason I started digging into this was because the do_debian_package.sh script did *not* work due to this error.

knnniggett:
Ah, fair enough. For some reason I thought it the debian readme was just a short paragaph.

MTres19:
@knnniggett Actually it seems that the ubuntu1604 scripts don't install the symlinks anymore anyway. If you want I can add a commit to remove those symlinks if they exist though.

knnniggett:
If you are willing, we should remove the symlink verbiage from README.Debian and then look at the abort logic in zoneminder.preinst that checks for non standard symlinks. We don't need that anymore.

If you would have asked me yesterday if I had done this already, I would have sworn I did. Sometimes I think I am losing my mind, having thought I've finished things I haven't.

MTres19:
@knnniggett Do you also want the apache-related symlink handling removed?

knnniggett:
@MTres19 No, I think we need to leave that. I believe symlink checking for the apache config is still valid.

Going to run this through our buildsystem against Ubuntu 1604 today to double check we don't create any new issues merging this into a single set of packaging scripts.

knnniggett:
Ubuntu 1604 packages build fine with the new packaging scripts:
```
------------------------------------------------------------------
Debian packages are ready
-------------------------------------------------------------------
15084 /build/zoneminder-dbg_1.31.1~20180309.381-xenial_amd64.deb
 5288 /build/zoneminder-doc_1.31.1~20180309.381-xenial_all.deb
   12 /build/zoneminder_1.31.1~20180309.381-xenial.debian.tar.xz
    4 /build/zoneminder_1.31.1~20180309.381-xenial.dsc
 3152 /build/zoneminder_1.31.1~20180309.381-xenial_amd64.build
    4 /build/zoneminder_1.31.1~20180309.381-xenial_amd64.changes
 6184 /build/zoneminder_1.31.1~20180309.381-xenial_amd64.deb
    0 /build/zoneminder_1.31.1~20180309.381.orig.tar.xz
--
```

I want to merge these changes into the ubuntu1604 folder but I was to talk to @connortechnology first to see how he feels about that.

knnniggett:
Note that the proposed changes in this PR were integrated into the ubuntu1604 folder rather than spawn a duplicate set of build scripts.

PR #2052, opened 2018-03-06
Add D-Link DCS-3415 to database

habibapp1365:
Hi there
This is my small contribution to your great project.


knnniggett:
Thanks for your work, but I'm closing this pr out because it wasn't set up quote right. I will assist with setting things right in #2053

PR #2053, opened 2018-03-06
Add D-Link DCS-3415 Script Control

habibapp1365:
Hi there
This is my small contribution to your great project.


knnniggett:
Thank you for your contribution, but we need to do a couple of things before this is merge-able.

You appear to be doing this from the web-gui so I will attempt to show you how to fix this from the web portal rather than commandline:

- From your fork, click the dropdown to change to your branch called "patch-3".
- Find web/skins/classic/views/video.php and set it back to the way it was. One way to do that would be to open a second browser and then open video.php in our master branch. Copy the contents and then paste them into your copy. Save changes. This should make the conflict go away in your pr. 
- Take the line of sql from the pr I just closed and copy it: https://github.com/ZoneMinder/zoneminder/pull/2052/commits/b5778a136478dfd0629f071c75cd6f9ca3587e18
- Verify you are still in your branch called "patch-3" in *your* fork, now paste that line of code into zm_create.sql.

Any changes to your "patch-3" branch will cause them to automagically appear over here in this pull request.

knnniggett:
I think the problem is that your master branch is out of sync with the zoneminder master branch.
In any case, I'll go ahead and fix it on my end. Thanks for trying.

habibapp1365:
Actually, this is my first contribution, I'd like to add DCS3415.pm and change the zm_create.sql.in but I think I messed up.

Sorry about my English.

Issue #2059, opened 2018-03-10
alarm_cause in SharedData


pliablepixels:
As of today, SharedData has a field called 'trigger_cause' that holds a text value when you create an external trigger, or 'ForcedWeb' if you force it from the web. There is no equivalent field for 'alarm cause' when its not triggered (i.e. regular alarm). The notes are directly written to the DB. Would it be useful to either overload 'trigger_cause' with 'alarm_cause' or create a new field? I'm not sure what the latter involves as we are changing the size of SharedData and not sure if other areas depend on this. If its the former, I can do a PR, but I do agree it would be breaking the core idea of 'trigger_'. That being said, I don't see a reason why there should only be a trigger_cause for triggered events and not for regular alarms.

Motivation: zmeventserver (used by zmninja) relies on shared memory to get alarm details and this would be useful as a push notification. 

(ref issue: https://github.com/pliablepixels/zmNinja/issues/593)

connortechnology:
Sounds like a good idea to me.  I don't see any technical reasons not to.

pliablepixels:
@connortechnology do you mean adding a new field or overloading trigger_cause? `SharedData` has [comments](https://github.com/ZoneMinder/zoneminder/blob/dd70eb3e5f80edbf87280e5ccaf7680b26b8e021/src/zm_monitor.h#L98) about expecting an exact size. Is adding a new field a good idea?

connortechnology:
Adding new fields isn't a bit deal.  It just changes the size of the shared data.  Which isn't a big deal.  It means that you can't mix older binaries with new ones... which you already can't because we don't allow it.  That's why we die when the version doesn't match or # of config items doesn't match what is expected.  

I would add a new field.  The code that manages the notes is a little messy and may get messier when using a simple char field in shared mem.  So be it.

PR #2060, opened 2018-03-11
2059 alarm cause

pliablepixels:
https://github.com/ZoneMinder/zoneminder/issues/2059 for review
@connortechnology


pliablepixels:
Another follow up, [Memory.pm throws](https://github.com/ZoneMinder/zoneminder/blob/master/scripts/ZoneMinder/lib/ZoneMinder/Memory.pm.in#L355) a `Fatal` error if we do a `zmMemRead` for an attribute that doesn't exist.Is it ok if we change this to `Error`? If you are ok with this, I can add it to the PR.

pliablepixels:
the PR is now complete from my side
@knnniggett @connortechnology for review. I've done local testing on ubuntu 16.

Issue #1813, opened 2017-03-16
Problem with API - /host/XXX reports 'not authenticated' but all other APIs work after authentication


pliablepixels:
There seems to be an increasing number of people who are reporting issues that a part of the APIs are failing when authentication is enabled. Specifically, everything under the `/host/` route fails with a 401.

See [here](https://github.com/ZoneMinder/ZoneMinder/issues/1813#issuecomment-287156063) for help needed to debug this issue.

**Occurrence**
Seems to be only on some systems. Don't have a pattern yet

**Steps to reproduce**

a) login to ZM (OPT_AUTH is on) using a browser, Open a new tab
b) Go to  `http://server/zm/api` - shows green
c) Go to `http://server/zm/api/monitors.json,` `/events.json` etc all checkout fine - they return valid JSON data
d) Go to `http://server/zm/api/host/getVersion.json` and pretty much any `/host/XXXX.json` function returns `401 Not authenticated `

I don't have a sense on what exactly is causing this - whether its an OS version problem, a Cake PHP problem, a PHP problem, a webserver config problem or something else.

**Why its failing - the cause (but not root cause)**
What we did figure out is the reason this happens is because the HostController.php file gets an empty Session Object, while all other Controllers get a valid session object. The API auth layer relies on the Session object to validate if you have logged in.

Why is the Session vars empty just for this one controller, I haven't a clue. I am setting up this thread so that I can invite folks who are facing this problem to report their OS/cake/PHP versions and see if we can debug something. I'll soon be tagging folks to contribute.

**Temporary Workaround (has security issues, see below)**
The workaround, for now is this: disable Auth checking on HostController. This will however expose the following details to any user who can access your server (as there is no auth in this controller). Note that you can keep Basic Auth enabled - it will continue to work and will honor basic auth. The problem is only with ZM Auth, for those who are affected.
a) API version
b) System load
c) Your server Time Zone (only applicable for ZM 1.30.2 and above)
d) disk percent (used/free) on a per monitor basis (this API also slows your system down significantly in addition to exposing your list of monitor names. You can comment out this function - see note below)

**Note** - if you are facing this problem, you can [completely comment out ](https://github.com/ZoneMinder/ZoneMinder/blob/master/web/api/app/Controller/HostController.php#L37-L100)the getDiskPercent API in HostController.php to remove open access. zmNinja does not use this API and neither does ZoneMinder - so there is no functionality loss

[...]

PR #2062, opened 2018-03-13
Fix intermittent API authentication problem caused by whitespace in CakePHP file

mcdamo:
Remove whitespace causing problems with session authentication. This fixes issue #1813


connortechnology:
I don't see how this could fix anything, but I'm all for removing useless whitespace.

Issue #2061, opened 2018-03-13
Video looks normal, but saved videos are stretced / wide


joseph85750:
I'm running ZoneMinder Version 1.30.4

I have 2 Amcrest cameras, and an older Foscam all exhibiting the same issue:  When video is created, the video x/y dimensions seem 'swapped' from the original JPG images--  giving a stretched wide video.  All 3 cameras are configured with their images rotated 90 degrees.
If I view the saved images on the server, they look normal.  But if I try to create a video via the web interface, the width and height seem to get swapped, as if the video creation is trying to use the dimensions of the original image orientation.

More details:
   https://forums.zoneminder.com/viewtopic.php?f=36&t=27039

Linux Distro / Version info:

NAME="Ubuntu"
VERSION="16.04.4 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.4 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial

More info needed?  I can provide sample images along with the resulting video if needed.

knnniggett:
Thanks for reporting this. @connortechnology has merged a fix to the master branch, and I'd expect this to show up in our development packages in a few days.

PR #2063, opened 2018-03-14
Use Event.Width and Event.Height instead of Monitor.Width and Monitor…

connortechnology:
….Height when generating video of an event.

Event.Width and Event.Height are rotated already, and furthermore, the monitor resolution may have been changed since the event happened.


knnniggett:
fixes #2061

PR #2065, opened 2018-03-16
New Monitor Type - Website

knnniggett:
This PR adds the ability to add an arbitrary website as a non-recordable, *fully interactive*, monitor in ZoneMinder. This is entirely a front-end change. By design, the C source backend knows nothing about monitors with a type of WebSite.

The intended use case for this feature is a simple way to integrate any third party web application into a tile on the montage screen. Websites can be viewed from the watch view, but that is less useful.

Features:
- Some websites can add an X-Frame option in the header. ZoneMinder will detect this and present a warning in the log. The end user must then install a browser plugin to ignore X-Frame header options.
- End user can manually turn off the warning about X-Frame options Options -> Web -> ZM_WEB_XFRAME_WARN
- Each website monitor can optionally have its own refresh timer to reload static content
- Resolution is arbitrary. Pick a size, big or small, that looks good on your monitor.
- since neither zmc nor zma are running, negligible resources are consumed

Caveats:
- If the website is self-signed, the end user must first manually navigate to the site and accept the certificate. If someone knows how to detect this condition in php, or better yet work around it, please let me know and I'll fix it.

Notes:
The complicated part of this PR wasn't adding the feature itself. Rather, it was turning off the normal features (streaming, ajax calls, playback controls, etc) that ZoneMinder automatically implements.

I've been testing this over the past few weeks to shake out several javascript & php errors in the logs. I'm at the point where I need some additional hands testing this out to make sure I got them all. Merging this to master should accomplish that.

@connortechnology this will likely conflict with the work you are doing in your fork. The dB stuff certainly will, and likely some of the other php changes. I had to touch a lot of files. Let me know how we can coordinate our work.

@pliablepixels Let me know if this is a simple thing to add to zmninja, or if there is something I can do to make this easier to implement.


connortechnology:
I'm curious to see what others say about this.  

I get the desire to record a website.. that's kinda neat.. but to just view another website in ZM? That seems like an odd thing to add in ZM. To me anyways. So this is the first step towards full recording, right?

I don't see anything that will mess too much with my branch too much.  I'll manage the merging.    

My only real thoughts on how this is done is I don't like the all the if type==Website added to all the scripts.  I wonder if instead we could build a datastructure and indicate to the scripts whether a given type supports recording.. but maybe that's nitpicking.

knnniggett:
Yeah, I agree about all the "type == website" stuff. That is annoying. 
A lot of that was to prevent zoneminder from starting or stopping zma/zmc that don't exist. I could instead do that check in zmdc.pl and just return success w/o doing anything. I'm open to suggestions.

The use case is really intended for always-on (overhead) monitors continuously running montage. If you've got this nice monitor always displaying information, this pr makes it easy to display additional information from other applications. a.k.a. this is a "poor-man's" integration. For me, that means showing the Moteino Gateway web portal on the montage screen. 

As I went down this path, I realized that a fully interactive web page was more beneficial (to me anyway) than recording. These two features seem mutually exclusive. Adding the ability to record would mean rendering the html into a steam, which in turn would no longer be interactive. I supposed we could try to do both... display the html website on the front-end, while rendering the page to a stream on the backend, but that might work better as yet another monitor type.  In addition, there are technical issues with the only C library I could find which rendered html -> jpeg.  Picking apart the Debian source package, I learned the intentionally excluded the library & header files because they don't think it is good enough to package. Ultimately, I decided not to pursue recording at this time.

knnniggett:
Here are some screenshots to both get an idea of how this looks in ZoneMinder, and a comparison to doing the same thing in the Avigilon thick client. The Avigilon client does not record websites either.

ZONEMINDER
screenshot from 2018-03-16 13-55-32

AVIGILON THICK CLIENT
avigilon_with_websites

Notice how the thick client embeds an entire browser into the view. I could look into that if we can think of a reason why that might be better somehow.

connortechnology:
I had a thought about record... How about just save the html to disk?

On March 16, 2018 3:53:53 PM EDT, Andrew Bauer <notifications@github.com> wrote:
>Here are some screenshots to both get an idea of how this looks in
>ZoneMinder, and a comparison to doing the same thing in the Avigilon
>thick client. The Avigilon client does not record websites either.
>
>ZONEMINDER
>screenshot from 2018-03-16
>13-55-32
>
>AVIGILON THICK CLIENT
>avigilon_with_websites
>
>Notice how the thick client embeds an entire browser into the view. I
>could look into that if we can think of a reason why that might be
>better somehow.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/ZoneMinder/zoneminder/pull/2065#issuecomment-373827016

-- 

pliablepixels:
That won’t help if the html is sourcing let ve data or live feed. It will change.

knnniggett:
Here is some additional data on rendering sites to an image.

Since the wkhtmltoimage C library had availability issues, I also looked into using the wkhtmltoimage binary to render a site to jpeg file or buffer, and then pipe that to the File camera type (a.k.a. decodeJpeg function).  Relatively simple sites would work okay, but sites with lots of javascript would cause the rendering process to take seconds. 

Take the case of a site with the very long or never ending scroll-down page... when you try to render that,  wkhtmltoimage does its best to render every last bit of content, but at some point it has to decide enough is enough. We could probably tweak the settings to either timeout faster or just crop the page to a certain size, but every site would be different.... so that means creating more form fields and hope the end user can figure out how to make his/her favorite website work.

Try it, so you can get an idea of how this tool works. The name of the package to install on your system is called *wkhtmltopdf* and then call the following binary like so:
`wkhtmltoimage www.yahoo.com yahoo.jpeg`

Rendering times can vary. This is one of the faster attempts:
```
$ time wkhtmltoimage -q www.yahoo.com yahoo.jpeg

real	0m2.271s
user	0m1.620s
sys	0m0.181s
```

Now imagine trying to do this as fast as possible to try and create a semi-fluid "stream".

pliablepixels:
Oxford dictionary 2019:

“Solitarus Syndrome”, noun:
Widely attributed  to Andrew Bauer. The sudden decline in theft monitoring productivity of CCTV surveillance operators in 2018. It turns out Zoneminder allowed them to play the very popular card game, solitaire while they feigned staring at a montage of cameras.

knnniggett:
Ahh, right. As expected this conflicted with @connortechnology 's work. 
So I need to:
- manually fix all the conflcits
- change this pr to bump zm to 1.31.42
- write the start, stop, and restart methods to the montior object

I've got an emergency to deal with at work tomorrow, but I promise to deal with this this week.

knnniggett:
The merge conflicts in this pr are a nightmare to resolve. I need to start over with a fresh master branch. It will be simpler.

connortechnology:
I can do it if you want.

knnniggett:
In the interest of time, I'm going to punt on the request to objectify the monitor start/stop/restart requests. 
I still want to do it, but there is more that can be done than what we've talked about here. Consequently, I think it makes more sense to create a separate PR due to the scope creep.

web/includes/functions.php
```diff
@@ -1966,253 +1987,4 @@ function jsonDecode( $value ) {
...
-  $user     = isset($parsed_url['user']) ? $parsed_url['user'] : ''; 
-  $pass     = isset($parsed_url['pass']) ? ':' . $parsed_url['pass']  : ''; 
-  $pass     = ($user || $pass) ? "$pass@" : ''; 
-  $path     = isset($parsed_url['path']) ? $parsed_url['path'] : ''; 
-  $query    = isset($parsed_url['query']) ? '?' . $parsed_url['query'] : ''; 
-  $fragment = isset($parsed_url['fragment']) ? '#' . $parsed_url['fragment'] : ''; 
-  return "$scheme$user$pass$host$port$path$query$fragment"; 
-}
-?>
+      if ( $value[$i] =
```
connortechnology: I'm concerned by this loss of code here, is it just github screwing up?

knight-of-ni: fixed with commit ff6b561

web/skins/classic/views/montage.php
```diff
@@ -204,7 +204,11 @@
     }
   }
 
-              echo getStreamHTML( $monitor, $monitor_options );
+              if ( $monitor->Type() == "WebSite" ) {
+                  echo getWebSiteUrl( 'liveStream'.$monitor->Id(), $monitor->Path(), reScale( $monitor->Width(), $scale ), reScale( $monitor->Height(), $scale ), $monitor->Name() );
```
knight-of-ni, line 208: @connortechnology This is the line of code I am most concerned with. It almost certainly needs to be fixed. It is based on the previous way of doing things in montage. This is what I need to test.

connortechnology: Well it looks like the old line needs to be the else condition...

connortechnology: oh never mind, there it is below, looks fine to me.

knnniggett:
@connortechnology If you are willing to merge this in the interest of forward progress, then so am I.

Issue #2064, opened 2018-03-16
Postinstall script of latets update fails


jzielke-nli:
After updating zoneminder the postinstall script runs insto an error

**Relevant log lines:**
```
Setting up zoneminder (1.31.1~20180315113240-xenial) ...
/var/lib/dpkg/info/zoneminder.postinst: 50: /etc/zm/zm.conf: tk#S: not found
dpkg: error processing package zoneminder (--configure):
 subprocess installed post-installation script returned error exit status 127
Errors were encountered while processing:
 zoneminder
E: Sub-process /usr/bin/dpkg returned an error code (1)

```

MTres19:
Hmmm... seeing as I wrote the updates to the postinst script, I feel a bit responsible.
However, the error makes it look like you have the line "tk#S" somewhere in your zm.conf. Is that the case?

connortechnology:
There is an eerie if you don't have mariadb installed that causes the script to die. More testing and improvement needed

MTres19:
Ah, got it. :/ I let a bashism sneak into the script. In bash, if a command
in a command substitution fails it won't kill the script even in set -e
mode. Switching from #!/bin/sh to #!/bin/bash would work, though it's
non-ideal and Ubuntu probably wouldn't like it. One way to fix it would be
to use systemctl status mariadb.service >/dev/null 2>&1 ||
MARIADB_STATUS=$? instead. Not sure if there are other places with this
problem though.

MTres19:
A (relatively) quick look into this seems to indicate that there are two problems: the error when there's no mariadb and the error that the OP mentioned. My error seems to be from the aforementioned bashism, but looking at the systemctl manpage there's a better way to do this anyway. I'll try to get this fixed and submit a PR.

MTres19:
@connortechnology I tested my changes that are in #2066 and the installation seems to work fine on my end. Sorry for causing this bug.

jzielke-nli:
Never mind, we all make mistakes :-) Thanks for fixing it so fast 👍

jzielke-nli:
I actually digged into my config file again and noticed, that the post-install script had problems with a `#`character in my passphrase, since the phrase was unquoted.

MTres19:
[...]

PR #2066, opened 2018-03-17
Fix the second part of #2064

MTres19:
This also uses `systemctl` in a better way, so that the script doesn't rely on undocumented return codes. I have verified that this postinst script is able to start mysql or mariadb if needed.

I would need feedback though on whether this fixes the original problem in #2064, which I'm not sure is exactly related due to the reference to `/etc/zm/zm.conf`.

### DIFF
--- a/distros/ubuntu1604/zoneminder.postinst
+++ b/distros/ubuntu1604/zoneminder.postinst
@@ -37,9 +37,9 @@ if [ "$1" = "configure" ]; then
           DBSERVICE="mysql.service"
         fi
         if systemctl is-failed --quiet $DBSERVICE; then
-          echo $DBSERVICE is in a failed state; it will not be started.
-          echo If you have already resolved the problem preventing $DBSERVICE from running,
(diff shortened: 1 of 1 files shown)


distros/ubuntu1604/zoneminder.postinst
```diff
@@ -30,25 +30,27 @@ if [ "$1" = "configure" ]; then
         #
         # Get mysql started if it isn't running
         #
-        $(systemctl status mysql.service >/dev/null 2>&1); MYSQL_STATUS=$?
-        $(systemctl status mariadb.service >/dev/null 2>&1); MARIADB_STATUS=$?
         
-        # 3 = inactive, 4 = nonexistant
-        if [ "$MARIADB_STATUS" != "4" ]; then
+        if [ -e "/lib/systemd/system/mariadb.service" ]; then
```
connortechnology, line 34: This looks to see if mariadb is installed, but it could be disabled.  Perhaps we should check instead in /etc/systemd/system/multi-user.target.wants/mariadb.service

PR #2082, opened 2018-04-19
Fix signalcheck

connortechnology:
Moves it from the config to a per-monitor setting, available no matter what type of monitor it is.

Since we always called SignalCheck, this will be a performance improvement for netcams.  This might also be useful to detect pink slime cameras.


knnniggett:
Where/How do we assign a value to SignalCheckPoints?

connortechnology:
In the monitor

knnniggett:
Ah, okay.

Issue #2089, opened 2018-04-29
getAuthHash API issues


pliablepixels:
@connortechnology, need to revisit how `getAuthHash` API works

1. When OPT_AUTH is enabled, it returns an auth_hash token
2. When OPT_AUTH is disabled, the API throws an error `"name": "Call to undefined function generateAuthHash()",` --> should it not return `auth_hash = ""`  or something like that? I suppose this is happening because auth.php may not be included in this case
3. How will this API react if a user has `"&user=&pass="` ? I forgot what setting of auth results in this output so I can't test it.

I plan to use this new API if ZM version > 1.31.41, so I need to ensure its as predictable as scraping the auth key from the watch page (which is what I am doing now)

pliablepixels:
Ok, I think this API should basically replicate what functions.php is doing:
```
if ( ZM_OPT_USE_AUTH ) {
  66     if ( ZM_AUTH_RELAY == 'hashed' ) {
  67       $args[] = 'auth='.generateAuthHash( ZM_AUTH_HASH_IPS );
  68     } elseif ( ZM_AUTH_RELAY == 'plain' ) {
  69       $args[] = 'user='.$_SESSION['username'];
  70       $args[] = 'pass='.$_SESSION['password'];
  71     } elseif ( ZM_AUTH_RELAY == 'none' ) {
  72       $args[] = 'user='.$_SESSION['username'];
  73     }
  74   }
```

fennec622:
Hi

if i understand now i can't use user pass in api request from watch page or snapshot ?

thanks

pliablepixels:
@fennec622  this is a new API which never existed before. Its not related to the watch page.

connortechnology:
I agree it should return '' if auth is disabled.  I will be merging a fix later today. 

I do not agree that it should return user=&pass= at least not how it is named, etc.

PR #2090, opened 2018-04-29
auth_key api for different situations

pliablepixels:
#2089 

@connortechnology  - take a look.

Seems to work for me, but I noticed that `User.Password` actually stores the MD5 encoded password, not the real password. So `&user=&pass=` pbly won't work - I suppose we need to load some other model for the password?


pliablepixels:
ok, so `&pass=` is by design, not supposed to send the password, as its never stored in raw form. I'll modify the PR to return a token `REPLACEPASSWORD` so the user knows he needs to replace it with their own. Better then not returning `&pass=` because then the API recipient needs to figure out if it was plain setting.

pliablepixels:
@connortechnology still an issue

When we use this from Curl (or zmNinja) I see this error (and then the array)

```
Notice (8): Undefined index: passwordHash [/usr/share/zoneminder/www/includes/auth.php, line 107]
Code Context

function generateAuthHash( $useRemoteAddr ) {
  if ( ZM_OPT_USE_AUTH and ZM_AUTH_RELAY == 'hashed' and isset($_SESSION['username']) and $_SESSION['passwordHash'] ) {
generateAuthHash - /usr/share/zoneminder/www/includes/auth.php, line 107
HostController::getAuthKey() - APP/Controller/HostController.php, line 43
ReflectionMethod::invokeArgs() - [internal], line ??
Controller::invokeAction() - CORE/Cake/Controller/Controller.php, line 491
AppController::invokeAction() - APP/Plugin/Crud/Lib/CrudControllerTrait.php, line 31
Dispatcher::_invoke() - CORE/Cake/Routing/Dispatcher.php, line 193
Dispatcher::dispatch() - CORE/Cake/Routing/Dispatcher.php, line 167
[main] - APP/webroot/index.php, line 107
{ "auth_key": "auth=", "append_password": 0 }
```

pliablepixels:
There are other issues with this API, that don't show up in the browser. But shows up on curl and zmNinja.

Calling this API multiple times results in other errors like "session already started, ignoring session_start()" and other stuff.

Needs more testing. Not ready for us

connortechnology:
I don't think you need to do all the Config lookups.  I think you can use the defines just like in the regular UI.  

If auth=plain, then we store the password in the session, unhashed.  If we are using auth=hashed then we store passwordHashed in the session as the hashed password.  If we switch auth methods after logging in from plain to hashed, then the session will not have the unhashed password in the session.  

auth=plain is inherently unsafe.  We should really get rid of all of it, and use actual public key cryptography to encrypt a hash.

PR #2098, opened 2018-05-06
Limit search api

pliablepixels:
EventsController has a search action this is so far unused. Its pretty powerful as a general purpose search. This PR adds a limit feature to it, with the following motivation:

1. zmNinja today uses `find ('neighbors')`  for navigation of events --> this feature is deprecated for Cake 3, I need to move away from it. This action can replace it, but for performance, I need to add a limit

2. Its backwards compatible - doesn't break anything and lets me replace 'neighbors' find with search more efficiently

3.  query params seems more flexible than passed params https://www.dereuromark.de/2013/05/04/passed-named-or-query-string-params/

### DIFF
--- a/web/api/app/Controller/EventsController.php
+++ b/web/api/app/Controller/EventsController.php
@@ -101,8 +101,8 @@ class EventsController extends AppController {
  * @return void
  */
 	public function view($id = null) {
+ 
     $this->loadModel('Config');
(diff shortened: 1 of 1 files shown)


pliablepixels:
I'll re do this - pagination is the right way it seems

PR #2101, opened 2018-05-10
Munge the config in the global configvals into the configs array 

connortechnology:
When using the api to get the config, things like PATH_ZMS are no longer returned because they don't exist in the Config table anymore.

This code munges the values from the /etc/zm/conf.d settings into the array before returning it.


knnniggett:
So I'll push the following to this PR:
```
{
  name        => 'ZM_PATH_ZMS',
  default     => '',
  description => 'Web path to zms streaming server',
  help        => q`
    The ZoneMinder streaming server is required to send streamed
    images to your browser. It will be installed into the cgi-bin
    path given at configuration time. This option determines what
    the web path to the server is rather than the local path on
    your machine. Ordinarily the streaming server runs in
    parser-header mode however if you experience problems with
    streaming you can change this to non-parsed-header (nph) mode
    by changing 'zms' to 'nph-zms'. IMPORTANT: The value for this
    variable is no longer read from the database. Instead, this
    variable should be set in a config file under the conf.d
    folder.
    `,
  type        => $types{rel_path},Graz, Austria
  category    => 'hidden',
},
```

@pliablepixels This should fix the viewByName function. 
The only question I have is, is this the only variable we want to put back in? What about the others? 
See: https://github.com/ZoneMinder/zoneminder/commit/bacf65ae9a826461b1e62f5567614448a6edd71f#diff-66f1fd548f32344e70c3b02e426e0508

connortechnology:
Having the entries in config table will mean that my munging code will need to change.  So no merging till I do that.

I think they should all come back mainly just to get the help text back and for completeness.

knnniggett:
Would it? The munging code just updates the key/value pairs with the entries from the config files, right? Isn't that what we want?

connortechnology:
Yeah, the api Config just returns an array of db rows.  So I just appended to it.  Instead we will need to loop through it and find the entry with the right name.

knnniggett:
Oh, I see.

pliablepixels:
from zmNinja's perspective, I use
`/configs/viewByName/ZM_PATH_ZMS.json` for the wizard functionality - important for me
`/configs/viewByName/ZM_EVENT_IMAGE_DIGITS.json` for direct image access, which I really don't need anymore (it was a legacy detail that I've since addressed using `view=image` as the alternate

connortechnology:
So I've updated this a bit.  Please forgive the commit log strangeness... rebase never works well for me.  The change is simple, just consider the Files changed. 

@pliablepixels is zmNinja expecting an array of config entries or a Name key'd hash of config entries?

Which do we think is better?   What I have done here is output the entire db Config table indexed by Name, with file based config values super-imposed or added.  

Comments?

connortechnology:
I've added support for viewByName

knight-of-ni:
@pliablepixels says it works for him, so merging....

PR #2106, opened 2018-05-14
fixes #2105

### LINKED ISSUE #2105: PHP Errors when setting up filters
I just upgraded ZoneMinder (from zmrepo-testing) and have found that the filters page no longer works.  Filters themselves appear to do their thing in the background, I just can't maintain them or use them through the front end.

There are two PHP errors that are happening, both appear in ssl_error_log:

1. PHP Fatal error:  Can't use method return value in write context in /usr/share/zoneminder/www/skins/classic/views/filter.php on line 346, referer: https://zoneminder...
2.  PHP Parse error:  syntax error, unexpected '(' in /usr/share/zoneminder/www/includes/Filter.php on line 23, referer: https://zoneminder...

The first error is the same as issue #1977, just in a different place.  The second error occurs after resolving the first error.

Both of these are due to the old version of PHP used by CentOS 7.  If the resolution is to drop CentOS support, I'm actually OK with that.

**ZoneMinder Version (`zmaudit.pl -v`):**

```
# zmaudit.pl -v
1.31.43
```

**Are you using a development snapshot / git checkout?  If so, what is the latest commit? (`git rev-parse HEAD`):**

```
# rpm -qi zoneminder
Name        : zoneminder
Version     : 1.31.43
Release     : 1.0.20180426gitg86b2f6a12.el7.centos
[snip]
```

**Linux Distribution and Version (`cat /etc/os-release` or `cat /etc/redhat-release`):**

```
# cat /etc/redhat-release
CentOS Linux release 7.5.1804 (Core)
```

Since this is a PHP error:
```
# php --version
PHP 5.4.16 (cli) (built: Apr 12 2018 19:02:01)
Copyright (c) 1997-2013 The PHP Group
Zend Engine v2.4.0, Copyright (c) 1998-2013 Zend Technologies
```

knnniggett:
Should work, but I don't have a running development system in front of me at the moment to verify this is indeed correct.

### DIFF
--- a/web/includes/Filter.php
+++ b/web/includes/Filter.php
@@ -20,10 +20,12 @@ public $defaults = array(
     'limit'           =>  100,
     'Query'           =>  array(),
     'sort_field'      =>  ZM_WEB_EVENT_SORT_FIELD,
-    'sort_asc'        =>  (ZM_WEB_EVENT_SORT_ORDER == 'asc' ? 'asc' : 'desc'),
+    'sort_asc'        =>  '',
 );
 
   public function __construct( $IdOrRow=NULL ) {
+    # Set the proper value for sort_asc here because we can't do it above
+    $this->defaults{'sort_asc'}  =(ZM_WEB_EVENT_SORT_ORDER == 'asc' ? 'asc' : 'desc');
     $row = NULL;
     if ( $IdOrRow ) {
       if ( is_integer( $IdOrRow ) or is_numeric( $IdOrRow ) ) {


connortechnology:
The only reason I don't like this, is that the defaults array is supposed to provide the default values, and should be query-able from elsewhere theoretically. I think instead of all this, we should just assign it the value in the Config define.

knnniggett:
Superseded by commit ead9959  .
Closing

PR #2112, opened 2018-06-01
Fast syntax fix for Option menu displaying

blackPantherOS:
This patch fix the broken option page  
Cheers


connortechnology:
Hey thanks for this, didn't know there was an issue.

Your fix is actually incorrect, as it should be </td>, but I pushed the correct fix to master.

PR #2113, opened 2018-06-01
Do not require mariadb/mysqld in systemd service file - database serv…

opoplawski:
…er may be remote

Without mariadb installed locally, on EL7 you get the very unhelpful error message:
```
Failed to start zoneminder.service: Unit not found.
```
If the database is installed locally, After=mariadb.service is sufficient to do what you want - start zoneminder after mariadb.

### DIFF
--- a/distros/redhat/nginx/zoneminder.service.in
+++ b/distros/redhat/nginx/zoneminder.service.in
@@ -4,7 +4,7 @@
 [Unit]
 Description=ZoneMinder CCTV recording and security system
 After=network.target mariadb.service nginx.service php-fpm.service fcgiwrap.service
-Requires=mariadb.service nginx.service php-fpm.service fcgiwrap.service
+Requires=nginx.service php-fpm.service fcgiwrap.service
 
 [Service]
 User=@WEB_USER@
--- a/distros/redhat/systemd/zoneminder.service.in
+++ b/distros/redhat/systemd/zoneminder.service.in
@@ -3,7 +3,7 @@
 [Unit]
 Description=ZoneMinder CCTV recording and security system
 After=network.target mariadb.service httpd.service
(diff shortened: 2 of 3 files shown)


knnniggett:
Yeah, it would certainly be nice if systemd actually spelled out which dependent unit file it could not find, but other than that, the message does indeed tell you why it couldn't start zoneminder.

This comes down to you can't make everyone happy. The zoneminder rpms are configured out of the box to run with a local database. The type of user targeted for this kind of installation is the kind of user who is not familiar with linux and may not be comfortable editing config files. This kind of user would also have difficulty understanding what is wrong with zoneminder if it were started before the database server was started (or if zoneminder did not stop automatically if the database crashed).  This is why the zoneminder service file "requires" mariadb, and this is also why this will not change.

Think of this as a once-size-fits-most kind of thing. If your use cases falls outside that category, then you'll need to do some work to get things going. 

In your case, you should make a copy of the zoneminder service file and put it under /etc/systemd/system. Then customize it to your liking. This will override the default service file that ships with the zoneminder package, and your custom serivce file will never be overwritten during a package upgrade.

opoplawski:
Well, right now when installed from the rpmfusion EL7 rpm, which does not require mariadb (as it shouldn't), one has to create a custom zoneminder service file to get it to work with a remote database.  If one wanted to use a local database, one would simply install it and start it.  Start ordering would be handled by the After clause.  And if the database wasn't started - the zoneminder logs would certainly indicate the problem clearly.  So there is an unnecessary disconnect between the rpm dependencies and the service file dependencies.

I'll bug rpmfusion about it: https://bugzilla.rpmfusion.org/show_bug.cgi?id=4924

knnniggett:
I'm also the rpmfusion package maintainer. ;-)

Issue #2110, opened 2018-05-31
Issues exporting / downloading video


frankc420:
So I've upgraded to 1.31.43 and after doing so I'm now unable to export or download a video of any events.  If I try to 'download video' as Zip, it fails instantly.  If I choose Tar, it outputs a file but it is empty.

Export continuously says 'Exporting video...', so my guess is that it is instantly failing.

I'm not sure what logs to look into to find the culprit, hoping you guys can point me in the right direction.

connortechnology:
1.31.43 is pretty old.  I have fixed several issues with Export since then. Please update to 1.34.44, or more specifically the latest commit.

frankc420:
I'm using the zmrepo-testing on CentOS 7 and I'm not seeing 1.34.44.  

Sorry for my ignorance here but what is the best method of updating on CentOS if the commit isn't part of the zmrepo-testing repository?

frankc420:
I did a git checkout of the master and the version file shows 1.31.44, is your 1.34.44 a typo?  I'm assuming so.

connortechnology:
1.34.44 was a typo.

frankc420:
@knnniggett Thanks!  I actually cloned github last night so that I could get to 1.31.44 and built out using cmake instructions.  After updating the system I was unable to start zoneminder.  I had to completely remove zoneminder and re-install using zmrepo.  I'm back to 1.31.43.  I'm not sure if this is related to the issue you are troubleshooting or not, but thought I would mention it.

knnniggett:
@frankc420 I don't know if you've upgraded to the latest development rpm. We have identified the issue preventing zoneminder from starting and will push out a fix shortly. In the meantime, the workaround to get things going again is to edit /usr/lib/systemd/system/zoneminder.service and comment out the PIDFIle statement.  It seems we have a race condition at startup and systemd is looking for the pidfile before it gets created.

knnniggett:
@frankc420 This appears to be working for me on the latest master branch. Please verify after performing a yum update to the latest package, which was just synced to the mirror servers last night.
zoneminder-1.31.44-1.232.20180613gitg4b78d85b1

frankc420:
@knnniggett Thanks!  I'll keep you updated!

knight-of-ni:
If you ever have a doubt that your machine is not picking up some rpm, you can manually verify what is in the affected repo with your webrowser.  In this case the url is:
[...]

PR #2114, opened 2018-06-04
create the pid file before doing anything else

knnniggett:
There is a race condition with systemd to create the zoneminder pid file before systemd tries to read it. There does not appear a way to completely avoid the race condition. Rather, we should create the pid file as soon as possible to keep systemd from refusing to start zoneminder.

This fixes the issue described in the latter part of #2110

### DIFF
--- a/scripts/zmdc.pl.in
+++ b/scripts/zmdc.pl.in
@@ -246,7 +246,7 @@ our $zm_terminate = 0;
 
 sub run {
 
-  # Call this first to win the race condition with systemd
+  # Call this first otherwise stdout/stderror redirects to the pidfile = bad
   if ( open(my $PID, '>', ZM_PID) ) {
     print($PID $$);
     close($PID);
--- a/scripts/zmdc.pl.in
+++ b/scripts/zmdc.pl.in
@@ -245,6 +245,16 @@ our %terminating_processes;
 our $zm_terminate = 0;
 
 sub run {
+
+  # Call this first to win the race condition with systemd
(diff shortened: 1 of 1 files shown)


knnniggett:
I have tested this on only one machine so far. We should test this further.

knnniggett:
Unfortunately the second machine I am testing this on still cannot start zoneminder so it looks like this pr is insufficient.

knnniggett:
This PR is likely not needed anymore. After further troubleshooting, we believe we fixed the problem starting up ZoneMinder via commit a25f5f1 . I'll leave this pr open for a few more days just in case.

knnniggett:
The alternative fix has been tested working on two machines now, so I'm going to close this pr out as it is no longer needed.

knnniggett:
Reopening.
After new data was collected, it appears we need this PR afterall, just for a slightly different reason.

It was observed that stderror was redirected to the pidfile, which explains why systemd was complaining the pidfile was invalid. It wasn't race condition. The pidfile was indeed created in time but it had a bunch of extra junk in the file that should not have been there e.g. 
```
5482DBD::mysql::st execute failed: Lost connection to MySQL server during query at /usr/share/perl5/vendor_perl/ZoneMinder/Logger.pm line 554, <CLIENT> line 1.
Use of uninitialized value in concatenation (.) or string at /usr/share/perl5/vendor_perl/ZoneMinder/Logger.pm line 564, <CLIENT> line 1.```

PR #2122, opened 2018-06-11
Fix startup db connection

connortechnology:


knnniggett:
I'm ok with closing this if you are. Pretty sure we fixed the startup bug.

PR #2130, opened 2018-06-17
Grant access to the new cache directory in Apache on Ubuntu

mnoorenberghe:
This fixes a regression from #2083 and aligns with generate_apache_config.pl and the changes for Debian


mnoorenberghe:
This issue was reported many times on the forums e.g. https://forums.zoneminder.com/viewtopic.php?f=37&t=27200

PR #2131, opened 2018-06-19
Change langfile's charset to UTF-8 as in #583

cpasqualini:
Hi!
As i have not seen any changes in #583 ande received a Stale advice from github, I spent a little bit of time on this and hope this will help this great project.

What I have done, in the directory web/lang is to run (in bash):

```
W=$(mktemp -d); for F in $(file *|grep ISO-8859|awk '{gsub(":","",$0);print $1}'); do
    cp -v $F $W/$F; 
    iconv -f ISO8859-15 -t UTF-8 < $W/$F > $F; 
done
```

This process had left a few files without change them to UTF-8:

```
web/lang$ file * | grep -v UTF-8
en_gb.php:       PHP script, ASCII text, with very long lines
en_us.php:       PHP script, ASCII text
it_it.php:       PHP script, ASCII text, with very long lines
updateLangs.php: PHP script, ASCII text
```

Let me know if these ASCII files needs to be converted to UTF-8 too

thanks


knight-of-ni:
Awesome. Thanks for your help.

I know very little about this so googling seems to indicate we don't need to be concerned with those remaining warnings. @connortechnology what do you think?

connortechnology:
I think straight ASCII is ok. it is a subset of UTF-8 so without any specific utf-8 characters in it, I don't think there is any way to tell the difference.  I don't really care so much about long lines either.  It is good practice in actual code, but I'm less fussy in a language definition file.

knight-of-ni:
Thanks for thee feedback, merging....

PR #2132, opened 2018-06-20
Allow selection of different source column filtering on the console

knight-of-ni:
This PR adds a new variable WEB_FILTER_SOURCE to Options -> Web.

This option filters the information shown on the web console source column in different ways.

It presents three levels of filtering to choose from:
- None
- NoCredentials
- Hostname

Selecting None does no filtering as you might expect. Selecting NoCredentials filters in the same way that the master branch filters right now, and selecting Hostname filters the same way the previous 1.30.4 release does its filtering.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
@@ -2964,6 +2964,37 @@ our @options = (
     category    => 'web',
   },
   {
+    name        => 'ZM_WEB_FILTER_SOURCE',
+    default     => 'Hostname',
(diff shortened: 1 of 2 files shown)


connortechnology:
I'm cool to merge this, but I was thinking of going another way, like adding a specific SourceName field to the Monitor that would default to whatever we do now, but could be edited to say anything that the user wants. Some cameras you might want path and query info, and others you might not.  So I kinda think it should be per-monitor, not a global config option.

knight-of-ni:
My goal was to keep things as simple as I could so we don't end up pushing our next release further back.  
How about we merge this pr as it is now and then expand it further for the release after?

PR #2135, opened 2018-06-25
Rework generateAuthHash to take a force parameter so that it can be u…

connortechnology:
…sed to generate auth hashes for zmu


knight-of-ni:
Let's give it a shot.

Issue #2144, opened 2018-07-06
Broken regex to match `arp -a` output / monitor probe detects no network cameras


onlyjob:
[Regex parsing output of the `arp -a` command](https://github.com/ZoneMinder/zoneminder/blob/master/web/skins/classic/views/monitorprobe.php#L324) is broken:

~~~~
$ /usr/sbin/arp -a | perl -nE 'say "$1, $2" if m/^.*([\d.]+).*([0-9a-f:]+).*/'
4, 0
7, 0
9, 0
3, 0
4, 0
~~~~

The following regex fixes the problem:

    /usr/sbin/arp -a | perl -nE 'say "$1, $2" if m/(\d+\.\d+\.\d+\.\d+).*(([0-9a-f]{2}:){5})/'

Note that in Perl correct/proper way to match IP and MAC would be by using
the [Regexp::Common](https://metacpan.org/pod/Regexp::Common::net) library:

    /usr/sbin/arp -a | perl -MRegexp::Common -nE 'say "$1, $2" if m{ ($RE{net}{IPv4}) .* ($RE{net}{MAC}) }x'

I do not know the equivalent approach for PHP.

Here are regexes used by _Regexp::Common_ to find IPv4 and MAC values:

    $ perl -MRegexp::Common -E 'say $RE{net}{MAC}'
    (?:(?:[0-9a-fA-F]{1,2}):(?:[0-9a-fA-F]{1,2}):(?:[0-9a-fA-F]{1,2}):(?:[0-9a-fA-F]{1,2}):(?:[0-9a-fA-F]{1,2}):(?:[0-9a-fA-F]{1,2}))

    $ perl -MRegexp::Common -E 'say $RE{net}{IPv4}'
    (?:(?:25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{1,2})[.](?:25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{1,2})[.](?:25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{1,2})[.](?:25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{1,2}))

onlyjob:
Worth noticing that original regex is equally broken for the output of `ip neigh`...

PR #2145, opened 2018-07-06
fix regex to match IP and first numbers of MAC in the `arp -a` output (Closes: #2144)

onlyjob:


knight-of-ni:
In order to satisfy the requests of others, we have made it possible for the end user to substitue their own arp command for the modprobe feature, particularly the ip neigh command.

Can you test this regex against the `ip neigh` command as well as the default `arp -a`?
To do that, set ZM_PATH_ARP to `/path/to/ip neigh`.

onlyjob:
It is working with `ip neigh` of course as new regex is crafted to match _only_ IPv4 and MAC addresses, as explained in #2144.

    ip neigh | perl -nE 'say "$1, $2" if m/(\d+\.\d+\.\d+\.\d+).*(([0-9a-f]{2}:){5})/'

onlyjob:
Thanks. :)

PR #2151, opened 2018-07-07
JSON::Any --> JSON::MaybeXS (Closes: #770).

onlyjob:
Straightforward change as per https://metacpan.org/pod/JSON::MaybeXS#CONVERTING-FROM-JSON::Any

It should work but I'm not sure how to test...

Please review.


connortechnology:
I don't think Json::MaybeXS is available for older distributions so may have to package it as well.

PR #2154, opened 2018-07-11
Removed "chosen" web site; "prism.js" is not distributable.

onlyjob:
No need to distribute sub-component's web site.


onlyjob:
Thanks.

PR #2155, opened 2018-07-11
Arp tool

### LINKED ISSUE #2146: Excessive warnings about failing attempts to find `arp` utility
https://github.com/ZoneMinder/zoneminder/blob/master/web/skins/classic/views/monitorprobe.php#L300

I get this warning all the time: `Unable to determine path for arp command, type -p arp returned '127'`

Because location of `arp` command can not be found like this:

https://github.com/ZoneMinder/zoneminder/blob/master/web/skins/classic/views/monitorprobe.php#L298

There are many problems with that approach.

~~~~
$ sh -c 'type -p arp'; echo $?
-p: not found
arp: not found
127
~~~~

That's because default non-interactive shell on Debian is _dash_ which have no `-p` option.

Let's try again:

~~~~
$ sh -c 'type arp'; echo $?
arp: not found
127

$ bash -c 'type -p arp'; echo $?
1
~~~~

Neither _dash_ nor _bash_ could find `arp` because it is not in user's PATH.
On Debian, `arp` is in `/usr/sbin/arp` (hence only `sudo sh -c 'type arp'` could find it).

But back to the original problem: there is no point to warn about failing attempts to locate `arp` utility if last fallback to hard-coded path works.

I suggest to remove all the warnings and replace them with only one, just [before the following line](https://github.com/ZoneMinder/zoneminder/blob/master/web/skins/classic/views/monitorprobe.php#L317) to warn only if `arp` hasn't been found after all attempts.

_Only one warning is enough._

Current situation is misleading because false warning led me to believe that monitor probe is failing because of missing `arp` utility while it was actually entirely unrelated problem: #2144.

knight-of-ni:
Fixes #2146 

Probe works w/o throwing errors on my test vm, but I don't have any hardware the probe will detect.


web/skins/classic/views/monitorprobe.php
```diff
@@ -206,7 +207,7 @@ function probeVivotek($ip) {
       'Width'    => 352,
       'Height'   => 240,
     ),
-  );
+  );t
```
connortechnology, line 210: This t should not be here. I will merge and just fix it though.

connortechnology:
I think you converted a lot of spaces to tabs.  I will just fix it though.

knight-of-ni:
I converted a lot of tabs to spaces, but only in the areas I worked in. In hindsight, I should have done the whole file.

Issue #2156, opened 2018-07-12
Add login/logout via API


pliablepixels:
This PR will build on top of @connortechnology 's session fixes via API and introduce login/logout APIs.

~~I plan to keep the AppController user=&pass= intact - doing it via any API is also a convenient option~~(now removed)

~~As a future note, we should investigate what it takes to make this a post and move user=&pass= to the body as data to avoid them showing up on logs.~~(done)

pliablepixels:
I can confirm the same problem with `login.php?user=admin&pass=whatever` - does not work with curl, but works with postman and zmN

```
curl -X POST \
  'https:/myserver/zm/api/host/login.json?user=admin&pass=xxxx' \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/x-www-form-urlencoded'
```

connortechnology:
Um, this isn't how you do post with curl.  When you stick the data in the url like that it is still a get.  There is another parameter -d for passing the post data.

pliablepixels:
Not sure what you are referring to. I am not putting data in the url for post.

connortechnology:
uh I don't know where it has gone but I had an email where you posted a curl command line that didn't work and you had ?user=&pass= in the url  passed to curl... I swear...

pliablepixels:
I posted two examples in the PR. One with POST and -d (new way) and one with GET and url params (your original  commit). Both exhibited the same curl issue.  It was posted here https://github.com/ZoneMinder/zoneminder/pull/2157

PR #2157, opened 2018-07-12
2156 api login

pliablepixels:
Implements #2156

Summary:
a) Adds a login function that also returns version, credentials in one go
b) Adds a logout function that removes the session
c) Fixes auth.php when user puts in an invalid username or password  - response is corrupted with PHP error notices 

Note:
a) As of now, if you do a successful login, than a bad login, the old session continues. Not sure if we should clear out the session if this happens

### DIFF
--- a/web/api/app/Controller/AppController.php
+++ b/web/api/app/Controller/AppController.php
@@ -70,76 +70,9 @@ class AppController extends Controller {
       throw new UnauthorizedException(__('API Disabled'));
       return; 
 		}
-		
-    $options = array('conditions' => array('Config.' . $this->Config->primaryKey => 'ZM_OPT_USE_AUTH'));
-    $config = $this->Config->find('first', $options);
-    $zmOptAuth = $config['Config']['Value'];
+    
+  
 
-    if ( $zmOptAuth == '1' ) {
-      require_once "../../../includes/auth.php";
-
-      global $user;
-      $user = $this->Session->read('user');
-
-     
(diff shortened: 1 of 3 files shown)


pliablepixels:
@connortechnology please review, but don't merge yet. I've done browser tests. Need to test with zmNinja. Will post when ready to merge.

pliablepixels:
ready to merge

pliablepixels:
TBD - working on:
a) converting login to POST
b) Moving user & pass to body 
c) removing user=&auth= from other APIs

pliablepixels:
This is complete from my side @connortechnology 

The only issue so far is I've not so far had success trying to login with curl. I don't believe its related to POST changes, but I haven't tested the old params method. 

Works fine with zmNinja as well as [postman](https://chrome.google.com/webstore/detail/postman/fhbjgbiflinjbdggehcddcbncdddomop?hl=en). Not sure why curl is not working. I even tried copy pasting the code generated from postman.
What happens in curl is that login works, but [this](https://github.com/ZoneMinder/zoneminder/blob/master/web/api/app/Controller/AppController.php#L117) check fails.

pliablepixels:
okay, seems to work reliably across all platforms at this stage.

pliablepixels:
@connortechnology can you review and if ok merge? I have folks running 1.31.44 who beta test zmN - before I send the next version to test, they will need these APIs as my app will use them if the version is >= 1.31.44

PR #2158, opened 2018-07-12
Add export to filter & fix exports

connortechnology:
This is a big effort to update and improve exporting.

The first job was to add an export button to filters so that you can just download the results of the filter.
The second step was to fix exports.  Some of this may be controversial.  I altered the structure of the resulting .zip or .tar to be shallow, and put the master.html in the root of the archive so that people can quickly find it.  It's the only file there, so people will naturally click on it, which will open it up in their browser. Instead of the full path to the event as we would see in /var/cache/zoneminder/events I just stuck them in a single dir referenced by event id. 

I roughed in some support for viewing .mp4 but this is incomplete at this time.

The export popup has been turned into a full page, which allows us to show some info about what the export will contain.

I changed the way that the archive is generated. Instead of passing tar a file list, I create a directory and build a tree inside of using symlinks to the real files and then tar that up.

I added the use of a unique key to prevent problems with concurrent exports.

Further todo:
The html in the export needs to be vastly improved. 
video tag support
export event metadata into a .json file for later re-import?
default selections to all images and video files instead of required things to be clicked.
testing

comments?


knight-of-ni:
That reminds me I was going to ask you about the export folder structure. I was thinking the same thing that is should be shallow. 

Building the archive from a list of symbolic links is the perfect solution for performing large exports on machines with little memory, like the Raspberry Pi. How would we handle this if the event to be exported is on a different zm server?

I've always sucked at making pretty html so I'll leave that up to someone else.

PR #2161, opened 2018-07-17
Added new control script for iPhone ip camera software

Daedilus:
Software for iPhones found below
https://itunes.apple.com/ca/app/ipcamera-high-end-network-camera/id570912928?mt=8

More details can be found here
See https://forums.zoneminder.com/viewtopic.php?f=9&t=27386 for details.


knight-of-ni:
@connortechnology @Daedilus This commit is missing the bits of sql that cause the control script to show up in the web console.  Please submit it.

See: https://github.com/ZoneMinder/zoneminder/blob/master/db/zm_create.sql.in#L779

Note you can use zmcamtool.pl to export the line of sql from an existing system.

PR #2162, opened 2018-07-17
Control/ipcamios

Daedilus:
Added the required default settings for control script IPCAMIOS.pm as per pull #2161


knight-of-ni:
Thanks!

PR #2168, opened 2018-07-24
fix_2167

connortechnology:
Populate a global  from the session on every request. Use the  object instead of using allowedMonitors in session. This is consistent with how we do it in the web ui.


pliablepixels:
I am testing today

pliablepixels:
Overall seems to work ok.
`Servers` API is broken - needs `$user` definition

pliablepixels:
next line `$user['system']` needs to be `$user['System']` - works after that

pliablepixels:
tested a few scenarios with/without auth in curl, zmN and browser. seems ok to me

PR #2173, opened 2018-07-31
retrieve geoip data rather than the ip

knight-of-ni:
We don't really need or want to know the user's ip address. Instead, get their approximate location via geoip database.

I've tested it and it returns success without @kylejohnson making adjustments to the server endpoint.

### DIFF
--- a/scripts/zmtelemetry.pl.in
+++ b/scripts/zmtelemetry.pl.in
@@ -33,6 +33,7 @@ use LWP::UserAgent;
 use Sys::MemInfo qw(totalmem);
 use Sys::CPU qw(cpu_count);
 use POSIX qw(strftime uname);
+use JSON::MaybeXS;
 
 $ENV{PATH}  = '/bin:/usr/bin:/usr/local/bin';
 $ENV{SHELL} = '/bin/sh' if exists $ENV{SHELL};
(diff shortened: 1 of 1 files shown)


knight-of-ni:
@pliablepixels to eventually create a map of ZoneMinder, just like this map for the open-source weewx project: http://www.weewx.com/stations.html

This will make it easy to answer the question "Where is ZoneMinder used"?

What is the concern?
In terms of personal data, how is the latitude and longitude any different than storing the city, state, and country?  In this case, both have the same level of accuracy, which is not very accurate at all.

pliablepixels:
The concern is granularity. Today browsers support lat long as well via the[ Geolocation APIs](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API) which are far more than city level granularity. This page identifies my location to < 1 mile from my browser. Is GeoIP simply translating an IP to a lat long or is it using browser GeoLocation? If its the latter, then its much more precise

pliablepixels:
Ok, I just looked at https://ipinfo.io/geo - it doesn't use browser location, so all good. Ignore I complained

knight-of-ni:
Thanks for checking. I am only looking for city-level accuracy. The free version of ipinfo seems to give that, but we can switch to something else if we find a need to.

When I paste my own coordinates from https://ipinfo.io/geo into Google, it comes up with a location 6.9 miles from my home.  

I chose to store both the human friendly city, sate, country and the computer friendly lat & long coordinates because I don't yet know what will be easier for the endpoint. Processing the location and making it look pretty is one of @kylejohnson skill sets, and he has been tied up with his dayjob.

PR #2176, opened 2018-08-07
Add Privacy Statement

knight-of-ni:
This PR adds a privacy statement that appears in-line to the admin user the first time ZoneMinder is started. This statement must be accepted or declined before continuing to the normal console screen.

If the statement is accepted:
- The Privacy statement will not show again until the next zoneminder update
- The Telemetry daemon is enabled (if you recall telemetry defaults to off)

If the statement is declined:
- The Privacy statement will not show again until the next zoneminder update
- The Telemetry daemon is disabled

@pliablepixels looking for feedback with the verbiage
@connortechnology looking for feedback with presentation and formatting

See lang/en_gb.php. I don't particularly like the long strings of text passed through the translator function, but I don't have a better idea at the moment to make this easier, while at the time have it possible to be translated.

Once we come to agreement on the verbiage, I need to duplicate the text (in English) to the other non-English language files. We will leave this up to other non-English speakers to update the text to their native language.

I am not opposed to changing how this works, or the look & feel (from anyone), so long as any feedback requesting a change is specific on how it can be improved. 

Here is how this looks in my demo environment:

privacy


pliablepixels:
Looks good overall. I'm out for the next 5 days, but some quick feedback:

- Add a point of contact if someone needs to reach out to you for privacy questions
- The text says that collecting telemetry helps answer questions from new users who ask. I don't believe there is a way for them to know today - so either provide a way for them to see this data, or ask you about it
- please add more clarity to the telemetry data, specifically lat/long. Without clarification, that text is implying that you know _exactly_ where a user is, which is not true. You are simply converting IP address to lat/long which is much wider. Would be good to address this in more detail. Granularity is a big concern
- It may be a good idea (in future) just to have a user exclude his lat/long/city but provide the rest. You still get what you want (mostly) and they don't have to fear pinpointing.
- Please provide a contact mechanism for users to reach out to you post enabling telemetry to have their data removed (they may change their mind)

connortechnology:
I've got a few tweaks, I'll try to get them done tonight or tomorrow. Overall looks great.

knight-of-ni:
Privacy text has been updated per the suggestions. Here is a new screenshot:

privacy2

connortechnology:
Sorry I thought this looked ready, but then noticed some oddities.

connortechnology:
I had thought that if you put something into en_gb, it would automatically get picked up for other languages... clearly I was wrong about that.  

I'm trying to think of some efficient way of doing this.  What I am thinking of is having the language files get generated programatically where it basically goes through a list of the available keywords and if an entry for the selected language isn't defined, it uses en_gb.  Not on every request, but maybe during build time.

knight-of-ni:
Yeah, what you said only works when item in question is just one word... e.g. translate(hello). If the key "hello" does not exist then it returns the name of the key. 

I went briefly went down the path you are considering. Trouble is, each language file loads all its keys into the same $SLANG array... so the question becomes, how do we load only some of the $SLANG keys from en_gb w/p overwriting all the other keys?  Maybe the answer is to create a different array that only contains the privacy text. $SLANG_PRIVACY ???  I have not thought through the implications of this.

pliablepixels:
Maybe the first time the language is loaded in [`loadLanguage`](https://github.com/ZoneMinder/zoneminder/blob/aa25936b3d07264e85012207dc4dad6484c620be/web/includes/lang.php#L30) one can do a one time load of the gb_lang file and populate missing keys and discard the duplicate lang file after that? Since it will happen on startup only, this should be acceptable.

PR #2185, opened 2018-08-22
Add Control plugin for Trendnet TVIP450PI

jof:
I did some experimentation and got this control plugin working with my Trendnet TVIP450PI outdoor dome PTZ camera.

The control protocol is based on the Physical Security Interoperability Alliance, IP Media Device API Specification

If I add a new row to the controls database, should I also maybe increment some database or software version number somewhere.


connortechnology:
Take a look at  zoneminder/scripts/ZoneMinder/lib/ZoneMinder/Control/Trendnet.pm for a version that contains a bunch of fixes in authentication.

I might suggest that we rename your file to something like PSIA.pm as it should work with any camera that implements the PSIA api.

connortechnology:
oh and no... no version #'s need updating.  PTZ control lines are easily updated.

PR #2191, opened 2018-08-28
Add Control plugin for PTZ cameras supporting the PSIA IP Media Device API

jof:
I did some experimentation and got this control plugin working with my Trendnet TVIP450PI outdoor dome PTZ camera.

The control protocol is based on the Physical Security Interoperability Alliance, IP Media Device API Specification.

This PR is a rename and largely similar to #2185

Issue #817, opened 2015-04-16
conflicting license statements


onlyjob:
I've noticed that some files have conflicting license statements. For instance the header of file [scripts/ZoneMinder/lib/ZoneMinder/Trigger/Channel.pm](../blob/master/scripts/ZoneMinder/lib/ZoneMinder/Trigger/Channel.pm) claims GPL-2+ but in the end of the same file I found [the following](../blob/master/scripts/ZoneMinder/lib/ZoneMinder/Trigger/Channel.pm#L161) statement:

> This library is free software; you can redistribute it and/or modify
it under the same terms as Perl itself, either Perl version 5.8.3 or,
at your option, any later version of Perl 5 you may have available.

Please clarify whether such `.pm` files are double-licensed or if only one particular license applies.

All `.pm` files in `scripts/ZoneMinder/lib/ZoneMinder/Trigger` are affected, some in `scripts/ZoneMinder/lib/ZoneMinder/Control` as well as all `*.pm` and `*.pm.in` in `scripts/ZoneMinder/lib/ZoneMinder`.

knnniggett:
@kylejohnson can you get a hold of Phil and see what he thinks about this one?

web2wire:
I'm happy to put everything on the same licence, probably GPL3 nowadays. I suspect the perl ones just had whatever I found in similar modules from elsewhere.

onlyjob:
So at the moment some Perl components (modules and scripts where Perl license is mentioned) are dual-licensed but we can upgrade to GPL-3+ (version 3 or above), right?

Since everything is already under GPL-2+ I suppose it already includes GPL-3+...

If everyone agrees we can just drop Perl license statement or (in case we choose to preserve dual-licensing) move them towards the beginning of files...

SteveGilvarry:
Review and fix all licenses

onlyjob:
No activity doesn't mean that the problem is gone.

996b:
Yeah not really knowing much about this trying to learn something

SteveGilvarry:
Thanks for the report (and apologies for the decade-long delay). Opened #4911 to resolve this.

**Clarification: only GPL-2-or-later applies.** ZoneMinder as a whole is GPL-2+ (see `COPYING`), and every affected file already carries the GPL-2+ notice in its header. The trailing POD footer — *"the same terms as Perl itself"* — is leftover `h2xs` module-scaffolding boilerplate, not an intentional dual-license election.

[...]

PR #2192, opened 2018-08-28
Control plugin for Dahua PTZ cameras

jof:


knight-of-ni:
Thanks for including the line of sql that allows this to show up in the web ui. 

In various places we've got conflicting license statements in the code. See #817 for details.

Your two open PR's you have open have the same issue. It's certainly not your fault since you most likely used existing scripts as examples.  Since you identify yourself as the author of these control files, technically speaking only you can modify the license. Would you please modify your PR's to reflect only one license. Simplest thing would be to remove the Perl license statement at the bottom or change it to match the GPL license in the header.

connortechnology:
Please see https://github.com/ZoneMinder/zoneminder/blob/master/scripts/ZoneMinder/lib/ZoneMinder/Control/Trendnet.pm as a more up to date example.

jof:
Thanks for the reviews, folks!

I updated the license information and also added handling for http or https-style URIs, based on the comment that came up in #2191

connortechnology:
Looking good.  Nitpicking now: We don't need the license at top and bottom.  I recommend removing the license at the top.

I would like to see more consistent conformance to google code style but I won't be too hard on that.

jof:
No worries. Nitpicks are what keeps a codebase clean.

Top-license removed.

scripts/ZoneMinder/lib/ZoneMinder/Control/Dahua.pm
```diff
@@ -0,0 +1,366 @@
...
+
+    # The Dahua camera firmware API supports the concept of having multiple
+    # channels on a single IP controller.
+    # As most cameras only have a single channel, and there is no similar
+    # information model in Zoneminder, I'm hardcoding the first and default
+    # channel "0", here.
+    $self->{dahua_channel_number} = "0";
+
+    my ($protocol, $username, $password, $address)
+       = $self->{Monitor}->{ControlAddress} =~ /^(https?:\/\/)?([^:]+):([^\/@]+)@(.*)$/;
```
connortechnology: This line doesn't work.  If the protocol isn't there, the returned values are shifted so you username in the protocol variable, etc.  Please see what I did in Trendnet.pm.

 if ( ( $self->{Monitor}->{ControlAddress} =~ /^(?<PROTOCOL>https?:\/\/)?(?<USERNAME>[^:@]+)?:?(?<PASSWORD>[^\/@]+)?@?(?<ADDRESS>.*)$/ ) ) {
    $PROTOCOL = $+{PROTOCOL} if $+{PROTOCOL};
    $USERNAME = $+{USERNAME} if $+{USERNAME};
    $PASSWORD = $+{PASSWORD} if $+{PASSWORD};
    $ADDRESS = $+{ADDRESS} if $+{ADDRESS};

connortechnology: Dang for some reason my cut n paste removed the <USERNAME> fields etc, probably because they look like HTML.  Please see Trendnet.pm.

jof: That's really good feedback. I guess there is no validation on the stored ControlAddress (since there can be so many applicable protocol schemes).

I used your scheme from Trendnet.pm in Dahua.pm. I'll also do the same change for #2191

Also, I removed some trailing whitespace from Trendnet.pm

PR #2197, opened 2018-08-31
README file location and name, README.https file

larandvit:
Made more accurate location and name of README file and mentioned about README.https file

### DIFF
--- a/docs/installationguide/redhat.rst
+++ b/docs/installationguide/redhat.rst
@@ -53,7 +53,9 @@ ZoneMinder releases are now being hosted at RPM Fusion. New users should navigat
 
 Note that RHEL/CentOS 7 users should use yum instead of dnf.
 
-Once ZoneMinder has been installed, it is critically important that you read the README file under /usr/share/doc/zoneminder. ZoneMinder will not run without completing the steps outlined in the README.
+Once ZoneMinder has been installed, it is critically important that you read the README.Redhat7 file under /usr/share/doc/zoneminder-[version #] folder. ZoneMinder will not run without completing the steps outlined in the README.Redhat7 file.
+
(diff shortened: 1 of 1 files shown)


docs/installationguide/redhat.rst
```diff
@@ -53,7 +53,9 @@ ZoneMinder releases are now being hosted at RPM Fusion. New users should navigat
 
 Note that RHEL/CentOS 7 users should use yum instead of dnf.
 
-Once ZoneMinder has been installed, it is critically important that you read the README file under /usr/share/doc/zoneminder. ZoneMinder will not run without completing the steps outlined in the README.
+Once ZoneMinder has been installed, it is critically important that you read the README.Redhat7 file under /usr/share/doc/zoneminder-[version #] folder. ZoneMinder will not run without completing the steps outlined in the README.Redhat7 file.
```
knight-of-ni, line 56: The trouble is, this will mislead Fedora users, which need to read README.Fedora.

docs/installationguide/redhat.rst
```diff
@@ -53,7 +53,9 @@ ZoneMinder releases are now being hosted at RPM Fusion. New users should navigat
...
+
+if you wish to get rig of a warning from your browser regarding certificate, you can follow suggestions in README.https file.
```
knight-of-ni, line 58: I get it you are trying to be helpful, but this is already stated in step 6 of the README. Let's keep this note in just one place. If you want to modify step 6 of both README's to make that more clear, I'd be in favor of that.
The README's live here: https://github.com/ZoneMinder/zoneminder/tree/master/distros/redhat/readme

larandvit: Thank you. I agree with you. I missed it.

PR #2198, opened 2018-09-03
optionhelp - support links markdown style

knight-of-ni:
This one-liner adds support for html links, markdown style, embedded in ZoneMinder's Option Help text.

Post 1.32 release, this could be expanded to use a php markdown library if we want make full use of markdown, but I don't want to open up a bigger can of worms than I have to right now.

This PR is being driven by a need to allow users to re-read the new Privacy text, after they have already accepted/declined it, by clicking on a link embedded in the ZM_TELEMETRY_DATA help text.

@connortechnology @pliablepixels Thoughts? I am all for any other ways to navigate to the privacy view.


connortechnology:
I like the idea of full markdown later on.  

I don't feel that we need to make it too accessible to get back to.  If someone really wants to get to it they can just type view=privacy into the url bar.  Let's just document it so that google searches find how to get to it.

PR #2201, opened 2018-09-09
use short header on montage

knight-of-ni:
This PR makes an attempt to recover some screen real estate from the montage view. 

Two things have been done:
1) The ZoneMinder navbar has been stripped down to just a few items
2) A hamburger icon has been added to the right, which allows the user to collapse the layout editing fields.

The decision to keep some navbar items and remove others was mostly subjective. 
Some items, like the running status and version of zoneminder, didn't make it because they required additional logic, which muddied up montage.php even further.

Feedback is welcome. 
Optionally, I could remove the remaining navbar items and just have a "back" link to go back to the console.

Screenshot showing the new menu with layout fields expanded (default):
short-header-expanded

Screenshot showing the new menu with layout fields collapsed:
short-header-collapsed

### DIFF
--- a/web/skins/classic/views/montage.php
+++ b/web/skins/classic/views/montage.php
@@ -132,10 +132,21 @@ foreach( $displayMonitors as &$row ) {
 xhtmlHeaders(__FILE__, translate('Montage'));
 ?>
 <body>
-  <div id="page">
-    <?php echo getNavBarHTML() ?>
+  <div id="page">   
+    <div class="navbar navbar-inverse navbar-static-top"> <!-- Begin Short Header -->
+      <div class="navbar-brand">
+        <a href="<?php echo ZM_HOME_URL?>" target="<?php echo ZM_WEB_TITLE ?>"><?php echo ZM_HOME_CONTENT ?></a>
+      </div>
+      <ul class="nav navbar-nav">
+        <li><a class="nav navbar-nav" href="?view=console"><?php echo translate('Console') ?></a></li>
+        <li><a href="?view=options"><?php echo translate('Options') ?></a></li>
+      </ul>
+      <button data-toggle="collapse" data-target="#collapseHeader" class="btn btn-default navbar-btn pull-right"><span class="glyphicon glyphicon-menu-hamburger"></span></button>
+    </div> <!-- End Short Header -->
+
     <div id="header">
-      <div id="headerButtons">
+      <div id="collapseHeader" class="collapse in">
+        <div id="headerButtons">
 <?php
 if ( $showControl ) {
 ?>
@@ -188,8 +199,9 @@ if ( $showZones ) {
         </form>
       </div>
     </div>
-    <div id="content">
-      <div id="monitors">
+  </div>
+  <div id="content">
+    <div id="monitors">
 <?php
 foreach ( $monitors as $monitor ) {
   $connkey = $monitor->connKey(); // Minor hack


connortechnology:
I don't really like the loss of links at the top.  I don't really like Montage being different from every other page.  I think we should use the same header function for all pages.  Long term we should make it totally customizable in config. 

I assume that once hiding the second layer header with all the filters that it doesn't remember it on refresh?  I just don't see any logic in there to store it in session or anything else.

knight-of-ni:
Would you agree that we need to do something for the montage view? Previously there was no header at all and the one there now takes up 1/3 of the screen. 

Initially I tried to make the entire header bar collapsible everywhere.

The normal header uses a bootstrap navigation bar. The trouble is, that conflicts with the bootstrap collapse feature.... apparently one can't use both at the same time. I spent several hours trying to do just that and was not successful.

I also explored trying to force the navigation bar hamburger button to simply stay visible all the time. That is not possible in bootstrap 3 without hacking/overriding the bootstrap css. It looks like it is possible to in bootstrap 4, but I don't want to do that before we release.

Perhaps you can see now why I went down the path I did. There really isn't another good choice at the moment but to create a different header bar. We could move the code in this pr into a function called  getShortNavBarHTML. I'm not opposed to putting back some of the links, just as long as we can get it to fit on one line.

connortechnology:
I agree that something needs to be done long term, however I don't agree that it needs to be done right now to appease one user, no matter that we like him.

Individual users can alter the skin/css to their hearts content to achieve whatever look they desire.  

I like the ability to hide the headers, I just don't like removing all the links at the top.  I put those there because _I_ find them useful.  Others will have different opinions.  Which is why long term it has to be configurable.  However for now, if we click the hamburger and they go away, I think that will appease most people, and it's something I would use too.  

Maybe we can't do it with bootstrap.  We can implement our own hamburger button that maybe shifts between 2 or more states.  Maybe full view, minimal view, half way view, etc...

knight-of-ni:
This isn't about one user's preference. Recall, the header was originally added by you w/o asking anyone else, and I'm sure you wouldn't like it if I returned the favor by making my own changes on my own. Rather than go down that path, I created this PR to avoid doing just that.

So let's keep this conversation about the objective facts concerning how much room the header takes on the screen and how best to go about addressing that.

Just for clarification, when you refer to the links at the top, do you mean just the links on the topmost line? That would include Console, Options, Log, Groups, the rest of the menu links, the name of the user who is logged in, and the Running status?

I can add those back in if that is what you mean. If the montage is full screen, those will not increase the vertical size of the header as shown in my previous screenshots.  If montage isn't full screen, or the resolution is sufficiently low, then we could have a problem since that automagic navbar stuff is not enabled. 

I wonder if there is a way to collapse the header w/o an actual button.... for example, double clicking the empty space in the header would trigger the function.

connortechnology:
We can definitely do double click however I think people are now used to click on hamburgers.

I think we need to focus on discussing what should be visible in the minimized view.  Everything can be hidden.

I think we also need to consider that this effort should also be useful for the other views as well, not just montage.

connortechnology:
I taken a stab at this as well and it gets more complicated with the chosen styles.  If the selects are hidden when the chosen scripts run then the resulting styled dropdowns have width:0px;  I'm working on figuring out how to get them to refresh.

knight-of-ni:
superseded by #2203. Closing.

PR #2207, opened 2018-09-14
Fix #2205

### LINKED ISSUE #2205: Wrong path for event images sent to filter scripts
Version 1.32, Ubuntu updated via apt yesterday.

The event path sent to the bash scripts in my filters are sending the old path structure to the scripts, and not the new structure.

eg.: /var/cache/zoneminder/events/6/18/09/14/08/03/17
and not: /var/cache/zoneminder/events/6/2018-09-14/44442/00069-capture.jpg

Making all my filterscripts incompatible with the current version.

connortechnology:
deprecate old event code in General. Using event object methods will result in the correct paths being returned.


knight-of-ni:
Assigning a bug label so we remember to include this in 1.32.1.

PR #2217, opened 2018-09-21
Touch lockfile

connortechnology:
zms clutters up ZM_SOCK_DIR with .lock files.  This code touches these files every hour while zms is running, so that we can clean out anything older than that.

### DIFF
--- a/src/zm_eventstream.cpp
+++ b/src/zm_eventstream.cpp
@@ -755,6 +755,12 @@ void EventStream::runStream() {
     // commands may set send_frame to true
     while(checkCommandQueue());
 
+    // Update modified time of the socket .lock file so that we can tell which ones are stale.
+    if ( now.tv_sec - last_comm_update.tv_sec > 3600 ) {
+      touch(sock_path_lock);
+      last_comm_update = now;
+    }
+
     if ( step != 0 )
       curr_frame_id += step;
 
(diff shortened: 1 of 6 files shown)


knight-of-ni:
I tested this last weekend, but got distracted on another issue.

What I discovered was the second part to this (as discussed in our slack channel) does not work. It turns out the tmpfiles.d "r", a.k.a. remove file, flag does not support the age parameter. Oui vey.

The "d" flag supports the age parameter, but it cleans out the entire folder contents older than the age we set.

So at the moment, our two options appear to be:
- Use something else to periodically clean the lock files
- Put the lock files into a subfolder, say ZM_SOCK_DIR/lock for example

We could use cron, but unfortunately BSD's don't have an "hourly" folder we can just drop a config file into like we have on Linux. We would have to script inserting a record into the web user's cron file under /var/spool. Yuck.

connortechnology:
On ubuntu trusty we don't have systemd.  So cron is the way to go there.  
there is always /etc/crontab although putting it in www-data's crontab is sorta better.

knight-of-ni:
I'll go ahead and merge, but I'm still thinking about how best to actually delete the lock files.
The problem with using the web server's crontab is you'd have to fully parse the existing crontab and search for duplicates. That's certainly not an impossible task, but that does seem overly complicated, and prone to error if the end user edits the cron entry, given the situation.

PR #2222, opened 2018-09-26
added note to make sure headers is on

pliablepixels:


knight-of-ni:
a2enmod is only available on Debian based Linux distros. Can you make the message more generic?

knight-of-ni:
Perfect. Thank you.

PR #2225, opened 2018-09-28
don't zmMemInvalidate when reloading monitors

knight-of-ni:
Zmtrigger has been observed to miss monitors that have been recently restarted as discuess in this forum thread and others: https://forums.zoneminder.com/viewtopic.php?p=107334

This PR resolves that by removing a redundant call to zmMemVerify in loadMonitors(). Calling zmMemVerify is handled in the main loop.

This PR should be tested as it seems possible this could cause other problems during different or abnormal conditions.

### DIFF
--- a/scripts/zmtrigger.pl.in
+++ b/scripts/zmtrigger.pl.in
@@ -323,12 +323,6 @@ sub loadMonitors {
   my $res = $sth->execute( $Config{ZM_SERVER_ID} ? $Config{ZM_SERVER_ID} : () )
     or Fatal( "Can't execute: ".$sth->errstr() );
   while( my $monitor = $sth->fetchrow_hashref() ) {
-# Check shared memory ok
-    if ( !zmMemVerify( $monitor ) ) {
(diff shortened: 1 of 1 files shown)


pliablepixels:
Is this reliably working? I have a user who is reporting more issues with these changes. I've asked him to post logs so I can investigate. This is using the changes in this PR applied to my notification server.

knight-of-ni:
Yes, it is.

knight-of-ni:
This pr fixes the same kind of errors reported by the user in the eventserver project. Notice this pr modifies the loadMonitors() sub. Also notice that the eventserver project is using the same loadMonitors() sub, which means it will exhibit the same problem zmtrigger had.

pliablepixels:
I've updated the code to model zmtrigger but the SHM errors for this user continue. I'll keep investigating and continue over at the event notification [issue](url
https://github.com/pliablepixels/zmeventserver/issues/54) that refers to this PR.

PR #2231, opened 2018-10-03
allow username&password even if AUTH_HASH is enabled

pliablepixels:
fixes https://github.com/ZoneMinder/zoneminder/commit/efda26121bb862263afb23ba1da8cd5b5999c713


connortechnology:
This needs something to differentiate a login attempt versus saving a user for example. We need an and $_REQUEST['action'] == 'login'.

Also if the GOOGLE recaptcha is enabled, we need to test for it as well.

PR #2242, opened 2018-10-09
Cleanup auth

connortechnology:
The gist of this is to move the guts of authentication logic from actions.php into auth.php.  Also move it from index.php.

Along the way we fixed a bug where you can't have php tags in your username or password.


connortechnology:
I think I might be able to clean out some code from api now too... let me see if I can add that to this.

connortechnology:
I think auth changes to api can come later.

PR #2248, opened 2018-10-11
rough in for alarmed frame retrieval

pliablepixels:
This PR allows us to retrieve the first alarmed frame in an event (critical for the object detection stuff I added to the event server)

### DIFF
--- a/web/views/image.php
+++ b/web/views/image.php
@@ -76,6 +76,22 @@ if ( empty($_REQUEST['path']) ) {
       return;
     }
 
+    # if alarm, get the fid of the first alarmed frame if available and let the
+    # fid= code continue processing it
+    if ( $_REQUEST['fid'] == 'alarm' ) {
+      $Frame = Frame::find_one(array('EventId'=>$_REQUEST['eid'], 'Type'=>'Alarm'));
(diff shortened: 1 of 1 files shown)


web/views/image.php
```diff
@@ -76,6 +76,22 @@ function imagescale($image, $new_width, $new_height = -1, $mode = 0) {
       return;
     }
 
+    # if alarm, get the fid of the first alarmed frame if available and let the
+    # fid= code continue processing it
+    if ( $_REQUEST['fid'] == 'alarm' ) {
+      $Frame = Frame::find_one(array('EventId'=>$_REQUEST['eid'], 'Type'=>'Alarm'));
```
connortechnology: Still have the problem that because we aren't doing any ordering, we could get the last, the first or any alarm frame in between.

pliablepixels: oh I thought it would always be ordered because of their occurrence in sequence?

connortechnology: Shouhld be $Frame = Frame::find_one(array('EventId'=>$_REQUEST['eid'], 'Type'=>'Alarm'),array('order'=>'Id DESC'));

PR #2250, opened 2018-10-12
check for __ARM_PCS_VFP instead of __armel__

knight-of-ni:
According to the Debian wiki, `__ARM_PCS_VFP` flag is set for armhf but not armel, so let's use this pre-processor macro instead of `__armel__` .
https://wiki.debian.org/ArchitectureSpecificsMemo

Please don't merge until we heard from @onlyjob this works.

I checked my raspi3 running fedora 28 and the __ARM_PCS_VFP flag is set also (what we want) so it does not appear this will exclude neon instructions in situations where the cpu supports it. @mastertheknife if you could spare a moment, it would be nice to know if you agree.

### DIFF
--- a/src/zm_image.cpp
+++ b/src/zm_image.cpp
@@ -3317,11 +3317,11 @@ __attribute__((noinline)) void std_fastblend(const uint8_t* col1, const uint8_t*
 }
 
 /* FastBlend Neon for AArch32 */
-#if (defined(__arm__) && !defined(__armel__) && !defined(ZM_STRIP_NEON))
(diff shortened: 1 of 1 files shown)


knight-of-ni:
The Debian build log reports the armel build succeeded with this change so I will merge this. It will be part of the 1.32.2 release which I am creating now.

onlyjob:
It works and __armel__ build [succeeded](https://buildd.debian.org/status/fetch.php?pkg=zoneminder&arch=armel&ver=1.32.1-3&stamp=1539431003&raw=0). Thank you, @knight-of-ni.

Issue #2254, opened 2018-10-15
Adding a few D-link camera presets


azend:
I have some old D-link cameras that don't have camera presets yet. I thought it would be useful to make them and share.

knight-of-ni:
Closing, since the associated pr was merged.

PR #2255, opened 2018-10-15
Added monitor preset for D-link DCS-930L

azend:
#2254


connortechnology:
One issue here is that the db update script won't be run ... it needs to be 1.32.3

azend:
I didn't realize we were on that version already. I'll move to the new version code.

PR #2256, opened 2018-10-15
Zoneminder bosnian translation

dado-ca:
Hi,

this is a bosnian translation for zoneminder, but it can be used as croatian and serbian translation too


dado-ca:
Need to make some corrections to translation.

Issue #2109, opened 2018-05-23
zmc restarting


oyhaare:
zmc keeps being restarted by zmwatch:

>zmwatch[19571]: ERR [Shared data size conflict in shared_data for monitor Monitor-1, expected 608, got 604]
> zmwatch[19571]: INF [Restarting capture daemon for Monitor-1, shared data not valid]

I suspect this might have to do with commit cc7756182699d1193d7828b21a039b9b78cc9472 and the fact that I am running on a 32bit system and/or perl vs cpp struct memory alignment/padding.

This is using the latest storage-area and master snapshot from iconnor ppa's

connortechnology:
c and perl should match.  I run both 32bit and 64bit and some odd things and they all work for me.  Is it latest storageareas or master snapshot?  Can't be both...

dpkg -s zoneminder

oyhaare:
I've tested with both storageareas and master.

Must be something wonky on my system then.... but what? Any ideas?
I'm running ubuntu bionic on a virtualbox vm.

I'm running ubuntu

connortechnology:
The correct value is 608.  These kinds of errors can happen when there are old libraries installed in a different location...

oyhaare:
Closing this as it seems something was messed up on my system.
Ended up reinstalling ubuntu from scratch...

thigger:
Also getting this issue, also on Ubuntu Bionic (just upgraded from 16.04). Any idea which libraries?

cipibad:
Same issue here, fresh install 32bit Ubuntu 18.04.1 LTS, zoneminder 1.32.2-bionic1 installed from ppa:iconnor/zoneminder-1.32

cipibad:
I managed to solve the error, see: https://github.com/ZoneMinder/zoneminder/pull/2258
You can manually apply the change directly to current install modifying /usr/share/perl5/ZoneMinder/Memory.pm file.

PR #2258, opened 2018-10-16
align to native instead hardcoded value

cipibad:
Solved error "zmwatch[5843]: ERR [Shared data size conflict in shared_data for monitor demisol, expected 608, got 604]" pn 23bit, change looks to be 64bit compatible, but someone needs to check it first.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Memory.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Memory.pm.in
@@ -207,7 +207,8 @@ sub zmMemInit {
           || $member_data->{type} eq 'uint64'
           || $member_data->{type} eq 'time_t64'
           ) {
-        $member_data->{size} = $member_data->{align} = 8;
+        $member_data->{size} = 8;
+        $member_data->{align} = $native;
       } elsif ( $member_data->{type} eq 'int32'
           || $member_data->{type} eq 'uint32'
           || $member_data->{type} eq 'bool4'


thigger:
Tested on 18.04.1 LTS Ubuntu 32-bit (on 64-bit processor) and solves my issue ( #2109 )

knight-of-ni:
I just need time to test this under a few different environments. If all is well, I will merge this.

knight-of-ni:
Unfortunately, this PR does not work on armv7 architecture. It causes a log full of shared memory errors, which is probably what you are seeing. I cannot merge it as it is currently.

Try running the perl script below on your machine. It uses four different methods to determine the "bitness" of your environment. If one of the four methods produces the right 32/64 answer in *all* cases, then perhaps we could change the way $arch is calculated to the alternative method. The assumption with this attepted fix is that a too small value of $member_data->{size} will still work.

```perl
#!/usr/bin/perl

use Config;

my $use64bitint = 0;
my $arch = 32 + 32*( qx(uname -m) =~ /64/ );
my $ivsize = 32 + 32*(qx(perl -V:ivsize) =~ /8/);
my $ptrsize = 32 + 32*(qx(perl -V:ptrsize) =~ /8/);

if ($Config{use64bitint} eq "define") {
    $use64bitint = 64;
} else {
    $use64bitint = 32;
}

print "Perl bitness using existing ZM method: $arch \n";
print "Perl bitness using use64bitint: $use64bitint \n";
print "Perl bitness using ptrsize: $ptrsize \n";
print "Perl bitness using ivsize: $ivsize \n";
```

Here is what that script outputs on x86_64, running a 64bit os:
```
Perl bitness using existing ZM method: 64 
Perl bitness using use64bitint: 64 
Perl bitness using ptrsize: 64 
Perl bitness using ivsize: 64
```

And here is what the same script outputs on armv7, running Armbian:
```
Perl bitness using existing ZM method: 32 
Perl bitness using use64bitint: 64 
Perl bitness using ptrsize: 32 
Perl bitness using ivsize: 64 
```

So compare the results of the script on your machine with the ones above. Does any one method produce the right 32/64 bit value for all three cases?

cipibad:
I don't think the problem is not "bitness" detection, but alignment of int64, uint64 and time_t64. 

In this case: $member_data->{size} = 8 is right, but I had to change $member_data->{align} from 8 to $native - which is 4 in 32bit. I suppose this 4 is not the right value, but something working in my case.

I suppose is not worthing spending more time on this, my hardware is to slow for ZM anyhow.
Bellow the output of test code:

Perl bitness using existing ZM method: 32
Perl bitness using use64bitint: 64
Perl bitness using ptrsize: 32
Perl bitness using ivsize: 64

knight-of-ni:
Agreed, the issue is alignment, not bitness, but the way we are calculating bitness is affecting the alignment. 

Now that I see the output from the Perl script from your machine, I think we can fix the problem, without breaking anyone else's system, by changing [line 139](https://github.com/ZoneMinder/zoneminder/blob/master/scripts/ZoneMinder/lib/ZoneMinder/Memory.pm.in#L139) from:
`our $arch = 32 + 32*( qx(uname -m) =~ /64/ );`

to:
`our $arch = 32 + 32*(qx(perl -V:ptrsize) =~ /8/);`

Technically we could get rid of $arch and just compute $native directly, but what I've shown for now is good enough for testing.

Try that and let me know. I'll test on my end as well.

knight-of-ni:
Ah no. Didn't work.

knight-of-ni:
So it would seem we need some thing else, other than bittness, to determine what our alignment should be.

knight-of-ni:
Better yet, try this perl script from the command to verify whether or not !L and Q are the same:
```perl
#!/usr/bin/perl

printf "format L! is %d\n",
	length pack("L!");

printf "format Q is %d\n",
	length pack("Q");
```

knight-of-ni:
@cipibad 
It looks like this problem began about 7 months ago with this commit:
https://github.com/ZoneMinder/zoneminder/commit/cc7756182699d1193d7828b21a039b9b78cc9472

That commit increased the shared memory by 4 bytes, which is the difference in shared data size conflict message you reported.

The commit itself isn't the problem. Rather, the underlying Perl script for the uint64 type does not seem to work correctly on some systems. We only now uncovered it when we tried to use that type for first time in ZoneMinder's history. It's like an old, mummified bug we just released into the wild.

I will be out of town this week for day job, but will attempt to make further progress, time permitting, in the evenings.

cipibad:
Yes, the q type is available:
format l! is 4
format L! is 4
format Q is 8

I changed in Memory.pm all l!->q and L!->Q and undo initial change, I got back to initial error.

I suppose it is a mix of them ...

knight-of-ni:
We believe we have fixed the issue described here in pr #2307. 
Please report success of failure.

cipibad:
Error changed, but still some issues are there, while streaming:
Got unexpected memory map file size 73733500, expected 73733504

knight-of-ni:
> Got unexpected memory map file size 73733500, expected 73733504

@cipibad That is coming from zms, the C side of our code rather than Perl. We are not seeing that on our end. Can you confirm whether you rebuilt zoneminder with all the changes shown in #2307 ?

knight-of-ni:
@techrockedge was also experience this issue and reported that it was fixed for him in the 1.32.3 release. Please try it when you have an opportunity.

knight-of-ni:
I am going to close this PR out as we have developed a different fix and have not seen any new reports with the symptoms described here following 1.23.3 release. If it does occur, please create a new issue and we will look into it.

PR #2269, opened 2018-10-23
API - Disable E_NOTICE from php error reporting in cake debug

ratmole:
Using zmNinja, the API reports E_NOTICE errors 
```
Notice (8): compact(): Undefined variable: subject [CORE/Cake/Utility/ObjectCollection.php, line 128]
Notice (8): compact() [<a href='http://php.net/function.compact'>function.compact</a>]: Undefined variable: subject [CORE/Cake/Utility/ObjectCollection.php, line 128]
Notice (8): compact() [<a href='http://php.net/function.compact'>function.compact</a>]: Undefined variable: subject [CORE/Cake/Utility/ObjectCollection.php, line 128]
Notice (8): compact() [<a href='http://php.net/function.compact'>function.compact</a>]: Undefined variable: subject [CORE/Cake/Utility/ObjectCollection.php, line 128]
```

and zmNinja will not work...
there is a better way, but i think disabling E_NOTICE error is way easier

### DIFF
--- a/scripts/zmfilter.pl.in
+++ b/scripts/zmfilter.pl.in
@@ -576,7 +576,8 @@ sub uploadArchFile {
       $host .= ':'.$Config{ZM_UPLOAD_PORT} if $Config{ZM_UPLOAD_PORT};
       Info('Uploading to '.$host.' using SFTP');
       my %sftpOptions = (
-          host=>$Config{ZM_UPLOAD_HOST}, user=>$Config{ZM_UPLOAD_USER} 
+          host=>$Config{ZM_UPLOAD_HOST},
+          user=>$Config{ZM_UPLOAD_USER},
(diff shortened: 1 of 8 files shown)


pliablepixels:
1. On the debug 0 part, I had a separate discussion with @connortechnology and the conclusion was that people may want to enable Debug messages in their local instance and APIs should continue to work. Like you correctly said, this leaks non JSON messages that breaks zmN. I recently implemented an interceptor to filter out anything that starts with `<php class='cake-error'>` and ends in a `<pre>`. Not ideal, but zmN should not break now.

2. Can you post a sample of E_NOTICE errors? Are they inside the same tags?

knight-of-ni:
The verbiage in this PR fails to mention that the code also disables the privacy popup forever. This is not in compliance with GDPR and thus this pr is unacceptable.

PR #2274, opened 2018-10-25
Workaround for php built without semaphore support

ASerbinski:
Not all php builds include support for semaphores.
Since zm now depends on semaphore support, it is necessary to
add facility to support this where it is not possible
to replace the distribution's build of php.


connortechnology:
This isn't needed.  I already added them to includes/functions.php

PR #2282, opened 2018-10-29
align API diskPercent with PHP code. Do away with du as it kills CPU

pliablepixels:


connortechnology:
Looks ok, except that it changes the API.  The old function took a mid as a parameter, this one takes a path.  Which I'm ok with, because you can request the amount of space taken up by a monitor from the monitors side of the API.  If I was going to be pedantic though I would say that we shouldn't change the API without a major version change.

connortechnology:
Of course, you should also be able to request that info from the Storage Model side of the api....
Perhaps we should just mark getDiskPercent deprecated and add this better functionality to the Storage Controller

pliablepixels:
Agree, the `/storage.json` API gives much more detailed info. I'll close this PR.

Issue #2304, opened 2018-11-12
Add the ability to view first alarmed frame while an event is recording without requiring frames to be stored


pliablepixels:
https://github.com/ZoneMinder/zoneminder/pull/2248 implements `fid=alarm` but while the event is being recorded, requires frames storage to be enabled to view the first alarmed frame which is an overkill for H264 storage users. The ask is to store the alarmed frame without needing frames enabled.

(Motivation: real time object detection on alarmed frame as the event records)

PR incoming.

PR #2305, opened 2018-11-12
Save first alarm

pliablepixels:
#2304

### DIFF
--- a/src/zm_event.cpp
+++ b/src/zm_event.cpp
@@ -117,6 +117,7 @@ Event::Event(
   alarm_frames = 0;
   tot_score = 0;
   max_score = 0;
+  alarm_frame_written = false;
 
   char id_file[PATH_MAX];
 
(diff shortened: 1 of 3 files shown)


knight-of-ni:
Sorry for the delay. Having been out of town, I'm trying to get caught up with all the open pull requests. 
Your friendly voice has been missing from our Slack channel this week, so I'm guessing you may be out this week due to the holidays.

This PR writes a jpeg, alarm.jpg, to the event folder. This is in addition to snapshot.jpg, which is also being written to the event folder. I'm wondering if, instead, we ought to just write the alarm frame to the event folder and use that for both purposes. Maybe @connortechnology could speak to this as well.

I don't intend to make a drawn out discussion over this, but I did want to ask the question before I merged this PR as it is currently written.

connortechnology:
The problem is that the snapshot is the frame with the most motion, not necessarily the frame that triggered the alarm.

pliablepixels:
@knight-of-ni I am out of town, sorry for the late response. Purely from an object detection perspective I'd suggest we keep both `alarm` and `snapshot`. I have users who invoke both methods to make sure they don't miss detection as depending on how their zones are setup, one of these will be useful.

PR #2307, opened 2018-11-13
remove epadding2, to restore the 64bit alignment of startup_time.  

connortechnology:
A proper fix instead of #2258

Startup_time is a 64bit so can be aligned to a 64bit offset.


knight-of-ni:
Thanks for figuring this one out.

PR #2313, opened 2018-11-17
Huh

regularnmae:
-Added video generate job, running in the background independet from the browser
-Fixed archive download, when large file (Poorly done but works for centos)
-Disabled missing files check (Filtering events takes forever on utilized disk storage)
-Fixed montagereview player showing NO DATA (Works only with bulk frames on)


regularnmae:
> 
> 
> Thank you for taking the time to write this.
> This may conflict with some of the work that @connortechnology has done so we may need to coordinate that.
> 
> For the short term, please fix the hard coded paths in your PR.
> Hard coded paths to files, binaries, and folders must be replaced with an existing ZM_XXXX variable.
> See other areas of the code to determine what those values should be.

Yes, there is hard coded everything not only paths, but I'm not sure anyone need this? Should I continue development or just leave as it is? Because storing frames as images is not efficient way of storing huge amount of data. So I'm concerned about future of this project

knight-of-ni:
In order to get others to test your pr, you need to change it so that those not running centos can use it. That means fixing the hard coded paths to ffmpeg, the temp folder, etc. This should be pretty a simple task, compared to the work you have already done. I can't say what we will do with your pr after that. We need to able to test it first.

Sure, jpeg storage is inefficient. No one disputes that, but storage is cheap. We've already incorporated mp4 storage into ZoneMinder, so it should be fairly obvious the direction the project is headed. If you really want to be a big help, send us a PR that resolves the jpeg requirement for some of the auxillary functions in ZoneMinder.

web/skins/classic/views/report_event_audit.php
```diff
@@ -102,12 +102,12 @@
     if ( $gap > $EventsByMonitor[$event['MonitorId']]['MaxGap'] )
       $EventsByMonitor[$event['MonitorId']]['MaxGap'] = $gap;
 
-  } # end if has previous events
-  if ( ! $Event->file_exists() ) {
-    $EventsByMonitor[$event['MonitorId']]['FileMissing'][] = $Event;
-  } else if ( ! $Event->file_size() ) {
-    $EventsByMonitor[$event['MonitorId']]['ZeroSize'][] = $Event;
-  }
+  } # end if has previous events # DISABLED takes very long time on busy disk sotarage when saving images
```
connortechnology: We won't be merging this bit as this code is actually useful and is the actual reason this view exists.

 I would accept a patch that added a toggle for whether or not to do these tests though.

regularnmae: Agree, and there is a help tip in the zmaudit options

> This is recommended for most systems however if you have a very large number of events the process of scanning the database and filesystem may take a long time and impact performance. In this case you may prefer to not have zmaudit running unconditionally and schedule occasional checks at other, more convenient, times. 

I need to disable zmaudit but not "missing files check" in the source, I'm feeling stupid now

regularnmae: If zmaudit is doing the same check about the missing files, than why it is nessesary to check files additionally from the report page? It can be stored in the database

regularnmae:
-Fixed hard coded variables except framerate, because can't figure out how to find actual framerate of the monitor
-Reworked archive download
-Removed missing files "fix"
-Removed bulk frames fix, because it should be option?

web/ajax/generate.php
```diff
@@ -12,7 +12,11 @@
   die('Request parameters missing');
 }
 
+require_once( '../includes/config.php' );
```
connortechnology: This shouldn't be necessary as it is included in index.php.  Doesn't hurt to have it though.

Oh wait, I'm wrong, I see now that you are using this view independenlty

web/includes/generate_job.php
```diff
@@ -68,27 +73,29 @@
...
-  $framerateStr = $framerate / 4;
+  $framerateStr = $framerate / RATE_BASE;
+  # How to get actual framerate?
+  $hardCodedFramerate = framerateStr * 5;
 
   # Command lines
-  $commandline1 = "cd /mnt/rec/events/" . $event["MonitorId"] . "/" . $date  .  "/" . $eid;
-  $commandline2 = '/usr/bin/ffmpeg -y -framerate ' . $framerate . ' -i %05d-capture.jpg -s ' . $sizeStr . ' ' . 
-    $encoderStr . ' -pix_fmt yuvj420p "Event-_' . $eid . '-r' . $framerateStr  . '-s' . $size  . '.avi"';
+  $commandline1 = 'cd ' . ZM_DIR_EVENTS . '/' . $event['MonitorId'] . '/' . $date  .  '/' . $eid;
```
connortechnology: We probably should have advised better on this.  ZM_DIR_EVENTS used to be the thing to do, but we can have multiple different storage locations.  The Event object has methods that can get you the correct path to the monitor. 
So you can do 
$Event = new Event($event); 
and then do 
= 'cd ' . $Event->Path();

regularnmae: I think better approach would be to use the function **createVideo**, which generates videos using the perl script called zmvideo.pl and set the encoder parameters in the options.

regularnmae:
OMG I REWROTE EVERYTHING, but it works..  and I like it :-P

PR #2314, opened 2018-11-17
rpm sub-packages

knight-of-ni:
Initial merge to split zoneminder rpm into sub-packages. 
Sub-packages allow support for nginx, in addition to apache:

zoneminder-common
zoneminder-httpd
zoneminder-nginx


knight-of-ni:
self merging, since this pr applies to rpm packaging only

PR #2321, opened 2018-11-23
Av logging

connortechnology:
This PR redirects the logging output from libavcodec into our own logging functions.
Along the way, I had to put a mutex in logPrint because multi-threaded codecs will log concurrently.  
This required some cleanup of various includes.


connortechnology:
This code seems to work pretty well, no crashes any more after adding the mutex.  

One problem is that a lot of cameras result in ffmpeg complaining about invalid golomb code and stuff like that which will now spam our logs.  We may have to adjust the levels a bit.

PR #2323, opened 2018-11-23
Fix triggers to sum event counts properly

knight-of-ni:
This pr creates a new update sql file by merging contents of 1.31.25.sql and 1.31.36.sql.

I did this rather quickly and needs verification.

For example, the top most UPDATE statement may be able to be removed or merged into the bottom UPDATe statement. @connortechnology to confirm.

### DIFF
--- a/db/zm_update-1.32.3.sql
+++ b/db/zm_update-1.32.3.sql
@@ -3,7 +3,184 @@
 --
 
 --
--- Add some additional monitor preset values
+-- Update our triggers to sum event counts properly
 --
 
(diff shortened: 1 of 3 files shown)


knight-of-ni:
hang out. Github is at it again.....

knight-of-ni:
Note: I am intentionally overtiring a ptz control present because adding these to sql update statements is no longer needed. zmcamtool.pl takes care of that now.

knight-of-ni:
Nope, this isn't right. After a while the counts shown on the web console go back to the incorrect values.

knight-of-ni:
The latest changes to this PR appear to be working. I've got two ZoneMinder systems, one el7 the other f28, that are tracking events correctly now.

PR #2328, opened 2018-11-29
returns user=&pass= in credentials for auth_relay plain and none  #2327

### LINKED ISSUE #2327: Problem with login.json and authentication relay
**Describe Your Environment**
- Version of ZoneMinder: 1.32.2
- Built from source `do_debian.sh`
- Ubuntu 18

**Describe the bug**

Users who switched to the new version of zmNinja that uses `login.json` are complaining streaming breaks. In all cases, it is because they have AUTH_RELAY set to "none". One user told me today it was recommended they set it to "none" to fix some console error (not sure about this)

There is a problem with how login.json works today. The summary is if you use `OPT_AUTH` and `AUTH_HASH_LOGINS` but set `AUTH_RELAY` to `none`, the credentials return `user=username` instead of `user=username&pass=password` or `auth=token`. This is wrong because streaming breaks. 

Suggested solutions:

- We append password to the `authRelay=none` situation too [here](https://github.com/ZoneMinder/zoneminder/blob/master/web/api/app/Controller/HostController.php#L78)

- However, I am not sure why we are using authRelay as a check at all. Should we not be simply using `AUTH_HASH_LOGINS` instead and if so, return `&auth=` irrespective of what `AUTH_RELAY` is set to? This API was modeled after the code in `userLogin` in PHP land but I'm not sure why its this way now.

pliablepixels:


knight-of-ni:
Merging to include in 1.32.3 release. As soon as we release, I plan to bump up to 1.33.0 development. That will be a good time to implement a more comprehensive fix.

PR #2335, opened 2018-12-06
strip port from HTTP_HOST

connortechnology:
HTTP_HOST will contain port if it is other than default.  So we need to strip it.

### DIFF
--- a/web/includes/Server.php
+++ b/web/includes/Server.php
@@ -52,7 +52,12 @@ class Server {
       return $this->{'Name'};
     }
     # Use HTTP_HOST instead of SERVER_NAME here for nginx compatiblity
-    return $_SERVER['HTTP_HOST'];
+    # ICON: Hi, I just met you, and this is crazy, but I need to strip port.
+    $host_with_port_maybe = $_SERVER['HTTP_HOST'];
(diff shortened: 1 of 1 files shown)


heegard:
I tried the change to Server.php, it didn't work.  I found a second instance where $_SERVER['HTTP_HOST'] is used.  I introduced a similar change there also.  That seemed to
fix the problem.

These are the changes:
55c55,60
<     return $_SERVER['HTTP_HOST'];
---
>     # return $_SERVER['HTTP_HOST'];
>      # ICON: Hi, I just met you, and this is crazy, but I need to strip port.
>     $host_with_port_maybe = $_SERVER['HTTP_HOST'];
>     $heres_my_host = preg_replace('/([^:]+)(:\d+)?/','${1}', $host_with_port_maybe);
>     #Just use it baby
>     return $heres_my_host;
98c103,108
< 			$url .= $_SERVER['HTTP_HOST'];
---
> 			     # ICON: Hi, I just met you, and this is crazy, but I need to strip port.
> 			$host_with_port_maybe = $_SERVER['HTTP_HOST'];
> 			$heres_my_host = preg_replace('/([^:]+)(:\d+)?/','${1}', $host_with_port_maybe);
> 			#Just use it baby
> 			#$url .= $_SERVER['HTTP_HOST'];
> 			$url .= $heres_my_host;

PR #2337, opened 2018-12-07
allow one to set manpage destination

### LINKED ISSUE #2334: (FREEBSD) share/man is a wrong path for man files
To my surprise, ports must place man files to man instead share/man
Now we have path hardcoded in pod2man
I think:
add PATH to POD2MAN macro
pass 4th parameter here https://github.com/ZoneMinder/zoneminder/blob/10531df54312f52f0f32adec3d4720c063897b62/src/CMakeLists.txt#L24

Add some defaults, maybe CMAKE has something for this.
What do you think?

knight-of-ni:
fixes #2334 
Needs testing. Let's include this in the 1.33.x development.


cmake/Modules/Pod2Man.cmake
```diff
@@ -58,7 +58,7 @@ MACRO(POD2MAN PODFILE MANFILE SECTION)
 
 	INSTALL(
 		FILES ${CMAKE_CURRENT_BINARY_DIR}/${MANFILE}.${SECTION}.gz
-		DESTINATION share/man/man${SECTION}
+		DESTINATION ${MANPAGE_DEST_PREFIX}/man${SECTION}
     	)
 ENDMACRO(POD2MAN PODFILE MANFILE SECTION)
```
abishai: I believe we need to reflect macro changes here as well

abishai:
It definitely works, I believe CMake allows simple endmacro(), so probably my comment can be ignored. https://cmake.org/cmake/help/latest/command/macro.html#command:macro
Thanks!

cmake/Modules/Pod2Man.cmake
knight-of-ni: Agreed. This has been updated.

Issue #2339, opened 2018-12-09
(Reverse Proxies) URL generation troubles


abishai:
I think we really have some issues with reverse proxies.
Some non-critical urls are broken in web console (for example fps request).

There are 2 problems:
https://github.com/ZoneMinder/zoneminder/blob/master/web/includes/Server.php#L65
https://github.com/ZoneMinder/zoneminder/blob/master/web/includes/Server.php#L75

1. $_SERVER['HTTPS'] is not set and there are no way to set in in proxy configuration (it works as intended)
2. $_SERVER['SERVER_PORT'] is 80 (obviously)

To overcome this issue I set up prepend script for PHP-FPM
```
if(isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https'){
        $_SERVER['HTTPS']='on'; 
        $_SERVER['SERVER_PORT'] = '443';
}
```

Sure thing, the solution is awful and temporary.
We already have https://github.com/ZoneMinder/zoneminder/blob/17551eaceed33c821b0e3eb76676d01408f2ff51/web/includes/Monitor.php#L454 Probably, we can extend it to Server.php ?
And, we need port override.

connortechnology:
That line needs to be $url = $Server->UrlToApi().'/monitors/'.$this->{'Id'}.'.json';

This has been updated in master.

abishai:
On first page load for monitor I see a bunch of errors, for example Loading mixed (insecure) display content “http://xxxxxxxxxxxxx/cgi-bin/nph-zms?scale=50&width=1016px&height=768px&mode=jpeg&maxfps=30&buffer=1000&monitor=1&auth=dacc6765eb1f452876631d07628813da&connkey=34360&rand=1544556291” on a secure page
or http://xxxxxxxxxxxxxx/zm/index.php?view=request&request=status&entity=events&id=1&count=20&sort=Id%20desc

Stream is constantly restarting without mu hack with Detected streamWatch of type: stream stopped, restarting

abishai:
```
root@localhost [zm]> select * from servers;
Empty set (0.00 sec)
```
Should I register my server there and attach cameras to it ? If I understand Servers.php correctly, incorrect URL detection will be overrided in constructor.

connortechnology:
You can't mix http content on a https page.  All urls have to have https in front.  The outside world can't see your internal http reverse proxy.  They only know about the external https side.  

[...]

PR #2343, opened 2018-12-13
support for forwarded proto/port in Server.php

mrosack:
I've been having the same problems as in #2339, where I can't view streams behind a reverse proxy.  I've updated the Protocol and Port functions in Server.php to respect HTTP_X_FORWARDED_PROTO and HTTP_X_FORWARDED_PORT and everything works OK for me (although, it'd be nice if HTTP_X_FORWARDED_PORT wasn't necessary because it's rarely set in a reverse proxy situation).

### DIFF
--- a/web/includes/Server.php
+++ b/web/includes/Server.php
@@ -62,7 +62,12 @@ class Server {
     if ( isset($this->{'Protocol'}) and ( $this->{'Protocol'} != '' ) ) {
       return $this->{'Protocol'};
     }
-    return (!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] !== 'off' || $_SERVER['SERVER_PORT'] == 443) ? 'https' : 'http';
+
+    return  ( 
+              ( isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on' )
+              or
+              ( isset($_SERVER['HTTP_X_FORWARDED_PROTO']) and ( $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https' ) )
+            ) ? 'https' : 'http';
   }
 
   public function Port( $new = '' ) {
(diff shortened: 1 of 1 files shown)


vajonam:
I am still seeing these in the logs

foo.com is a internal only hostname that resolves properly the setup is the I am connecting to the zm which is running on http://foo.com:8080/zm  and the reverse proxying that via https. 

The end result is that I cannot see the current FPS or Force An Alarm. 

mootools-core.js:1 Mixed Content: The page at 'https://foo.com/zm/?view=watch&mid=1' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://foo.com:8180/zm/index.php?view=request&request=stream&connkey=354685&command=99'. This request has been blocked; the content must be served over HTTPS.

mrosack:
Is your proxy server setting HTTP_X_FORWARDED_PROTO and HTTP_X_FORWARDED_PORT?  From my nginx.conf:

```
location / {
			proxy_set_header	HOST $host;
			proxy_set_header 	X-Forwarded-Proto $scheme;
			proxy_set_header 	X-Forwarded-Port 443;
			proxy_set_header	X-Real-IP $remote_addr;
			proxy_set_header	X-Forwarded-For $proxy_add_x_forwarded_for;
			proxy_pass		http://zoneminder;
		}
```

vajonam:
No that doesn't work. Its reading both the http and the port number and sending it to the JS code to make hte AJAX call.. ZM is running on port 8180 on server, but being reverse proxied via port 443

vajonam:
Got it working using the following Apache directives

```
      <Location "/zm">
           ProxyHTMLEnable On
           ProxyHTMLExtended On
           ProxyPass http://foo.bar:8180/zm
           ProxyPassReverse /zm
           RequestHeader    unset  Accept-Encoding
           SetOutputFilter  proxy-html
           ProxyHTMLURLMap http://foo.bar:8180/zm /zm
           ProxyHTMLURLMap  /zm /zm
        </Location>
```

PR #2348, opened 2018-12-16
make shebangs universal

abishai:
Make shebangs BSD compatible (we have perl in /usr/local)


knight-of-ni:
Sorry. Both RPM and Debian distros completely disallow using env to guess the interpreter.

knight-of-ni:
Not to leave you hanging.... the correct way to solve the issue of a perl bin in an less common folder is to create a new make macro, maybe call it PERL_BINDIR, then replace the shebang in all the perl ".in" files with `#!@PERL_BINDIR@/perl -wT`

PR #2349, opened 2018-12-16
Added FreeBSD startup script

abishai:
FreeBSD rc.d startup script for ZoneMinder


abishai:
Accidently pulled shebangs

PR #2350, opened 2018-12-16
added FreeBSD startup script

abishai:
rc.d startup script for ZoneMinder

### DIFF
--- a/distros/freebsd/zoneminder.in
+++ b/distros/freebsd/zoneminder.in
@@ -1,0 +1,38 @@
+#!/bin/sh
+#
+# $FreeBSD$
+#
+# PROVIDE: zoneminder
+# REQUIRE: LOGIN mysql
+# KEYWORD: shutdown
+#
+# Add these lines to /etc/rc.conf.local or /etc/rc.conf
+# to enable this service:
+#
+# zoneminder_enable (bool):     Set to NO by default.
+#                               Set it to YES to enable zoneminder.
+
+. /etc/rc.subr
+
+name=zoneminder
+rcvar=zoneminder_enable
+
+load_rc_config $name
+
+: ${zoneminder_enable="NO"}
+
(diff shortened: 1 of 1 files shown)


knight-of-ni:
Good idea, but I am not sure if you intend for this to be a static file or to be configured by cmake. The ".in" extension implies it needs to be processed, but you have not defined a CMakeLists.txt to do the configuring.

If you intend this to be just a static file, then maybe we ought to put this under /misc.

abishai:
Ports framework uses .in files as well. However if you think the same can be achieved with CMake, I can try it.  
I need to place file this in ${STAGE_DIR}/etc/rc.d. So, I should create CMakeLists.txt with install(), right ?

knight-of-ni:
I think a good start would be to just configure the file with cmake, then install it with the ports framework.

To do that, you need to add `add_subdirectory(distros/freebsd)` to the main CMakeLists.txt file.
You can optionally add the subdirectoy only when ZM_TARGET_DISTRO equals FreeBSD. This will become required if we ever want cmake to install any FreBSD specific file(s) in question.

Create a new CMakeLists.txt file in dostros/freebsd. It should have content similar to this:
```
configure_file(zoneminder.in ${CMAKE_CURRENT_SOURCE_DIR}/zoneminder @ONLY)
```

Then update your zoneminder.in file to use make macros for the paths you are using for `command` and `pidfile`. There are existing make macros for the paths you are using. I just forget what they are at the moment.

Do you have any other files you want to include? Why just this one? I wonder if if makes sense to host all the packaging files and scripts you might use to build the port.

abishai:
Well, I have a readme as well :P Thanks for suggestions, I'll definitely look into existing cmake scripts to make this one and shebangs in proper way!

knight-of-ni:
If you have any questions, feel free to ask in our slack channel. I'm sure we can work this out.

PR #2356, opened 2018-12-20
Swresample

connortechnology:
in ubuntu 1810 libavresample is deprecated.  

This implements optional support for libswresample which is required for audio encoding from raw to aac


connortechnology:
Must have based it off storageareas.

PR #2376, opened 2019-01-01
Fixes #2375

### LINKED ISSUE #2375: Docs reference specific PHP version and file location on Ubuntu/Debian Installation instructions regarding the PHP config file
I noticed in the docs [here](https://zoneminder.readthedocs.io/en/stable/installationguide/ubuntu.html) and [here](https://zoneminder.readthedocs.io/en/stable/installationguide/debian.html), that in step 9 it references a specific version of PHP and a specific location for the php config file, but some might not have that version of PHP, or their PHP may be loading a different config file.

This might throw some people off, so to help them (in addition to having more correct and thorough documentation), this step in the documentation should also describe how to find the config file that PHP will actually be loading.

SalmonMode:
Fixes #2375 

I noticed this because my install of PHP was referencing a config file in `/etc/php/7.2/cli/`, instead of `/etc/php/7.0/apache2/`.


docs/installationguide/debian.rst
```diff
@@ -108,6 +108,21 @@ required apache modules.
 
 **Step 11:** Edit Timezone in PHP
 
+To find the config file that PHP will be using, you can run this
+
+::
+
+    php --ini | grep 'Configuration File'
```
connortechnology: The problem is that this will give you the path to the config file used when running from the command line, not the one that is in use under apache.

We need to look in /etc/apache2/mods-enabled to see which php module is loaded.  On mine we have php7.2.conf and php7.2.load so we our php config is /etc/php/7.2/apache2/php.ini

SalmonMode: Ah, interesting. I'm not actually super familiar with PHP and apache. This actually clarified a lot of things for me, thanks!

So if I understand correctly, one should always be able to expect the PHP config file for apache2 to be in an `apache2` folder inside `/etc/php/<version>/` or `/etc/php5/` (if version 5). So you could determine the version by checking `/etc/apache2/mods-enabled`, and then piece together where the config file should be, correct?

SalmonMode:
Changed per the convo in #2375. I just added a duplicate of the 16.04 guide and changed the Ubuntu version and where the PHP config file could be found.

docs/installationguide/debian.rst
connortechnology: This php --ini still gives you the path to the config file used when running php from the command line. NOT the one that is used by apache.

SalmonMode: Sorry, this file was supposed to be reverted to what it was originally. I think I see how I missed this. I'll make sure it gets reverted this time

SalmonMode:
Ok, definitely fixed that issue with the debian instructions. It should be all set now. Sorry about that :sweat_smile:

PR #2377, opened 2019-01-02
Fix 2253

connortechnology:
Add the default Server entry and use it when figuring out the path index.php


knight-of-ni:
dang it. you beat me to it.

PR #2378, opened 2019-01-02
Timezone check

connortechnology:
This is my old PR that adds a check for php timezone being different than the system time zone.

Feel free to not merge yet, we can discuss.  

If we put the timezone setting into the config, then we could move these checks to the options page and only do them there.

However we can always just set the timezone to the system time zone.


knight-of-ni:
Since the majority of our user base is running a Linux distro with systemd, we should probably check `tiemdatctl` first if the binary exists.
```
$ timedatectl
      Local time: Thu 2019-01-03 19:07:25 CST
  Universal time: Fri 2019-01-04 01:07:25 UTC
        RTC time: Fri 2019-01-04 01:07:25
       Time zone: America/Chicago (CST, -0600)
     NTP enabled: yes
NTP synchronized: yes
 RTC in local TZ: no
      DST active: no
 Last DST change: DST ended at
                  Sun 2018-11-04 01:59:59 CDT
                  Sun 2018-11-04 01:00:00 CST
 Next DST change: DST begins (the clock jumps one hour forward) at
                  Sun 2019-03-10 01:59:59 CST
                  Sun 2019-03-10 03:00:00 CDT
``` 

This PR does not appear to be compatible with *BSD. I am still looking for a solution. /etc/localtime exists but it is binary data. `date "+%Z"` gives the three character timezone. Perhaps we can convert that to the same format php expects.

While we are at it, we ought to try to either verify the timezone in mysql or set it for the session. Both seem to be possible. We know that if the timezone differs in the dB then some of the fields don't display the time correctly when running a filter. I forget if it is all the fields or just some.

PR #2385, opened 2019-01-04
Split actions.php into individual files per view

connortechnology:
I think this is good but have done only rudimentary testing.

It splits actions.php into individual files per view.  This enhances performance and is so much easier to understand when trying to figure out which bits are relevant to the current view.


knight-of-ni:
Here goes nothing.....

PR #2387, opened 2019-01-04
implement timezone check function

### LINKED ISSUE #2354: Changing timezone breaks just about everything
In order to submit a bug report, please populate the fields below. This is required.

**Describe Your Environment**
- Version of ZoneMinder: 1.32.3
- How you installed ZoneMinder: PPA
- Ubuntu Server 18.04 LTS

**Describe the bug**
I realized after completely setting up 1.32.3 that my OS time was not correct. This causes the video feed and events to show the wrong time. I used dpkg-reconfigure tzdata to set it to the correct timezone (America/New_York) and then rebooted into the BIOS to set the hardware clock to the correct time. After rebooting back into the OS, Zoneminder would not load any video feeds at all. 

Apparently this is a known issue with AUTH_RELAY and the system time being off by a few hours,
therefore, I disabled this feature so the feeds would work again but all previous events are now unplayable.  Now the feed itself shows the correct time and the Events show the correct time on my desktop browser but ZMNinja on my phone shows the same event is off by -5 hours and the Log file in Zoneminder is off by +5 hours. On top of that enabling AUTH_RELAY causes the feeds to stop working entirely. 

zmtzbug3
zmtzbug1
zmtzbug2

**To Reproduce**
1. Have an OS set to the wrong timezone, or change to another timezone
2. Use dpkg-reconfigure tzdata to set it to the correct time zone. 
3. Change the BIOS clock to match the timezone that you set. You can do this with the hwclock command as well. I don't think it makes a difference
4. Witness that enabling AUTH_RELAY breaks your streams, check your log and the time at the top be off by a number of hours. 
5. If you use ZMNinja, the event times will be off by a negative amount of hours instead of positive like the Log above.

**Expected behavior**
You should be able to change just ONE variable for time. I've spent 3+ hours trying to resolve what seems like such a simple issue at first only to come to the conclusion that I'll likely have to start from scratch and make damn sure the timezone and clock are set correctly before installing anything to the base OS. This would also make moving your server physically from one timezone to another a true PITA.

knight-of-ni:
fixes #2354 
supersedes #2378 

This implements a timezone check function that compares the system timezone to php and mysql timezones. It does this in a way that is compatible with any Linux or BSD type system without looking at specific file paths on the filesystem. 

If a discrepancy is found between the three timezones, an appropriate error is logged to alert the end user to fix the problem.

This is meant as a short term solution to #2354. A redesign of how times are stored in ZoneMinder is planned for a future release.

### DIFF
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -2276,6 +2276,9 @@ function check_timezone() {
   if ( $sys_tzoffset != $mysql_tzoffset )
         Error("ZoneMinder is not installed properly: mysql's timezone does not match the system timezone! Event lists will display incorrect times.");
 
+  if (!ini_get('date.timezone') || !date_default_timezone_set(ini_get('date.timezone')))
+    Fatal( "ZoneMinder is not installed properly: php's date.timezone is not set to a valid timezone" );
+
 }
 
 function unparse_url($parsed_url, $substitutions = array() ) { 
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -2261,7 +2261,6 @@ function csrf_startup() {
 function check_timezone() {
   $now = new DateTime();
 
-  # Probably need to check the results of these for errors
   $sys_tzoffset = trim(shell_exec('date "+%z"'));
   $php_tzoffset = trim($now->format('O'));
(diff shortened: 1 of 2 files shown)


connortechnology:
Merging but I think we will want to move this out to options or login or console so that it isn't being done on every page hit.

knight-of-ni:
Sounds like a good idea. Thanks for merging. I can look at this when I am back in town.

zd59:
I'm from Slovenia.
I use English on my computers.
Please take a look at this post:
[https://forums.zoneminder.com/viewtopic.php?f=8&t=27858&p=108598&hilit=date+format#p108598](url)

A large part of technicians worldwide use English on computers and wish to set up date view by local standard.
Please correct date display as suggested in my post above.
Thank you.

PR #2401, opened 2019-01-08
add a logging callback to the libvlc camera

connortechnology:
This does the same thing as I did for ffmpeg.  It causes libvlc to output it's logging to the ZM logging functions so that they appear in the ZM log.


knight-of-ni:
So we will probably want a want to quite this, like we also did for ffmpeg, right?
Maybe we ought to tie this into the flag we created for ffmpeg, and rename it to something more generic.

connortechnology:
I wrote this so that I could try to figure out why the libvlc input wasn't working for me.  

After investigating the libvlc api more... this is always going to be a substandard input method.  I think we should just deprecate it.

PR #2410, opened 2019-01-13
Replace onclick inline event handlers for createPopup

mnoorenberghe:
I'm working on adding a CSP policy to disallow 'unsafe-inline' scripts via Content Security Policy to help mitigate potential XSS attacks. This is the first step in removing a popular usage of `onclick`. If there is agreement that disallowing unsafe-inline scripts is a worthy goal then I can continue with the other inline event handlers.


connortechnology:
I think we generally want to be as safe as possible.  My mind is always a little blown every time I come across a new XSS attack.

PR #2413, opened 2019-01-16
Add a CSP script-src policy with nonce-source and convert more inline event handlers

mnoorenberghe:
* Add a CSP script-src policy to mitigate XSS on some views
  * The nonce attribute is required for all inline `<script>` blocks for stricter blocking with CSP 2
  * For pages where all script violations are fixed, enforce the policy (blocking other scripts), otherwise only report issues to the browser's console.
* Converted some `onchange` handlers and more `onclick`

### DIFF
--- a/web/skins/classic/includes/export_functions.php
+++ b/web/skins/classic/includes/export_functions.php
@@ -263,12 +263,12 @@ function exportEventImages( $event, $exportDetail, $exportFrames, $myfilelist )
 <div id="imagevideo" align="center"></div>
 <br>
 <div align="center">
-<button onclick="stepbackward()">< Step</button><button 
-	id="btnrwd" onclick="rewind()" >Rwd</button><button
-	id="btnplay" onclick="playstop()">Stop</button><button
-	onclick="stepforward()">Step ></button><button
-	id="btnspeedup" onclick="speedup()">speedup</button><button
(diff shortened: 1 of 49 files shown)


web/skins/classic/includes/export_functions.php
```diff
@@ -263,12 +263,12 @@ function exportEventImages( $event, $exportDetail, $exportFrames, $myfilelist )
 <div id="imagevideo" align="center"></div>
 <br>
 <div align="center">
-<button onclick="stepbackward()">< Step</button><button 
-	id="btnrwd" onclick="rewind()" >Rwd</button><button
-	id="btnplay" onclick="playstop()">Stop</button><button
-	onclick="stepforward()">Step ></button><button
-	id="btnspeedup" onclick="speedup()">speedup</button><button
-	id="btnspeeddown" onclick="speeddown()">slowdown</button>
+<button data-on-click="stepbackward">< Step</button><button
```
connortechnology, line 266: While we are in here, should that < Step be &lt; Step?

connortechnology: ahem... I mean & lt ;

mnoorenberghe: Yes, definitely. I'll make a PR

PR #2414, opened 2019-01-17
Update Dahua.pm 

kobold81:
Hello,

I ran into trouble using the provided Dahua.pm because it simply did not work out of the box. I tried to use bobylapointe69300's code but no chance either. Having not even beginner knowledge of perl, I simply copied relevant methods in the Dahua.pm and who would have guessed - ptz control works like a charm.

I would be happy if you could include this into the zoneminder release. I do not care to be named, I just want to make using this great application easier for normal guys.

Best regards 
Marc

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Control/Dahua.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Control/Dahua.pm
@@ -1,3 +1,24 @@
+# ==========================================================================
+# 
+# CGI Control Protocol
+#
+# This module is using code from bobylapointe69300 
+# https://forums.zoneminder.com/viewtopic.php?f=9&t=27000&p=104601&hilit=dahua#p104601
+# 
+#Control Address field on Control tab should be like :
+#   Field 'Controll Address' in Monitor Options 
+#    -> contains:  admin:password@192.168.1.XX:80
+#    -> must not contain:  http:// or https://
+#    -> adjust with your own parameters (IP, port, username and password)
+#   Control Capabilities for 'Dahua' has 
+#   -> in field 'Protocol' the value: Dahua
+# 
+# This module works if URL of camera control contains '/cgi-bin/ptz.cgi?'
+# for commands see 'DAHUA HTTP API FOR IPC Version 1.67'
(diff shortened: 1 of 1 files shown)


knight-of-ni:
Please coordinate with @jof . He created the initial PR that brought this script into zoneminder, and it worked for him. Perhaps there is just a simple misunderstanding.

Note that the addition of control subroutines MoveConUpLeft, MoveConDownRight, etc requires one to modify the line of sql in db/zm_create.sql.in that enable these features. Otherwise they won't ever get used.

jof:
Hey, cool to see some continued contributions on this Control module.
For context, I only implemented the features that I needed for the particular camera and environment I was last working with. We only required relative motion commands and support for presets.

Seems like the major changes in this patch are:
* Switching to only supporting http instead of both http and https. Seems like maybe an oversight? The `?` coming after the `s` in `https` should make the `s` character optional there for the group capture.
* Adding support for continuous motion commands. These didn't make sense for my environment, and I just didn't add them. However, my momentary motion commands are essentially just sending the continuous command, quickly followed by a stop command. So, I think what you have here really ought to work.
However, unfortunately, I don't have access to this Dahua camera anymore, so I can't really confirm that your changes would work on the platform I last developed this against.

I think I can only recommend just leaving in the `https?`, keeping the new motion commands, and then also updating the SQL schema so that the UI actually sends the right PTZ control commands to the backend.

kobold81:
Hi,

you are completely right, it should have been https?. Due to my lack of knowledge in both perl and zoneminder, I was fuzzy testing to get my cam working and cobbled together a version that would work with my camera. My objective was just to get to do the job with ptz. There is also the string 'test' somewhere, that should be deleted also.

Basically, I just added some methods to make the script work with the settings in zoneminder, even if they are not configured correctly. 

Feel free to adjust the code as you are far more knowledgeable than I am.

cnighswonger:
@jof I'm working on a few enhancements, etc. to the Dahua.pm control module. I am testing on a Dahua SD49225TN-HN. I'll be glad to incorporate the work by @kobold81 into my work and test it. If everything works out, I'll be submitting a pull request in the next few weeks.

connortechnology:
Superseded be #2552

connortechnology:
Superseeded by #2552

PR #2418, opened 2019-01-18
Fix duplicate 'class' attribute in options

mnoorenberghe:
I actually think this looks worse visually but it seems like this was the intended markup. If you prefer I can delete the small/medium/large classes


connortechnology:
In general I hate the bootstrap way of specify layout in the classes.  I think I would prefer to remove size based classes and go for more semantic classes.

PR #2429, opened 2019-01-20
Replace MooTools usage for adding window event listeners

mnoorenberghe:
It's not necessary for modern browsers and adds indirection/noise in the browser developer tools when inspecting event listeners.


SteveGilvarry:
Was just about to ask if it could go and what do I see already here a PR to
remove it.

PR #2430, opened 2019-01-20
Upgrade jQuery version

mnoorenberghe:
According to the release notes there were no breaking changes. This is that latest 1.x version.

If we stop support for IE8 we can upgrade to new versions.


mnoorenberghe:
After some discussion with the team I also added a 2nd commit to drop IE8 support and upgrade jQuery to 2.2.4   
 
* 2.2.4 is compatible with 1.12.4
* This fixes a CSP violation on every page load due to jQuery testing of focusin support with a hidden element.

PR #2438, opened 2019-01-23
Filter improvements

connortechnology:
The new actions break up caused filter execution to not work. This fixes that.

### DIFF
--- a/web/includes/actions/filter.php
+++ b/web/includes/actions/filter.php
@@ -19,63 +19,64 @@
 //
 
 // Event scope actions, view permissions only required
-if ( canView('Events') ) {
+if ( !canView('Events') ) {
+  Warning('You do not have permission to view Events.');
+  return;
+}
 
-  if ( isset($_REQUEST['object']) and ( $_REQUEST['object'] == 'filter' ) ) {
-    if ( $action == 'addterm' ) {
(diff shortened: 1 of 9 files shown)


web/includes/actions/filter.php
```diff
@@ -19,63 +19,64 @@
...
-          executeFilter( $tempFilterName );
-        }
+      if ( $_REQUEST['Id'] and ( $action == 'Save' ) ) {
+        dbQuery('UPDATE Filters SET '.$sql.' WHERE Id=?', array($_REQUEST['Id']));
+      } else {
+        dbQuery('INSERT INTO Filters SET'.$sql);
+        $_REQUEST['Id'] = dbInsertId();
+      }
+      if ( $action == 'execute' ) {
+        executeFilter($_REQUEST['Id']);
```
phillijw, line 74: I think this may have broke filters. When I save a filter, it doesn't send an ID across so this seems like it could cause a problem. See #2498

connortechnology: not having an Id means that it acts like Save As.  Which should create a new Filter.

phillijw: Good point. And the action isn't execute in that case.

Issue #2478, opened 2019-01-25
Stack based buffer Overflow - /src/zm_user.cpp (zmu)


Loginsoft-Research:
**Describe Your Environment**
- ZoneMinder v1.33.1
- Installed from - ppa:iconnor/zoneminder-master 

**Describe the bug**
A Stack overflows occur when variable size data is copied into fixed length buffers located on the program stack without any bounds checking. Vulnerabilities of this class are generally considered to be of high severity since their exploitation would mostly permit arbitrary code execution or Denial of Service.

**To Reproduce**
  _Affected Binary_ - zmu

An attacker can exploit the buffer by smashing the stack and modifying the return address of the function. This can be used to call some other function, like pointing the return address to some custom shellcode, injected into the stack. 

_In function_: `zmLoadUser()`

_Vulnerable function_: `mysql_real_escape_string()`

_Detail_: The vulnerability exists in function `zmLoadUser()`, in `zm_user.cpp`, while authenticating the user. The vulnerability exists in the login functionality. Once a username & password is supplied to the zmu binary, the username & password is passed through `mysql_real_escape_string()` function in order to produce an escaped SQL string. Due to absense of any protection and limitation placed to the length of username & password, there exists a stack based buffer overflow. 

_Vulnerable code -_ 
```
char safer_username[65]; 
mysql_real_escape_string(&dbconn, safer_username, username, strlen( username ) ); 
```

```
char safer_password[129]; 
mysql_real_escape_string(&dbconn, safer_password, password, strlen( password ) ); 
```

<img width="742" alt="stack" src="https://user-images.githubusercontent.com/47003572/51722837-c84c7c00-207c-11e9-9cff-76a3a1a102e4.png">

<img width="673" alt="stack 2" src="https://user-images.githubusercontent.com/47003572/51722836-c7b3e580-207c-11e9-9216-2bd43f5e1568.png">

_Setup_ - 
We need to compile the binary from source, using custom flags, inorder to trigger & debug the vulnerability. 

_Steps to compile_ - 
- mkdir build (In zoneminder directory)
- cd build 
- cmake .. -DCMAKE_C_FLAGS="-fsanitize=address -g -O0 -fno-stack-protector" -DCMAKE_CXX_FLAGS="-fsanitize=address -g -O0 -fno-stack-protector" 
- cd src/

_Command to run_ - ./zmu -U $(python -c "print 'a'*645") -P admin

_Root Cause Analysis_ - 

[...]

PR #2482, opened 2019-01-25
Fixes 2478 Remove overrun possibility

SteveGilvarry:
Safer_username and safer_login should be based on the username and login (lengths * 2)+1. Control input lengths at user input


connortechnology:
Fixed #2478 ?

setharnold:
Does this introduce a memory leak? Do you care?

Thanks

connortechnology:
It did, we fixed it in a subsequent commit.

PR #2484, opened 2019-01-25
Validate zmu Username and Password lengths

SteveGilvarry:
Ensure user provided values are not larger than allowed and error if
they are, therefore further preventing overflow.


connortechnology:
This code needs to exist in zms as well.  Perhaps we should move it to zm_user and make a function called like check_username and check_password or something.

SteveGilvarry:
Ok I will look tomorrow

Issue #2477, opened 2019-01-25
Sessions last too long and persist browser restart


Loginsoft-Research:
**Describe Your Environment**
- ZoneMinder v1.33.1
- Installed from - ppa:iconnor/zoneminder-master 

**Describe the bug**
When user logged-in and stays idle, the session should be destroyed after a certain period of time. Also the reusability of old session is possible, even after we close the application tabs in browser, which can cause an unauthorized access to the application.

**To Reproduce**
  Affected URL :
http://localhost/zm/index.php

- Log into Zoneminder as any user. 
- Leave the session idle for longer period, the session never gets invalidated.

An attacker can take advent of the weakness and gain unauthorized access to the web application without providing any credentials.

<img width="581" alt="session" src="https://user-images.githubusercontent.com/47003572/51722616-02694e00-207c-11e9-8944-821e09f39d48.png">

**Expected behavior**
 - Always specify session expiration date, make sure it is not to long and that the user can reset the session using the log off functionality of the application 

**Debug Logs**
```

None

```

mnoorenberghe:
This is normal for modern web-apps e.g. Facebook keeps you logged in for months at a time. The administrator can control the session lifetime via PHP configuration and users can manually log out from our UI so I would consider this report invalid personally. I'll let others chime in.

Sjord:
> An attacker can take advent of the weakness and gain unauthorized access to the web application without providing any credentials.

I don't think this is as straightforward as this sentence implies. Even if sessions last forever it is pretty hard to obtain a valid session token.

I would find a session timeout of several days acceptable. I think PHP's gc_maxlifetime makes sure that sessions don't last forever, but I am not sure on the details of how that works.

mnoorenberghe:
This was fixed by session length UI in #2487

PR #2487, opened 2019-01-30
Improve session

connortechnology:
Addresses #2477 I believe.  I might want to do more work on this. Presenting the PR so that we can discuss it.


mnoorenberghe:
Thanks for working on this. I cherry-picked c9032d3cb41d3aef0fd21dc0bd47302816788c66 since it was great on its own. I also accidentally conflicted a bit with the `PHP_SELF` removals.

I think it would be helpful for me to review if we have a minimal patch for #2477, maybe less refactoring and not adding the session length config. Personally I'd rather we try to avoid adding more complexity to the front-end as it's not in a very maintainable state.

PR #2508, opened 2019-02-10
Set CSRF on as the default for new installs. Fixes #2507

### LINKED ISSUE #2507: Enable CSRF protection by default
I've been running with it on for years and never have problems, even using the API from HASS and zmNinja. We really should have it enabled by default to go with our other security fixes. I would suggest enabling it for everyone by default new installs and upgrades, but I would also be fine with just enabling it for new installs as a starting point to not potentially break compatibility. I think it's possible that the other security fixes will already break some compatibility so personally I'd rather do it at the same time and call this stuff out in release notes.

@SteveGilvarry, you were looking for things to help with. Maybe you're familiar with DB migrations / changing defaults?

SteveGilvarry:
As discussed by @mnoorenberghe, to set CSRF on by default. Not sure we can impact config on existing installations. We probably need to call it out in the release notes to get people to turn it on.

Please discuss...


connortechnology:
I run all my servers with it on.  There have been issues, mostly due to out of ram due to the output buffering.  I have fixed most of these however.

PR #2517, opened 2019-02-11
Fix API SQL Injection

### LINKED ISSUE #2099: API SQL Injection
API is vulnerable to SQL injections via so called "named parameters" ( https://book.cakephp.org/2.0/en/development/routing.html#named-parameters ).

In the `EventsController::index()` and `MonitorsController::index()` methods (located in `web/api/app/Controller/EventsController.php` and `web/api/app/Controller/MonitorsController.php`), `$this->request->params['named']` is directly being used as conditions for `Model::find()` calls, and unless named parameters are explicitly whitelisted ( https://book.cakephp.org/2.0/en/development/routing.html#controlling-named-parameters ), arbitrary values can be passed, which means that it's possible to define the key side of the conditions array, which is where CakePHP accepts plain SQL.

Here's a basic example for testing purposes:

    /zm/api/events/index/Id = 1 OR 1=:1.json

This will result in an conditions array that looks like

    [
        'Id = 1 OR 1=' => 1
    ]

which will subsequently generate the following SQL:

    ... WHERE Id = 1 OR 1=1

Named parameters should either be whitelisted on router level, either per route ( https://book.cakephp.org/2.0/en/development/routing.html#per-route-named-parameters ), or globally via `Router::connectNamed()`, ie every possible column/operator combination (like `Id >=`, `Id <=`, `Id =`, etc...) needs to be registered, or they should be validated against whitelists manually accordingly.

The above example has been tested against the latest release and the master branch, using the release and the development docker files ( https://github.com/ZoneMinder/zmdockerfiles ).

mnoorenberghe:
Fixes #2099.

1st commit is ready for review, the 2nd just has TODOs to remind me which other places to fix if the approach is fine.

### DIFF
--- a/web/api/app/Controller/Component/FilterComponent.php
+++ b/web/api/app/Controller/Component/FilterComponent.php
@@ -1,14 +1,88 @@
 <?php
 App::uses('Component', 'Controller');
+
 class FilterComponent extends Component {
+
+	/**
+	 * Valid MySQL operands that can be used from namedParams with two operands
+	 * where the right-hand side (RHS) is a literal.
+	 * These came from https://dev.mysql.com/doc/refman/8.0/en/non-typed-operators.html
+	 */
+	public $twoOperandSQLOperands = array(
+					      'AND',
+					      '&&',
+					      '=',
+					      //':=',
+					      //'BETWEEN ... AND ...',
+					      //'BINARY',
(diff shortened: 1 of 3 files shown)


connortechnology:
Looks good to me.

PR #2519, opened 2019-02-11
don't unset Event as its used later. Also fix message

pliablepixels:


pliablepixels:
Close in favor of https://github.com/ZoneMinder/zoneminder/commit/5ce681b463dc15dc45fd5f4d26ae981a61fc4dc0

Issue #2528, opened 2019-02-17
cannot insert new monitors in master (or likely update them)


mitchcapper:
returns sql error due to double quoting.

PR #2529, opened 2019-02-17
don't quote dbEscape values it will quote it already

mitchcapper:
Breaks ability to insert new monitors (and likely update existing ones).


mitchcapper:
closes #2528

PR #2535, opened 2019-02-19
fix slack join link

pliablepixels:
Slack allows its own join portal. I created a permanent link we can use instead of the 3rd party heroku app which is slow


EuphoriaCelestial:
I want to join slack group too, where can I find link?

PR #2552, opened 2019-03-12
Dahua control improvements

cnighswonger:
This series of commits incorporates code provided by @kobold81 in pull req #2414 as well as a few new methods and misc cleanup of the code in general. This has been tested on a Dahua SD49225TN-HN PTZ camera.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Control/Dahua.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Control/Dahua.pm
@@ -182,9 +182,10 @@ sub sendPtzCommand
     my $self = shift;
     my $action = shift;
     my $command_code = shift;
-    my $arg1 = shift;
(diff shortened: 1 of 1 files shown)


scripts/ZoneMinder/lib/ZoneMinder/Control/Dahua.pm
```diff
@@ -39,16 +39,26 @@ sub AUTOLOAD
     my $class = ref($self) || croak( "$self not object" );
     my $name = $AUTOLOAD;
     $name =~ s/.*://;
+    ## This seems odd... if the method existed would we even be here?
+    ## https://perldoc.perl.org/perlsub.html#Autoloading
```
connortechnology, line 43: This is a hash lookup, not a function call lookup.  So basically if we have the value stored in us, return it.

PR #2553, opened 2019-03-12
Camera reboot function

cnighswonger:
Dahua cameras support a "Reboot" function. Other's may as well. This series of commits adds a reboot function option to the camera control configuration as well as a reboot button to the control UI.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Control/Dahua.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Control/Dahua.pm
@@ -39,6 +39,8 @@ sub AUTOLOAD
     my $class = ref($self) || croak( "$self not object" );
     my $name = $AUTOLOAD;
     $name =~ s/.*://;
+    ## This seems odd... if the method existed would we even be here?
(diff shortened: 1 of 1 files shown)


web/skins/classic/includes/control_functions.php
```diff
@@ -18,344 +18,389 @@
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 // 
 
-function getControlCommands( $monitor ) {
-  $cmds = array();
+function getControlCommands( $monitor )
+{
```
mnoorenberghe: There seems to be a lot of unrelated style changes in this file. Can you revert them?

cnighswonger: My apologies. Something is really screwed up with those commits. I'll revert them and recommit the fixes.

cnighswonger: Okay. This problem was due to my convoluted workflow. My working branch was based off of an older version of master and so effectively reverted a number of changes made in master since the older version. I've opened pull request #2554 for corrected patches.

Thanks for drawing my attention to a real mess!

PR #2554, opened 2019-03-13
Camera reboot function

cnighswonger:
Dahua cameras support a "Reboot" function. Other's may as well. This series of commits adds a reboot function option to the camera control configuration as well as a reboot button to the control UI.

### DIFF
--- a/db/zm_create.sql.in
+++ b/db/zm_create.sql.in
@@ -68,6 +68,7 @@ CREATE TABLE `Controls` (
   `CanWake` tinyint(3) unsigned NOT NULL default '0',
   `CanSleep` tinyint(3) unsigned NOT NULL default '0',
   `CanReset` tinyint(3) unsigned NOT NULL default '0',
+  `CanReboot` tinyint(3) unsigned NOT NULL default '0',
   `CanZoom` tinyint(3) unsigned NOT NULL default '0',
   `CanAutoZoom` tinyint(3) unsigned NOT NULL default '0',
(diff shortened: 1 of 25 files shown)


knight-of-ni:
After building master today, I see the following in my logs:
```
2019-03-30 10:20:03	web_php		16643	WAR	Unknown function call Monitor->CanReboot from /usr/share/zoneminder/www/skins/classic/includes/control_functions.php:359	includes/Monitor.php	279
```

Doing some initial troubleshooting, this PR appears to be missing a zm_update-1.33.x.sql file to insert the CanReboot column into an existing database. We also need to define CanReboot in Monitor.php, perhaps here: https://github.com/ZoneMinder/zoneminder/blob/master/web/includes/Monitor.php#L116

@cnighswonger would you mind send a new PR to correct this? Looks like we will need to bump the version of zm to 1.33.5 for existing systems to capture this change.

cnighswonger:
> @cnighswonger would you mind send a new PR to correct this? Looks like we will need to bump the version of zm to 1.33.5 for existing systems to capture this change.

I'll take care of this first of next week. Thanks!

connortechnology:
I already took care of it.

cnighswonger:
Thanks. I've had a few scheduling issues the past week or so which have kept me away from the dev box.

knight-of-ni:
I am hit or miss myself with a new job so I completely understand. We are always interested in people who wish to hang around and participate in the project.

PR #2557, opened 2019-03-19
Add camera relative iris control methods

cnighswonger:
This set of methods invoke realtive iris size in the direction indicated by
    the <Large/Small> portion of their name. They accept no arguments.

    NOTE:

    This only just does work. The Dahua API specifies "multiples" as the input.
    We pass in a 1 for that as it does not seem to matter what number (0-8) is
    provided, the camera iris behaves the same.


knight-of-ni:
Shouldn't this PR enable the canIris and canIrisRel fields in Dahua control shown here or has that been done already?
https://github.com/ZoneMinder/zoneminder/blob/master/db/zm_create.sql.in#L792

If you have already done this to your local system, simplest thing to do might be to use zmcamtool.pl to export the control from your system and add it to your open PR. Otherwise you go cross-eyed trying to line up which "0" needs to be a "1".

cnighswonger:
That should do it, I think.

Thanks!

knight-of-ni:
Thanks for following up.
On another note, last weekend I replaced my Netcat PTZ with a Dahua PTZ so I can help test this and improve on it. I'd really like to be able to change the pan/tilt sensitivity on the fly, but that requires creating new widgets.

cnighswonger:
I'm planning on adding controls to allow rate-of-movement type adjustments by the user. I've added a few things to our local installation but not this one yet. I plan to push it all up as I have time.

The Dahua camera I work with seems to loose its reference (axis origin, etc.) over time which is frustrating. [Here](https://ipcamtalk.com/threads/dahua-api-and-relative-camera-movement.37783/#post-348606) is some discussion I was having on the ipcamtalk forum about relative movement. Dahua wanted me to sign a NDA to get help from them on various issues like this. I told them I could not unless they wrote in an exception for OSS work. Never heard back from them.

cnighswonger:
And thanks for taking time to help with this. I've only one camera and limited time.

PR #2575, opened 2019-04-12
Add shutdown capability

connortechnology:
This PR adds a power icon at the top right that pops up and allows the user to either shutdown or restart the ZM server.   This feature would mostly be useful for ZM appliances, but might be of interest to others as well.

powerbutton

shutdownpopup
shutdownhelp

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
@@ -466,7 +466,10 @@ our @options = (
     name        =>  'ZM_SYSTEM_SHUTDOWN',
     default     =>  'true',
     description =>  'Allow Admin users to power off or restart the system from the ZoneMinder UI.',
-    help        =>  'The system will need to have a polkit file installed to enable the apache user
+    help        =>  'The system will need to have sudo installed and the following added to /etc/sudoers~~
+    ~~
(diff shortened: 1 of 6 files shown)


knight-of-ni:
We already have polkit to call functions that require elevated privs. Why not use that?
One just needs to write a polkit policy file and a rules file similar to the ones in /misc

knight-of-ni:
I've got my foot half way out the door, but before I do, I thought I'd mention some other ideas if you really want to stick with sudo:
- Make sudo a package requirement in our packaging scripts
- Add `@WEB_USER@ ALL=NOPASSWD:@SBINDDIR@/shutdown` to a file called `zm-sudo` and put it in our /misc folder
- Modify packaging scripts to put `zm-sudo` into `/etc/sudoers.d`

- FreeBSD will probably require some tweaks to this
- Redhat puts shutdown in `/usr/sbin` while Debian puts it in `/sbin` so we need to handle that

connortechnology:
I went with polkit initially but failed to get it to work.  Agreed with the rest of your suggestions, however I would not install the sudoers file by default.  I don't think giving ZM power to shutdown the server it is running is in most people's interests.  So I think manually enabling it is the way to go.

knight-of-ni:
Can you at least fix the hard-coded path to `/sbin/shutdown`? That will only work on a Debian distro.

PR #2587, opened 2019-04-24
Use GREATEST function prevent negative values in event counts

connortechnology:
So yet another attempt to prevent negative values in the event counts columns.

The idea is that we use coalesce to give NULL a 0 value, but if a zero occurred, then we could end up with a negative count (Typically this would happen on a running system during an update).

So I have added the use of GREATEST to ensure that the counts never go below zero.

I also discovered a + that should have been a minus.


knight-of-ni:
Wow, nice find. I will take a closer look at this when I get to my hotel this evening.

PR #2591, opened 2019-04-30
Netcat ONVIF: adding ONVIF authentication

redaco:


knight-of-ni:
Thank you for maintaining backwards compatibility with the control address field

PR #2592, opened 2019-04-30
Migrate to Bcrypt, Add JWT

pliablepixels:
This PR will do two things:

1) Migrate from mysql password scheme to bcrypt for each user that logs into ZM (one time only)
2) **TBD** :Also return a JWT token for login as an added feature

As of now, this PR only does 1. I'd like folks to comment on portability issues/suitability before I do 2. This is all new to me.

Notes:
1. bcrypt and JWT are not related, but I think its a good idea to move to bcrypt anyway
2. Opens the door to migration to other hashing mechanisms in the future

Limitations:
- Bcrypt migration only if PHP >= 5.5. There is a compat library we can use, but do we need to bend back for < 5.5 or just let them continue with mysql
- Assumes a min of MySQL 4.1 (as mysqls hashing was diff. before that). If someone is using mysql < 4.1, this PR will fail on auth. But MySQL 4.1 was released in 2004. I seriously doubt it is supported.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -157,3 +157,4 @@ web/undef.log
 zm.conf
 zmconfgen.pl
 zmlinkcontent.sh
+externals/
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,6 +5,6 @@
 [submodule "web/api/app/Plugin/CakePHP-Enum-Behavior"]
 	path = web/api/app/Plugin/CakePHP-Enum-Behavior
 	url = https://github.com/ZoneMinder/CakePHP-Enum-Behavior.git
-[submodule "externals/bcrypt"]
-	path = externals/bcrypt
+[submodule "third_party/bcrypt"]
+	path = third_party/bcrypt
 	url = https://github.com/trusch/libbcrypt
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,3 +5,6 @@
 [submodule "web/api/app/Plugin/CakePHP-Enum-Behavior"]
 	path = web/api/app/Plugin/CakePHP-Enum-Behavior
 	url = https://github.com/ZoneMinder/CakePHP-Enum-Behavior.git
+[submodule "externals/bcrypt"]
+	path = externals/bcrypt
+	url = https://github.com/trusch/libbcrypt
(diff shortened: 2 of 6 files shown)


knight-of-ni:
Have you considered using the crypt function for php < 5.5? It says password_hash and crypt are backwards compatible with one another.

pliablepixels:
> Have you considered using the crypt function for php < 5.5? It says password_hash and crypt are backwards compatible with one another.

Yes - there is a long SO thread on the pros and cons of that. It seems the compat lib is recommended over `crypt()` for < 5.5 for some issues with its implementation, but I haven't researched yet. Will do so towards the end.

knight-of-ni:
Travis is failing the building because the subfolder "externals" does not exist in the PR.

```
$ git submodule update --init --recursive
fatal: No url found for submodule path 'externals/bcrypt' in .gitmodules
The command "eval git submodule update --init --recursive " failed. Retrying, 2 of 3.
```

However, @connortechnology and I have been discussing getting rid of submodules altogether. 

If we do go the submodule route, we need to link a specific commit so we can have some kind of revision control (I noticed the bcrypt project does not have any releases).

pliablepixels:
Right, I've changed "externals" to "third_party" - not sure why it does not show up, but I'll fix it. There will be a lot of wrong turns as I work this out.

That being said, this is going to be more work than I had originally anticipated but I'll keep plowing through. When its ready to test, I'll take off the not ready to merge.

In the mean time, I did want opinions on some approaches. We can discuss on slack as well:

1) My motivation to get rid of MYSQL password is that it uses SHA1 which is [insecure](https://shattered.io). Bcrypt seems to be the one to use now (with argon2 showing promise). The other motivation is if we move the hashing to the application layer, we get to decide how to evolve our security ourselves. This comes at a slight cost of performance I am told, as MySQL PASSWORD may be just a little faster. I still think it's worth it. Do you agree/disagree?

2) After I get bcrypt in, I plan to migrate AUTH_HASH to JWT (if possible). AUTH_HASH uses MD5 which is even worse (even though it uses a timestamp hour parameter to it). Do you have opinions on this? Should AUTH_HASH be left alone and we create a separate JWT token? I can't think of why we should do that.

3) On your comment on deprecating submodules, that's fine with me. What would you prefer? I use Cmake  [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html )?

pliablepixels:
I'm closing this. Will do separate PRs for crypt and JWT

PR #2594, opened 2019-05-02
FIFO support for zoneminder zone debugging

mitchcapper:
Adds fifo options for diagnostic images for much lower impact diagnostics mode.  Diagnostic images are only written when there is a client listening for them (otherwise they are skipped).  Also added a json stream for the detection data so you can see in real time the pixels or blobs detected for the motion.  This allows for easy real time stream of both delta and reference images (as video streams) along with the detection numbers.

Once you turn on `RECORD_DIAG_IMAGES` and the new `RECORD_DIAG_IMAGES_FIFO` in the logging options you can then use 3 new remote stream urls:

The delta or reference images as a MJPEG stream (great to see where it is seeing the motion!):
	cgi-bin/nph-zms?mode=jpeg&bitrate=2&buffer=0&connkey=185787&source=fifo&format=delta&monitor=1&maxfps=5
	cgi-bin/nph-zms?mode=jpeg&bitrate=2&buffer=0&connkey=185787&source=fifo&format=reference&monitor=1&maxfps=5

text json raw stream:
	cgi-bin/nph-zms?&buffer=0&connkey=185787&source=fifo&format=raw&monitor=1

This will output a text stream like:
{"zone":5,"type":"ALRM","pixels":778661,"avg_diff":50}
{"zone":5,"type":"FILT","pixels":762704}
{"zone":5,"type":"RBLB","pixels":728102,"blobs":5}
{"zone":5,"type":"FBLB","pixels":728021,"blobs":2}
{"zone":6,"type":"ALRM","pixels":130844,"avg_diff":44}
{"zone":6,"type":"FILT","pixels":128608}
{"zone":6,"type":"RBLB","pixels":86679,"blobs":5}
{"zone":6,"type":"FBLB","pixels":84206,"blobs":1}
{"zone":5,"type":"ALRM","pixels":977,"avg_diff":29}
{"zone":6,"type":"ALRM","pixels":493022,"avg_diff":47}
{"zone":6,"type":"FILT","pixels":480465}
{"zone":6,"type":"RBLB","pixels":455631,"blobs":10}
{"zone":6,"type":"FBLB","pixels":454611,"blobs":1}
{"zone":5,"type":"ALRM","pixels":689,"avg_diff":29}
{"zone":6,"type":"ALRM","pixels":568555,"avg_diff":40}
{"zone":6,"type":"FILT","pixels":555893}
{"zone":6,"type":"RBLB","pixels":538928,"blobs":13}
{"zone":6,"type":"FBLB","pixels":519478,"blobs":1}
{"zone":5,"type":"ALRM","pixels":174931,"avg_diff":55}
{"zone":5,"type":"FILT","pixels":167048}
{"zone":5,"type":"RBLB","pixels":72633,"blobs":7}
{"zone":5,"type":"FBLB","pixels":0,"blobs":0}
{"zone":6,"type":"ALRM","pixels":501729,"avg_diff":40}
{"zone":6,"type":"FILT","pixels":491671}
{"zone":6,"type":"RBLB","pixels":401359,"blobs":7}
{"zone":6,"type":"FBLB","pixels":397845,"blobs":1}
{"zone":5,"type":"ALRM","pixels":315746,"avg_diff":39}
{"zone":5,"type":"FILT","pixels":310313}
{"zone":5,"type":"RBLB","pixels":307017,"blobs":3}

There are four types of events right now: Alarm, Filter, Raw Blob and Filtered Blobs that correspond to those stages of analysis.  It will show the number of pixels detected (along with average pixel difference against the threshold) and number of blobs at each stage.

All of this can allow for some great client improvements like: http://mitchcapper.com/sample_fifo_streaming.png
That is showing the actual video stream on the left, and the diagnostic streams on the right (delta stream on top reference on the bottom). Below it shows in real time the analysis data.

This code is actually quite old, going back to my initial port in 2010 and updating it to MTK's branch in 2012 (https://github.com/mitchcapper/ZoneMinder-old/compare/master...mitchcapper:fifo_diag)  I updated it to our current master and added json as a less binary format.  

Assuming it gets merged in I will submit a second PR for documentation.


pliablepixels:
I don't see any UI integration in this PR. This is very useful for me. Do you plan to put in a UI integration?

mitchcapper:
It would be pretty easy to add an option to the browser UI when viewing a monitor to switch to one of the two diagnostic streams.   Note they are live only so you couldn't pause them.   The debug log data would be a larger UI change.     I am using it in a WPF application which I might release but adding to the main UI I am not sure if I would undertake.

As for request #2)
This patch primarily is to make diagnostic output easier to connect to and far lighter weight than before (especially when not actively connected to it).  While it would be possible to save diagnostic data for recorded events it would have a fairly decent performance hit and be pretty unrelated to this patch itself (minus maybe storing the diag log stream).   I would recommend if you want to see what happened on a recorded event you just create a test monitor and point it to the video file of the recorded event.   ZM will then effectively re-analyze that events video and you can see what happened in real time.

pliablepixels:
ok thx

pliablepixels:
@mitchcapper I am updating documentation to put in details of FIFO based on this PR comments. Is it possible for you to share and image that shows these streams that I can put into the doc? http://mitchcapper.com/sample_fifo_streaming.png no longer works.

pliablepixels:
actually never mind. I created a gif myself.

vajonam:
@pliablepixels / @mitchcapper  when using a NFS based storage area this I get a problem when trying to diagnose this.

mitchcapper:
@vajonam that combination may not be the best for fifo's.  Can you try doing a CLI fifo from inside the docker container and see if it works correctly? I  did see:

> NFS was designed as a multi-OS, lowest common denominator filesystem. As such, it does not support full Unix file system semantics. In particular, FIFOs/named pipes, if available at all, will not be shared across systems. (Two processes on the same host may be able to communicate via an NFS FIFO, though I would not recommend doing this).

Do you plan to use debug diag in fifo mode? The other option is to disable it and just have traditional debug output.

vajonam:
@mitchcapper  I did create a "diag" storage location pointed it to /tmp/, for that source and the diag worked fine!. 

Maybe a FR maybe we can specify the diag/fifo location to go somewhere other than the default storage location?

mitchcapper:
Yeah there is no reason for them to have to be next to the normal files other than was the obvious choice.  They don't take space up so something like /tmp has no real downside.  It is unclear if the failure is due to an NFS setting, kernel version, docker related, or docker capability privilege missing so it could be potential to get it working with current location as well.

vajonam:
Happy to help you debug if you need. I suspect its NFS related more than anything. 

auto mount NFS 

```
s_events                -rw,soft,rsize=98304,wsize=98304,async  server:/zfs/sahara/media/Storage/Security
```

```
    zm:
        image: quantumobject/docker-zoneminder:1.34
        container_name: zm
        environment:
          - PGID=2500
          - PUID=1006
          - ZM_DB_HOST=zmdbhost
          - ZM_DB_USER=zmuser
          - ZM_DB_PASS=zmpass
        depends_on:
          - mariadb
        privileged: true
        ports:
          - 8180:80
          - 6802:6802
          - 9000:9000
        shm_size: 4gb
        volumes:
          - /opt/docker/configs/zoneminder/zm:/etc/zm
          - /opt/docker/configs/zoneminder/etc/crontab:/etc/crontab
          - /opt/docker/configs/zoneminder/tokens.txt:/etc/private/tokens.txt
          - /opt/docker/configs/zoneminder/limits.conf:/etc/security/limits.conf
          - /net/sahara/Storage/Security:/var/cache/zoneminder:shared
          - /net:/net:shared
          - /opt/docker/configs/zoneminder/logs:/var/log/zm/
          - type: tmpfs
            target: /dev/shm
        restart: unless-stopped
```

vajonam:
@mitchcapper do you recommend I open a FR? my c/c++ skills aren't good enough to be able to provide a solid PR quickly.

mitchcapper:
While the exact cause is unclear I don't see a downside to moving these to /tmp I will put ta PR in the next week or two to do so.

connortechnology:
I would tend to think that it should go wherever ZM_SOCKDIR is, which in ubuntu is /run/zm/ which would be tmpfs.

vajonam:
@mitchcapper can you confirm this is correct url for the raw json.

`https://portal/zm/cgi-bin/nph-zms?&buffer=0&source=fifo&format=raw&monitor=1&<auth>` is that missing a mode=? or something like that? I cannot get this work.

vajonam:
@mitchcapper Looks like @connortechnology's latest commits moved these to ZM_SOCKDIR and it seems to be working for the most part, delta and reference are working.

however when trying to get raw stream, I get 
```
Flocking problem on /var/run/zm/dbgpipe-1.log.rlock: - Resource temporarily unavailable
```
when I cat the /var/zm/dbgpipe-1.log I see the json spewing out fine. any. These are not longer on NFS shares, they are simple docker volumes, which is just a mergeFS.

mitchcapper:
That is odd, so the lock file is just a normal file created and then locked with LOCK_EX.  you can try using the flock command line to see if it works on the tmpfs directory (i assume that is what it is actually for you).

For what it is worth I run zoneminder in docker containers currently without issues with a mounted storage directory.

vajonam:
just ran a quick test. not really used flock before, but I assume this is how its used? 

seems to run okay. 

```
root@6047ba4b8316:/var/run/zm# touch lock.test ; flock -x lock.test echo sleeping ;sleep 5s;
sleeping
```

vajonam:
however this doesn't work

```
root@6047ba4b8316:/var/run/zm# cat dbgpipe-1.log.rlock 
root@6047ba4b8316:/var/run/zm# flock -x dbgpipe-1.log.rlock 
flock: bad file descriptor: 'dbgpipe-1.log.rlock'
```

vajonam:
strange however the `diagpipe-d-1.jpg.rlock` seems to behave the same way but I can still see the stream via nph-zms

```
root@6047ba4b8316:/var/run/zm# flock -s diagpipe-d-1.jpg.rlock
flock: bad file descriptor: 'diagpipe-d-1.jpg.rlock'

```

vajonam:
sorry I should add this is not on tmpfs just regular fs inside a container.. whatever that is.

Issue #1235, opened 2016-01-08
Replace MD5 hash function


knight-of-ni:
ZoneMinder uses MD5 for its authentication. Problem is that it is known to be compromised.
I don't recall what the recommended hash function should be. We simply need to find what the library is, add support for it, and make that the default instead of other libraries that use md5 like openssl or gnutls.
##

kylejohnson:
We should be hashing and salting passwords.  For now, I say we can ignore the salt as these are local installs and aren't always publicly exposed (read: to save us a lot of time completely revamping the password functions).

As for the hash function, SHA256 is currently recommended.

knnniggett:
We also need to be careful with existing users when they upgrade to a version that uses the new hash. We don't want to lock them out of the web portal.

kylejohnson:
Something like this:

```php
$hashed_password =  hash('sha256', "$the_users_password");
```

connortechnology:
We will have to try to auth properly, then try to auth using old md5, and then auto-update their db.

Linwood-F:
This is perhaps too much history but... why doesn't Zoneminder just let the web server do authentication, and use zoneminder's database for authorization.  Stop rolling our own (even with libraries)?  At least that way when security patches are issued for the web, it is automatic here?  Or does that change too many things?

kylejohnson:
@Linwood-F You mean using something like auth_basic (or some other apache module)?

Linwood-F:
Yes.  I'm not engaged enough to know the risks of various ways choices, but it just strikes me that building it into Zoneminder, when (well, +/- the API) Zoneminder always lives behind a web server with authentication built in, seemed a bit redundant.

SteveGilvarry:
Previous related discussion https://github.com/ZoneMinder/ZoneMinder/pull/916

kylejohnson:
We'll want to verify that SHA256 is 'the best' for us right now.

As for not locking out previous users, I suggest we:

 *  Add a new column to the Users table called `hashing_scheme` with a default value of our current scheme (`md5`).
 *  When a user attempts to log in,  hash the password they submitted with whatever's in that user's `hashing_scheme`. 
  *  If the password was incorrect, fail.
  * If the password was correct, and this user's `hashing_scheme` value is different to the application's `current_hashing_scheme`, rehash the user's password with the new scheme, then update the `password` and `hashing_scheme` value in the database.

kylejohnson:
[...]

PR #2598, opened 2019-05-03
Replace MySQL Password() with bcrypt, allow for alternate JWT tokens

pliablepixels:
Addresses #2592 #1235 

This PR will address the following:
a) Offer a more secure alternative to our `auth=` algorithm, which uses MD5 and for some reason, also adds the user password to the key before hashing

b) Switch from MySQL PASSWORD to bcrypt
Motivation: SHA1 is considered unsafe. 
However  MYSQL uses `SHA1(raw(SHA1(password))` - I don't know how unsafe that is. I don't see any salt being used, which I don't think is a good idea. Folks who understand security better can comment.

c) Offer a JWT token based API access mechanism

 
**Tasks:**

General
- Demote logs after all features are complete ✅ 
- Need to discuss import strategy for third_party libs ❓ 
- New API documentation ✅ 
- Test against zmninja ✅ 
- Update EventServer to validate against bcrypt ✅ 

Bcrypt related:
-   When users login, migrate from MD5 to Bcrypt automatically ✅ 
-  Support Bcrypt for PHP >=5.3 <5.5 ✅ 
- Wrap all MD5 passwords with a  Bcrypt hash at first install  ✅ 

JWT related:
-  Allow API to use JWT tokens to login ✅ 
-  Implement both access and refresh tokens ✅ 
- Support JWT in C++ layer for zms and zmu ✅ 
- Add UI support to enable/disable APIs per user ✅ 
-  Add UI support to revoke tokens ✅ 

**Screenshot of UI change:**

Screen Shot 2019-05-13 at 3 03 44 PM


pliablepixels:
Umm, @connortechnology @SteveGilvarry @knight-of-ni  any of you have an idea why its complaining in the PR above? I built it locally and it works well. I even added all the flags it uses. My g++ version is 7.x (the travis VM seems to use 4.x of gcc, don't know about g++)

It seems to complain about a typecast issue (char */const char *) but also points to `openssl/evp.h` having that error 

```
cd "/build/zoneminder-1.33.8~20190505.26/dbuild/src" && ccache /usr/lib/ccache/x86_64-linux-gnu-g++   -DHAVE_SENDFILE4_SUPPORT=1 -g -O2 -fPIE -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2  -std=c++11 -Wall -D__STDC_CONSTANT_MACROS -O2 -I"/build/zoneminder-1.33.8~20190505.26/dbuild/src" -I"/build/zoneminder-1.33.8~20190505.26/src" -I"/build/zoneminder-1.33.8~20190505.26/dbuild" -I/usr/include/mysql -I"/build/zoneminder-1.33.8~20190505.26/src/../third_party/bcrypt/include/bcrypt" -I"/build/zoneminder-1.33.8~20190505.26/src/../third_party/jwt-cpp/include/jwt-cpp"    -o CMakeFiles/zm.dir/zm_timer.cpp.o -c "/build/zoneminder-1.33.8~20190505.26/src/zm_timer.cpp"
/usr/bin/cmake -E cmake_progress_report "/build/zoneminder-1.33.8~20190505.26/dbuild/CMakeFiles" 60
[ 85%] Building CXX object src/CMakeFiles/zm.dir/zm_user.cpp.o
cd "/build/zoneminder-1.33.8~20190505.26/dbuild/src" && ccache /usr/lib/ccache/x86_64-linux-gnu-g++   -DHAVE_SENDFILE4_SUPPORT=1 -g -O2 -fPIE -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2  -std=c++11 -Wall -D__STDC_CONSTANT_MACROS -O2 -I"/build/zoneminder-1.33.8~20190505.26/dbuild/src" -I"/build/zoneminder-1.33.8~20190505.26/src" -I"/build/zoneminder-1.33.8~20190505.26/dbuild" -I/usr/include/mysql -I"/build/zoneminder-1.33.8~20190505.26/src/../third_party/bcrypt/include/bcrypt" -I"/build/zoneminder-1.33.8~20190505.26/src/../third_party/jwt-cpp/include/jwt-cpp"    -o CMakeFiles/zm.dir/zm_user.cpp.o -c "/build/zoneminder-1.33.8~20190505.26/src/zm_user.cpp"
/usr/bin/cmake -E cmake_progress_report "/build/zoneminder-1.33.8~20190505.26/dbuild/CMakeFiles" 61
[ 86%] Building CXX object src/CMakeFiles/zm.dir/zm_utils.cpp.o
cd "/build/zoneminder-1.33.8~20190505.26/dbuild/src" && ccache /usr/lib/ccache/x86_64-linux-gnu-g++   -DHAVE_SENDFILE4_SUPPORT=1 -g -O2 -fPIE -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2  -std=c++11 -Wall -D__STDC_CONSTANT_MACROS -O2 -I"/build/zoneminder-1.33.8~20190505.26/dbuild/src" -I"/build/zoneminder-1.33.8~20190505.26/src" -I"/build/zoneminder-1.33.8~20190505.26/dbuild" -I/usr/include/mysql -I"/build/zoneminder-1.33.8~20190505.26/src/../third_party/bcrypt/include/bcrypt" -I"/build/zoneminder-1.33.8~20190505.26/src/../third_party/jwt-cpp/include/jwt-cpp"    -o CMakeFiles/zm.dir/zm_utils.cpp.o -c "/build/zoneminder-1.33.8~20190505.26/src/zm_utils.cpp"
In file included from /build/zoneminder-1.33.8~20190505.26/src/zm_crypt.h:27:0,
                 from /build/zoneminder-1.33.8~20190505.26/src/zm_user.cpp:32:
/build/zoneminder-1.33.8~20190505.26/src/../third_party/jwt-cpp/include/jwt-cpp/jwt.h: In function ‘std::string jwt::helper::extract_pubkey_from_cert(const string&, const string&)’:
/build/zoneminder-1.33.8~20190505.26/src/../third_party/jwt-cpp/include/jwt-cpp/jwt.h:77:87: error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]
    std::unique_ptr<BIO, decltype(&BIO_free_all)> certbio(BIO_new_mem_buf(certstr.data(), certstr.size()), BIO_free_all);
                                                                                       ^
In file included from /usr/include/openssl/evp.h:75:0,
                 from /build/zoneminder-1.33.8~20190505.26/src/../third_party/jwt-cpp/include/jwt-cpp/jwt.h:9,
                 from /build/zoneminder-1.33.8~20190505.26/src/zm_crypt.h:27,
                 from /build/zoneminder-1.33.8~20190505.26/src/zm_user.cpp:32:
/usr/include/openssl/bio.h:668:6: error:   initializing argument 1 of ‘BIO* BIO_new_mem_buf(void*, int)’ [-fpermissive]
 BIO *BIO_new_mem_buf(void *buf, int len);
      ^
```

pliablepixels:
@ZoneMinder/comitters, I'd like to open this PR up for comments. Only on functionality for now please (code reorg can come later) on how tokens are being used compared to what is the norm in this token approach.

Still not ready for merge, need to demote logs etc. Approach at a high level:

-  When you use `login.json` with user=&pass= it returns the existing auth hash (I decided not to remove it) but also returns two other tokens `access_token` and `refesh_token`. `access_token` lasts for 2 hrs or whatever the lifetime of AUTH_HASH_TTL is.
-  To get a new `access_token` you can do a `login.json?token=<refreshtoken>`. This will re-generate a new `access_token` Each `refresh_token` is valid for 24 hrs. The idea here is avoid sending `user=&pass=` over the wire. Using refresh tokens, you only need to pass u=&p= it once in 24 hrs. The only way to create a new `refresh token` is to login via username and password
-  zms now supports `&token=`. If will only accept the `access_token` for security purposes. refresh tokens can only be used with `login.json` to get a new `access_token`
- zmu now supports `-T` option with `access_token`
- he tokens will ignore `auth_hash_ip` - I don't think it should be enabled as most often consumers will be mobile (plus if you sign the token with remote IP, zmu will always fail in token auth mode, it won't know remote IP)
- This token mode is largely "stateless" - no PHP sessions are created so logging in via API tokens doesn't imply you also have a web interface token. Since zms/zmu/apis directly validate tokens, no need for one (or so I think)
- The core inner working of the ZM php layer still uses sessions. Haven't touched that. This token system is only meant for folks who use the API ecosystem
- If auth_token is "stolen" it will be valid for the duration of the token (AUTH_HASH_TTL). This is the only case where I see this to have an additional issue compared to the existing auth_hash. We can avoid this by signing the token with `remote_ip` if enabled, but the problem here is if you sign the token with `remote_ip`, zmu will always fail. I'm not sure this is important. I haven't seen a single token implementation that adds remote IP. The general commentary is `if a token is stolen, black list it` but that assumes you get to know it is stolen. We can black list it by simply upping the minimum time at which a token is generated. (haven't done this yet)
- If refresh_token is "stolen" its a bigger concern, because it can be used  to create new tokens  till its period of validity (24 hrs). The resolution is similar. If its stolen, black list it (same comments as access token). That being said "refresh_tokens" should be treated like "passwords". So I'd argue, this is is better than someone stealing your password.
- All tokens are validated against DB. However, DB validation is only to see if the user exists and if so, assign it latest permissions (Monitor/Events/etc). There is no concept of password checking. The token is sensitive data free (except for user name in the payload which anyone can view, which I don't consider sensitive)
- The "migrate all passwords to a bcrypt over-encoding" is TBD (not related to JWT) (edited) 


- Happy to hear concerns/thoughts/improvement suggestions

src/CMakeLists.txt
```diff
@@ -4,20 +4,27 @@
...
+set(ZM_BIN_SRC_FILES zm_box.cpp zm_buffer.cpp zm_camera.cpp zm_comms.cpp zm_config.cpp zm_coord.cpp zm_curl_camera.cpp zm.cpp zm_db.cpp zm_logger.cpp zm_event.cpp zm_frame.cpp zm_eventstream.cpp zm_exception.cpp zm_file_camera.cpp zm_ffmpeg_input.cpp zm_ffmpeg_camera.cpp zm_group.cpp zm_image.cpp zm_jpeg.cpp zm_libvlc_camera.cpp zm_local_camera.cpp zm_monitor.cpp zm_monitorstream.cpp zm_ffmpeg.cpp zm_mpeg.cpp zm_packet.cpp zm_packetqueue.cpp zm_poly.cpp zm_regexp.cpp zm_remote_camera.cpp zm_remote_camera_http.cpp zm_remote_camera_nvsocket.cpp zm_remote_camera_rtsp.cpp zm_rtp.cpp zm_rtp_ctrl.cpp zm_rtp_data.cpp zm_rtp_source.cpp zm_rtsp.cpp zm_rtsp_auth.cpp zm_sdp.cpp zm_signal.cpp zm_stream.cpp zm_swscale.cpp zm_thread.cpp zm_time.cpp zm_timer.cpp zm_user.cpp zm_utils.cpp zm_video.cpp zm_videostore.cpp zm_zone.cpp zm_storage.cpp zm_crypt.cpp)
+
+
 
 # A fix for cmake recompiling the source files for every target.
 add_library(zm STATIC ${ZM_BIN_SRC_FILES})
+link_directories(/home/pp/source/pp_ZoneMinder.git/third_party/bcrypt)
```
connortechnology: um.. this very specific path isn't going to exist anywhere else...

pliablepixels: oops typo will fix

src/zm_crypt.h
```diff
@@ -0,0 +1,32 @@
...
+// 
+
+#ifndef ZM_CRYPT_H
+#define ZM_CRYPT_H
+
+
+#include <string.h>
+#include <openssl/sha.h>
+#include "BCrypt.hpp"
+#include "jwt.h"
```
connortechnology: If these includes are only usined in the .cpp, then they should only go in the .cpp, else we slow down compilation unnecessarily.

pliablepixels: https://github.com/ZoneMinder/zoneminder/pull/2598/commits/2212244882ef6dca0acf51f0149fcaf6fd251857 fixes this

src/CMakeLists.txt
pliablepixels: https://github.com/ZoneMinder/zoneminder/pull/2598/commits/2212244882ef6dca0acf51f0149fcaf6fd251857 fixes this

src/zm_crypt.h
pliablepixels: `zm_crypt.h` is included by `zm_user.cpp` and `zm_crypt.cpp` - does this slow down compile compared to putting `BCrypt.hpp` and `jwt.h` in each of those cpp files?

pliablepixels: https://github.com/ZoneMinder/zoneminder/pull/2598/commits/4ab0c35962926e7149eccdc6bc415b048d5ef223 resolves this comment

knight-of-ni:
Didn't see a zm_update.sql to update existing users. Speaking of, do we have a migration path to get existing users onto the new scheme? Is it a simple matter of recreating each password from the web console?

Also, do the composer files under web need to live there? Need to know whether or not to tell our packaging scripts to delete those files during the build.

pliablepixels:
@knight-of-ni, on:

> Didn't see a zm_update.sql to update existing users. 
> Speaking of, do we have a migration path to get existing users onto the new scheme?

Update from `PASSWORD()` to `BCrypt` ? I don't believe this can be done by an SQL script ([ref](https://stackoverflow.com/a/20296312/1361529)). We need to do this at application level. That is what I need to discuss with the team. It seems package related scripts are either in Perl or are MYSQL commands. I don't want to add another dependent library in Perl for Bcrypt.  Right now, the migration I have in place is in PHP land. If you login using your `PASSWORD()` encoded password, it will automatically get migrated to Bcrypt. The limitation therefore is, "needs to be used to login"

 > Is it a simple matter of recreating each password from the web console?
I've modified the `users.php` form to ensure all new users use bcrypt. However, I'm not sure we should  enforce the user re-create as it adds a burden for folks with many logins (and suddenly stops working too after upgrade). 
My personal thought here is since we already have a migration in place for users who login (not just web console, any login via API or PHP will auto migrate), we don't really need to do more, but happy to hear other thoughts.

> Also, do the composer files under web need to live there? Need to know whether or not to tell our packaging scripts to delete those files during the build.

We can delete `composer.json` and `composer.lock` but we need to make sure the vendor directory is not deleted. We do need it checked into git though, so we can continue to add PHP modules in the future easily

knight-of-ni:
That first part was actually two questions. Your PR is missing a `zm_update-1.33.9.sql` to update the dB schema for existing users. Maybe you are holding off on that because the migration from old to new is still TBD. 

The second was indeed meant as an application question... After I upgrade an existing system to 1.33.9, all my passwords are still the old hash values, so that raises a few questions:
- Will the end user still be able to log in with the old hash values?
- Once logged in as admin, what is the process to the update all the old hash values to bcrypt? My guess is I will need to edit each user account and reenter the password. Is that it?

due to the shear number of files under `/web`, I'd prefer not to include non-essential files under it. All the packaging scripts I've seen glob the contents, which means we would have to add a special exception for those files. Not the end of the world, but if you could put the composer files in a folder that doesn't get bulk copied (even the root folder), it would save us packagers a little work.

Anyway, I'm off to the airport...

knight-of-ni:
Not to self: 
- find the rpm package equivalent for libdigest-bcrypt-perl  and update rpm specfile

pliablepixels:
> Your PR is missing a zm_update-1.33.9.sql to update the dB schema for existing users
Committed. Forgot to add it in

> Since you are adding columns to an existing table, you need to make this bit of sql "re-runnable" like we do here: https://github.com/ZoneMinder/zoneminder/blob/master/db/zm_update-1.31.16.sql

Updated, but I _think_ that procedure needs fixing. If you execute it manually, it invokes the command irrespective, even after it prints the desired "already exists" message. I compared it to [SO solutions](https://dba.stackexchange.com/a/199688), and they are a bit different.

> Will the end user still be able to log in with the old hash values?

You mean if they use `auth=<old auth hash value>` via ZMS, API, ZMU? Yes. That code still exists. All the new code uses `token=<>` and also why I did not overwrite auth.

>  Once logged in as admin, what is the process to the update all the old hash values to bcrypt? My guess is I will need to edit each user account and reenter the password. Is that it?

I've updated `zmupdate.pl` to go through the passwords and hash all MYSQL passwords with a bcrypt overlay (see https://github.com/ZoneMinder/zoneminder/pull/2598/commits/e9f843f29707f2f3c8f5c3792357d05e3ff64612), so this conversion happens automatically when you upgrade. Obviously, this is an overlay hash - I don't know the password, so in that same PR, I've added code to `auth.php` to do the real migration if login succeeds. All of this happens automatically, so you don't have to do anything.

> Due to the shear number of files under /web, I'd prefer not to include non-essential files under it. (stuff deleted)

Sounds good. Let me know what needs to move where, and I'll try and do it.
Also, you will have to advise on the third party libs (and I may need help here). That is still a ❓ in the top list of todos. The issues that need resolution there:
- I am using my own bcrypt fork and in static lib form. The reason I am using my own fork is because I couldn't figure out how to get  a shared lib working with debian dpkg-shlibdeps
- I am also using `jwt-cpp` as another submodule, but its dev is very responsive - updates were merged, so no fork needed.

knight-of-ni:
Note to self, these are the rpm package names that need to be added to the rpm specfile to get this to build:
- openssl-devel
- perl-Crypt-Eksblowfish
- perl-Data-Entropy

The good thing is these packages do exist in the rpm repos already.

distros/ubuntu1604/control
```diff
@@ -76,6 +79,9 @@ Depends: ${shlibs:Depends}, ${misc:Depends}, ${perl:Depends}
         ,rsyslog | system-log-daemon
         ,zip
         ,libpcre3
+        ,libssl
```
knight-of-ni: Not sure this is right for all versions of Ubuntu. It might need to be `libssl|libssl1.0.0` or `libssl|libssl1.1` . See: https://forums.zoneminder.com/viewtopic.php?f=39&t=28294

pliablepixels: Correct, just pulled ubuntu 12 and 16 dockers. `libssl1.0.0` works on both. committed.

pliablepixels: https://github.com/ZoneMinder/zoneminder/pull/2598/commits/33db7e1e35944668f3a5b9db9b424d6f4fd2967a

PR #2617, opened 2019-05-24
WIP: allow filters to be disabled

tolland:
I recently had an issue where a log file filled up the disk, and triggered the `PurgeWhenFull` filter which deleted some events I wanted to keep.

I'd like to be able to disable the filter from running automatically, but be able to run the filter manually if required.


connortechnology:
Why not just uncheck the Background checkbox?

tolland:
fair enough, I wasn't aware that

PR #2634, opened 2019-06-05
Completed Polish translation

fri-K:
I made update to Polish translation file


web/lang/pl_pl.php
```diff
@@ -87,29 +88,29 @@
 // Simple String Replacements
 $SLANG = array(
     '24BitColour'          => 'Kolor (24 bity)',
-    '32BitColour'          => 'Kolor (32 bity)',          // Added - 2011-06-15
+    '32BitColour'          => 'Kolor (32 bity)',
     '8BitGrey'             => 'Cz/b (8 bitów)',
     'Action'               => 'Działanie:',
     'Actual'               => 'Domyślna',
-    'AddNewControl'        => 'Add New Control',
+    'AddNewControl'        => 'Dodaj nową kontrolę',
```
connortechnology: Why use ą instead of y like in the other entries?

fri-K: This sentence is completely correct, but I changed it to be similar with other translations, both translations are equal

PR #2643, opened 2019-06-19
fix for zone overlay scaling issues in montage

tolland:
Currently the zone overlay svg is mis-sized due to extra "px" suffixed to the height and width

it's due to the extra "px" being appended, it messes up the values...

After fix....

### DIFF
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -206,7 +206,7 @@ function Monitor(monitorData) {
 
 /**
  * called when the layoutControl select element is changed, or the page
- * is rendered??
+ * is rendered
  * @return null
  */
 function selectLayout(element) {
@@ -270,10 +270,7 @@ function selectLayout(element) {
           // APPLET's and OBJECTS need to be re-initialized
         }
         streamImg.style.width = '100%';
-
-        updateZoneSvg(monitor, streamImg.getSize().x, streamImg.getSize().y);
       }
-
     } // end foreach monitor
   }
 } // end function selectLayout(element)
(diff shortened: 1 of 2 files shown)


tolland:
ah... this is still not scaling properly..



let me take another look

tolland:
ok. able to track the img height and width for layout, scaling and height size

connortechnology:
Looks pretty good!  Are you ready to have it merged?

tolland:
ah, I hit a problem, which is that the changeSize sets the src on the image which loads async... so the height for rescaling the zone is not available until some time later

tolland:
so it seems that scaling of the contents of an SVG can be achieved by setting the viewBox on the element.

knight-of-ni:
@tolland eslint is failing our Travis build with the following:

```
$ eslint --ext .php,.js .
/home/travis/build/ZoneMinder/zoneminder/web/skins/classic/views/js/montage.js
  207:1  error  Missing JSDoc for parameter 'element'                     valid-jsdoc
  210:4  error  Unexpected @return tag; function has no return statement  valid-jsdoc
  280:4  error  Unexpected @return tag; function has no return statement  valid-jsdoc
  327:4  error  Unexpected @return tag; function has no return statement  valid-jsdoc
✖ 4 problems (4 errors, 0 warnings)
The command "eslint --ext .php,.js ." exited with 1.
```

True, it is false trigger since the lines in question are just comments, but the simplest solution is to just remove the code. Not doing this will prevent us from seeing real javascript errors in any future pull requests. Thanks.

connortechnology:
The errors seem to be due the /** instead of /*.  If I remove the extra * then eslint doesn't complain anymore. Is the extra * intentional?

PR #2653, opened 2019-07-06
Add primary keys to Logs and Stats tables

bluikko:
Adds an auto_increment int(10) Id as PRIMARY KEY to Logs and Stats tables.

Closes ZoneMinder#2550 and makes ZoneMinder compatible with Galera writeset replication for InnoDB (Galera Cluster, Percona XtraDB Cluster).


bluikko:
This is a rather naive implementation but I did not think any of the existing columns could be used as primary keys.

I hope to get some feedback as I'm not totally sure I followed all ZoneMinder developer practices.

Also, some systems might have a very large `Logs` table and the migration could take a while. Should not be a big issue *except* for Galera 3. Since Galera was basically unsupported in the past anyways, I would not worry about that too much.

The exact issue with Galera 3 is that it does not support writeset splitting and a very large writeset (say. ALTERing a table of 10 GB to add a column) would cause the writeset to be too large. This is finally fixed in the latest Galera 4 that just came out.

db/zm_update-1.33.12.sql
```diff
@@ -0,0 +1,6 @@
+--
+-- Add primary keys for Logs and Stats tables
+--
+
```
connortechnology, line 4: close but we need these scripts to be able to be re-applied without failure, check other scripts for how to test for existence in the table first.

bluikko: I see, I think e35c165 should correct this.

bluikko: And e5104ac adds to actually prepare&execute the statements, oops.

Issue #2579, opened 2019-04-17
Deleting event deletes last event and not one clicked


kdhosting:
**Describe Your Environment**
- Version of ZoneMinder: 1.32.3
- How you installed ZoneMinder: PPA
- Full name and version of OS: Ubuntu 18.04
- Browser name and version (if this is an issue with the web interface):
Chrome

**Describe the bug**
When deleting events in the Feed page, only the top (last) event gets deleted regardless of which event is clicked. 

**To Reproduce**
1. Go to 'Console'
2. Click on any camera
3. Scroll down to list of events
4. Click 'X'
5. Observer error

**Expected behavior**
That only the event that is intended to be delete, gets deleted.

connortechnology:
This has already been fixed in master.

gaaf:
I could not find the fix in the master branch, so I created a fix myself for use in 1.32.3. It applies to master too. See PR #2660

PR #2660, opened 2019-07-10
Delete the selected event, not the most recent one (#2579)

gaaf:
The click event uses a closure to call deleteEvent with event.Id as parameter. At the end of the loop the value of event.Id is always equal to the id of the most recent event, so trying to delete any row always results in deleting the most recent event. 

Making the event block scoped instead of function scoped fixes this.

### DIFF
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -523,7 +523,7 @@ function getEventCmdResponse( respObj, respText ) {
     } );
 
     for ( var i = 0; i < dbEvents.length; i++ ) {
-      let event = dbEvents[i];
+      const event = dbEvents[i];
       var row = $('event'+event.Id);
       var newEvent = (row == null ? true : false);
       if ( newEvent ) {
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -523,7 +523,7 @@ function getEventCmdResponse( respObj, respText ) {
     } );
 
     for ( var i = 0; i < dbEvents.length; i++ ) {
-      var event = dbEvents[i];
+      let event = dbEvents[i];
       var row = $('event'+event.Id);
       var newEvent = (row == null ? true : false);
       if ( newEvent ) {


gaaf:
Well, I don't know what syntax is acceptable.

Code works for me.

knight-of-ni:
@gaaf To see why Travis failed the build, click on the the blue "Details" link. That takes you to the Travis CI portal, which shows you which build failed. We can see that eslint failed, so click on that to open the build log. We can then see exactly what is failing the build:
```
6.89s$ npm install -g eslint@5.12.0 eslint-config-google@0.11.0 eslint-plugin-html@5.0.0 eslint-plugin-php-markup@0.2.5
npm WARN deprecated circular-json@0.3.3: CircularJSON is in maintenance only, flatted is its successor.
/home/travis/.nvm/versions/node/v8.9.1/bin/eslint -> /home/travis/.nvm/versions/node/v8.9.1/lib/node_modules/eslint/bin/eslint.js
+ eslint-plugin-html@5.0.0
+ eslint-plugin-php-markup@0.2.5
+ eslint@5.12.0
+ eslint-config-google@0.11.0
added 134 packages in 6.041s
   ╭──────────────────────────────────────╮
   │                                      │
   │   Update available 5.5.1 → 6.10.0    │
   │      Run npm i -g npm to update      │
   │                                      │
   ╰──────────────────────────────────────╯
3.32s$ eslint --ext .php,.js .
/home/travis/build/ZoneMinder/zoneminder/web/skins/classic/views/js/watch.js
  526:7  error  Parsing error: The keyword 'const' is reserved
✖ 1 problem (1 error, 0 warnings)
The command "eslint --ext .php,.js ." exited with 1.
```

knight-of-ni:
If we can all agree this is a false trigger, it looks like the way to silence this is to create a .eslint file:
https://stackoverflow.com/questions/42706584/eslint-error-parsing-error-the-keyword-const-is-reserved

@mnoorenberghe

connortechnology:
One of our issues with js is that we want to support older ipads which are still serviceable (or at least I do) and they don't support modern js.  SO I think we configured eslint to warn us about it.

A better fix for this problem is to put the event_id in an attribute of the button and grab it from there when calling deleteEvent.  I will post an alternate PR in a moment.

gaaf:
But 'let' is used in a lot more JS files. So I assumed ECMAScript 2015 was supported.

connortechnology:
it shouldn't be.  I just havn't had time to go through them all.  It would be really nice to drop support for older devices... but I want to use my ipad with ZM.  Definitely a big problem.

PR #2661, opened 2019-07-10
Refresh event list instead of page reload

gaaf:
There seems to be no need to reload the page. Refreshing the event list seems sufficient and prevents flicker and/or scroll position changes.


connortechnology:
Please note that the removed code doesn't refresh the watch view, it would refresh the console. In the old days, watch was in a popup.  Nowadays it isn't in a popup, so the code should really have no effect. So I'm ok with it, but will wait a bit before merge so others can chime in.

PR #2667, opened 2019-07-19
Platerecognizer

connortechnology:
Making this public for comments

This adds web ui support for using the platerecognizer.com service with ZoneMinder

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
@@ -3983,6 +3983,40 @@ our @options = (
     type        => $types{boolean},
     category    => 'logging',
   },
+  {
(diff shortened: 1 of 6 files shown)


IgorA100:
The idea is certainly good, but I would look for an open source license plate recognition project instead of a proprietary one.

connortechnology:
We have support for both in eventserver.  This is mostly just the UI parts to make it easy to use. Way out of date though. 5 years.

IgorA100:
platerecognizer.com, in its free version, allows little number plate recognition.
I saw that this PR is 5 years old and has not been implemented.
Are you no longer planning to implement license plate recognition?
Maybe I should try using Tesseract OCR https://github.com/tesseract-ocr/tesseract?

connortechnology:
ALPR is hard... tesseract is the final step.  Anyways, ZMES supports both already.  This PR is just about adding elements to the web UI to configure it all.  We need a ton of work make nice UI to configure AI in the web ui.

PR #2688, opened 2019-08-28
Tweaks to the ubuntu installation instructions

raboof:
I recently walked a new user through the Ubuntu installation, and noticed some
things that were confusing to him that would be easy to fix.


raboof:
There are separate sections for Ubuntu 18.04 and for Ubuntu 16.04, so I don't think explaining the difference is helpful anymore - we can just have the correct instructions in each section.

PR #2714, opened 2019-10-02
Add setting of timezone to Options/Config instead of php.ini

connortechnology:
Request For Comment

This makes it more user friendly.  Not sure if there was a reason we didn't go this route before.


pliablepixels:
I'd strongly support this if it requires zero fiddling with `php.ini`. This is one of the top issues of new folks trying to make ZM work.

PR #2725, opened 2019-10-14
Add LIKE and NOT LIKE to filter options

externo6:
This is useful for filtering notes.
EG filtering detected objects from zmeventnofification;
WHERE notes LIKE %detect%
WHERE notes NOT LIKE %car%


externo6:
Perhaps there is a reason why this hasn't been added before, though this is useful for my setup with uploading based on the notes.

connortechnology:
Is it just me or have you not added it to the perl Filter module that actually implements the filter?

externo6:
I've not had to add anything to the perl filter for this to work? Its working without any modification.

connortechnology:
Won't work in background filters. Will work in web list matches....

externo6:
I can confirm its working as a background filter. Its uploading everytime a match is found.
Here are files been uploaded to my ftp server from this filter.
image

connortechnology:
Nope. I suspect that the term is simply being ignored and so it is matching more events than it should.

externo6:
Not sure what else I can say here.
Here are my filter options. 
image
and here is a direct sql statement from the db with the same id's from the ftp upload i posted.
image
If indeed the term was getting ignored, it would be uploading every event that occurred after 2019-09-28 10:00:00 as I use mocord, and I can confirm that is not the case.

externo6:
I'm not a perl programmer but I suspect its because of this line? 
https://github.com/ZoneMinder/zoneminder/blob/962049fa3171a335478645626a0eccb1c59c5416/scripts/ZoneMinder/lib/ZoneMinder/Filter.pm#L299
As far as I could tell it adds the raw operator and its value to the Sql string. Correct me if I'm wrong. Happy to add it here though if its standard practice.

externo6:
> yeah `LIKE` is a base mysql operator - I use a variant of it in zmninja too (`REGEXP` because `%` causes issues with URLs). That being said, should this not be called 'contains/ does not contain' at the filter display level?

Makes sense to rename to be Contains / Does not Contain. 

Code would need to change to include the % before and after the value something like:
```       ....
          case 'LIKE' :
          case 'NOT LIKE':
            $filter['sql'] .= ' '.$term['op'].' %'. $value .'%';
            break;
```

I'll look to change this tonight.

externo6:
I'm closing this to open a revised pull request.

connortechnology:
I'm not so sure I agree with the contains vs LIKE.  LIKE makes sense to us SQL'ers.  contains perhaps makes more sense to non techies..   We need to test for the existence of % in the value and if it doesn't exist, add it before and after.  Should also implement ILIKE.

PR #2726, opened 2019-10-15
More filter options

externo6:
Previous pull request https://github.com/ZoneMinder/zoneminder/pull/2725

This is useful for filtering by notes by using the new Contains / Does not Contain options.
EG filtering detected objects from zmeventnofification;

The filter scripts check if 'Like' or 'Not Like' is the operator and appends % before and after the value to allow Like or not like to act as Contains / Not contains with SQL.

I'm not a perl developer but I've added the options and its working with my setup. Please review and let me know if you have any feedback / issues.


scripts/ZoneMinder/lib/ZoneMinder/Filter.pm
```diff
@@ -295,6 +300,10 @@ sub Sql {
             $self->{Sql} .= ' IN ('.join(',', @value_list).')';
           } elsif ( $term->{op} eq '!~' ) {
             $self->{Sql} .= ' NOT IN ('.join(',', @value_list).')';
+          } elsif ( $term->{op} eq 'LIKE' ) {
+            $self->{Sql} .= " LIKE $value";
```
connortechnology, line 304: I would have expected value to need single quotes around it...

scripts/ZoneMinder/lib/ZoneMinder/Filter.pm
```diff
@@ -236,6 +236,11 @@ sub Sql {
                 || $term->{attr} eq 'Cause'
                 || $term->{attr} eq 'Notes'
                 ) {
+                if ( $term->{op} eq 'LIKE'
+                || $term->{op} eq 'NOT LIKE'
+                ) {
+                $temp_value = '%'.$temp_value.'%';
```
connortechnology: This adds % around it regardless of if it contains %

scripts/ZoneMinder/lib/ZoneMinder/Filter.pm
externo6: The reason I didn't add single quotes is that line 244 puts quotes around each value.
https://github.com/ZoneMinder/zoneminder/blob/4b71bc75eab113240cf5a69650c2149839083ac4/scripts/ZoneMinder/lib/ZoneMinder/Filter.pm#L244

scripts/ZoneMinder/lib/ZoneMinder/Filter.pm
externo6: Yeah, very true. During testing SQL didn't seem to care about if % was already in the string, so adding extra logic to detect it didn't seem efficient. Do you think it best to detect if % is at the beginning and at the end and skipping this?

Just like to say I'm not a programmer by trade and learn as I go so feedback is welcomed :)

connortechnology: To my mind, if I am specifying a % then I am interested in a specific case... like everything starting with something.. and if you add more %'s then I won't get that.  SO I think it is worth the extra processing which is pretty negligible.  
$temp_value = '%'.$temp_value.'%' if $temp_value !~ /%/;

externo6: Cool, I'll make the change, let it run to ensure the filters are working and push the change.

scripts/ZoneMinder/lib/ZoneMinder/Filter.pm
connortechnology: you are of course correct

scripts/ZoneMinder/lib/ZoneMinder/Filter.pm
externo6: @connortechnology I've added the line and worked fine in my setup. Please review :)

PR #2736, opened 2019-10-22
restricted montage to specified groups

saliyaw:
Restrict montage to specified groups by modifying link generation in console.php and report_event_audit.php


connortechnology:
This fix is not correct.  Further current master does not have the issue. It was fixed a long time ago.

Issue #2747, opened 2019-11-01
The start point of stored event video is not quiet right


selvanair:
ZM Version: _1.33.14_
OS: _GNU/Linux Debian 10 (buster)_

Some relevant parameters:

_analysis fps = 5
captured video is 15 fps nominal
pre-event-count = 10 (i.e. 2 seconds using 5 fps).
image buffer = 100_

**What happens:**
The stored video often starts less than 2 seconds before the alarm start time
**Expected:**
The pre-event segment should be at least 2 seconds (could be more because of the need to start at an IDR frame (keyframe)).

Looking at debug logs I see the captured packet queue is routinely cleared keeping at least 10 frames which leads to ~~11~~ 15 frames in the queue in the worst case to ~~40~~ 29 frames in the best case at any time. That is 1 sec to 2 sec of video, very often much less than the required 2 seconds (in my case).

```
Log:
10/31/19 20:53:22.104711 zmc_m1[26779].DB3-zm_packetqueue.cpp/131 [Clearing all but 10 frames, queue has 30]
10/31/19 20:53:22.104926 zmc_m1[26779].DB3-zm_packetqueue.cpp/190 [Deleted 15 packets, 15 remaining]
```

Consistent with this, the logs also show that `clear_unwanted_packets()` called by `CaptureAnd Record()` doesn't  find a point in the packet queue to start the event recording and ends up keeping the whole queue.

```
Log:
10/31/19 20:53:56.058295 zmc_m1[26779].DB1-zm_packetqueue.cpp/255 [Didn't find a frame before event start time. keeping all]
```

I think the pre-event frame count (10 in my case) should be scaled by capture_fps/analysis_fps while trimming the queue. Same logic should also apply with `clear_unwanted_packets()` called before event clip storage. Even better, convert that number internally to a time interval while manipulating packets as capture-fps is somewhat variable.

selvanair:
> a camera at 15fps means that 10 pre-event frames is 2/3 of a second. You should set it to 30 to get 2 seconds of pre-event.

That would retain enough packets in the video buffer, but will give incorrect number of pre-event frames. As far as I can see, jpeg frames are always defined using the analysis fps and that will save 30 frames lasting for 6 seconds. Even if jpegs are not saved, the frames data in the db will have 30 pre-event frames with an approximate delta time of 0.2 seconds (5 fps). Same with post event frames. 

And, frames recreated from the video will be off too as they are generated using the data in the db.

[...]

PR #2768, opened 2019-11-30
Correct video buffer length when analysis and capture fps differ

selvanair:
Summary of commits:
- Properly compute video buffer length when analysis fps is set
- Fix the recorded video start time
  This fixes a bug that surfaced when the video start got correctly extended to the past
- Update frame deltas in the database to be relative to the video start time

See commit messages for more details.

Addresses issue #2747

### DIFF
--- a/src/zm_ffmpeg_camera.cpp
+++ b/src/zm_ffmpeg_camera.cpp
@@ -744,6 +744,8 @@ int FfmpegCamera::CaptureAndRecord(
   }
   int ret;
 
+  struct timeval video_buffer_duration = monitor->GetVideoBufferDuration();
+
   int frameComplete = false;
   while ( !frameComplete ) {
     av_init_packet(&packet);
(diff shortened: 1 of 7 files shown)


src/zm_ffmpeg_camera.cpp
```diff
@@ -870,14 +872,23 @@ int FfmpegCamera::CaptureAndRecord(
           // since the last keyframe.
           unsigned int packet_count = 0;
           ZMPacket *queued_packet;
+          struct timeval video_offset = {0};
 
           // Clear all packets that predate the moment when the recording began
           packetqueue->clear_unwanted_packets(
-              &recording, monitor->GetPreEventCount(), mVideoStreamId);
+              &recording, 0, mVideoStreamId);
```
connortechnology, line 879: Wait, won't this make it clear too many packets?

selvanair: The original would have been correct if the start-time passed as `recording` did not include the pre-event count. But it does already include it (see below). So we should not keep an additional pre-event-count frames before the start-time (=`recording`).

Note: the value of `recording` passed to` CaptureAndRecording()` is
`videostore_data->recording = event->StartTime()`. Event is created with 
`start-time = *(image_buffer[pre_index].timestamp)`
where pre-index included pre-event frame count.

An exception is when a new recording segment is created due to segment length exceeding the limit. But in that case no pre-event count back-track required.

P.S. This bug went undetected in the past probably because there was never enough packets in the video buffer to backtrack by an additional pre-event count frames. It surfaced when my changes pushed the buffer start enough into the past.

connortechnology: right, I see now. Cool!

connortechnology:
Nice work.

Issue #2772, opened 2019-12-04
API Monitors issue


dragonsongohan:
API: _domain/zm/api/monitors/index/GroupId:1/.json_

This is function index() in zoneminder/web/api/app/Controller/MonitorsController.php
![Screenshot from 2019-12-04 15-13-14](https://user-images.githubusercontent.com/12947223/70125112-cd208a00-16a8-11ea-8b92-6502f1f2ee43.png)

This is function buildFilter() in zoneminder/web/api/app/Controller/Component/FilterComponent.php
![Screenshot from 2019-12-04 15-12-40](https://user-images.githubusercontent.com/12947223/70125173-ef1a0c80-16a8-11ea-9706-54af76a6621a.png)

When I passed the GroupId parameter, the array $conditions  never had the GroupId key, you turned the key into \`GroupId \`. Hope that the error can be corrected, thank you!

PR #2774, opened 2019-12-06
dirty fix filtering Monitors by GroupId.

connortechnology:
Change occurrences of GroupId to "`GroupId` " as that is what the key is in conditions.  Please note that other operators like != won't work.

This is a quick fix... but is far from ideal.  I might suggest that we not merge it except that we want to get 1.34 out.  

Ideally we would rewrite the FilterComponent to be a two stage process with an intermediary data structure.


connortechnology:
Have updated, with a more appropriate fix.  Fixes #2772 I think.

Issue #2797, opened 2020-01-05
Monitors.json API returns multiple copies of a monitor if that monitor belongs to multiple groups


neillbell:
**Describe Your Environment**
- Version of ZoneMinder: 1.33.16 (1.33.16~20191229135312-bionic)
- How you installed ZoneMinder: ppa:iconnor/zoneminder-master
- Full name and version of OS: Ubuntu 18.04.3 LTS

**Describe the bug**
When a monitor is included in multiple groups, the API returns multiple entries for that monitor in the monitors.json request.

**To Reproduce**
1. Create a ZoneMinder system with a single monitor
2. Create two groups, each containing your single monitor
3. Execute https://example.org/zm/api/monitors.json
4. Observe the data for your single monitor occur twice in the output

**Expected behavior**
The monitors.json API request should return one entry for each monitor without regard to the groups to which that monitor belongs.

**Debug Logs**

With a single monitor and no groups:
https://example.org/zm/api/monitors.json
```
[...]

PR #2809, opened 2020-01-15
Avoid duplicate monitors in API

gonzalezcalleja:
In a deploy with multiple monitors that share groups, in the API zm/api/monitors.json duplicate monitors will appears, with this PR the left join query will group by Monitor.Id


connortechnology:
Cool thank you!  I wasn't sure how I was going to fix this one.  Fixes #2797

PR #2813, opened 2020-01-20
Instead of passing query string in form post variables, store it in s…

### LINKED ISSUE #2811: Return to original url after login
**Describe Your Environment**
- Version of ZoneMinder : 1.34.0 (bug is also in 1.32.X)
- How you installed ZoneMinder : from sources
- Full name and version of OS : Debian Buster 10.2
- Browser name and version : Chrome Mac

**Describe the bug**
Every time there is an alarm, I have a filter which send me an email, including the event URL.
If I'm already logged in, of course I directly see the event when I click on the URL. But if I'm not logged in, it redirects to the login page, and after I log in, I'm redirected to the default page, not the event page.

**To Reproduce**
Url received in the email: http://my-domain.com/zm/?view=event&mode=stream&mid=10&eid=39127

**Expected behavior**
Redirect to the asked URL after correct login.

I remember this was working in old releases like 1.30 , and then stopped working in 1.32.X I think.
Thanks

connortechnology:
…ession so that it survives redirects.  Fixes #2811

### DIFF
--- a/web/index.php
+++ b/web/index.php
@@ -152,8 +152,6 @@ if (
   setcookie('zmCSS', $css, time()+3600*24*30*12*10);
 }
 
-# Only one request can open the session file at a time, so let's close the session here to improve concurrency.
-# Any file/page that sets session variables must re-open it.
 
 require_once('includes/lang.php');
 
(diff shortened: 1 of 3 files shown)


mnoorenberghe:
Note that I haven't looked closely at the patch or tested it so I'm mostly commenting on the commit message:

I really dislike when websites use the session for this instead of using the query string… it means that if I have two tabs open to different pages and the session times out then they both redirect to the one page instead of preserving the correct pages. Using the session also means we need to be careful to not redirect to the previous page when the user is just trying to login to the default landing page (console).

Issue #2831, opened 2020-02-11
timezone defaults to UTC on update


connortechnology:
The behaviour is just not ideal. When we moved the Timezone setting to Options, there wasn't a good way to set it to the current system timezone.  

Instead we should have just left it NULL and  not touched the timezone so that the previous settings stayed in effect.

PR #2832, opened 2020-02-11
Fix 2831

connortechnology:
Adds an unset value to TIMEZONE.  Default it to '' so that upgrades will continue to use the value set in php.ini


connortechnology:
fixes #2831

Issue #2833, opened 2020-02-15
Add controls for D-Link DCS-5020L and fix IR mode (pending)


zlodag:
- Version of ZoneMinder: 1.34.2
- How you installed ZoneMinder: PPA (zoneminder-master)
- Full name and version of OS: Ubuntu 19.10

**If the issue concerns a camera**
- Make and Model: D-Link DCS-5020L, Firmware Version | 1.16.01 (2019-01-25)
- ZoneMinder Source Type: Remote

**Describe the bug**
Control capabilities and script not built-in.
[Current control script on wiki ](https://wiki.zoneminder.com/D-Link_DCS-5020L_Control_Script) not working for IR mode (on, off, auto)

**To Reproduce**
Enable PTZ control as described [in the wiki](https://wiki.zoneminder.com/D-Link#DCS-5010L.2C_DCS-5020L_and_DCS-5025L). Click wake/sleep/reset -> camera does not respond 

**Expected behavior**
Camera mode changes to day, night, or auto

I will fix and submit pull request

PR #2837, opened 2020-02-19
#2833 Added monitor presets and controls for D-Link DCS-5020L, and minor bug fix in zmcamtool

zlodag:


zlodag:
Control script modified from the file found [here](https://wiki.zoneminder.com/D-Link_DCS-5020L_Control_Script), broken functions including IR mode fixed, added the ability to set preset camera positions

PR #2839, opened 2020-02-19
Add alarmed zone to filters

connortechnology:
Add AlarmedZoneId as an attr to filter by.  This allows us to take action on events that have alarms in only a section of the video.

Please note that due to the subquery used, we can only support a simple operator.


connortechnology:
This appears to contain work from another branch.  I will merge the other branch first to simplify this one.

PR #2850, opened 2020-02-24
mention other values for GPU

pliablepixels:


connortechnology:
Ideally we want this doc to apply to 1.34 as well... I'll do the legwork to get it in, but in future, if doing something that should go into 1.34, let's base the work off the release-1.34 branch

PR #2857, opened 2020-02-25
Fix montagereview end calendar 

### LINKED ISSUE #2853: Montage Review Calendar not updating
The end date calendar can't go before the start date calendar, but after adjusting the start date, the end date doesn't reflect the new setting.

connortechnology:
populate a js var minTime and maxTime and use it when setting up the calendar widget to set minTime for end calendar.

Fixes #2853


connortechnology:
Fixes #2853

PR #2888, opened 2020-03-14
Mp4 animations for objdetect

pliablepixels:
This PR replaces GIF changes made to image.php with MP4


connortechnology:
I wonder if we should leave the gig code for people to use in places that can't handle the MP4.

connortechnology:
Like embedding in email

pliablepixels:
Agree, I'll make the changes. GIF has its own benefits. This is what I will do:

`objdetect` -> mp4 first -> GIF next -> image next
`objdetect_gif` -> GIF
`objdetect_video` -> mp4
`objdetect_image` -> jpeg

Issue #253, opened 2013-11-19
Add vnc screen recording. [$100]


connortechnology:
##

SteveGilvarry:
Closing as part of 2017 cleanup, not saying not valid, just no one is working on this and don't believe anyone is planning on doing so.

connortechnology:
Re-Opening because there is a bounty on it, and it is a feature that I want.

connortechnology:
Getting tired of reopening this.  How do I tell stalebot to NOT auto-close an issue?

SteveGilvarry:
Label it I think

connortechnology:
label it with what?  It has three labels and a bounty.

shawnggraham:
Its so weird that i was just thinking about this the other day and googled it.  Why is this something that has not be done yet? This started in 2013?  wow.  Ill take a look at it.

eilerjc:
Just looking at setting up my first NVR and was checking to see what zoneminder's bug list looks like...anyways, does the VNC feature need to be completely integrated in zoneminder? A few years ago I used a utility called vncrec to make a movie of a vnc session before. There are several versions floating around, this is the latest I know of (and produced the best results) http://panthema.net/2014/vncrec-rgb/  It would seem like using -movie and ffmpeg but telling ffmpeg to send it to a localhost port like in https://trac.ffmpeg.org/wiki/StreamingGuide

Now I haven't tried it, but based on the linkes above maybe something like:
vncrec -movie out.vnc | ffmpeg `vncrec -ffinfo -movie out.vnc` -i -  -f rtsp -rtsp_transport tcp rtsp://localhost:8888/live.sdp

But you all probably understand ffmpeg better than I ever could.

jimender2:
@connortechnology Would a user agent install be acceptable? Where you go to a page, download and install something and it streams back to zoneminder?

connortechnology:
That might be nice, but I'd really like to be able to use existing vnc/apply rdp.

nasichassoulas:
This feature would be really beneficial to my use case, and will be very valuable over the next few years. I would be willing to contribute towards it if there is a good chance of success on it. Any idea what size bounty would be enough to get this going?

clbr:
I'd be willing to work on this for the mentioned 700 (Bountysource 200 + @connortechnology's mentioned 500). @connortechnology, is the offer still valid?

I'd use libvncclient, a GPL library, to create the monitor type.

connortechnology:
[...]

PR #2901, opened 2020-03-28
Add vnc as a camera type. 

hax0kartik:
This adds a new camera type, VNC. Code built and tested on ubuntu xenial.
Fixes issue #253


nasichassoulas:
I have tested this on an instance running Ubuntu 18.04 and VNC functionality appears to work perfectly!

PR #2913, opened 2020-04-16
Fix monitor loading

connortechnology:
@hax0kartik noticed some issues with the order of fields not matching the sql and a missing signalpoints.


hax0kartik:
Looks good to me.

PR #2920, opened 2020-04-23
Update zmupdate.pl.in

michaeleino:
updating the way executing a command with perl -T switch
I got below on freenas/freebsd plugin/jail
Insecure dependency in exec while running with -T switch at /usr/local/bin/zmupdate.pl line 992.
Ref: http://www.novosial.org/perl/backticks/index.html


michaeleino:
okay, it was just working..
I have created another pr #2921
Thanks

PR #2928, opened 2020-05-05
added association method to GroupsController to get Monitor Id/Name o…

pliablepixels:
…f monitors that belong to the group


connortechnology:
I should point out that it won't return the associations for subgroups, just direct group membership.

Issue #2659, opened 2019-07-08
MySQL reserved table name GROUP(S)


EdwinGH:
I'm using Fedora 29 with MySQL Community 8.0 and Zoneminder 1.33.11 from https://zmrepo.zoneminder.com/f/29/x86_64/

I upgraded to Fedora 29 and then MySQL upgraded from 5.7 to 8.0 and Zoneminder stopped working. I got many SQL errors, and one seems to lead to the use of 'Group' as (first part of) a table name, and is a reserved word in MySQL 8 (see https://dev.mysql.com/doc/refman/8.0/en/keywords.html )

Error log:
web_php[1388]: ERR [SQL-ERR 'SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'Groups' at line 1', statement was 'SELECT * FROM Groups ' params:]

Testing directly in mysql proves the issue:
mysql> use zm;
mysql> select * from Groups;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'Groups' at line 1
mysql> select * from `Groups`;
Empty set (0.00 sec)

Note that I find the same reserved word for MySQL 5.7, but somehow it does not work on MySQL 8 (I can't test it on MySQL 5.7)

EdwinGH:
Note that above in the second select statement there were single quotes around the Groups, but the rendering engine replaced that with code :-( 

Does anyone else experience ZoneMinder issues with MySQL 8? Is there a way to prevent these problems?

EdwinGH:
I'm sorry that I run into a problem every time, but also "Function" is now a reserved keyword according to https://dev.mysql.com/doc/refman/8.0/en/keywords.html , and I get now the following error:
`SQL-ERR 'SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'Function' at line 1', statement was 'describe Monitors Function' params:
`

PR #2975, opened 2020-07-10
Quote some more MySQL8 Reserved keywords

LocutusOfBorg:
See:
https://bugs.launchpad.net/ubuntu/+source/zoneminder/+bug/1859295/comments/1
https://github.com/ZoneMinder/zoneminder/issues/2659
https://bugs.launchpad.net/ubuntu/+source/zoneminder/+bug/1859295


LocutusOfBorg:
there might be some more, but I don't know how to test

PR #3003, opened 2020-08-05
Don't prefix manpage with zoneminder-

marillat:
man pages must match binaries names.


knight-of-ni:
Sorry for the delay. I had to weed through build infrastructure issues this morning.

Issue #3007, opened 2020-08-07
DiskSpace only calculated on the first refresh of Events


MarkBarbieri:
**Describe Your Environment**
- New install of Master on debian Buster (v10.5) from deb https://zmrepo.zoneminder.com/debian/master buster/
- 1.35.5~20200806.315
- Microsoft Edge Chromium Browser

**Describe the bug**
The Events listing does not display the correct DiskSpace for each recording when set to record continuously (MOCORD).  DiskSpace is calculated the first time the UI updates the event list, and the size is never updated again.

**To Reproduce**
1. Set to MOCORD
2. Click on Events listing
3. Repeatedly refresh the UI, note that the DiskSpace updates on the first refresh, but never again

**Expected behavior**
DiskSpace should be recalculated on each refresh

connortechnology:
True.  This should only affect in-progress events though.  Once an event is finished it's size shouldn't change significantly so there is no point updating it.

Should be a simple fix.

MarkBarbieri:
> True. This should only affect in-progress events though. Once an event is finished it's size shouldn't change significantly so there is no point updating it.
> 
> Should be a simple fix.

True, but once the in-progress recording completes the size is not updated, so the completed events are also incorrect.

PR #3008, opened 2020-08-07
Fix DiskSpace default arg

MarkBarbieri:
#3007 

The UI passes null to DiskSpace(), which means take the default arg, an empty string.

In this case, the folder_size() is only calculated on the first call, and not subsequent calls.

I still don't think this functions logic is correct as any value passed to DiskSpace() ($new != '') will be reported in the UI but won't be updated in the database.  Not quite sure of the use case here.

### DIFF
--- a/web/includes/Event.php
+++ b/web/includes/Event.php
@@ -262,7 +262,7 @@ class Event extends ZM_Object {
     return $streamSrc;
   } // end function getStreamSrc
 
-  function DiskSpace( $new='' ) {
+  function DiskSpace( $new=null ) {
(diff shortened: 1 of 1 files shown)


connortechnology:
That won't do it.  When we call $event->DiskSpace() we aren't passing null we aren't passing anything, so php sets it to ''.  When we want to refresh this value, we call $event->DiskSpace(null), which passes null and causes it to reload values and update the db entry.  

I've already pushed a fix to make it not commit to db when the event isn't finished.

PR #3012, opened 2020-08-09
remove legacy bandwidth and version popups

knight-of-ni:
This PR deletes the legacy version and bandwidth popups. This functionality has been moved into the navbar.

Note the action file version.php has been reused, so we don't delete that.

Merge only when ready


connortechnology:
I''m tempted to leave the bandwidth action for  use by 3rd party ui's..  Might be silly....probably no one uses it and never will... perhaps it should morph into some more generic cookie management api.

PR #3021, opened 2020-08-24
Utils: remove some old CI/package building related files

Carbenium:
As far as I can see these scripts/patches aren't referenced anywhere and their functionality is incorporated either directly in the docker images or `startpackpack.sh`.

This should make it easier to get an overview of the CI/build system for a newcomer.


knight-of-ni:
The extra files in the packpack folder are intentionally left there for reference. This is not debatable.

Travis is surprisingly temperamental, sometimes failing because of too much output and then failing other times with too little output. Changes in packpack cause us to patch it during the build process, or fall back to an older commit. The unused files shall remain in case something changes again, and we have to revert back to former behavior. 

While we do appreciate others willing to help us, you've made assumptions about the purpose of certain files without asking us first. Moving forward please talk to us in our Slack channel before making changes to the build system.

I'm closing this pr. I don't disagree that some of the files in this pr can be deleted, while others should definitely not. I'm not opposed to a new pr, but I think the best way to go about this is to have discussion first.

PR #3023, opened 2020-08-25
CI: Do not try to run deploy when building from a fork

Carbenium:
The deploy step fails on forks since the deploy key is obviously not present.


SteveGilvarry:
ON: Condition: expects bash string so this evaluates to false always. Confusing to have two usages of condition in same yaml, but it does.

https://docs.travis-ci.com/user/deployment#conditional-releases-with-on

I think you might need to use Stages  name = Deploy, which can then run with if: fork=false

PR #3026, opened 2020-08-25
Build: Make out-of-source builds not only a suggestion but enforce them

Carbenium:


knight-of-ni:
We can promote ooc builds, but we cannot fail in-source builds. Zoneminder is packaged by several distros, and the package maintainer has to build according to the distro's build system. Take redhat for example. This change would break every build except for fedora 33.

knight-of-ni:
Just to expand on this a little... for the case of rpm, the out of source is handled by the %cmake, %cmake_build, and %cmake_install macros. On a redhat machine, you can see what these macros do with `rpm --eval %cmake`

Carbenium:
> [...] Take redhat for example. This change would break every build except for fedora 33.

Not quite IMHO. I suppose the succeeding CI builds are the best proof.

The `__cmake_in_source_build`/`__cmake3_in_source_build` variable defines if rpm should attempt an in- or out-of-source build.
This PR manually undefines them (https://github.com/ZoneMinder/zoneminder/pull/3026/files#diff-f9086d321f65e75795f26d41afd15979R30-R31) so we have out-of-source builds.
The only change that is coming with F33 is that `__cmake_in_source_build` is undefined by default.

knight-of-ni:
For redhat, this is the build engine that would be the ultimate proof: http://koji.rpmfusion.org/koji/

What value does this change add? There is real potential negative, that this is going to hurt someone.

Carbenium:
The benefit I see is one build configuration/variation less to worry about.
The enforcement made it actually visible that the rpm builds weren't done out-of source.

If you'd like I can change the [`FATAL_ERROR`](https://github.com/ZoneMinder/zoneminder/pull/3026/files#diff-0a9a6f966b4c4af639c58b1e881243bdR8) to a `WARNING` if that sounds better.

knight-of-ni:
Ahh, initially I didn't see this:
```
%undefine __cmake_in_source_build
%undefine __cmake3_in_source_build
```
That's how you got it to build on redhat. Remove these and the builds fail as expected. That's a cool trick, however. I didn't know about that. 

Despite the fact that this does appear to work on redhat, I am still unwilling to merge this pr. What we have now works just fine, is something that we are all familiar with, and does not require much maintenance. Trying to fix something that isn't broken gives us more to worry about, not less.

I am even more concerned about the other less used distros... especially the ones I'm not thinking of right now. Zoneminder has always amazed us at how widely deployed it is. The last thing I want to do is force something, on some package maintainer, that we had no good reason to do in the first place. Then Isaac and I will be stuck dealing with the mess ...again, more work, not less. If we had a really good reason to change this, I'd be okay with it, but we simply don't.

You seem really sharp, so I hope you stick around. I'm sure there other things you can work on if you are willing.

Carbenium:
Sounds reasonable to me :)

The changes from https://github.com/ZoneMinder/zoneminder/pull/3026/commits/1c01a9d2e80902f23e661ad630b55c194a7597a5 might be still interesting for you. They ensure that the Fedora-specific files are actually generated in the build directory and not the source directory.

knight-of-ni:
Sure, we talk about it in slack. I like how you organized the longish statements in CMakeLists by the way.

PR #3028, opened 2020-08-25
Extract generated part of zm_config.h into zm_config_data.h

Carbenium:
With this change IDEs have it easier to correctly reference the
variable/class declarations. Additionally one does not have to
regenerate the zm_config.h file when changing the code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,7 +120,7 @@ src/CMakeFiles/
 src/cmake_install.cmake
 src/libzm.a
 src/nph-zms
-src/zm_config.h
+src/zm_config_data.h
 src/zm_config_defines.h
 src/zma
 src/zmc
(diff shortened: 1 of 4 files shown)


connortechnology:
I'm not sure I understand why.

I get having a file with ONLY substitutions in it, but you moved a bunch constant defines as well....

Or is it a separation between defines and function declarations?

Carbenium:
The plan was to separate the generated part from the static one. I wasn't sure if you would like to have all defines in one place or not. I'm happy to move the non-substituted defines back.

connortechnology:
I think that  makes sense.  Let's keep the subsitutions on their own.

connortechnology:
Of course we already have zm_config_defines.h which is generated from scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in

I am currently working on actually compiling in those defaults and loading only values that have changed from the /etc/zm/conf.d files and db Config table. I could make that branch available if interested.

Carbenium:
@connortechnology There you go.

If I can do anything to make these changes compatible with yours, I'm happy to do so. Also feel free to incorporate them into your patches if that's easier.

PR #3035, opened 2020-09-05
Update debian.rst

zemsten:
Updated debian buster instructions to be version independent for php, updated mysql package, included gnupg, added tee commands for easy one-liners.

### DIFF
--- a/docs/installationguide/debian.rst
+++ b/docs/installationguide/debian.rst
@@ -43,7 +43,7 @@ trigger their installation manually.
 
 ::
 
-    sudo apt install apache2 mysql-server
+    sudo apt install apache2 default-mysql-server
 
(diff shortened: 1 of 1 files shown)


docs/installationguide/debian.rst
```diff
@@ -116,12 +120,12 @@ required apache modules.
 Automated way:
 ::
 
-    sudo sed -i "s/;date.timezone =/date.timezone = $(sed 's/\//\\\//' /etc/timezone)/g" /etc/php/7.0/apache2/php.ini
+    sudo sed -i "s/;date.timezone =/date.timezone = $(sed 's/\//\\\//' /etc/timezone)/g" /etc/php/7.*/apache2/php.ini
```
connortechnology, line 123: This isn't needed anymore.  We can now set timezone from options

PR #3038, opened 2020-09-10
Fix saving function changes

mnoorenberghe:
Wait for the request to complete before reloading.


knight-of-ni:
The event view needs a similar change when delting events. This is on my to-do this.
I am planning to update all the new getJSON calls to use the .done and .fail callbacks, in addition to making sure the right ajax response is sent (i.e. add new php under /ajax as needed).

PR #3040, opened 2020-09-10
Don't log errors connecting to function=none monitors

mnoorenberghe:
Make one debug and only log the other when enabled.

Thoughts on this vs. a `zmu` change?

### DIFF
--- a/src/zm_monitor.cpp
+++ b/src/zm_monitor.cpp
@@ -490,7 +490,9 @@ Monitor::Monitor(
     //video_store_data->frameNumber = 0;
   } else if ( purpose == ANALYSIS ) {
     if ( ! (this->connect() && mem_ptr && shared_data->valid) ) {
-      Error("Shared data not initialised by capture daemon for monitor %s", name);
+      if ( Enabled() ) {
+        Error("Shared data not initialised by capture daemon for monitor %s", name);
+      }
       exit(-1);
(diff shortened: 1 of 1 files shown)


connortechnology:
So one thing we have to be careful of here, is that Enabled doesn't mean what you think it means.   Enabled turns on/off Motion detection, not capturing... What you really mean to check is Function.

If we are connecting for the purpose of ANALYZING then we must be zma (or something like it).  The web ui is smart enough to stop all the processes in the correct order.  We should not be seeing those errors and if we are, then we need to figure out why, not turn off the logging.

mnoorenberghe:
> So one thing we have to be careful of here, is that Enabled doesn't mean what you think it means. Enabled turns on/off Motion detection, not capturing... What you really mean to check is Function.

Interesting… I've always kinda wondered what it meant exactly… can we change the label to "Motion Detection Enabled" to make this clear for everyone?

> If we are connecting for the purpose of ANALYZING then we must be zma (or something like it). The web ui is smart enough to stop all the processes in the correct order. We should not be seeing those errors and if we are, then we need to figure out why, not turn off the logging.

This is from an API call to `/zm/api/monitors/alarm/id:4/command:status.json` but the same issue could be hit by running `zmu` from the command line so it seems better to me to fix it at the lower level.

PR #3044, opened 2020-09-13
don't throw exception if frame ID for max score not found

pliablepixels:
Should fix https://github.com/pliablepixels/zmNinja/issues/972
Events API errors out if it can't find a frame with max score. This condition may occur if there was an error during event recording - frame was never written, etc. This should not break all of events.json


connortechnology:
Ok I get the gist, however in the resulting code $frame['Frame'] won't be defined.  php will croak.  Need to still test $frame and return something instead of unconditionally access array members.

pliablepixels:
I don't think it croaks. it returns null which is the expected value.
```
php -r '$frame=null; print $frame["a"]["b"];'
```

Issue #3046, opened 2020-09-17
mariadb shouldn't be a hard requirement in the systemd unit file (multi-server)


dodexahedron:
**Describe Your Environment**
- zoneminder-httpd-1.35.6-1.291.20200914gitgba900dd9d.el8.x86_64
- zmrepo
- CentOS 8.2

**Describe the bug**
Multi-server installs will only have the database on one server or on a separate server. The systemd unit file lists mariadb.service as Requires. Thus, if mariadb-server is not installed on the local system, the service won't start.

**To Reproduce**
1. Install zoneminder from zmrepo, but don't install mariadb-server.
2. Attempt to start zoneminder

**Expected behavior**
I would think a user would expect it to work out of the box, as the documentation for multi-server doesn't actually say anything about messing with the unit file.

A couple simple solutions:

1. The unit file already lists mariadb.service as After. Change Requires to Wants so that it starts it if it's there, but doesn't fail if it isn't.
2. Provide a multi-server rpm (call it *-multiserver or something) which omits the Requires line from the systemd unit file, and change the normal rpms to have mariadb-server as a dependency.
3. Update the docs to mention needing to dump a modified unit file in /etc/systemd/system/

**Debug Logs**
N/A

knight-of-ni:
This is by design, and it absolutely should be a hard requirement. If mariadb were to stop or crash, then zoneminder must stop as well. Back in the sys v init days, this was a real problem. Hence the requires. 

The ZoneMinder documentation for redhat is designed around the least common denominator, which is a single server managed by a user with little to no knowledge of Linux.

The nice thing about ZoneMinder is you can craft it into many different system architectures. This is a bit of double edged sword, however, as we don't have the resources nor desire to document every single niche case.

The multi-server configuration is a special case, is a distro agnostic set of instructions, and is not intended to work out of the box, for any distro. It's something that only an advanced user should try and only after some manual tweaking.

If you want to submit a PR to add a documentation to the multiserver config, then feel free. The systemd service file, however, will not change nor will a multiserver subpackage be created. This is not negotiable.

dodexahedron:
Well, then it's a bug anyway, if that's the design intent, because that's not how Requires works. Requires is only checked on service start.
If you want the behavior you're describing, it needs to be using BindsTo, rather than Requires.
mariadb failing will not cause systemd to stop zoneminder, and zoneminder will just crash on its own, with Requires.

[...]

PR #3047, opened 2020-09-17
Update multiserver instructions for systemd

dodexahedron:
Include step to remove the Requires directive for the unit file, so that zoneminder can start on servers that do not have mariadb-server installed (addresses #3046).
Also removed reference to mysql, since mysql would not satisfy the Requires directive anyway.


knight-of-ni:
Close, but keep in mind these particular instructions are not specific to Redhat or the systemd service file that ships with zonemidner rpms. We need to leave in references to mysql.

knight-of-ni:
superseded by 57b0a96

PR #3069, opened 2020-11-02
Join Monitors as M as well in total and totalNotFiltered queries

connortechnology:
Adds JOIN Monitors AS M


knight-of-ni:
Tested. This is working. Thanks.

PR #3072, opened 2020-11-05
Rework code to use Filter methods to generate advanced search 

connortechnology:
Reduces sql queries to 1 or 2, using count() to populate [total] and [totalNotFiltered].  Does pagination using array_splice instead of SQL queries.  Will use more ram, but reduces db load.

This also fixes post sql conditions.

### DIFF
--- a/web/ajax/events.php
+++ b/web/ajax/events.php
@@ -1,5 +1,4 @@
 <?php
-
 $message = '';
 $data = array();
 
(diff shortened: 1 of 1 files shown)


knight-of-ni:
Let me test this.

knight-of-ni:
This looks good. Testing passed.

PR #3084, opened 2020-11-29
Improve group permissions (mode 660)

BlueMax:
Creates mode 660 sockets and /dev/shm files.
Allows processes with different UIDs to share sockets/shm.

Allows systems like Alpine Linux to keep default configs and improves process separation.

```
  UID USER       PID RGROUP   COMMAND
    0 root         1 root     /sbin/init
    0 root       242 wheel    /sbin/syslogd -t
    0 root       364 root     /sbin/getty 38400 console
    0 root     10773 root     nginx: master process /usr/sbin/nginx -c /etc/nginx/nginx.conf
  101 nginx    10774 nginx    nginx: worker process
  101 nginx    10775 nginx    nginx: worker process
  101 nginx    10776 nginx    nginx: worker process
  101 nginx    10777 nginx    nginx: worker process
    0 root     10809 root     php-fpm: master process (/etc/php7/php-fpm.conf)
  102 fcgiwrap 10836 www-data /usr/bin/fcgiwrap -c 4 -s unix:/run/fcgiwrap/fcgiwrap.sock
  102 fcgiwrap 10841 www-data /usr/bin/fcgiwrap -c 4 -s unix:/run/fcgiwrap/fcgiwrap.sock
  102 fcgiwrap 10842 www-data /usr/bin/fcgiwrap -c 4 -s unix:/run/fcgiwrap/fcgiwrap.sock
  102 fcgiwrap 10843 www-data /usr/bin/fcgiwrap -c 4 -s unix:/run/fcgiwrap/fcgiwrap.sock
  102 fcgiwrap 10844 www-data /usr/bin/fcgiwrap -c 4 -s unix:/run/fcgiwrap/fcgiwrap.sock
  103 zonemin+ 11677 www-data php-fpm: pool zoneminder
  103 zonemin+ 11681 www-data php-fpm: pool zoneminder
  100 mysql    11926 mysql    /usr/bin/mariadbd --basedir=/usr --datadir=/tmp/mysql
    0 root     11927 root     logger -t mysqld -p daemon.error
  103 zonemin+ 11984 www-data /usr/bin/perl -wT /usr/bin/zmdc.pl startup
  103 zonemin+ 12012 www-data /usr/bin/zmc -m 1
  103 zonemin+ 12021 www-data /usr/bin/perl -wT /usr/bin/zmfilter.pl --filter_id=1 --daemon
  103 zonemin+ 12022 www-data /usr/bin/zma -m 1
  103 zonemin+ 12027 www-data /usr/bin/perl -wT /usr/bin/zmfilter.pl --filter_id=2 --daemon
  103 zonemin+ 12032 www-data /usr/bin/perl -wT /usr/bin/zmfilter.pl --filter_id=5 --daemon
  103 zonemin+ 12042 www-data /usr/bin/perl -wT /usr/bin/zmwatch.pl
  103 zonemin+ 12049 www-data /usr/bin/perl -wT /usr/bin/zmstats.pl
  103 zonemin+ 12257 www-data php-fpm: pool zoneminder
```

### DIFF
--- a/src/zm_monitor.cpp
+++ b/src/zm_monitor.cpp
@@ -557,7 +557,7 @@ bool Monitor::connect() {
   Debug(3, "Connecting to monitor.  Purpose is %d", purpose );
 #if ZM_MEM_MAPPED
   snprintf(mem_file, sizeof(mem_file), "%s/zm.mmap.%d", staticConfig.PATH_MAP.c_str(), id);
-  map_fd = open(mem_file, O_RDWR|O_CREAT, (mode_t)0600);
+  map_fd = open(mem_file, O_RDWR|O_CREAT, (mode_t)0660);
   if ( map_fd < 0 ) {
     Error("Can't open memory map file %s, probably not enough space free: %s", mem_file, strerror(errno));
     return false;


BlueMax:
Accidentally declared `my $old`. But this is php not perl. It was correct in the first place. : )

I will bring this to Alpine in any case as web services are unified by group permissions there.
Maybe this is interesting for upstream as well.

connortechnology:
Do we actually need to call umask?  We are explicitly setting the permissions when creating the file...

BlueMax:
I haven't found any other way. Default umask is 022 stealing the `g+w`.
`zm_monitor.cpp`: Without umask `/dev/shm/zm.mmap.1` gets `640` (when `mode_t` is set to `0660`).
`stream.php`:  Without umask the `w.sock` is `755`.
The `s.sock` is correct though. 
The lock files don't get pruned. Not sure if that's a permission issue as well. Haven't seen any errors yet.
```
$ ll /run/zoneminder/
total 4
drwxrwxr-x    2 zonemind www-data       200 Dec  1 00:52 .
drwxr-xr-x    9 root     root           280 Dec  1 00:49 ..
-rw-r--r--    1 zonemind www-data         4 Dec  1 00:49 zm.pid
srwxr-xr-x    1 zonemind www-data         0 Dec  1 00:49 zmdc.sock
-rw-------    1 fcgiwrap www-data         0 Dec  1 00:51 zms-077153.lock
-rw-------    1 fcgiwrap www-data         0 Dec  1 00:52 zms-527700.lock
srwxrwxr-x    1 fcgiwrap www-data         0 Dec  1 00:52 zms-527700s.sock
srwxr-xr-x    1 zonemind www-data         0 Dec  1 00:52 zms-527700w.sock
-rw-------    1 fcgiwrap www-data         0 Dec  1 00:51 zms-848466.lock
-rw-------    1 fcgiwrap www-data         0 Dec  1 00:51 zms-991330.lock
```

connortechnology:
yes lock files have to be cleaned up later due to races..

knight-of-ni:
There are two issues here: 
- socket file permissions
- mem mapped file permissions

I'm not sure we need to explicitly set the permission on the sock files. This is already done via systemd tmpfiles. 

For example, here is how we do it for redhat distros:
https://github.com/ZoneMinder/zoneminder/blob/master/distros/redhat/httpd/zoneminder.httpd.tmpfiles.in

Please test this on your system and report back.

I don't see any issue with the idea of specifying different permissions on the mem mapped files, but if we are going to change it, I think would be better to make this a configurable option, rather than hard code it. That way the end user can change the perms to whatever suits.

connortechnology:
The tmpfiles thing specifies permissions on the directory that they are in, not on the sockets themselves, so I suspect that the umask stuff is required for the sockets.  

I'm not entirely sure I want to let people shoot themselves in the foot by making it configurable either.   

I think in general I like the group read/write idea.  I wonder though if the group shouldn't be zoneminder, and fcgidwrap would be in the zoneminder group.  A matter of preference I suspect.

src/zm_monitor.cpp
```diff
@@ -138,7 +138,10 @@ bool Monitor::MonitorLink::connect() {
 
     Debug(1, "link.mem.size=%d", mem_size);
 #if ZM_MEM_MAPPED
-    map_fd = open( mem_file, O_RDWR, (mode_t)0600 );
+    /* mode 660 to allow group www-data */
+    mode_t oldmask = umask(0002);
```
connortechnology: Here we are just opening the file, not creating it, so do we need to change the umask here?

connortechnology: I think the mode being passed in open is ignored as well

connortechnology:
I'm concerned that the socket created by zms hasn't been addressed.  As near as I can tell it should need similar treatment

src/zm_monitor.cpp
```diff
@@ -557,7 +560,10 @@ bool Monitor::connect() {
   Debug(3, "Connecting to monitor.  Purpose is %d", purpose );
 #if ZM_MEM_MAPPED
   snprintf(mem_file, sizeof(mem_file), "%s/zm.mmap.%d", staticConfig.PATH_MAP.c_str(), id);
-  map_fd = open(mem_file, O_RDWR|O_CREAT, (mode_t)0600);
+  /* mode 660 to allow group www-data */
+  mode_t oldmask = umask(0002);
+  map_fd = open(mem_file, O_RDWR|O_CREAT, (mode_t)0660);
```
connortechnology, line 560: I did a test here.  the umask change was not required.  On my machine the 0660 took effect.

knight-of-ni:
> The tmpfiles thing specifies permissions on the directory that they are in, not on the sockets themselves, so I suspect that the umask stuff is required for the sockets.
> 

Setting the group setuid bit on the directory will cause files under that folder to inherit the group permissions of the parent. 

The relevant line in systemd tmpfiles would look like this:
D @ZM_SOCKDIR@ **2**755 @WEB_USER@ @WEB_GROUP@

src/zm_monitor.cpp
BlueMax: Agreed. I can remember i just focused on 600 and a "shouldn't harm". Regarding `man 2 open` it gets ignored.

src/zm_monitor.cpp
BlueMax: Here its mandatory. Global `umask` might come from the init system. Alpine has a custom init system and OpenRC.
```
$ ll /dev/shm/
total 502208
drwxrwxrwt    2 root     root            60 Dec  2 03:12 .
drwxr-xr-x    4 root     root           440 Dec  1 01:39 ..
-rw-r-----    1 zonemind www-data 514259144 Dec  2 03:12 zm.mmap.1
$ umask
0022

```

BlueMax:
> Setting the group setuid bit on the directory will cause files under that folder to inherit the rw group permissions of the parent.

I think its ownership only, not permissions.

```
The effective mode is modified by the process's umask in the
usual way: in the absence of a default ACL, the mode of the
created file is (mode & ~umask).
(man 2 open)
```

There's kinda difference between upstream and Alpine. Maybe its better to handle that isolated. On the other hand umask makes sure the permissions are correct. You never know what default ACL, init system or whatever comes between Zoneminder and its file permissions.

src/zm_monitor.cpp
BlueMax: Found a way in OpenRC to set `process umask` (start-stop-daemon-args). Will remove the umask.

BlueMax:
All `umask` calls have been removed. The `s.sock` and `w.sock` umask can be handled via init systems (systemd, openrc).
> I'm concerned that the socket created by zms hasn't been addressed. As near as I can tell it should need similar treatment

The `s.sock` inherits `mode_t` from the parent folder. I've checked that again. Should be enough i guess.

On Alpine Linux `php-fpm7` has wrong process umask. That's an upstream bug on their side. That's why the `w.sock` was different. So that's not needed anymore either.

Left over is a 1-byte patch for /dev/shm files (600->660).

PR #3087, opened 2020-12-01
add thumbnail support on the backend for watch view

knight-of-ni:
This adds php backend support for event thumbnails in the watch view. Frontend javascript & php will be added after this is merged.

@connortechnology Please review. Status.php is used by more than just the watch view. I'm looking for a second set of eyes before merging.

### DIFF
--- a/web/ajax/status.php
+++ b/web/ajax/status.php
@@ -329,6 +329,17 @@ function collectData() {
         foreach ( dbFetchAll($sql, NULL, $values) as $sqlData ) {
           foreach ( $postFuncs as $element=>$func )
             $sqlData[$element] = eval('return( '.$func.'( $sqlData ) );');
+
+          // Create parameters used to build the event thumbnail html
+          $event = new ZM\Event($sqlData['Id']);
+          $scale = intval(5*100*ZM_WEB_LIST_THUMB_WIDTH / $event->Width());
+          $imgSrc = $event->getThumbnailSrc(array(),'&amp;');
+          $streamSrc = $event->getStreamSrc(array(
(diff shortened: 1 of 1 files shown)


connortechnology:
I think this might be no good.  status.php can be working on things other than events as you say. So sqldata['Id'] might not be an event id.  So I think we need to do something a little different.

knight-of-ni:
Certainly, but what I'm asking for is a way to fix it, that is acceptable to you. I can think of a number of different ways to go about this. 

A quick fix might be to just check `$entitySpec == 'events'` .

Alternatively, I could create a new ajax/watch.php with all new logic. This is the way I'd go about it, if there wasn't already logic to handle this request in ajax/status.php

Thirdly, I could create the thumbnail html entirely in javascript. This is more efficient in terms of information we transmit to the client with each ajax call, but that's not how we've been doing this for the other views that have  thumbnails. So doing it this ways begs we go back and change the other thumbnail cases as well.

The code in ajax/status.php is largely your work, and I don't want to start changing things around without coordinating with you.

connortechnology:
Actually I've done very little in status.php. This stuff predates us.  

I don't actually like it.  What we are listing is event data, so I'd be tempted to use ajax/events.php and try to be more consistent in view with the main events list. 

I'm ok with creating it in javascript as well.  Eventually I'd like to have some nice js libraries that people who like to use node.js etc can use to interact with ZM.  So that approach has merit too.  

Anyways I'll try to cogitate on it today.

connortechnology:
I'm leaning towards just doing this in processRows.

knight-of-ni:
I like your idea about using event.php. I hadn't considered that, and it shouldn't require any backend changes. I just need to refactor the Ajax call. Closing this PR since it is no longer relevant...

PR #3099, opened 2020-12-22
save frame scores to trigger_data for use with linked monitors

vertigeaux:


connortechnology:
I think this is missing some commits.  trigger_score isn't listed in the trigger_data definition

vertigeaux:
Is that not line 170 of zm_monitor.h?

connortechnology:
Oh I see, you re-used the existing trigger stuff. Hmm... I kinda had in mind adding a new element separate from triggers called something like motion_score.  because triggers are a separate thing than motion detection.  Technically we should have analysis_score or event_score which is trigger_score + motion_score + whatever else score.
trigger score could get overwritten if someone is using triggers.

vertigeaux:
I see now, I will continue to work on it. I suppose for now I will just add this score property to SharedData. As far as I can tell, the code already sums motion_score + trigger_score + linked_monitor_scores to arrive at a total frame score as you suggest. It's just that it used the arbitrary value of 50 rather than a real score from the linked monitor(s), which completely blows up the timeline.

PR #3102, opened 2020-12-23
fix api monitor save daemon control

connortechnology:
Move daemonControl logic into the model.  daemonControl needs to be called on the previous monitor data so that it knows which daemons to stop.

fixes 3101


connortechnology:
Fixed #3102

PR #3113, opened 2021-01-23
new api to control daemon

pliablepixels:
API to control daemons - zmNinja needs this to restart zmeventnotification


connortechnology:
I could have sworn we already had this.  Oh well.

One thing, can't allow just any user to control daemons.  Need to check monitor view/edit permissions etc.

pliablepixels:
done

PR #3119, opened 2021-02-02
Add fmt library for type-safe string formatting

Carbenium:
Using this library we can add type safety for all string formatting done and avoid a whole class of possible bugs.

Additionally with this library we can transition form
the printf-style formatting to the new "{}"-style introduced
in C++20 and the time is ripe switch to std::format.

For more infos on the library: https://github.com/fmtlib/fmt

This PR includes some simple initial usage to showcase the new functionality.

Feel free to ping me on Slack if there are questions.


knight-of-ni:
Is there a minimum version of gcc that this will compile against?
FYI, CentOS 7 uses gcc 4.8.5 so we need to make sure zm will build in that environment.

Carbenium:
According to the [docs](https://fmt.dev/latest/index.html) only a small subset of C++11 features is needed which is provided by GCC 4.8.
So from this information I'd say it should work. Although I don't have anything that old around to test...

knight-of-ni:
Thanks, I'll test this pr offline.

I noticed some other, potentially large issues.

There seems to be a lot of js style and formatting needed:
https://travis-ci.org/github/ZoneMinder/zoneminder/jobs/757247924

Also, this PR is loading old bootstrap 3.3 and old jquery 1.11. Zoneminder uses bootstrap 4.5 and jquer 3.5.1.

Carbenium:
Regarding the JS stuff: Those are just the doc files of the library which were included in the tarball, so they should be excluded from the analysis.
I would just add the whole `src` directory to the `.eslintignore`. Or even better call: eslint only on the `web` subdirectory. That might even speed up the analysis.

Opinions?

connortechnology:
Do we have to suck this library into our git?  I'd really rather leave it as an external dependency.

Carbenium:
Hmm, yeah, I thought about this as well...

CentOS 7 is the culprit. They only ship 3.0.2 whereas 4.0.0 would be nice. To not ship the lib in the repo, CMake's `FetchContent` module could be used but that is only available since CMake 3.11, which is also not in CentOS 7.
So maybe keep it in-tree until CentOS 7 support can be dropped and the remove it from the repo as well?

knight-of-ni:
With all the turmoil with CentOS 8/Stream/Rocky linux, CentOS 7 is the only sure bet at the moment. This will change with time, but for the moment I don't see any clear EOL date for centos 7 support.

Cmake 3.17 is available for el7:
https://src.fedoraproject.org/rpms/cmake3

That's what we currently use to build zoneminder. However, linux distros do not allow internet access in their build environments, so FetchContent would not work.

Package builders can treat it as an additional source, however. We do this currently for Crud library in our rpm specfile:
https://github.com/ZoneMinder/zoneminder/blob/master/distros/redhat/zoneminder.spec#L46

So basically, that leaves it up to the pacakger to fetch the appropriate tarball before starting the build.

There is lots to think about. I'm travelling for day job the rest of this week. My responses may be few.

Carbenium:
Right, there is a certain wind of change...

I'm totally fine with dropping the in-tree code and relying on the packagers to supply the library. Would it be okay for you to require 4.0 as minimal version?
Then I would drop the in-tree copy.

knight-of-ni:
I have submitted a bug report to get fmt 6.2.1 (same package as el8) added to epel 7 repos:
https://bugzilla.redhat.com/show_bug.cgi?id=1925932

Let's see where this goes.

Carbenium:
@knight-of-ni Wow, thank you very much for your efforts! Much appreciated.

knight-of-ni:
@barjac we are considering the addition of the fmt package as a build dependency for ZoneMinder. Can you verify whether mageia repo has fmt >= 6.2.1 ?

barjac:
> @barjac we are considering the addition of the fmt package as a build dependency for ZoneMinder. Can you verify whether mageia repo has fmt >= 6.2.1 ?

We have it at 6.2.1 in current stable (Mga7) and 7.1.3 in Cauldron. However Cauldron is now in release freeze so we will be sticking with zoneminder-1.34.22 for the Mageia-8 release.
Thanks for the heads up.

Carbenium:
> Cmake 3.17 is available for el7:
> https://src.fedoraproject.org/rpms/cmake3

knight-of-ni:
So far I'm getting the silent treatment from the fmt package maintainer. The good news is there is a written procedure I can follow to more aggressively ping the package maintainer. If the maintainer still does not respond, then I will be granted by the admins the ability to manage the fmt package myself. The only down side to this is that it takes time, sometimes a lot of time, for this process to work itself out.

Based on these circumstances and the likelihood this PR will need a lot of testing in different environments, I think it best we give this a milestone of the release after the next, e.g. v1.38.

knight-of-ni:
Good news. After a little prodding via email, the package maintainer has submitted fmt 6.2.1 build to epel 7 repo:
https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2021-d5fe34a16a

Now we wait 2 weeks for it to show up in stable repo.

knight-of-ni:
whoo hoo... the package gained enough karma while in testing to make it to stable repo early:
```
[abauer@vCentOS7 ~]$ rpm -ql fmt
/usr/lib64/libfmt.so.6
/usr/lib64/libfmt.so.6.2.1
/usr/share/doc/fmt-6.2.1
/usr/share/doc/fmt-6.2.1/ChangeLog.rst
/usr/share/doc/fmt-6.2.1/README.rst
/usr/share/licenses/fmt-6.2.1
/usr/share/licenses/fmt-6.2.1/LICENSE.rst
[abauer@vCentOS7 ~]$ rpm -qi fmt
Name        : fmt
Version     : 6.2.1
Release     : 2.el7
Architecture: x86_64
Install Date: Sun 21 Feb 2021 03:35:25 PM CST
Group       : Unspecified
Size        : 407691
License     : BSD
Signature   : RSA/SHA256, Thu 18 Feb 2021 08:25:38 AM CST, Key ID 6a2faea2352c64e5
Source RPM  : fmt-6.2.1-2.el7.src.rpm
Build Date  : Thu 18 Feb 2021 08:18:19 AM CST
Build Host  : buildhw-x86-04.iad2.fedoraproject.org
Relocations : (not relocatable)
Packager    : Fedora Project
Vendor      : Fedora Project
URL         : https://github.com/fmtlib/fmt
Bug URL     : https://bugz.fedoraproject.org/fmt
Summary     : Small, safe and fast formatting library for C++
Description :
C++ Format is an open-source formatting library for C++. It can be used as a
safe alternative to printf or as a fast alternative to IOStreams.
```

@Carbenium Do you want to refactor this PR to use the fmt package or just create a new PR?

Carbenium:
@knight-of-ni Oh, this great!

I'll rework this PR as soon as I've got some time to spare.

Carbenium:
Alirght, this is cleaned up. We now only use fmt supplied by the system.

**Note**: I just realized that Xenial does NOT ship fmt. So this should only be merged when dropping Xenial support is final.
CI is failing since the Bionic package is built with `ubuntu_16_04/control`, which has no dependency entry for `libfmt`.

src/zm_image.cpp
```diff
@@ -2001,7 +2001,7 @@ void Image::Annotate(
   const uint8_t bg_b_col = BLUE_VAL_RGBA(bg_colour);
   const uint8_t bg_bw_col = bg_colour & 0xff;
   const Rgb bg_rgb_col = rgb_convert(bg_colour, subpixelorder);
-  const bool bg_trans = (bg_colour == RGB_TRANSPARENT);
+  const bool bg_trans = (bg_colour == kRGBTransparent);
```
connortechnology: this stuff seems unrelated to fmt.  What's the k signify?

Carbenium: I rebased the PR, so those changes are gone now since https://github.com/ZoneMinder/zoneminder/pull/3181 is merged.
The "k" prefix is used to distinguish constants (see Google styleguide: https://google.github.io/styleguide/cppguide.html#Constant_Names)

knight-of-ni:
Looks good. Much cleaner.

Note that we shouldn't need to Require the fmt package in the rpm specfile. Rpmbuild should sniff out the libfmt.so.6 dependency automatically.

connortechnology:
So I reverted because I'm having too much trouble packaging libfmt for bionic/focal/buster etc.

knight-of-ni:
libfmt is in debian and ubuntu repos already. Is there a particular version you are looking for?
https://packages.debian.org/search?searchon=names&keywords=libfmt

connortechnology:
libfmt-dev is in, but not the actual lib.  libfmt7 is in impish, but focal has nothing.

PR #3127, opened 2021-02-04
Cleanup and reorganize includes 

Carbenium:
First of all, apologies for dropping such a large change-set without any forewarning.

**Short summary of the changes**
 * The deprecated C-style headers (`string.h`, `stdio.h`, ...) have been replaced with the C++11 conform ones (`cstring`, `cstdio`, ...)
 * Major cleanup of includes
 * Unify some include guards

**Regarding the include cleanup**

With this commit a unified structure for includes is introduced.
**The general rules**:
 * Only include what you need
 * Include wherever possible in the cpp and forward-declare in the header

 The includes are sorted in a local to global fashion. This means for the **include order**:
   1. If cpp file: The corresponding h file and an empty line
   2. Includes from the project sorted alphabetically
   3. System/library includes sorted alphabetically
   4. Conditional includes

**Tests performed**

I compiled with different optional package combinations to make sure I didn't remove necessary includes.


src/zm_analysis_thread.h
```diff
@@ -1,10 +1,9 @@
 #ifndef ZM_ANALYSIS_THREAD_H
 #define ZM_ANALYSIS_THREAD_H
 
-#include "zm_thread.h"
-#include <signal.h>
-
 #include "zm_monitor.h"
+#include "zm_thread.h"
+#include <csignal>
```
connortechnology: we don't actually use any of the signal stuff in analysis thread.  It can all go.

src/zm_frame.h
```diff
@@ -20,12 +20,16 @@
...
 #define ZM_FRAME_H
 
+#include "zm_time.h"
 #include <sys/time.h>
-#include <sys/types.h>
-class Frame;
 
-#include "zm_event.h"
-#include "zm_time.h"
+typedef uint64_t event_id_t;
```
connortechnology: why move this to here?  It is defining the event id type.  I think it should stay in zm_event.h
I'm guessing it means that we don't need to include zm_event.h .  Feeling this is a bit short sighted.  Frames might need to know more about events in future.

src/zm_videostore.cpp
```diff
@@ -952,11 +949,15 @@ int VideoStore::writePacket(ZMPacket *ipkt) {
   if ( ipkt->packet.stream_index == video_in_stream_index ) {
     return writeVideoFramePacket(ipkt);
   } else if ( ipkt->packet.stream_index == audio_in_stream_index ) {
-    return writeAudioFramePacket(ipkt);
+    if ( monitor->RecordAudio() )
```
connortechnology: So... I don't like doing extra function calls and if's on super frequently called code.  If we aren't recording audio, then audio_in_stream_index = -1.  Also we have a check in Monitor::Capture for this as well.  So it SHOULD never happen that we are given a packet matching audio_in_stream_index.

Carbenium: Apologies, theses changes shouldn't have landed in this PR. I removed them. They are very much WIP.

src/zm_frame.h
Carbenium: Good catch. I moved the typedef now back and solved the circular dependency which triggered this changes differently.

src/zm_analysis_thread.h
Carbenium: Alirght, I removed the references to `<csignal>` from `AnalysisThread`.

PR #3133, opened 2021-02-06
ffmpeg: exit dumpPacket early if debug logging is not enabled

Carbenium:


connortechnology:
I can't help feel that a better way to go would be to turn dumpPacket into a define.  Start with the if logDebugging.  Instead of do procedure call, then if and return. Plus then we could include the line # it is being called from.

Carbenium:
Here we go.

PR #3134, opened 2021-02-06
Image: Replace qsort with std::sort

Carbenium:
std::sort has stricter complexity requirements than qsort and can be better optimized by the compiler due to templating.


connortechnology:
This may actually have a measurable performance impact... that's cool.

PR #3136, opened 2021-02-07
DB: Make connection initialization more predictable and finalize mysql library correctly

Carbenium:
Remove calls to `zmDBConnect` from various places to avoid possible side-effects/double initialization.
The function should be called once from the main thread of the daemon.

Also split config loading into 2 steps: static and DB config loading. Load the static config before `zmDBConnect` is called so it has a chance to succeed.

The call to `mysql_library_end` is re-introduced. No more leaks from mysql on shutdown.

### DIFF
--- a/src/zm_config.cpp
+++ b/src/zm_config.cpp
@@ -36,21 +36,21 @@ void zmLoadStaticConfig() {
 
   // Search for user created config files. If one or more are found then
   // update the Config hash with those values
-  DIR* configSubFolder = opendir(ZM_CONFIG_SUBDIR);
-  if ( configSubFolder ) { // subfolder exists and is readable
+  DIR *configSubFolder = opendir(ZM_CONFIG_SUBDIR);
+  if (configSubFolder) { // subfolder exists and is readable
     char glob_pattern[PATH_MAX] = "";
     snprintf(glob_pattern, sizeof(glob_pattern), "%s/*.conf", ZM_CONFIG_SUBDIR);
 
     glob_t pglob;
-    int glob_status = glob(glob_pattern, 0, 0, &pglob);
-    if ( glob_status != 0 ) {
-      if ( glob_status < 0 ) {
+    int glob_status = glob(glob_pattern, 0, nullptr, &pglob);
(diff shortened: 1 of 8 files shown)


connortechnology:
I think the original idea was that zmDbConnect would manage the not double initializing mysql.  So the global boolean would really only be known about by zmDbConnect. ANd when you get down to it, the dbConn handle works just as well for boolean purposes.  

And I don't like making every executable do 4 lines of code, when we know that they MUST do all four lines.  It's just boilerplate.  It should be 1 line that performs all four tasks.

One other thing that I was working on but got distracted from, was actually compiling in the default values for config loaded from the db.  Then we could only load config values that differ from the default.  

Further I was planning to add some sort of categorization to the config db so that we only load & parse config that is used in each executable.

src/zm_config.cpp
```diff
@@ -30,77 +30,81 @@
...
+  if (!zmDbConnected) {
+    Fatal("Not connected to the database. Can't continue.");
   }
   config.Load();
   config.Assign();
 
   // Populate the server config entries
-  if ( !staticConfig.SERVER_ID ) {
-    if ( !staticConfig.SERVER_NAME.empty() ) {
+  if (!staticConfig.SERVER_ID) {
```
connortechnology, line 70: This should probably be moved to LoadStaticConfig

connortechnology: ALthough it needs db so never mind

Carbenium:
I see your point regarding boilerplate code. Since I'm at least planning to revamp the logging code once we have fmt support, I'm expecting the API to change a bit. Same goes for the DB API. Some type-safe wrappers which handle locking, maybe even prepared statements, etc. would be nice. That's why I didn't spend much time on thinking about that. I don't expect a new daemon gets introduced every other day where this might become a maintenance burden ;)

Regarding the calling locations of `zmConnectDb`: I for myself really don't expect a harmless looking logging initialization function call to start messing around with the DB connections. If the DB connection exists before logging is initialized, then let's also write to the DB. If not, well then don't.

Does that clarify a bit the choices I made?

PR #3145, opened 2021-02-09
AnalysisThread: Move it over to std::thread

Carbenium:


dodexahedron:
Love all this work.
Thanks for this and related stuff. 👍

PR #3151, opened 2021-02-09
AnalysisThread: Make the class un-movable 

Carbenium:
We depend on an unchanged value of "this" after the creation of the thread otherwise undefined values would be accessed within the std::thread.


connortechnology:
Merged but I'm a little unhappy with the way this code is going.  It is completely unfathomable to me and everyone else involved.  Going to need some helpful code comments.

PR #3154, opened 2021-02-10
Monitor: Don't push packets into the packet queue when in Monitor mode

Carbenium:
There is no consumer for the packets in this mode, so don't queue them up. This saves quite a bit of RAM for mode of operation.

### DIFF
--- a/src/zm_monitor.cpp
+++ b/src/zm_monitor.cpp
@@ -2613,7 +2613,8 @@ int Monitor::Capture() {
       shared_data->last_write_time = packet->timestamp->tv_sec;
       image_count++;
 
-      if ( packetqueue.packet_count(video_stream_id) or packet->keyframe or event ) {
+      if (GetFunction() != Function::MONITOR
+          and (packetqueue.packet_count(video_stream_id) or packet->keyframe or event)) {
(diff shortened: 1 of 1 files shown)


connortechnology:
Crap I merged and then realized that this is flawed.  Have to include whether we are doing rtsp serving.  Also NODECT.  Also when in the scope of the class, can just write function.  Don't need to do GetFunction.

connortechnology:
Actually NODECT is iffy.  We are supposed to be able to take alarm commands, so technically should have a primed queue to start writing from.

Carbenium:
Ah yes, the RTSP case is valid. So an additional check for `rtsp_server` should do it, right?

connortechnology:
Yeah I think so, but maybe we should put this logic elsewhere and have it set a simple variable like do_queue_packets.

Carbenium:
What would be nice is letting the queue do this. Meaning it should maintain a list of subscribers/consumers which get notified if new packets are available and if there are none, discard all incoming packets. If you agree, I'd quick-fix this for the moment and then we can think about and discuss a proper implementation in the meanwhile.

connortechnology:
agreed.  Should be simple. if ( ! iterators.empty() )

PR #3157, opened 2021-02-10
Fix a possible deadlock on zmc shutdown and reconnect

Carbenium:
Close the monitor and thus drain the packet queue before trying to stop/join the analysis threads since they might hang while waiting for the next packet to arrive.

Tested with various combinations of signal loss / reconnection / shutdown.

### DIFF
--- a/src/zm_analysis_thread.cpp
+++ b/src/zm_analysis_thread.cpp
@@ -9,7 +9,7 @@ AnalysisThread::AnalysisThread(std::shared_ptr<Monitor> monitor) :
 }
 
 AnalysisThread::~AnalysisThread() {
-  terminate_ = true;
+  Stop();
   if (thread_.joinable())
     thread_.join();
 }
--- a/src/zm_analysis_thread.h
+++ b/src/zm_analysis_thread.h
@@ -13,6 +13,8 @@ class AnalysisThread {
   AnalysisThread(AnalysisThread &rhs) = delete;
   AnalysisThread(AnalysisThread &&rhs) = delete;
 
+  void Stop() { terminate_ = true; }
+
  private:
   void Run();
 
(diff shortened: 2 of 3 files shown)


connortechnology:
It won't work.  the analysis thread can be waiting in the queue.  You won't be able to clear the queue until the analysis thread is done.

You have to stop the analysis thread.  Then close, then join.

Carbenium:
Can you point me to a place where this could happen?

As far as I can see, by calling `PacketQueue:clear()` all potentially blocking calls get non-blocking since `PacketQueue::deleting = true` and `PacketQueue::condition` is notified.

N.B.: For testing I even removed the `zm_terminate` part of the checks from `PacketQueue` to make sure that doesn't interfere (We should do that permanently at some point. Global state makes things often just more complicated).

connortechnology:
If we havn't told the threads to stop, they will wakeup, run and immediately go back into the packetqueue.  They will generally return null because deleting will be true.  They will run at 100% looping going in and out of the packetqueue.  

I've gone through several iterations of the code so the issues are a little jumbled in my head.  Originally the packetqueue was dynamically allocated.  That 100% cpu would continue until I deleted the packetqueue.  Then segfault.  Which is why it is now a member of Monitor instead of a pointer.  But the 100% cpu race of the threads while we clear the packetqueue is not ideal.

This isn't such an issue with the analysis thread because it sleeps for 5 second on receiving a nullptr.  However the rtsp thread doesn't have this sleep. It goes nuts.

Is there a reason why we can't stop the threads first?

connortechnology:
I don't think we need to pass terminate to Analyze().  The thread will be waiting in the packetqueue.  It gets stopped, which sets terminate to true.  Then we call Close which will set the packetqueue's deleteing=true.  Then it signals the analysis which will wake up and return null for the packet.  Analyze() will return.  Notice the terminate and quit.
Close will finish, then we can clear() the threads.

PR #3180, opened 2021-02-28
ALERT also mean warming up with alarm_frame_count value before return…

rabsym:
…ing to ALARM....


rabsym:
Issac, got it working and tried with different values and scenarios, maybe you prefer make the "Info" line introduced a "Debug" line.

PR #3189, opened 2021-03-04
Make last_alarm_count last alarmed frame while only in ALARM state to…

rabsym:
… make post_event_count counts since here not alone alarmed frames that don't shot alarms.


rabsym:
Tested and working in several configs. This is needed because post_event_count is computed after last_alarmed_frame so it should carry last_alarmed_frame belonging to an ALARM state.

PR #3190, opened 2021-03-06
Fix SQL queries with missing lock

Carbenium:
Replace raw `mysql_query` calls with the `zmDb*` functions.
With this we can make sure we have proper locking of our DB connection at all times.

Also replace `stringtf` with a type-safe version that can't overflow.

### DIFF
--- a/src/zm_user.cpp
+++ b/src/zm_user.cpp
@@ -94,33 +94,25 @@ bool User::canAccess(int monitor_id) {
 // Function to load a user from username and password
 // Please note that in auth relay mode = none, password is NULL
 User *zmLoadUser(const char *username, const char *password) {
-  char sql[ZM_SQL_MED_BUFSIZ] = "";
   int username_length = strlen(username);
-  char *safer_username = new char[(username_length * 2) + 1];
 
   // According to docs, size of safer_whatever must be 2*length+1
   // due to unicode conversions + null terminator.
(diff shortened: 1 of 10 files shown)


src/zm_user.cpp
```diff
@@ -94,33 +94,25 @@ bool User::canAccess(int monitor_id) {
...
+  std::string escaped_username((username_length * 2) + 1, '\0');
 
-  MYSQL_RES *result = mysql_store_result(&dbconn);
-  if ( !result ) {
-    Error("Can't use query result: %s", mysql_error(&dbconn));
-    exit(mysql_errno(&dbconn));
-  }
+
+  size_t escaped_len = mysql_real_escape_string(&dbconn, &escaped_username[0], username, username_length);
+  escaped_username.resize(escaped_len);
```
connortechnology, line 105: Is the resize necessary? We are just going to discard this a moment anyways.

connortechnology:
Love this.

src/zm_user.cpp
Carbenium: Sorry for the delay. Somehow I missed your comment...

It's there to discard the surplus `\0` which are still there from the initial construction of the string. Since we call `c_str` afterwards (so we expect a `\0` terminated string) it shouldn't really matter, it's just cleaner IMHO. Performance wise that's negligible.

PR #3195, opened 2021-03-17
Created new control script

therealmichaelberna:
The netcat control script wasn't working for some of the ONVIF PTZ cameras that I was testing, so I made this one. It uses the same structure as commands issued by ONVIF Device Manager, so if your camera works with ONVIF device manager, it will probably work with this.


knight-of-ni:
Thanks for the contribution, but for this to be useful to the average user we also need the sql that goes with it.
Use zmcamtool.pl to export the relevant sql from your system, then add it to this PR in the dbcreation file here:
https://github.com/ZoneMinder/zoneminder/blob/master/db/zm_create.sql.in#L923

therealmichaelberna:
@knight-of-ni 
Thanks Andrew. I added that now.

therealmichaelberna:
@knight-of-ni 
Let me know if anything else is needed on my end and if I did this properly.

connortechnology:
Thanks for your efforts... however... look we have at least 3 or 4 onvif scripts, half of which aren't even onvif.  We were trying to clean them up, and now we have yours as well, with no information on which is better and why.  It is all way too condusing to anyone out there trying to actually use any of them.

I'm not inclined to merge this at this time.  I'd like to see some discussion as to the merits and approach.  Also instead of just adding another script, how about patches to the existing onvif script in small chunks that we can actually read and understand in small bytes.

Otherwise we will have to wait until after releasing 1.36.

therealmichaelberna:
> Thanks for your efforts... however... look we have at least 3 or 4 onvif scripts, half of which aren't even onvif. We were trying to clean them up, and now we have yours as well, with no information on which is better and why. It is all way too condusing to anyone out there trying to actually use any of them.
> 
> I'm not inclined to merge this at this time. I'd like to see some discussion as to the merits and approach. Also instead of just adding another script, how about patches to the existing onvif script in small chunks that we can actually read and understand in small bytes.
> 
> Otherwise we will have to wait until after releasing 1.36.

This one is based off of the SOAP commands.

https://www.onvif.org/wp-content/uploads/2016/12/ONVIF_WG-APG-Application_Programmers_Guide-1.pdf

As to why it is better, it worked for some Chinese onvif cameras that I was trying to use when the other scripts wouldn't. I believe that this is in accord with the official ONVIF standard since ONVIF says they use soap on the PDF guide above.

PR #3196, opened 2021-03-18
Fix build under clang

Carbenium:
Also enable `-Wconditionally-supported` on GCC so we catch this there as well.

### DIFF
--- a/src/zm_remote_camera_http.cpp
+++ b/src/zm_remote_camera_http.cpp
@@ -289,7 +289,7 @@ int RemoteCameraHttp::ReadData(Buffer &buffer, unsigned int bytes_expected) {
     total_bytes_to_read -= bytes_read;
   } while (total_bytes_to_read);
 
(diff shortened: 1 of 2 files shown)


src/zm_remote_camera_http.cpp
```diff
@@ -289,7 +289,7 @@ int RemoteCameraHttp::ReadData(Buffer &buffer, unsigned int bytes_expected) {
     total_bytes_to_read -= bytes_read;
   } while (total_bytes_to_read);
 
-  Debug(4, "buffer: %s", buffer);
+  Debug(4, "buffer size: %d", static_cast<int>(buffer));
```
connortechnology, line 292: just a note: The purpose of this debug line was to view the contents of the buffer, not it's size.

Carbenium: Oh I see. Then I misinterpreted that one. I'll try to restore the functionality in a cross-compiler way.

Issue #3197, opened 2021-03-20
Event Close Mode Time not working on RECORD


SirLouen:
**Describe Your Environment**
* Zoneminder: 1.35.21 (current master)
* From Source
* Ubuntu Focal

**If the issue concerns a camera**
N/A

**Describe the bug**
If I set the Function on RECORD, and set a Length, it doesn't stop the event on the given Event in case the EVENT_CLOSE_MODE is set to time. This only works on MOCORD

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Options, set EVENT_CLOSE_MODE to time
2. Create a Monitor with function RECORD
3. Set any time in Misc -> Event Length, for example 3600s. First event will be longer than 3600 because it rounds up to the first 3600 modulus from the timestamp (for example: at 9:00:00am 10:00:00am 11:00:00am, etc...

**Expected behavior**
Similarly to MOCORD, it should stop on event length if Event Close Mode is set to time.

I will be providing a patch for this

SirLouen:
One of my solutions have been merged, but I think it should be fully acting according to the docs

https://zoneminder.readthedocs.io/en/stable/userguide/options/options_config.html

> EVENT_CLOSE_MODE - When a monitor is running in a continuous recording mode (Record or Mocord) events are usually closed after a fixed period of time (the section length). However in Mocord mode it is possible that motion detection may occur near the end of a section. This option controls what happens when an alarm occurs in Mocord mode. The ‘time’ setting means that the event will be closed at the end of the section regardless of alarm activity. The ‘idle’ setting means that the event will be closed at the end of the section if there is no alarm activity occurring at the time otherwise it will be closed once the alarm is over meaning the event may end up being longer than the normal section length. The ‘alarm’ setting means that if an alarm occurs during the event, the event will be closed once the alarm is over regardless of when this occurs. This has the effect of limiting the number of alarms to one per event and the events will be shorter than the section length if an alarm has occurred.

In fact the new conditions regarding to the MOCORD option, I'm not 100% sure if they are well coded.

Theorically the 3 scenarios are

CLOSE_TIME = regardless if RECORD or MOCORD, on TIME, close. This doesn't affect the MOCORD scenario with the current conditional
CLOSE_IDLE = for this conditional, this is equivalent to TIME, for RECORD it should remind unaffected, but for MOCORD, its identical to RECORD
[...]

PR #3198, opened 2021-03-20
Issue #3197 Add RECORD to Event_Close_Mode time

SirLouen:
This is the solution related to the Issue #3197


SirLouen:
I think that RECORD shall be related to CLOSE_TIME  event_close_mode while I'm not sure about MOCORD logic and why it was placed there, so I will leave it untouched as it was originally.

PR #3200, opened 2021-03-22
Issue #3197 Improvements according to the diagram

SirLouen:
In the #3197 I've placed a diagram to explain the improvements in this conditionals according to the docs logic


connortechnology:
The code has changed so much, this PR is no longer needed and could never apply.

PR #3204, opened 2021-03-30
Amcrest PTZ script error

alabamatoy:
First time I have attempted this process, so if I am doing it wrong, please enlighten me.

In my distro, the file is /usr/share/perl5/ZoneMinder/Control/Amcrest_HTTP.pm

Line 78 changed from:   
my $res = $self->{ua}->get($$self{base_url}.'cgi-bin/ptz.cgi');

to:
my $res = $self->{ua}->get($$self{base_url}.'cgi-bin/magicBox.cgi?action=getDeviceType');

The cameras do not like the ...ptz.cgi call with no values passed, and return a error 500 or error 400.  This new call does nothing but return the type of camera, which is not saved.

Also change line 130 from:
    Error("Failed to get $$self{base_url}cgi-bin/ptz.cgi ".$res->status_line());
to:
    Error("Failed to get $$self{base_url}cgi-bin/magicBox.cgi?action=getDeviceType ".$res->status_line());
This is to make the error message match the actual call being made, no functionality is changed.

I have tested this on 2 different systems with several different Amcrest cameras.


connortechnology:
Not sure what you did with this PR but it's no good.  I applied your changes manually though.

PR #3205, opened 2021-04-03
Cleanup zm_utils and add unit tests

Carbenium:
Unit tests were added before the cleanup to ensure no semantic changes were introduced (except for the rework of one `Split` overload, which is unused).


connortechnology:
I can't believe we actually have tests now.  So awesome.  Thanks so much!

PR #3210, opened 2021-04-10
FifoSource: Adapt to changed xop::AVFrame API

Carbenium:
See https://github.com/ZoneMinder/RtspServer/pull/3 for the API change.

I'll update this PR with the correct submodule update/packpack update once the referenced PR is merged.


connortechnology:
I personally was thinking of adding another constructor that takes the buffer size and type so that we could reduce it to one line of code....

Carbenium:
Yeah, thought about that one as well. I guess since we are changing RtspServer's API already we can go that way. I'll whip up something in a moment.

Carbenium:
https://github.com/ZoneMinder/RtspServer/pull/4

PR #3211, opened 2021-04-11
comms: Fix Woverloaded-virtual warnings

Carbenium:
Give `zm_comms.{h,cpp}` an auto-format treatment and add some basic unit tests.
Also enable `Woverloaded-virtual` warnings on GCC (they are on per default on clang).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -18,5 +18,5 @@ task:
     - make install
 
   test_script:
-    - cd build
-    - CTEST_OUTPUT_ON_FAILURE=1 make test
+    - cd build/tests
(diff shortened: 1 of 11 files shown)


connortechnology:
This is great, but I'm not sure how much we actually use zm_comms...  maybe in future check with me to determine what might be deprecated code before you go spending time on it.

Carbenium:
It's only used in the RTP client implementation (which I know will be removed). Anyway it gave me a chance to take again a look at the POSIX socket API. So no wasted time :smiley:
I'm working towards a warning free build on all platforms (so we can switch on -Werror on CI jobs).

PR #3212, opened 2021-04-11
SessionDescriptor: Fix a ffmpeg deprecation warning

Carbenium:
AVStream::codec is deprecated. The replacement AVStream::codecpar is initialized at the end of the method from the codec context.


connortechnology:
We are still supporting versions of linux that used older ffmpeg libs.  I think we still need codec for those.

connortechnology:
Oh, I think I see what you mean.  I expect this to cause crashes due to other bad code but that is ok.

PR #3214, opened 2021-04-11
Fix warnings reported by Wtype-limits and re-enable the warning on GCC

Carbenium:


src/zm_image.cpp
```diff
@@ -2051,10 +2051,6 @@ void Image::Annotate(
       for ( unsigned int y = lo_line_y, r = 0; y < hi_line_y && r < char_height; y++, r++, ptr += width ) {
         unsigned char *temp_ptr = ptr;
         for ( unsigned int x = lo_line_x, c = 0; x < hi_line_x && c < line_len; c++ ) {
-          if ( line[c] > 0xFF ) {
-            Warning("Unsupported character %c in %s", line[c], line);
-            continue;
-          }
           uint64_t f = font_bitmap[(line[c] * char_height) + r];
```
connortechnology, line 2054: Why are you removing this, it serves a purpose.

Carbenium: The check was introduced in https://github.com/ZoneMinder/zoneminder/commit/bd287d7c7c6caac34a73679f9469e4e513d1b12a where it had the form
```
            if ( (line[c] * ZM_CHAR_HEIGHT * size) + r > sizeof(bigfontdata) ) {
```
In https://github.com/ZoneMinder/zoneminder/commit/528699f942c2dbbde607e4ae0c56ab1703f98ef3 this was changed to
```
          if ( line[c] > 0xFF ) {
```
Excerpt from the commit messages: 

> line[c] is a char. There is no possibility for it to be larger than 0xff.

I might be missing something, but I can't see what they should be protecting us from in the current code.

connortechnology:
It is entirely likely that it is bogus.  But we do need some check for implemented characters.  In 528699f @hax0kartik implemented a new font data structure and supposedly faster font rendering.  We will have to look closer.

PR #3234, opened 2021-05-16
Rework API of geometry classes

Carbenium:
This patch series cleans up the public interfaces of `Vector2` (previously `Coord`), `Box` and `Polygon`. Additionally the `Line` and `LineSegment` classes are introduced. Unit tests are also provided.
The Sutherland-Hodgman algorithm has been implemented to clip convex and concave polygons with convex boundary boxes.

Additionally this fixes some bugs found on the way:
 * We now actually clip zone polygons extending past the image size
 * Fix a memory overflow in the polygon fill algorithm
 * Clip polygons provided on the CLI when dumping zone images via `zmu -z`


connortechnology:
I don't require you to change it but Vector2 means nothing to me.  Coord at least is a very definite thing.  I might have been tempted to just provide an alias when using it for Vector or call it CoordOrVector.  Maybe discuss name changes before doing the work. Naming can be hard.

Super work. I see a couple of warnings in there so I'll wait a bit before merging in case you want to address them.

Carbenium:
The reasoning for changing to `Vector2` (see the commit message from the appropriate commit): 
```
The class is not only used to represent coordinates but also lengths in XY.
Vector2 is a more fitting/general name for this purpose.
```
So it should be interpreted purely in the mathematical sense. And the `2` since it is a 2-component vector. That comes more or less from computer graphics conventions since often there you distinguish between 2D and 3D objects. Does that sound okay? (And yes, naming is the hardest problem in IT :smile: )

Where do you see warnings? CI doesn't complain and it builds locally without warnings as well.

connortechnology:
I'm fine with the name now you've done the work, but I might have pushed for something more obvious to ME.  Anyways.


Check warning on line 56 in src/zm_vector2.h

Code scanning

Multiplication result converted to larger type
Multiplication result may overflow 'double' before it is converted to 'long double'.
Show more details

Carbenium:
Ah, yes indeed. Let me fix those. Weird that those code scan results are not presented a bit more prominently...

Carbenium:
Looks better now.

PR #3235, opened 2021-05-17
Remove variable-length arrays from the codebase

Carbenium:
VLAs are a C99 feature that never was specified for C++ (due to better alternatives). Remove VLA usage from the codebase and switch on `Wvla` so we get notified on accidental introduction.

### DIFF
--- a/src/zm_logger.cpp
+++ b/src/zm_logger.cpp
@@ -529,15 +529,17 @@ void Logger::logPrint(bool hex, const char *filepath, int line, int level, const
   if (level <= mDatabaseLevel) {
     if (zmDbConnected) {
       int syslogSize = syslogEnd-syslogStart;
-      char escapedString[(syslogSize*2)+1];
-      mysql_real_escape_string(&dbconn, escapedString, syslogStart, syslogSize);
+      std::string escapedString;
+      escapedString.reserve((syslogSize * 2) + 1);
+      mysql_real_escape_string(&dbconn, &escapedString[0], syslogStart, syslogSize);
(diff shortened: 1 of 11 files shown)


src/zm_logger.cpp
```diff
@@ -529,15 +529,16 @@ void Logger::logPrint(bool hex, const char *filepath, int line, int level, const
   if (level <= mDatabaseLevel) {
     if (zmDbConnected) {
       int syslogSize = syslogEnd-syslogStart;
-      char escapedString[(syslogSize*2)+1];
-      mysql_real_escape_string(&dbconn, escapedString, syslogStart, syslogSize);
+      std::string escapedString;
+      escapedString.reserve((syslogSize * 2) + 1);
+      mysql_real_escape_string(&dbconn, &escapedString[0], syslogStart, syslogSize);
```
connortechnology, line 534: Isn't this considered bad form writing to the array address of the std::string?

Carbenium: Since C++11 this is actually supported since the standard guarantees a continuous internal buffer. C++17 even added a non-const overload for `.data()` which makes this even nice to read.

I added one small thing (would have worked without it in our use-case): Shrink the string to `strlen` of the escpaed one so that `size()` works as well.

connortechnology:
There is a problem when using gnutls, I zm_rtsp_auth line 145.  we pass a pointer to md5len which you made constexpr

Carbenium:
Indeed. I'll try to come up with something.

PR #3241, opened 2021-05-18
Fix 3239

### LINKED ISSUE #3239: The Zoneminder API does not work correctly for alarm operations when using username/password authentication
**Describe Your Environment**
- Version of ZoneMinder: 1.36.0 (1.36.0-bionic1)
- How you installed ZoneMinder: PPA
- Full name and version of OS: Ubuntu 18.04.5 LTS

**Describe the bug**
The Zoneminder API does not work correctly for alarm operations when using username/password authentication. Operations fail silently.

**To Reproduce**
Make an API request for the status of a monitor using username/password authentication.

```
curl 'https://example.org/zm/api/monitors/alarm/id:4/command:status.json?username=username&password=********'
{"status":""}
```

**Expected behavior**
The API call should return the monitor status rather than a blank. This works correctly when using token-based authentication.

```
https://example.org/zm/api/monitors/alarm/id:4/command:status.json?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJab25lTWluZGVyIiwiaWF0IjoxNjIxMjk1OTU1LCJleHAiOjE2MjEzMDMxNTUsInVzZXIiOiJuZWlsbCIsInR5cGUiOiJhY2Nlc3MifQ.8f3FodI9lDYk1Vhu2OZxlmL5cd_Hurfl2YB9kC3MJqY
{"status":"0"}
```

**Additional information**
The API operations to set a monitor alarm state on or off also fail silently when using username/password authentication.

**Debug Logs**
No errors are generated in the debug logs.

connortechnology:
Fixes #3239

### DIFF
--- a/web/api/app/Controller/MonitorsController.php
+++ b/web/api/app/Controller/MonitorsController.php
@@ -266,7 +266,7 @@ class MonitorsController extends AppController {
       if ( $mToken ) {
         $auth = ' -T '.$mToken;
       } else if ( ZM_AUTH_RELAY == 'hashed' ) {
-        $auth = ' -A '.generateAuthHash(ZM_AUTH_HASH_IPS);
+        $auth = ' -A '.calculateAuthHash(ZM_AUTH_HASH_IPS?$_SERVER['REMOTE_ADDR']:'');
       } else if ( ZM_AUTH_RELAY == 'plain' ) {
         # Plain requires the plain text password which must either be in request or stored in session
         $password = $this->request->query('pass') ? $this->request->query('pass') : $this->request->data('pass');;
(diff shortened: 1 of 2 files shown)


mattkasa:
@connortechnology :wave: this MR appears to have broken some integrations that relied on this API:

https://github.com/home-assistant/core/issues/50949

It looks like we're getting back a string `Ok` for `status` from the status endpoint instead of the integer return code from the `exec` like we used to. Is that the intended behavior going forward and am I reading it correctly that it can it now be either string or int?

PR #3255, opened 2021-05-23
Enable gitlab CI to build debian packages

anon8675309:
Adding a .gitlab-ci.yml file so my infrastructure can build .deb packages automatically.

I wanted to reuse as much of the existing code as possible, so I added some options to do_debian_package.sh and made sure they defaulted to values which ensure that everything functions as it did before if they are not specified.

The `yes ""` is to accept defaults when prompted by things like apt, asking if it's okay to install dependence.  If apt detects that the signature doesn't match, it will make the default choice "no".  That is why I chose to always accept the default, rather than always answering "yes".

The || true at the end is required either because of a [bug/limitation in GitLab's CI](https://stackoverflow.com/questions/65412154/gitlab-runner-throws-cleaning-up-file-based-variables-0001-error-job-failed), or there's some command in do_debian_package.sh which is returning a non-zero exit status.  To be clear, do_debian_package.sh builds everything correctly whether run from GitLab CI or in a standalone environment.  The .deb files are produced, they install fine, the dependencies get installed when the package is installed, the database gets set up correctly... really, it's fine.

Instead of changing a bunch of code in do_debian_package.sh to try to make sure everything returns with an exit code of zero (or hacking around limitations of GitLab CI, if that's the problem), so I just slapped the `|| true` on the end.

Finally, the reason the CI specifically builds branch 1.36.1 is because there was some problem of automatically detecting the branch based on `--snapshot=stable`.  More specifically, `$(git rev-list --tags --max-count=1)` returned an empty string on [line 114](https://github.com/ZoneMinder/zoneminder/blob/master/utils/do_debian_package.sh#L114).  Again, this may be because of some limitation of GitLab CI.  I have not yet attempted to debug it and didn't want to muddy up the "add a CI file" pull request up with other changes which were not strictly necessary.  I will likely chase that down in a future release with another small pull request.

Thanks in advance for allowing me to contribute. :heart:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,17 @@
+default:
+  image:
+    name: ubuntu:latest
+  before_script:
+    - apt-get update -yq
+    - DEBIAN_FRONTEND=noninteractive apt-get install -yq devscripts sudo
+
+deb:
+  stage: build
+  tags:
+    - docker
+  script:
+    - yes "" | ./utils/do_debian_package.sh --snapshot=stable --branch=1.36.1 --type=binary --interactive=no --dput=no --debbuild-extra=--no-sign || true
+  artifacts:
+    paths:
+      - '*.deb'
+    expire_in: 1 week
--- a/utils/do_debian_package.sh
+++ b/utils/do_debian_package.sh
@@ -58,6 +58,14 @@ case $i in
     PACKAGE_VERSION="${i#*=}"
     shift
     ;;
+    -x=*|--debbuild-extra=*)
+    DEBBUILD_EXTRA="${i#*=}"
+    shift
+    ;;
+    --dput=*)
+    DPUT="${i#*=}"
+    shift
+    ;;
     --default)
     DEFAULT=YES
     shift # past argument with no value
@@ -306,9 +314,11 @@ EOF
   if [ "$DEBSIGN_KEYID" != "" ]; then
     DEBUILD="$DEBUILD -k$DEBSIGN_KEYID"
   fi
+  # Add any extra options specified on the CLI
+  DEBUILD="$DEBUILD $DEBBUILD_EXTRA"
   eval $DEBUILD
   if [ $? -ne 0 ]; then
-  echo "Error status code is: $?"
+    echo "Error status code is: $?"
     echo "Build failed.";
     exit $?;
   fi;
@@ -344,8 +354,10 @@ EOF
         dput $PPA $SC
       fi;
     else
-      echo "dputting to $PPA";
-      dput $PPA $SC
+      if [ "$DPUT" != "no" ]; then
+        echo "dputting to $PPA";
+        dput $PPA $SC
+      fi;
     fi;
   fi;
 done; # foreach distro


connortechnology:
I'm certainly happy to merge the do_debian-package bits.  I'd like a little more information about what/how this integration can be used.  I'm currently looking at automatically building .debs for debian on all the various architectures and I kinda hate docker and packapack etc.  Might like to get away from ubuntu's launchpad as well.

anon8675309:
# Automation
If someone has their own GitLab server, they can import the ZoneMinder repo and create a webhook to fire when there is activity on the GitHub repo ([docs](https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html)).  I've only every set up CI when everything is in GitLab, so I might have some of the details surrounding the webhooks wrong.  I'm confident that it's possible, based on the documentation, but it may take some more code changes.

# Jobs
When it's time for the CI to run (either automatically, or when manually triggered) GitLab will create a pipeline with the one job I have configured in .gitlab-ci.yml.  A GitLab CI runner will see this and pick start running the commands.  I am doing so in a fresh docker container each time to make sure it'll fail if there are any missing dependencies.

The CI runner will do its thing, and produce [artifacts](https://docs.gitlab.com/ee/ci/pipelines/pipeline_artifacts.html#pipeline-artifacts).  It's just a zip file containing all the .deb files.  If the .deb files are going to be signed, it'd be reasonable to have the CI runner have the PGP key to do the signing, although that's not the only option.

# More automation
It would be possible to write a script that watches the GitLab instance for artifacts, downloads them, and then sends them to one's private repo.

# Putting it all together
The ultimate goal, in my mind, is to allow anyone to set up their own GitLab server and GitLab Runner to have .deb files just automatically appear in their private debian/ubutntu repo.  As I understand it, it'd go like this:

1. Tags would be created on the repo in GitHub (core developers)
2. The user's webhook would fire, notifying their GitLab server
3. Their GitLab CI job to build the latest version would fire
4. Another machine (possibly the deb repo itself) will run a daily script looking for artifacts, download them, put them into the repo, re-run `dpkg-scanpackages`
5. Any ZoneMinder machines would get the new version the next time they `apt update && apt upgrade`

If do_debian_package.sh has flags to build packages for other architectures (arm64, MIPS, etc.), they could be generated with jobs just like the one I set up to generate the amd64 package.  So instead of just building one package in step 3 above, it'd build all of them.

# Conclusion
All this is really about the automation I want.  I'm not sure which parts of this are also interesting to others.

My motivation is that I don't want to rely on 3rd party services and I want things to work in as many environments as practical.  I have a copy of the code, I can build it, don't need 3rd party debian/ubuntu repos, and the package can be installed in a VM, in docker, on a physical machine, etc.  Furthermore, I want others to be able to easily reproduce my work, if they are so inclined.

If someone really wanted to go wild, they could set up a cron job to use terraform to spin up a fresh VM and install ZoneMinder from the repo on a regular basis.  That'd provide end-to-end testing to verify that the .deb packages aren't somehow messed up.

Let me know if there's anything you'd like to accomplish and I'll see what I can do to help solve it. :slightly_smiling_face:

PR #3268, opened 2021-05-30
CI: Remove Travis CI

Carbenium:
Travis is substantially slower than GitHub Actions. Let's use GitHub Actions to run ESLint and test our packaging scripts (currently Buster).

@connortechnology I didn't port the `rsync_xfer.sh` invocation. Is it still used?


SteveGilvarry:
Was going to try using Actions for packaging instead of PackPack, but if planning to do something along these lines I will hold off.

connortechnology:
The rsync_xfer stuff is to copy the .deb's to zmrepo.  But now zmrepo builds it's own packages using a different script, so i think we are ok.

Carbenium:
@SteveGilvarry No further plans on my side. This was really just to move away from Travis.

PR #3277, opened 2021-06-04
Remove libavresample support

Carbenium:
libavresample hasn't been maintained for a long time by FFmpeg since libswresample superseded it.
In 2018 it was officially deprecated [1].

Let's remove the support for it since there is no need to maintain this option further.

[1] https://patchwork.ffmpeg.org/project/ffmpeg/patch/20171225175335.18183-1-atomnuker@gmail.com/


connortechnology:
I actually like it better, but yeah time to remove it.  

Same with xenial and other older ffmpeg support.

Issue #3311, opened 2021-07-02
Longitude DB column doesn't allow three digits left of the decimal point.


KentigernEnnis:
**Describe Your Environment**
v1.36.5
PPA
Ubuntu 18.04.5 LTS
Version 91.0.4472.124 (Official Build) (64-bit) on Windows 10

**Describe the bug**
I upgraded to the latest version today and went to save my latitude/longitude for a camera and consistently got a 500 error.  The logs indicated that the Longitude value was too large to fit.  If I changed the value to be between -100 and 100, it saves no problem. 

**To Reproduce**
1. Edit a monitor
2. Go to Location
3. Enter a Longitude that's west of Winnipeg (Denver, for example: 39.73939286, -104.9848040)
4. Save
5. See 500 error

**Expected behavior**
Technically, since the Longitude can go from -180 to 180, these values should be permitted.

**Debug Logs**
This didn't get logged properly because of a parsing error in the javascript for the rows, but I was able to get at this with a breakpoint:
```
[...]

PR #3312, opened 2021-07-02
Updating the size of the Longitude field

KentigernEnnis:
To address https://github.com/ZoneMinder/zoneminder/issues/3311

I'm unfamiliar with this particular system for DB migrations (First time contributing to an open source project).
I think this should work but I am unsure of how best to test this to ensure that it's effective.


connortechnology:
Mmostly good, just that the zm_update has to be current_version + 1.  I'll fix it up.  Will also merge to 1.36.

PR #3315, opened 2021-07-06
Misc. cleanup of DB related code

Carbenium:
Most important: getting rid of `static` char buffers in function scopes. `static` variables are thread-unsafe. Let's only use them when really needed.


connortechnology:
very nice.

PR #3319, opened 2021-07-18
Gitlab ci

anon8675309:
This is a change to have GitLab CI build the latest version instead of always building v1.36.1. The original PR had it hard-coded because the automatic detection of the latest tag wasn't working in the CI environment (despite working just fine when running the same commands manually).

The workaround is to search distros/fedora/zoneminder.spec for the latest version number. This is admittedly inferior to using git tags, and so it is only ever used if the git tags fail to get the version number. The fallback method should always work, as that .spec file seems to be a reliable format and updated on each release.

Most importantly it doesn't require any extra work for the maintainers (i.e., you), such as requiring someone to update a VERSION file on each release.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,7 +10,7 @@ deb:
   tags:
     - docker
   script:
-    - yes "" | ./utils/do_debian_package.sh --snapshot=stable --branch=1.36.1 --type=binary --interactive=no --dput=no --debbuild-extra=--no-sign || true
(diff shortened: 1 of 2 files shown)


knight-of-ni:
Wouldn't it be easier to get the version number from the version file in the root of the repo?
https://github.com/ZoneMinder/zoneminder/blob/master/version

Sometimes we forget to bump the version in the rpm specfile at the same time we bump the version file.

anon8675309:
It would be far easier and more reliable to use the version file. I must have missed it somehow when I was trying to find places where the version number was in a file. I'll put in a pull request to change it; that is way cleaner. Ty!

PR #3322, opened 2021-07-29
Add support for Accu-Cams (Eufy for example)

StefCGN:
Hi,
is it please possible, to add some technicial stuff for the Eufy Cams, which are battery based and only send, when an event is triggered?

II mean, ZoneMinder is recording, but also floods my logfile, cause of the cams are not reachable.
Furthermore, i mostly miss the first 7 seconds of a record, because ZM is polling the cam every 10 seconds.
Could this be increased somehow?

Thank you!


SteveGilvarry:
I think you meant to log an issue not a Pull Request (Which is used to merge changes you have made to the project, but this has no code from you). But also we don't use Guthub issues for requests. Please discuss in forums or slack channel.

PR #3335, opened 2021-08-21
Fixed "Call to undefined function" Exception in control_functions.php

r01k:
Fixed bug causing "Call to undefined function Error()" in control_functions.php. Problem was due to logger.php not being included in control_functions.php.


connortechnology:
I don't think the require_once is necessary but doesn't harm anything.  And technically you should include anything you use so that the code can be reused externally . So just the ZM\ was really necessary.  Thanks!

PR #3355, opened 2021-09-16
Set mysql character set to utf8 explicitly

ColorfullyZhang:
Set mysql character set to utf8 explicitly to support chinese characts (or other special characters).

Zm will report errors if it enconter chinese characters in storage path

<html>
<body>

Can't mkdir /mnt/a/????/events/1: No such file or directory | zm_event.cpp | 130
-- | -- | --

</body>
</html>

 In fact, the `????` here in the path is actually the chinese characters "中文目录". I can set the path contains chinese characters correctly, but zm will no longer be able to save the records due to the charset problem.

After set the charset as utf8 explicitly in zm_db.cpp, evething works fine.


connortechnology:
We will give this a test for a bit in master and if all is good, put it in 1.36.8.

PR #3358, opened 2021-09-19
<distrorel> typo fix

5472qaywsx:
Original text had the parameter <distrorel> and later on <distrorev>; as it means *release*, I corrected <distrorev> into <distrorel>.


Carbenium:
Thanks! :shipit:

PR #3361, opened 2021-09-29
Added Debian 11 (bullseye) installation instructions

criadoperez:


docs/installationguide/debian.rst
```diff
@@ -3,6 +3,92 @@ Debian
...
+
+**Step 4:** Setup permissions for zm.conf
+
+To make sure zoneminder can read the configuration file, run the following command.
+
+::
+
+    sudo chgrp -c www-data /etc/zm/zm.conf
+
+**Step 5:** Edit Timezone in PHP
```
connortechnology: This section is no longer required.  You set the timezone in options now.

docs/installationguide/debian.rst
```diff
@@ -3,6 +3,92 @@ Debian
...
+
+Run the following commands.
+
+::
+
+    sudo apt update
+    sudo apt upgrade
+    sudo apt install zoneminder
+
+**Step 3:** Initialize database
```
connortechnology: If you install mysql/mariadb first, then the db will be automatically created and this section will no longer be necessary.

docs/installationguide/debian.rst
criadoperez: Great! I removed this in a new commit.

docs/installationguide/debian.rst
criadoperez: Great! I updated this in a new commit.

PR #3404, opened 2022-01-04
Adds ONVIF motion Detection

jp-bennett:
Not ready for deployment, proof of concept only.


connortechnology:
My only comment on this at this time is that Poll and PollThread is a little generic.  Would prefer ONVIFEventPoll or something a little more specific about what it is.

jp-bennett:
> My only comment on this at this time is that Poll and PollThread is a little generic. Would prefer ONVIFEventPoll or something a little more specific about what it is.

That's fair. My thought was that this thread would eventually support APIs other than ONVIF, like the Hikvision API. Though we could split that out into separate threads if that makes more sense. Up to you.

src/zm_monitor.cpp
```diff
@@ -66,6 +66,11 @@
 #define MAP_LOCKED 0
 #endif
 
+struct Namespace namespaces[] =
```
jp-bennett, line 71: This is a workaround for RHEL and derivatives.

src/zm_monitor.cpp
```diff
@@ -1713,6 +1750,47 @@ void Monitor::UpdateFPS() {
...
 
+//Thread where ONVIF polling, and other similar status polling can happen.
+//Since these can be blocking, run here to avoid intefering with other processing
+bool Monitor::Poll() {
+  if (ONVIF_Healthy) {
+    set_credentials(soap);
+    int result = proxyEvent.PullMessages(response.SubscriptionReference.Address, NULL, &tev__PullMessages, tev__PullMessagesResponse);
+    if (result != SOAP_OK) {
+      if (result != -1) //Ignore the timeout error
+        Warning("Failed to get ONVIF messages! %i", result);
```
jp-bennett, line 1771: We may need to add busy loop detection or prevention here.

PR #3409, opened 2022-01-15
Refactor Montage view js -- Janus in MonitorStream

jp-bennett:


src/zm_monitor.cpp
```diff
@@ -1120,7 +1121,9 @@ bool Monitor::connect() {
 #if HAVE_LIBCURL    //janus setup. Depends on libcurl.
     if (janus_enabled && (path.find("rtsp://") !=  std::string::npos)) {
       if (add_to_janus() != 0) {
-        Warning("Failed to add monitor stream to Janus!");
+        if (add_to_janus() != 0) {
```
connortechnology, line 1124: What's this all about? Just trying twice?  Maybe comment why so that someone doesn't remove it later.

jp-bennett: Hoping to eventually replace this with something better, but noted. Next PR will either include a comment, or a better solution.

PR #3415, opened 2022-01-26
Adds Amcrest MD, makes libcurl a hard requirement, and removes a redundant JS define.

jp-bennett:


PrplHaz4:
This will be a great feature for ZM!

Issue #3417, opened 2022-01-31
Remove TravisCI Badge


baileytj3:
**Describe Your Environment**
- Version of ZoneMinder
  - Latest
- How you installed ZoneMinder
  - NA
- Full name and version of OS
  - NA
- Browser name and version
  - NA

**Describe the bug**
Readme badge indicating failing TravisCI build when no longer using TravisCI.

**To Reproduce**
1. Go to 'https://github.com/ZoneMinder/zoneminder/blob/master/README.md'
2. View Badges

**Expected behavior**
No TravisCI badge.

PR #3418, opened 2022-01-31
Remove TravisCI Badge from README

baileytj3:
Remove the TravisCI badge from the README since it is no longer valid.

Fixes: #3417


baileytj3:
Fixed by another commit.

PR #3421, opened 2022-02-02
Update MariaDB Bullseye info in docs

JedMeister:
Specifically note that setting a root password is not required (or recommended; unix socket authentication is better than simple password auth).

### DIFF
--- a/docs/installationguide/debian.rst
+++ b/docs/installationguide/debian.rst
@@ -35,7 +35,7 @@ Run the following commands.
     sudo apt install mariadb-server
     sudo apt install zoneminder
 
-When mariadb is installed for the first time, it doesn't add a password to the root user. Therefore, for security, it is recommended to run ``mysql secure installation``.
+By default MariaDB uses `unix socket authentication`_, so no root user password is required (root MariaDB user access only avaialble to local root Linux user). If you wish, you can set a root MariaDB password (and apply other security tweaks) by running `mariadb-secure-installation`_.
 
 **Step 3:** Setup permissions for zm.conf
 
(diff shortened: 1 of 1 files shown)


docs/installationguide/debian.rst
```diff
@@ -35,7 +35,7 @@ Run the following commands.
     sudo apt install mariadb-server
     sudo apt install zoneminder
 
-When mariadb is installed for the first time, it doesn't add a password to the root user. Therefore, for security, it is recommended to run ``mysql secure installation``.
+By default MariaDB uses `unix socket authentication`_, so no root user password is required (root MariaDB user access only avaialble to local root Linux user). If you wish, you can set a root MariaDB password (and apply other security tweaks) by running `mariadb-secure-installation`_.
```
connortechnology: There's a typo in there.  Also debian does the --defaults-file=/etc/mysql/debian.cnf thing.  I have meant to do some reading on people's points of view on best practices.  redhat don't do that for example.  We should probably mention the --defaults-file thing though.

JedMeister: Oops, apologies on the typo.

I've fixed that now.

> I have meant to do some reading on people's points of view on best practices.

IME since the switch to MariaDB (IIRC Debian Jessie/8?), the Debian defaults are much more sane (and therefore secure) than the old MySQL defaults. Plus blindly adding a root password is likely reducing overall security (generally unix socket auth is considered more secure and incidentally blocks remote root user access as well).

Having said that, I get that this sort of thing can be contentious, so rather than being too opinionated, I included the links to the relevant MariaDB docs so users can make up their own minds on which path to go.

PR #3438, opened 2022-02-17
Rework the state machine code

connortechnology:
Move blending code up to motion detection area where it should be.  Remove event start/stop from the state machine transition logic.  Redo the event open/close logic after the state machine logic using switches.

Todo: finish removing TAPE. Move event_close_mode into each monitor, add an event start time modulus.


connortechnology:
This is a major change, would love some input on whether people find it easier to follow and if I've missed anything.

PR #3442, opened 2022-02-22
Lazy load images

BHMSD-bdailey:
We are planning to run what I would consider a rather "large" zoneminder deployment. We are currently run 64 cameras on a 128 monitors across 2 servers. I final install will be about 180 cameras on 360 Monitors across 5 servers. This is a simple fix that will help with page load times and distributing the creation of event snapshots as the events page scrolls insteads of loads. I am also using this PR as an opportunity to work out the contribution process. Thank you!


connortechnology:
Sweet!  Thank-you so much for your contribution!  Your timing is good, I am working on 1.36.13 and will include this.

connortechnology:
First thoughts... in event list.. this REALLY slows down thumbnail loading... as in nothing appears for at least 10 seconds and then they trickle in...

I have a branch that I never completed that turns off images that are off-view on montage...if you end up going down that rabbit hole let me know and I'll dig it up.

BHMSD-bdailey:
Excellent! Which browser are you testing in? Before I put in future work is there a good place to talk about features? May I get added to slack? I was a little confused by the contribution guidelines. The talked about opening an issue to discuss and then making branches based of the issue number but the issues and templates state they are bugs only. Sorry for the confusion.

connortechnology:
We are not that formal.  As the number of contributors goes up, more formal procedures are needed.  Right now, you are doing fine as it.  Slack shouldn't need an invite, just follow from the link on zoneminder.com.  I'll see if I can send you an invite though.
We actually have two Slack's, which we should really merge.  Anyways, baby steps.

PR #3453, opened 2022-03-09
Show recording status to console

gonzalezcalleja:
Console shows Capturing and Analysing status in function column, I think "Recording" status is also important in this column.


connortechnology:
Agreed, but what I'd like to see is current status, and colour code based on setting (it can be changed at runtime).  Also instead of capturing, it should say capturing on motion etc.  A lot more work to be done here.

PR #3482, opened 2022-05-12
Remove Cambozola variable and Dependency

jp-bennett:
This seems to be what has broken the RPM builds.


knight-of-ni:
Thanks for looking into this @jp-bennett. Due to my work travel schedule I have not been around as much lately.

connortechnology:
I swear I remember removing those lines.  Good work.

PR #3504, opened 2022-06-08
Set en_us.php date/time locale to the proper mm/dd/yyyy format

andrewvillano:
Changed the en_us.php locale file to use the en_US date/time format of mm/dd/yyyy HH:MM:SS, which is expected when the Locale is set to en_US

### DIFF
--- a/web/lang/en_us.php
+++ b/web/lang/en_us.php
@@ -70,6 +70,7 @@ require_once( 'lang/en_gb.php' );
 // setlocale( LC_TIME, 'en_GB' ); Date and time formatting 4.3.0 and after
 
 // Simple String Replacements
+setlocale( LC_TIME, 'en_US' );
 $SLANG['24BitColour']          = '24 bit color';
 $SLANG['32BitColour']          = '32 bit color';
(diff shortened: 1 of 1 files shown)


connortechnology:
Yes, very new.  php8.1 requirements forced us to finally address this.  So we did.  Will get merged to 1.36.20 once we do some testing and iron out exactly the issues that you are having, namely what are some good examples of things to enter in these fields.

andrewvillano:
As far as a regex goes ^(0[1-9]|1[012])[\/](0[1-9]|[12][0-9]|3[01])[\/](19|20)\d\d$

01/01/2000

05/20/2022

06/09/2022

Two Digit Month, Two Digit Day, Four digit Year



[EASTERN DISTRICT OF NEW YORK_color3]
Mr. Andrew V. Villano , CCNA, RHCSA
Linux Administrator
United States District Court
225 Cadman Plaza East, Room 304S
Brooklyn, NY 11201-1818
(718) 613 - 2682
***@***.******@***.***>

***@***.***  ***@***.***

PR #3506, opened 2022-06-15
Create ieGeek.pm

clipo1979:
ieGeek PTZ camera control module


connortechnology:
What does this do differently than the existing onvif ptz scripts?

clipo1979:
This is specifically for the ieGeek brand as they have only partially implemented ONVIF functions.

PR #3524, opened 2022-07-13
created es_la.php with all messages up to 1.37.17

ovargasp:
1. es_la.php new language file, 
2. added translate prompts on options.php (uses es_la.php file), 
3. montage review label translation.


web/skins/classic/views/options.php
```diff
@@ -467,6 +467,7 @@ function format_timezone_name($name) {
   foreach ($configCat as $name=>$value) {
     $shortName = preg_replace('/^ZM_/', '', $name);
     $optionPromptText = !empty($OLANG[$shortName])?$OLANG[$shortName]['Prompt']:$value['Prompt'];
+    $optionPromptText = translate(optionPromptText);
```
connortechnology: I think we are missing a $ here on optionPrompText

Also... OLANG is already translated I think.  So it's like a double translate.

ovargasp: Hey Issac you're right about the missing $, sorry still working on my github skills (did the edit locally and then change it on repo manually)... 

About the $OLANG, it doesn't translate without that line, at least using language files, but if you want I can check why is not translating.

src/zm_monitor.cpp
```diff
@@ -1714,7 +1714,9 @@ bool Monitor::Poll() {
         Debug(1, "Got Good Response! %i", result);
         for (auto msg : tev__PullMessagesResponse.wsnt__NotificationMessage) {
           if (msg->Topic->__any.text != NULL &&
-          std::strstr(msg->Topic->__any.text, "MotionAlarm") &&
+	  // need to modify this for something more standardized
+          // std::strstr(msg->Topic->__any.text, "MotionAlarm") && org. line
+	  std::strstr(msg->Topic->__any.text, "CellMotionDetector") &&
```
connortechnology: What does this have to do with language changes?

PR #3529, opened 2022-07-16
Fix printf formats & enable large file support for 32-bit

dougnazar:


src/zm_logger.cpp
```diff
@@ -82,8 +82,8 @@ Logger::Logger() :
     smSyslogPriorities[PANIC] = LOG_ERR;
 
     char code[4] = "";
-    for (int i = DEBUG1; i <= DEBUG9; i++) {
```
connortechnology, line 85: I'm not entirely in agreement with this one.  Yes DEBUG1 to DEBUG9 are > 0, but in general the log levels are integers. The other values are negative.  I don't see any reason to do this.  Got further insights?

dougnazar: Just to silence the warning in the commit message from GCC 12. The other option is to add another condition `i > 0` or force the format to a single digit `%.1d`.

I don't know why GCC can't see that `i` always starts above 0 and believes that `i` can be *in the range [-2147483647, 9]*.

connortechnology: ok, good enough for me.

PR #3535, opened 2022-07-19
Allows users to set an specific "Onvif alarm text" for each monitor

ovargasp:
With this the cameras that use strings different to "MotionAlarm" such as Dahua, will be able to get Onvif alarms. To make it easier, the change sets a default "Alarm Text"  value to monitors, and allows to modify it according to the particular necessities of each brand. It consists of a new column on table Monitors, changes on Monitors.h, Monitors.cpp and a changes in UI to manage the value of Alarm Text on each monitor.


src/zm_monitor.cpp
```diff
@@ -93,9 +93,13 @@ std::string load_monitor_sql =
 "`ImageBufferCount`, `MaxImageBufferCount`, `WarmupCount`, `PreEventCount`, `PostEventCount`, `StreamReplayBuffer`, `AlarmFrameCount`, "
 "`SectionLength`, `MinSectionLength`, `FrameSkip`, `MotionFrameSkip`, "
 "`FPSReportInterval`, `RefBlendPerc`, `AlarmRefBlendPerc`, `TrackMotion`, `Exif`,"
-"`RTSPServer`, `RTSPStreamName`,"
+"`RTSPServer`, `RTSPStreamName`, `ONVIF_Alarm_Txt`,"
 "`ONVIF_URL`, `ONVIF_Username`, `ONVIF_Password`, `ONVIF_Options`, `ONVIF_Event_Listener`, `use_Amcrest_API`, "
 "`SignalCheckPoints`, `SignalCheckColour`, `Importance`-1, ZoneCount FROM `Monitors`";
+//"`SignalCheckPoints`, `SignalCheckColour`, `Importance`-1, ZoneCount FROM `Monitors`";
```
connortechnology: Why are you commenting out this stuff?

ovargasp: An old habit of mine, when I change a line of code, I leave the original line commented out... I was going to put the ONVIF-Alarm-Txt at this line, so I commented out that line and changed my mind and forgot to delete the commented line... as you can see is the same... sorry

src/zm_monitor.cpp
```diff
@@ -1714,7 +1722,9 @@ bool Monitor::Poll() {
         Debug(1, "Got Good Response! %i", result);
         for (auto msg : tev__PullMessagesResponse.wsnt__NotificationMessage) {
           if (msg->Topic->__any.text != NULL &&
-          std::strstr(msg->Topic->__any.text, "MotionAlarm") &&
+//          std::strstr(msg->Topic->__any.text, "MotionAlarm") &&
+//          std::strstr(msg->Topic->__any.text, "CellMotionDetector") &&
```
jp-bennett: Same here. These commented out lines shouldn't be in the final PR.

jp-bennett (changes requested):
Just need those commented lines pulled out. That's a decent habit if you aren't using git, but with version control, it's just extra noise. Otherwise looks great!

src/zm_monitor.cpp
ovargasp: Agreed, I'm  working on my github skills... meaning I don't know if I can do it now and how... I need to continue studying git documentation in how to do certain things... I'm still a github newbie...

jp-bennett: I think you just need to make a second commit on your fork, and this PR will automatically pick it up. Then there's an option to squash it to a single commit on this end.

jp-bennett (approved):
Looks good to me. @connortechnology can squash and merge, if he also approves.

ovargasp:
> Looks good to me. @connortechnology can squash and merge, if he also approves.

Well I learned something, I just need to push the same file and github does the rest automatically... It's a very simple change but it gets dahua and I think some others cameras working...

Issue #695, opened 2015-02-05
When editing a monitor, using RTSP, host cannot contain random punctuation


connortechnology:
However the password might have punctuation in it.

We need to be consistent on all the various methods, including the control tab.  

I think username/password should be separate fields, that way we can actually do useful input validation.
##

SteveGilvarry:
Use Username and Password fields already in the db and just do this.

connortechnology:
I hate you stale bot.

GlenCooper:
Just commenting because I've reached this point, but need sleep (sorry).  Hopefully this comment will keep the stale-bot at bay for a little longer.

parvit:
Hi all,
i'm working on this and i've just got the system working from source.
I'll let you know as i have a change ready.

parvit:
@connortechnology I've added the PR responding to your request.

parvit:
Hello, is there any feedback on this?

parvit:
@connortechnology 
Thanks for merging the PR, if we may close this ticket it would allow me to claim the bounty.

As a side note: I would like to work also on #355, but it's locked so i don't know if i can proceed, can we unlock that one? thanks.

connortechnology:
I don't see anything locked about #355.  You might want to discuss your approach on it before digging in too far.  Would be great to have it done though.

I've modified your code a bit.  in php8 passing null to a lot of functions generates errors.  Also I'm not sure strlen is the most efficient way to determine if a string is empty or not.  I also took out the setting of auth_type in ffmpeg.  It shouldn't be necessary and is just generating errors in the logs.

parvit:
This is what i see from my side on #355
![image](https://user-images.githubusercontent.com/93098480/186004199-745178f5-6478-4e17-aad9-4743e612a1aa.png)

connortechnology:
Okay it is unlocked.

parvit:
It worked, thanks.

PR #3537, opened 2022-07-22
Added configuring of authentication for rstp cameras

parvit:
This responds to #695.

The change is not too complex and builds on support that was already present in the code.
It simply introduces the required fields in the monitor edit page to allow to specify the user and password for the camera. The values are already handled and saved/loaded by the backing php monitor code.

At this moment I have not inserted any explicit validation seen that the authentication is mediated by rtsp headers and not in the url so any value should be allowed.

I've tested this on my local machine with a "remote" type camera and it seems to authenticate correctly with the correct username/password containing special characters not allowed in url (without encoding).

Please review and test.


jp-bennett:
This changeset should also update how Janus gets username and password for RTSP streams: https://github.com/ZoneMinder/zoneminder/blob/master/src/zm_monitor_janus.cpp#L44

Good to see this being worked on!

parvit:
> This changeset should also update how Janus gets username and password for RTSP streams: https://github.com/ZoneMinder/zoneminder/blob/master/src/zm_monitor_janus.cpp#L44
> 
> Good to see this being worked on!

Thanks, i have not seen the part for janus, but doesn't your change split the data _in_ url? So it would fail all the same with special characters, or maybe i'm missing something?

jp-bennett:
> but doesn't your change split the data _in_ url?

Yes. We currently split username/password out of the given URL so Janus can authenticate. When we pull this changeset, and switch to using a username and password, the Janus code should drop the URL parsing and just use user and pass. Just pulling this code as-is would break Janus.

parvit:
> Just pulling this code as-is would break Janus.

Should i then investigate this kind of integration with the janus straming too?

jp-bennett:
> Should i then investigate this kind of integration with the Janus streaming too?

Yes please. It should be straightforward to drop the URL parsing code, and set the user/pass sent to Janus using the new options.

parvit:
Sure i'll look into it as soon as i can.

connortechnology:
The way that I was going to go was to actually change the Source tab UI to sync it up between remote and ffmpeg types.  Because we don't want to be able to actually see the password so the idea is you could either cutnpaste a full url and it would auto populate the split out parts (scheme, username, password, host, path, query etc).  Because when you switch between ffmpeg and remote they use the parts differently It would be nice to be able to switch back  and forth.

jp-bennett:
> cutnpaste a full url and it would auto populate the split out parts (scheme, username, password, host, path, query etc).

Would be nice to have that capability anyway, to transition from the old all-in-one URL to the new split-out interface.

parvit:
@connortechnology @jp-bennett 
I'll see the janus integration first than i'll check how to implement the (browser-side) sync.

parvit:
@jp-bennett 
I was trying to test my change with janus, but i'm getting an error that the "_javascript/janus/janus.js_" library is not found when opening the live view, can you point me to how i should find / generate that lib? 

Also is it correct that the user and password given in the janus monitor are actually the ones from the parent monitor? 

thank you for the support.

jp-bennett:
> "_javascript/janus/janus.js_" library is not found when opening the live view, can you point me to where can i find / how to generate that lib?

It depends on what distro you're using, but for Ubuntu, installing libjs-janus will get the javascript library in place (Might require an a2enconf, can't remember). It's contained in the Janus source (https://github.com/meetecho/janus-gateway/blob/master/html/janus.js) but the js lib needs to match the version of the Janus binary.



> Also would it be correct to assume that the user and password to be used for janus monitor are the same as the parent monitor?

Yes. What this is doing is sending the proper credentials to Janus, so Janus can connect to the camera directly.

jp-bennett:
Getting a crash with this code. 
```
terminate called after throwing an instance of 'std::regex_error'
  what(): Unexpected end of regex when escaping.
```

parvit:
Thanks for testing i'm having serious trouble setting up janus (something about the ICE servers not being defined), i'll check now.

jp-bennett:
> Thanks for testing i'm having serious trouble setting up janus (something about the ICE servers not being defined), i'll check now the cause of the error.

That's a misleading error. You don't need ICE servers. It's a pain to set up the first time, that's true. What's worse, browsers have this habit of not playing the streams if they're not specifically h264 baseline. I'm also trying to figure out exactly what regex is complaining about.

parvit:
Found, it was the backslashes at the end

jp-bennett:
> Found, it was the backslashes at the end

I think we figured that out at the same moment. =)

parvit:
ops i forgot to add back the _std::_ prefix

jp-bennett:
That part looks good. Now we just need to get the bits wired up for ffmpeg and vlc capture. And it would be really slick to have a database update that pulls any auth strings out of the source path, and populates username/password.

parvit:
Should the url parsing happen only when the fields are empty? because otherwise you might accidentally overwrite the user values.

jp-bennett:
> Should the url auth parsing happen only when the fields are empty? because otherwise you might accidentally overwrite the user values.

That sounds reasonable. If either field is populated, leave everything as-is. If they are empty and the URL appears to have a username and/or password, extract to the appropriate fields.

parvit:
@jp-bennett @connortechnology 
Hi all, i've integrated the request for updating dynamically from the db and ui the update of the path, user and password values.

On the UI side:
* Added the user and password fields to Remote type monitors
* Edit of the path field will cause the local update of the user and password fields _only if_ they were empty, moreover if they are equal it will update the corresponding field dynamically (ie. updating the username field in the url will cause the username form field to update)
* On losing focus for any of the fields, a local sync of the values if performed, meaning that if the fields in both the url and the username/password are equal than they are removed from the url

On the backend side:
* On update from the clientside and on loading of the monitors from the database, the same rules are enforced

So this means for example that loading a path value already in the db with both username and password set will populate the corresponding fields and remove them from url

I hope i've interpreted correctly what was requested. 

Is there anything else that should be included? thanks.

jp-bennett:
> Is there anything else that should be included?

The only other thing I know of is that ffmpeg cameras should use the username/pass fields, too. That's how most users set up for RTSP. It's likely a simple pair of options to set in https://github.com/ZoneMinder/zoneminder/blob/master/src/zm_ffmpeg_camera.cpp

parvit:
@jp-bennett 
I've added the ffmpeg and (looking at the previous messages from @connortechnology) the libvlc support for the auth. 

Please check if everything necessary for the feature is present and test it if you can.

jp-bennett:
> @jp-bennett I've added the ffmpeg and the libvlc support for the auth.
> 
> Please check if everything necessary for the feature is present and test it if you can.

I'll give it a thorough test over the weekend. Excited to get this pulled in.

jp-bennett:
Tested and working here. @connortechnology you see any problems with this? I'm inclined to pull it.

connortechnology:
I'll take a look tonight

PR #3538, opened 2022-07-22
Fix ci

dougnazar:
Should fix FreeBSD-12 build due to some ports being compiled against 12.3.
12.3 is first version to introduce `strerror_l`

(`libperl.so.5.32: Undefined symbol "strerror_l@FBSD_1.6"`)

Debian Stretch just exited LTS and uses GCC 6 which doesn't support features in the current code base.


dougnazar:
CentOS 7 still has 2 years of support but the GCC (4.8) doesn't support the current code base so I created a docker image with all the requisites & GCC 8 and used that as the basis to run the build. I assume that you'll want to build your own image and use that instead.

connortechnology:
The Centos7 stuff is cool... we were basically planning to drop support for it.  Leaving them stuck on 1.36.  We will probably be forced to figure out a support path for them though.

dougnazar:
As far as I can tell, the resulting binaries work fine on a CentOS system without the updated dev kit installed so normal users shouldn't care. It's just required when building the RPMs, but I'll let somebody more familiar with that environment verify that.
```
[root@sb-pc0900 nazard]# cat /etc/centos-release
CentOS Linux release 7.9.2009 (Core)
[root@sb-pc0900 nazard]# rpm -q zoneminder
zoneminder-1.37.18-1.el7.x86_64
[root@sb-pc0900 nazard]# /usr/bin/zmc --version
1.37.18
```

PR #3545, opened 2022-07-27
Misc fixes

dougnazar:
- Add padding to shared_data to re-align fields on 32bit
- Don't hardcode db in upgrade script.


baudneo:
Sick, 776 bytes was what I was getting in python when polling shm. 2 bytes for 32 and then extra 4 for 64 bit.

PR #3546, opened 2022-07-27
Switch from av_init_packet() to av_packet_alloc()

dougnazar:
Remove all uses of deprecated av_init_packet() and switch any stack
based AVPackets to unique_ptrs allocated with av_packet_alloc().

Ensure that all code paths call av_packet_unref() after use to reset
before next usage.

---

- This API has been available since v3.0 so that covers all the supported systems.
- I checked that it properly initializes all fields so can drop the setting of packet.size & packet.data
- Debating creating an exit_guard to handle cleanup

Thoughts?

### DIFF
--- a/src/zm_ffmpeg_output.cpp
+++ b/src/zm_ffmpeg_output.cpp
@@ -84,13 +84,12 @@ AVFrame *FFmpeg_Output::get_frame( int stream_id ) {
   Debug(1, "Getting frame from stream %d", stream_id );
 
   int frameComplete = false;
-  AVPacket packet;
-  av_init_packet( &packet );
+  av_packet_ptr packet{av_packet_alloc()};
   AVFrame *frame = zm_av_frame_alloc();
   char errbuf[AV_ERROR_MAX_STRING_SIZE];
 
   while ( !frameComplete ) {
-    int ret = av_read_frame( input_format_context, &packet );
+    int ret = av_read_frame( input_format_context, packet.get() );
     if ( ret < 0 ) {
       av_strerror(ret, errbuf, AV_ERROR_MAX_STRING_SIZE);
       if (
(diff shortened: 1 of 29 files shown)


connortechnology:
Wow thanks for tackling this!  It has been on my todo list for a while.

In general we like good RAII concepts.  I'll have to review this and think on it.

src/zm_ffmpeg.cpp
```diff
@@ -504,3 +504,10 @@ int zm_get_samples_from_fifo(AVAudioFifo *fifo, AVFrame *frame) {
   zm_dump_frame(frame, "Out frame after fifo read");
   return 1;
 }
+
+
+
+void zm_free_av_packet::operator()(AVPacket *pkt) const
+{
+    av_packet_free(&pkt);
```
Carbenium: I would move that into the header so it can be inlined.

connortechnology:
That all looks pretty straight forward to me. I like the idea of an exit_guard.

src/zm_ffmpeg_output.cpp
```diff
@@ -84,13 +84,12 @@ AVFrame *FFmpeg_Output::get_frame( int stream_id ) {
...
   int frameComplete = false;
-  AVPacket packet;
-  av_init_packet( &packet );
+  av_packet_ptr packet{av_packet_alloc()};
   AVFrame *frame = zm_av_frame_alloc();
   char errbuf[AV_ERROR_MAX_STRING_SIZE];
 
   while ( !frameComplete ) {
-    int ret = av_read_frame( input_format_context, &packet );
+    int ret = av_read_frame( input_format_context, packet.get() );
```
Carbenium, line 102: If you are already touching these lines, feel free to adjust the code style:
```suggestion
    int ret = av_read_frame(input_format_context, packet.get());
```

src/zm_packet.h
```diff
@@ -23,6 +23,7 @@
 #include "zm_logger.h"
 #include "zm_time.h"
 #include "zm_zone.h"
+#include "zm_ffmpeg.h"
```
Carbenium: alphabetical ordering

Carbenium (commented):
Looks good! Then I can drop the branch I had stashed away since quite a while ;)

One thing I can think of: what if `av_packet_alloc` returns `NULL` and we dereference?

src/zm_ffmpeg_output.cpp
dougnazar: I decided early on to ignore the screaming in my head as we kept switched formatting. 

Next project on my list is to run clang format over the code, and then maybe setup the CI to check it.

Carbenium: :+1:

PR #3552, opened 2022-07-31
Misc fixes

dougnazar:
A few fixes I came across and some codeql cleanups.


baudneo:
Damn bro, where did you come from? Lol.

You're really hitting the repo hard with fixes, good job and thank you!

src/zm_monitor.cpp
```diff
@@ -2743,23 +2743,27 @@ void Monitor::TimestampImage(Image *ts_image, SystemTimePoint ts_time) const {
   while (*s_ptr && ((unsigned int)(d_ptr - label_text) < (unsigned int) sizeof(label_text))) {
     if ( *s_ptr == config.timestamp_code_char[0] ) {
       bool found_macro = false;
+      int rc = 0;
       switch ( *(s_ptr+1) ) {
         case 'N' :
-          d_ptr += snprintf(d_ptr, sizeof(label_text)-(d_ptr-label_text), "%s", name.c_str());
+          rc = snprintf(d_ptr, sizeof(label_text)-(d_ptr-label_text), "%s", name.c_str());
```
Carbenium: Maybe pull out the `sizeof(label_text)-(d_ptr-label_text)` calculations. GCC with -O2 seems to generate shorter assembly and quick benchmarking shows a 10% improvement there.

Carbenium (approved):
See one comment. Otherwise LGTM.

src/zm_packetqueue.cpp
```diff
@@ -314,7 +314,7 @@ void PacketQueue::clearPackets(const std::shared_ptr<ZMPacket> &add_packet) {
   } // end while
 
   if (keyframe_interval > pre_event_video_packet_count) {
-    Warning("Max Image Buffer setting is too low! Needs to be great than keyframe interval. Keyframe interval is at least %d", keyframe_interval);
```
connortechnology: This isn't correct.  The message is correct, I suspect the use of pre_event_video_packet_count is wrong.  Please remove this and I will take a look and correct this properly.  The intention here is to detect and warn that the keyframe interval being sent from the camera is greater than the max queue size, so it is impossible for ZM to queue enough packets.  It is new code I added this just a few days ago.

connortechnology:
Regarding the: Cast operand to larger destination size to avoid possible overflow is

I kinda hate these random casts.  I get why they might be needed, I'm just wondering if there is a nicer way to do it.

dougnazar:
For most of them looks like an inline function would probably work, silently promoting the operands. Not sure if it would work on 32bit for the first one, since that's specifically an `unsigned long long`.

`size_t calculate_image_buffer_size(size_t width, size_t height, size_t colours);`

src/zm_packetqueue.cpp
dougnazar: np. I wasn't sure, but if I increased max queue size from 0 (to 120), I was getting other errors. Keyframe interval was around 60 as I remember it.

connortechnology:
I guess what I am saying is that.... we already made imagesize an unsigned long long, presumably with the idea that that would fix the overflow problem.  Your casting here seems to suggest that it doesn't.  That we need to upgrade one of the operands to make it do 64bit multiplication before assigning to imagesize.  We can be fairly sure for at least a few more years that width and height won't be very large, but we could just make them 64 integers too.  I suppose another thought I have is that perhaps we should be using size_t instead of unsigned long long.

dougnazar:
Right, the issue (and warning) is that the calculation happens before the promotion.
`size_t` should be fine on all platforms. that still handles 8k images on 32bit (4 bytes), and is 8 bytes on 64bit.

Using `size_t` for size, width, height, etc. should also fix the issue. If that's your preference, I can look at doing that. A quick grep shows only about 500 lines.

Carbenium:
Wouldn't it make sense to use the fixed-width typedefs (`[u]int{8,16,32,64}`) at least in those places where we don't want this platform-dependent ambiguity? ``

PR #3574, opened 2022-08-13
db: start the processing thread after all fields are initialized

peat-psuwit:
Initializing `zmDbQueue::mThread` with initializer list makes the thread
start in the middle of object initialization. This means `zmDbQueue::
process()` could start and reach mCondition.wait(lock); before `std::
wait_condition`'s constructor, which causes deadlock.

To make sure thread won't start before object initialize, move
`mThread`'s initialization to constructor's body.

This _at last_ solves the random zmu deadlock issue that has plagued my
setup for almost a year.

### DIFF
--- a/src/zm_db.cpp
+++ b/src/zm_db.cpp
@@ -227,9 +227,10 @@ zmDbRow::~zmDbRow() {
 }
 
 zmDbQueue::zmDbQueue() :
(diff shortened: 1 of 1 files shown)


connortechnology:
OMG you are amazing!  Thank you so much for this!

Issue #3516, opened 2022-07-01
LDAP auth vs database auth case sensitivity leads to HTTP 500


megasaturnv:
To allow Zoneminder to authenticate with LDAP an admin will set AUTH_TYPE to "remote" and configure apache to authenticate with the LDAP server. Then, the same username is created in Zoneminder. Problem is, LDAP authentication is case insensitive a user can login with a valid account which doesn't match the case used for Zoneminder's user.

e.g. user in Zoneminder's database might be "John.Doe" but LDAP would allow "john.doe" to authenticate

After authenticating with the correct username but doesn't match the exact case used for the user in the Zoneminder database, the user is given a HTTP error 500

A temporary solution we have is to change
https://github.com/ZoneMinder/zoneminder/blob/2e22cd5c3b489ee05a2cf6ff749df8995f7e6d38/web/includes/auth.php#L323
to `$user = dbFetchOne($sql, NULL, array(strtolower($_SERVER['REMOTE_USER'])));` and then only create lower-case usernames in Zoneminder's database. But, we have to change this file manually after every update. I was thinking of making a pull request for this, but it would break users that just use simple "htpasswd" authentication (and don't use lower case usernames)

Are there any long-term solutions for this? Thanks

Extra info, to follow the issue template:
* Zoneminder version: 1.36.19
* OS: CentOS 7
* Installed binary from Repo - https://zoneminder.readthedocs.io/en/latest/installationguide/redhat.html

connortechnology:
I think we would simply add an additional configuration option, telling ZM to use this code.  Something like ZM_CASE_INSENSITIVE_USERNAMES. Then people can do whatever they need to do.   Pretty trivial. Do you want to fund an hour of my time to get it into the next ZM release?

megasaturnv:
Thanks. I'll fork it and do a PR

megasaturnv:
Thanks!

megasaturnv:
Just re-opening this issue to check something. Do I need to do a PR specifically to merge these changes into release-1.36 (rather than master), or are these changes scheduled to be in the next major release?

I don't mind doing a fresh PR for this and testing again if needed, since there are major differences between master and release-1.36.

Thanks!

connortechnology:
It has been merged to release-1.36 and will be released in 1.36.29 today.

PR #3582, opened 2022-09-02
Issue #3516 case insensitive login

megasaturnv:
Pull request for ZoneMinder/zoneminder#3516 to add the option to match mixed case usernames to lower case usernames in database. Useful for LDAP auth which is case-insensitive. Has been compiled and tested

Initially created option `ZM_AUTH_CASE_INSENSITIVE_USERNAMES` but this was too long, so renamed to `ZM_CASE_INSENSITIVE_USERNAMES`. Maintainers may wish to change where this goes in the options list, since it no longer begins with AUTH_

Feel free to write a better description for the docs. Translations may also be wanted for `web/lang/es_la.php` and `web/lang/ru_ru.php` as these have descriptions for similar options


web/includes/auth.php
```diff
@@ -60,7 +60,12 @@ function migrateHash($user, $pass) {
 function validateUser($username='', $password='') {
   $sql = 'SELECT * FROM Users WHERE Enabled=1 AND Username=?';
   // local user, shouldn't affect the global user
-  $user = dbFetchOne($sql, NULL, array($username));
+  $user = null; // Not global
+  if (ZM_CASE_INSENSITIVE_USERNAMES) {
+      $user = dbFetchOne($sql, NULL, array(strtolower($username)));
+  } else {
+      $user = dbFetchOne($sql, NULL, array($username));
+  }
```
connortechnology, line 65: This requires the value stored in the db to be lowercase as well.  I think we should alter the sql as well to use tolower(Username).

connortechnology:
Looks good, just a couple comments.  I don't think we should say that this is for AUTH_REMOTE.  It applies to all auth really.  It's just a choice to be case insensitive on usernames.  A lot of people use their actual names as usernames with regular auth and of course they have to type it in case correct.  I think this is a great feature in general.I would actually prefer the config name to start with AUTH but as you say it gets too long so I don't really care.

megasaturnv:
Agreed, didn't feel right to me either only matching against lower case usernames in the database.

As I was changing the code, I realised the readability was much better if the tolower() was done entirely in SQL, rather than one half in PHP and the other in SQL. Makes covering all login options easier too.

Currently testing a build of this on my end

megasaturnv:
No problems with my build from source here. If there are two users with the same name but different cases (e.g. Test and TEst), it will login with the first user with a matching name in the table.

One quirk of PHP - ```If``` statements in don't have their own scope. Let me know if you'd prefer me to add:
```
$sql = null;
```
before each:
```
if (ZM_CASE_INSENSITIVE_USERNAMES) {
  $sql = 'SELECT * FROM Users WHERE Enabled=1 AND LOWER(Username)=LOWER(?)';
} else {
  $sql = 'SELECT * FROM Users WHERE Enabled=1 AND Username=?';
}
```

PR #3585, opened 2022-09-05
Adds Janus PIN and secret

jp-bennett:
In an effort to make our Janus integration a little safer to expose to the internet, this adds a randomly generated key that is required to view and enumerate Janus mounts. Uses the shared memory and zmu to push into the browser.

### DIFF
--- a/src/zm_monitor_janus.cpp
+++ b/src/zm_monitor_janus.cpp
@@ -18,6 +18,7 @@
 //
 
 #include "zm_monitor.h"
+#include "zm_crypt.h"
 #include <regex>
 
 std::string escape_json_string( std::string input );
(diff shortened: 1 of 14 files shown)


src/zm_monitor_janus.cpp
```diff
@@ -51,6 +52,9 @@ Monitor::JanusManager::JanusManager(Monitor *parent_) :
   } else {
     rtsp_path = parent->path;
   }
+  parent->janus_pin = generateKey(16);
+  Warning("%s", parent->janus_pin.c_str());
```
connortechnology: This should be a Debug(1

src/zm_monitor_janus.cpp
```diff
@@ -51,6 +52,9 @@ Monitor::JanusManager::JanusManager(Monitor *parent_) :
...
+  strcpy(parent->shared_data->janus_pin, parent->janus_pin.c_str()); //copy the null termination, as we're in C land
```
connortechnology: I would think this should be strncpy

web/includes/Monitor.php
```diff
@@ -268,7 +268,12 @@ public static function getStatuses() {
     'ArchivedEvents' =>  array('type'=>'integer', 'default'=>null, 'do_not_update'=>1),
     'ArchivedEventDiskSpace' =>  array('type'=>'integer', 'default'=>null, 'do_not_update'=>1),
   );
-
+  public function Janus_Pin() {
+    $cmd = getZmuCommand(' --janus-pin -m '.$this->{'Id'});
```
connortechnology, line 272: Doesn't have to be fixed now, but this misses the multi-server case, in which we would have to use the API, so the API needs support for Janus_Pin too.

PR #3589, opened 2022-09-16
Instar control for 720p CGI devices

VikingGer:
Onvif just yielded the absolute basic functions, but especially no presets. This control fills the gap.


connortechnology:
Hey Thanks for this!  I'll merge as is, but a couple of notes in case you want to go further.

In the authorization headers, if the camera responds with multiple options, like auth basic, OR digest, then then www-authenticate header will be an array.  See the code in AxisV2 for an example. This function should realy be moved up to Control.pm to make it available everywhere.

I've been working on code to get/set config in cameras so that I can apply settings en-mass.  See an example in Grandstream.pm. 

One last thing, you might want to include an entry in zm_create.sql.in so that it appears as an available control protocol.

PR #3594, opened 2022-09-17
Update reports.js

VikingGer:
Function name missing? Assumed to be correct according to the rest.


connortechnology:
You don't have to have a function name, it's a lambda.  eslint was likely complaining about the space between function and (). giving it a name fixes it though.

PR #3609, opened 2022-10-13
Move relevant code out of includes/actions/auth.php into includs/auth…

connortechnology:
Move the guts out of includes/actions/login.php and into includes/auth.php.

The action only happens on post, and now only handles redirecting.

So now everything is in auth.php and this means that google recaptcha breaks api logins.  Not sure we can do anything about that.

### DIFF
--- a/web/includes/actions/login.php
+++ b/web/includes/actions/login.php
@@ -23,90 +23,16 @@ if ( ('login' == $action) && isset($_REQUEST['username']) && ( ZM_AUTH_TYPE == '
 
   // if true, a popup will display after login
   // lets validate reCaptcha if it exists
-  if (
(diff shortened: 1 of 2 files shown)


pliablepixels:
In previous versions, this would not break the API so I'd think this is undesirable from that perspective. Are we able to exclude this check if the user is using login.json  as a mechanism to log in?

connortechnology:
I think this doesn't break the api.  The api uses user=&pass= to do initial authentication and tokens thereafter.
It only breaks the use of username=&password= when also wanting to use recaptcha. Which was broken before anyhow.

pliablepixels:
Ah ok, if it has 0 API impact, lgtm

PR #3638, opened 2022-11-23
zm_ffmpeg_camera: Add support for Jetson Nano/Xavier NVMPI

kholk:
Add support for the NVMPI hardware decoders (not CUDA decoders!) of the Tegra SoCs found in the Jetson Nano/TX/TX2 and Xavier NX/AGX.

This platform supports hardware decoding of H264, HEVC, MPEG2, MPEG4, VP8 and VP9.


connortechnology:
Thanks for your contribution!

I will merge it, but I really think we should give the user an option on which to try to open, similar to how I've done it in videostore.cpp for encoding.

PR #3645, opened 2022-12-12
Docs/criadoperez

criadoperez:
Improvements on the docs:

* Added install instructions for Ubuntu 22.04LTS (verified on a clean VM)
* Noted that building from source is discouraged
* Removed pre-fix of "easy-way" on all debian installs as there is no "hard-way" or "other-way" anymore.

### DIFF
--- a/docs/installationguide/ubuntu.rst
+++ b/docs/installationguide/ubuntu.rst
@@ -32,14 +32,8 @@ To use this repository instead of the official Ubuntu repository, enter the foll
 ::
         sudo apt install -y zoneminder
 
-**Step 4:** Set permissions
 
-::
-
-        sudo chown -R www-data:www-data /usr/share/zoneminder/
-
-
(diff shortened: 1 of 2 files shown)


docs/installationguide/ubuntu.rst
```diff
@@ -3,8 +3,64 @@ Ubuntu
...
+**Step 3:** Install Zoneminder
+
+::
+        sudo apt install -y zoneminder
+
+**Step 4:** Set permissions
+
+::
+
+        sudo chown -R www-data:www-data /usr/share/zoneminder/
```
connortechnology: Why are you doing this?  This is bad for security and is not needed. Otherwise these updates are great.

criadoperez: Well observed!

I re-tested all the steps without running this command and it's works fine, so yes better to remove it.

PR #3648, opened 2022-12-22
Servers Modal POST form correction

jonnio:
In Options->Servers the form post is broken.

* Add correct hidden value to servers modal form
* Add IntelliJ Idea exclusion to .gitignore for users of IntelliJ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -156,3 +156,4 @@ zmlinkcontent.sh
 zm_config_data.h
 zm_config_defines.h
 **/.DS_Store
+.idea/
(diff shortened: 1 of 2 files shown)


jonnio:
Issue is at : https://github.com/ZoneMinder/zoneminder/issues/3649

connortechnology:
Are you sure about this? The actual Save button is a submit type, whose name is action.  So another hidden input shouldn't be necessary.  Also, the change to .gitignore is not desired...

connortechnology:
Ah, I see.  1.36.  There was more to fix there and was already fixed in master. I have cherry-picked  f96c90ffa790c4516ade9ef5ee7be4cdf7087a75 which should handle it, making this redundant. It will be included in 1.36.33.

Thanks for your efforts though!

PR #3651, opened 2022-12-31
Make event button toolbar wrap-able

DKnight110:
For small screens (i.e. phones), it's
useful to split the event button toolbar
on multiple rows, otherwise the video event
becomes useless (i.e. too small to be useful)

Signed-off-by: Alex Porosanu <dknight110@gmail.com>


DKnight110:
As an example of how it looks on my phone, see here: 

Screenshot_20221231-115739

PR #3652, opened 2023-01-10
Add janus rtsp session timeout

baudneo:
# What
This allows to override RTSP session timeout if it is set, or it can be set if the RTSP server does not send `;timeout=\<timeout seconds>` in its `RTSP SETUP` Session header response.

# Symptoms
I have a camera that times the RTSP session out within 90 seconds but, it does not send that data in its `SETUP` session header response. 
I would have little blips in the stream and eventually, it would fail to stream altogether. Janus has a default timeout of 5s once it can no longer stream from the RTSP source, so I would get a `401 Authorization Required` every 94–96 seconds.

# Configuration
- Configure by going to the 'Viewing' tab in the monitor config modal, where other Janus settings are.
- The input is in seconds and only accepts integers, a sane default is `60`.
- Setting to `0` (zero) will **disable** overriding/setting the timeout.
    - If the RTSP server (source; camera) has a correct reply with timeout set, it will use that. If it has an incorrect reply without a timeout and the session eventually times out, the `401 Authorization Required` responses will start happening.

# Examples

### Repeated 401 response examples pulled from janus.log:

```log
[janus-ZM] [Mon Jan  9 00:03:27 2023] [WARN] [1] 5s passed with no media, trying to reconnect the RTSP stream
[janus-ZM] [Mon Jan  9 00:03:27 2023] DESCRIBE answer:RTSP/1.0 401 Authorization Required

# Roughly 95 seconds later

[janus-ZM] [Mon Jan  9 00:05:03 2023] [WARN] [1] 5s passed with no media, trying to reconnect the RTSP stream
[janus-ZM] [Mon Jan  9 00:05:03 2023] DESCRIBE answer:RTSP/1.0 401 Authorization Required
```

### Example of RTSP SETUP reply that does not contain timeout
```
[janus-ZM] [Mon Jan  9 00:05:03 2023] SETUP answer:RTSP/1.0 200 OK
CSeq: 3
Date: Mon Jan  9 00:05:03 2023 GMT
Transport: RTP/AVP;unicast;client_port=18766-18767;server_port=38912-38913
Session: 618750c3
```
### What it should look like:
According to [RFC 2326 Section 12.37](https://www.rfc-editor.org/rfc/rfc2326.html#section-12.37 "RTSP Session Headers")
```
[janus-ZM] [Mon Jan  9 00:05:03 2023] SETUP answer:RTSP/1.0 200 OK
CSeq: 3
Date: Mon Jan  9 00:05:03 2023 GMT
Transport: RTP/AVP;unicast;client_port=18766-18767;server_port=38912-38913
Session: 618750c3;timeout=90
```


web/includes/actions/monitor.php
```diff
@@ -93,6 +93,7 @@
       'JanusEnabled' => 0,
       'JanusAudioEnabled' => 0,
       'Janus_Use_RTSP_Restream' => 0,
+//       'Janus_RTSP_Session_Timeout' => 0,
```
connortechnology, line 96: Why is this commented out?

connortechnology: Oh because it isn't necessary because it's a numeric input not a checkbox

PR #3680, opened 2023-03-28
Update installation procedure for RHEL8

yelnis:
Need to add repository for RPM Fusion
https://rpmfusion.org/

### DIFF
--- a/docs/installationguide/redhat.rst
+++ b/docs/installationguide/redhat.rst
@@ -59,10 +59,16 @@ RHEL/CentOS 7 & 8 users must enable the EPEL repo:
     
 RHEL/CentOS 8 users must also enable the PowerTools repo:
 
+.. sidebar :: Note
+
+    PowerTools may not work on RHEL8. Refer https://rpmfusion.org/Configuration#Command_Line_Setup_using_rpm
+
 ::
 
     sudo dnf install dnf-plugins-core
+    sudo dnf install --nogpgcheck https://mirrors.rpmfusion.org/free/el/rpmfusion-free-release-$(rpm -E %rhel).noarch.rpm https://mirrors.rpmfusion.org/nonfree/el/rpmfusion-nonfree-release-$(rpm -E %rhel).noarch.rpm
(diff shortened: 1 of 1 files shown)


knight-of-ni:
While we certainly appreciate those willing to take the time update our documentation (there is always a need for this), this PR needs some adjustment.

We already instruct the user to navigate to rpmfusion and install the repo. 

See the first sentence below the section title:
https://zoneminder.readthedocs.io/en/latest/installationguide/redhat.html#how-to-install-zoneminder

However, we do need to advise RHEL9 users to enable the CRB repo. Redhat decided to remove powertools and add crb when they released rhel9.

The RHEL8 instructions should not change.

Really all we need to do here is add the following note, just after the rhel8 note:

```
RHEL/CentOS 9 users must also enable the crb repo:

::

    sudo dnf install dnf-plugins-core
    sudo dnf config-manager --set-enabled crb
```

At some point we should probably remove all reference to centos, since the centos this documentation was written around no longer exists.

yelnis:
@knight-of-ni 
Thanks for your confirmation. My PR is just notify that I couldn't install zm to RHEL8 with this document. Also I found the PowerTools was renamed to powertools, but my company's RHEL8 doesn't work lower case of powertools too. My RHEL is managed & modified by company IT or some vendors, so it would not normal condition. Let me leave comment as just information here.

By the way, RHEL installation section doesn't describe to configure database for mysql. Ubuntu section describes mysql configuration. Could you add mysql configuration to RHEL section?  It would be very helpfull for beginer.

Cheers.

knight-of-ni:
@yelnis 
Sounds like you missed this step in the documentation:
```
Once ZoneMinder has been installed, it is critically important that you read the README file under /usr/share/doc/zoneminder. ZoneMinder will not run without completing the steps outlined in the README.
```
The README steps you through dB config along with several other steps.

knight-of-ni:
Superseded by [ce92b14](https://github.com/ZoneMinder/zoneminder/commit/ce92b14247e22783d6dbdece4b0e8a687748c6e6)

PR #3682, opened 2023-04-07
fix(alarmSound): Use Audio.play() for alarm sound

Simpler1:
I reworked the alarm sound when viewing a monitor.  The old implementation seemed to be more complicated broken.  (At least I couldn't get it to work.)

I also removed the specific Internet Explorer (IE) checks since IE has not been supported by Microsoft for quite a while now.

Once the monitor view is open, the user has to interact with it (click anywhere on the page) and then the alarm sound will activate when a new alarm is triggered.  (See [this page](https://developer.chrome.com/blog/autoplay/) for details.)


connortechnology:
Again, thanks!  That's been on the todo list for a long time.

PR #3683, opened 2023-04-07
feat(events): Shift delete with no confirm

Simpler1:
This is an enhancement to the events page.  Multiple events can be selected and if the Shift key is held while pressing the delete button on the page, the confirmation dialog will not come up and the events will be deleted.


connortechnology:
Awesome, well done and thank you!

connortechnology:
I will likely merge this back to 1.36 as well.

PR #3685, opened 2023-04-07
feat(event): Shift delete with no confirm

Simpler1:
This is an enhancement to the event page. If the Shift key is held while pressing the delete button on the page, the confirmation dialog will not come up and the events will be deleted.


Simpler1:
Pulled the trigger too fast on this one...  The second commit has the working code.

PR #3704, opened 2023-05-11
feat(event): Add keyboard control

Simpler1:
Adds keyboard control from the "event" view.  The user can use the space bar to play/pause, the left/right arrows to move to the previous/next event, delete to bring up the delete dialog.

note:
For some reason the eventRenameModal does not behave like the eventDetailModal and deleteConfirm dialogs.  It doesn't have the "fade" class like the other two.


connortechnology:
Cool thanks for this!  I'm actually considering getting rid of the rename modal and merging it into the edit details modal. Why have two?

Simpler1:
I agree.  Combining the eventRenameModal into the eventDetailModal would be a good idea.

PR #3706, opened 2023-05-12
feat(sound): Include default alarm sound

Simpler1:

PR #3707, opened 2023-05-12
feat(sound): Include default alarm sound (again)

Simpler1:
Resubmitting this pull request.  For some reason the first one was inadvertently closed.

Original pull request was #3706


Simpler1:
The biggest advantage of this change is to create the sounds directory and provide some "subtle" sound by default.  Then it will be easier for admins to supply their own sound if they desire.   Currently, we don't create the sounds directory and the docs just say to put the sound file in the sounds directory, but don't say where the sounds directory should be created.

connortechnology:
I'm all for including this, I'm just not sure about the license compatibility on that sound file. If I can get a reading on that, I'll merge today.

@onlyjob I know if I merge this and there is a problem, you will let us know.  Want to weigh in early?

Simpler1:
I removed the internet sound a recorded a new sound just for ZoneMinder.  I release all right to this snap.ogg sound.

connortechnology:
Just FYI you could have just updated the old PR.  If you push to the branch it was generated from it automatically updates.  Handy.  Anyways no biggy.  Thanks for this!

PR #3719, opened 2023-06-03
(feat): Tags

Simpler1:
This PR implements Tags on Events.

Each tag is one word (no spaces) and will be created on space, comma, or enter.
The same tag can be added to many events.
The same event can have many tags.

Down arrow in the Tags input field to show the available tags or start typing to show matching available tags.
Available tags are shown in the order that they were applied (most recent at the top).
Ctrl-Down arrow to apply the last tag to this event.
Left/Right arrow to move to the previous/next event.

Removing a tag from the last event that had that tag also deletes the tag from the available tags.

Basic (equals) filtering works.


SteveGilvarry:
Lots of eslint errors, which are just formatting but should be fixed.
@connortechnology eslint was passing until an unused variable you introduced a little while ago in dnsmasq ip_parts.

web/ajax/event.php
```diff
@@ -167,6 +167,36 @@
...
+    $response = dbFetchAll($sql, NULL, $values);
+
+    $sql = 'UPDATE Tags SET LastAssignedDate = NOW() WHERE Id = ?';
+    $values = array($_REQUEST['tid']);
+    dbFetchAll($sql, NULL, $values);
+
+    ajaxResponse(array('response'=>$response));
+    break;
+  case 'removetag' :
+    dbFetchAll('DELETE FROM Events_Tags WHERE TagId = ? AND EventId = ?', NULL, array($_REQUEST['tid'], $_REQUEST['id']));
```
connortechnology: Why dbFetchAll instead of just dbQuery?

db/tags.sql
```diff
@@ -0,0 +1,22 @@
+DROP TABLE IF EXISTS `Events_Tags`;
+DROP TABLE IF EXISTS `Tags`;
+
+CREATE TABLE `Tags` (
+  `Id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+  `TagName` varchar(64) CHARACTER SET latin1 COLLATE latin1_bin NOT NULL DEFAULT '',
```
connortechnology: I would call this column just Name instead of TagName.  Simply to be consistent with our other tables.

web/ajax/event.php
Simpler1: I was having trouble with dbQuery while I was trying to figure out the code.  It may work just as well in this case.

Simpler1:
> Lots of eslint errors, which are just formatting but should be fixed. @connortechnology eslint was passing until an unused variable you introduced a little while ago in dnsmasq ip_parts.

All of the eslint errors are in existing code.  If we want to fix these (and we should), we should have a concerted effort to submit pull requests to just fix all (many) of the eslint errors so that we are starting from a clean slate for new code. :)

db/tags.sql
Simpler1: Good point.  I was trying to be as consistent as possible with existing code.  I'll change this to Name.

connortechnology:
The existing eslint error has been fixed.  The remaining I _think_ are in your additions to event.js.  Looks like mostly trailing spaces etc.

web/ajax/event.php
Simpler1: Changed from dbFetchAll to dbQuery for remove tags.

Simpler1 (commented):
Arg!  I'm obviously running the wrong lint checker.    What lint checker are you using?

Simpler1:
Finally a clean eslint check.  :)

connortechnology:
This is looking pretty good.  We probably would want cakephp api support as well.

Simpler1:
I'm working through the filters now.  I'm not familiar with cakephp, but I'll look into that once the filter stuff is working better.

Simpler1:
What is needed for cakephp?  Can you point me in the right direction?   What else is required to merge this PR?

Thanks

web/lang/en_gb.php
```diff
@@ -76,6 +76,7 @@
     'AlarmRefImageBlendPct' => 'Alarm Reference Image Blend %ge',
     'AlarmRGBUnset'         => 'You must set an alarm RGB colour',
     'AllTokensRevoked'      => 'All Tokens Revoked',
+    'AllTags'               => 'All Tags',
```
connortechnology: There is actually no problem with the lookup key having spaces, so I'd like to get away from using AllKeys and instead just use All Keys. Then we don't have to clutter up en_gb because if there isn't a translation, we just pass back the key.

web/skins/classic/views/js/event.js
```diff
@@ -1258,9 +1272,203 @@ function initPage() {
...
...haie|hcit|hd\-(m|p|t)|hei\-|hi(pt|ta)|hp( i|ip)|hs\-c|ht(c(\-| |_|a|g|p|s|t)|tp)|hu(aw|tc)|i\-(20|go|ma)|i230|iac( |\-|\/)|ibro|idea|ig01|ikom|im1k|inno|ipaq|iris|ja(t|v)a|jbro|jemu|jigs|kddi|keji|kgt( |\/)|klon|kpt |kwc\-|kyo(c|k)|le(no|xi)|lg( g|\/(k|l|u)|50|54|\-[a-w])|libw|lynx|m1\-w|m3ga|m50\/|ma(te|ui|xo)|mc(01|21|ca)|m\-cr|me(rc|ri)|mi(o8|oa|ts)|mmef|mo(01|02|bi|de|do|t(\-| |o|v)|zz)|mt(50|p1|v )|mwbp|mywa|n10[0-2]|n20[2-3]|n30(0|2)|n50(0|2|5)|n7(0(0|1)|10)|ne((c|m)\-|on|tf|wf|wg|wt)|nok(6|i)|nzph|o2im|op(ti|wv)|oran|owg1|p800|pan(a|d|t)|pdxg|pg(13|\-([1-8]|c))|phil|pire|pl(ay|uc)|pn\-2|po(ck|rt|se)|prox|psio|pt\-g|qa\-a|qc(07|12|21|32|60|\-[2-7]|i\-)|qtek|r380|r600|raks|rim9|ro(ve|zo)|s55\/|sa(ge|ma|mm|ms|ny|va)|sc(01|h\-|oo|p\-)|sdk\/|se(c(\-|0|1)|47|mc|nd|ri)|sgh\-|shar|sie(\-|m)|sk\-0|sl(45|id)|sm(al|ar|b3|it|t5)|so(ft|ny)|sp(01|h\-|v\-|v )|sy(01|mb)|t2(18|50)|t6(00|10|18)|ta(gt|lk)|tcl\-|tdg\-|tel(i|m)|tim\-|t\-mo|to(pl|sh)|ts(70|m\-|m3|m5)|tx\-9|up(\.b|g1|si)|utst|v400|v750|veri|vi(rg|te)|vk(40|5[0-3]|\-v)|vm40|voda|vulc|vx(52|53|60|61|70|80|81|83|85|98)|w3c(\-| )|webc|whit|wi(g |nc|nw)|wmlb|wonu|x700|yas\-|your|zeto|zte\-/i.test(navigator.userAgent.substring(0, 4))) {
```
connortechnology: This mobile detection code should go in skin.js or something more generic as a function.

connortechnology:
I'm ok with leaving the cakephp stuff till after merge.  What is required is to setup models/controllers to get the tags, event tags etc.  And setup the relationship so that when you ask for an event you get the tags as well. Mostly just a little boilerplate.

web/lang/en_gb.php
Simpler1: Isn't in beneficial to have all of the strings present in the main file (en_gb.php) so that it's easy to see what needs to be translated to other languages?

None of the existing keys have spaces. 

Should we keep this consistent and then update all of the keys to have spaces in the future?

connortechnology: There are lots of potential translations that I didn't add because I used spaces in the key already.  It would be nice for people doing the translations to have a master list yes, but at this point, we are trying to minimise developer time and maximise performance.  So if we were going to do it, I would do it in a template file that is not generally used (because the php compiler has to go through all this stuff).  So if you want to put the effort in, go ahead.

Simpler1: I agree.  Performance should be the main driver.  I'll make this change.

web/skins/classic/views/js/event.js
Simpler1: Fixed in commit https://github.com/ZoneMinder/zoneminder/pull/3719/commits/4b95ca345c9c8cd6611585fcdb6e4f6cd042d1d8

web/lang/en_gb.php
Simpler1: Fixed with commit https://github.com/ZoneMinder/zoneminder/pull/3719/commits/145b254ace2a913b42da446d85b1b45bc28b4885

PR #3720, opened 2023-06-04
Fix Builds

SteveGilvarry:
-Remove docker images that don't exist yet
-Swap back to packpack dockerhub, dependencies should be installed by packaging so we shouldn't need our own dockers. I think the history was timeouts in travis, but github actions that is 360 hours.

### DIFF
--- a/.github/workflows/create-packages.yml
+++ b/.github/workflows/create-packages.yml
@@ -16,16 +16,12 @@ jobs:
             dist: buster
           - os: debian
             dist: bullseye
-          - os: debian
-            dist: bookworm
           - os: ubuntu
             dist: bionic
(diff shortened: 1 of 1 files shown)


connortechnology:
So, the reason to use my own dockers was twofold.  One, packpack wasn't keeping up with newer releases, and second, I wanted to pre-install most of the dependencies.  These builds were happening on my server, and taking forever to download everything. In fact they were timing out and being killed.  

So I don't just want to switch back to packpack dockers.  We need a better solution.

connortechnology:
Okay, you raise the point that this is github ci, not travis, so hey yeah, let's give it a try.

PR #3722, opened 2023-06-05
distros/ubuntu2004: Update libgsoap dependency for Debian 12

mgulick:
Allow libgsoap-2.8.124, as that is what is shipped with Debian 12. This fixes do_debian_package.sh on Debian 12.


SteveGilvarry:
Is it possible to just make this => a particular version?

mgulick:
It looks like debian's libgsoap packaging incorporates the version number as part of the package name.  The package is not libgsoap version 2.8.124, the package is libgsoap-2.8.124 version 2.8.124.  I don't know if there's a way to do a wildcard or regex match against the package name itself.

PR #3727, opened 2023-06-14
Fix(event): Consistent cookie timeframe

Simpler1:
Sets cookie timeframe from one hour to two years (consistent with other cookie settings).


Simpler1:
image

Looks like we are using cookies where we really should be setting preferences which are indefinite.
It's painful to have to constantly set these back every hour to what is desired.

The only ones that I can think of that should be shorter (maybe even as short as 30 minutes) is the page # that is displaying.  I have had issues where I forgot that I had paged beyond page 1 earlier and when I looked for current events they did not show up at the top (descending order) because I was on page 2 or page 3 without realizing it.

PR #3748, opened 2023-07-24
Refresh table on show

Simpler1:
After opening an event from the events table and editing some properties, returning back to the events table would show the cached version of the table without the new values.  

This change will refresh the events table each time it's shown. 

It is a bit of a performance hit, but it's more important to show accurate information after a quick delay than to show bad/inaccurate information fast.

### DIFF
--- a/web/skins/classic/views/js/events.js
+++ b/web/skins/classic/views/js/events.js
@@ -420,6 +420,11 @@ function initPage() {
     }
   });
 
+  window.onpageshow = function(evt) {
+    console.log('Refreshing table');
+    table.bootstrapTable('refresh');
+  };
+
   table.bootstrapTable('resetSearch');
   // The table is initially given a hidden style, so now that we are done rendering, show it
   table.show();


connortechnology:
When you return back to the events table, what do you mean exactly? Back button? Back button above table?  Some other form of navigation?

The double hit on loading the events is annoying and I'm having trouble re-creating the situation you describe here.

Simpler1:
Here are the steps:

- Comment out the lines from this PR
- Refresh the events page
- Open one of the events by clicking the event Id in the Id column
- Make a change (e.g. Archive the event, add a note, add a tag)
- Go back to the events page (use either the back button at the top left of the page or the browsers back button or Alt-LeftArrow)
- Notice that the change that you made to the event does not show up in the corresponding row until you explicitly refresh the list or refresh the page.

With this PR uncommented, the changes are immediately seen.

It doesn't matter how you got to the events page (e.g. clicking the event count from the Console page, from the filters page, etc.)

PR #3765, opened 2023-09-09
Fix #3764 - improve ffmpeg camera init

### LINKED ISSUE #3764: Potential memory leak in FfmpegCamera OpenFfmpeg()
**Describe Your Environment**
- ZoneMinder master branch commit b44216d3da88f84a9150ae9dde1e970b4d1ba09d
- compiled from source
- own custom OS

**Describe the bug**
in FfmpegCamera class, in OpenFfmpeg() function, the second_input pointer gets initialized with FFmpeg_Input class that allocates ressources for the AVCodecContext and AVFormatContext. The second_input object life time is only for a small scope and never gets deleted that leads to a potentiel memory leak.

**Expected behavior**
Use unique_ptr for second_input so all ressources get freed correctly on end of FfmpegCamera object life time.
Use already initialized context from FFmpeg_Input instead initializing the audio context again.

d-bum:
Fixes issue #3764

### DIFF
--- a/src/zm_ffmpeg_camera.cpp
+++ b/src/zm_ffmpeg_camera.cpp
@@ -124,6 +124,7 @@ FfmpegCamera::FfmpegCamera(
   mOptions(p_options),
   hwaccel_name(p_hwaccel_name),
   hwaccel_device(p_hwaccel_device),
+  mSecondInput(nullptr),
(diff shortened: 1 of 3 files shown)


connortechnology:
love it! Thanks!

PR #3769, opened 2023-09-24
Don't attach images in a nested multipart message

akgrant43:
This causes Thunderbird and Apple iPhone email to display the images as attachments instead of inline.

Tested that it appears correctly in web gmail, iOS iPhone email, fastmail web mail and fastmail iOS email.

### DIFF
--- a/scripts/zmfilter.pl.in
+++ b/scripts/zmfilter.pl.in
@@ -750,7 +750,7 @@ sub substituteTags {
             my $filename = fileparse($path);
             my $attachment =  { type=>'image/jpeg', path=>$path, content_id=>uri_encode($filename) };
             push @$attachments_ref, $attachment if !is_in_attachments($path, $attachments_ref);
-            $text =~ s/%EI1%/<img src="cid:$$attachment{content_id}"\/>/g;
+            $text =~ s/%EI1%/$$attachment{content_id}/g;
           } else {
             Warning("Path to first image does not exist at $path for image $first_alarm_frame");
             $text =~ s/%EI1%/No image found for EI1/g;
@@ -763,7 +763,7 @@ sub substituteTags {
             my $filename = fileparse($path);
             my $attachment =  { type=>'image/jpeg', path=>$path, content_id=>uri_encode($filename) };
             push @$attachments_ref, $attachment if !is_in_attachments($path, $attachments_ref);
-            $text =~ s/%EIM%/<img src="cid:$$attachment{content_id}"\/>/g;
+            $text =~ s/%EIM%/$$attachment{content_id}/g;
           } else {
             Warning("No image for EIM at $path");
             $text =~ s/%EIM%/No image found for EIM/g;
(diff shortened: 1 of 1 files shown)


akgrant43:
This PR makes two basic changes:

- Changes the substituted text for the image macros from a link, e.g. `` to simple text, e.g. 00001-capture.jpg`.  The link in this case appears to be broken, it's just the source that CakeEmail expects.  I'm not sure what is supposed to be generated here (the image is included and displayed anyway).
- Removes the mime multipart nesting of the images so that they are displayed inline in Thunderbird and iPhone email.

connortechnology:
But we don't want them as attachments.  We want them inline.  That was the entire point of that code.  Now.... if you google this problem, you will find lots of people trying hard to figure out how to structure an email to make it work everywhere. It's a hairy problem. When I last worked on this, I was pretty sure it worked for iphone, android, thunderbird and gmail.  So more testing will need to be done, but removing the ability to have inline images is not likely to get merged.

akgrant43:
I think I worded it badly.  The current code nests multipart messages, which causes the images to be displayed as attachments, not inline.

My patch removes the nesting which means they are displayed inline.

Sorry for the confusion.

connortechnology:
But it also removes the image tags.  We will need to do a lot more testing.

akgrant43:
By image tags are you referring to the link that was included in the email body?

That's the bit that I'm not so sure about.  In Thunderbird the link is invalid and simply points to, e.g. `cid:00847-capture.jpg`, which is obviously invalid.  The relevant portion of the .eml is:

```
ZM alarm detected

Monitor: Kam66
Time: 2023-09-25 16:03:36

First alarm image: <img src="cid:00322-capture.jpg"/>

Highest score image: <img src="cid:00323-analyse.jpg"/>
```



Maybe it would be better if I split this out into two PRs:

- One to flatten the MIME structure, which results in the images being displayed inline.
- One to look at the image tags

Would you prefer two PRs?

Thanks,
Alistair

connortechnology:
the cid thing is how you refer to an attached image, when using in an image tag in an html email.  So that it shows both inline and as an attachment.

This all works for me... I use it regularly with thunderbird on linux.

connortechnology:
I'd like to close this, as I won't be merging it without more work.

I was looking at an email last night and I noticed that k9-mail starts off showing the images as attachments, and there is a button to download images for display. Once clicked, the images appear inline but no longer as attachments. However if I click on the images I can still download them just like as attachments.

Is this something like what you are seeing?

akgrant43:
In Thunderbird I get a link for each of the attachments:



Clicking on the link opens the image in the default OS image viewer (gnome's default image viewer, loupe, in my case).

Closing the issue is fine, thanks for letting me know.  I'd still like to see the modification, but understand the need for wide testing.

Just FYI, I'm applying my patched version of zmfilter.pl after each upgrade so i get the images in-line.  The filter definition is:

```
ZM alarm detected

Monitor: %MN%
Time: %ET%

First alarm image: %EI1%

Highest score image: %EIMA%

V014
```

Thanks!
Alistair

connortechnology:
I think I may understand the issue a little more. I use html emails.  When I cutnpasted your body into a filter and ran it, I get  tags which is useless because the email body is text. However the images are listed as attachments and nicely displayed inline after the body in thunderbird 115.6.0 on linux.
 
 So I can turn off the img tag substitution when we aren't doing an html email.

Maybe you should try making your body html to see what happens.

akgrant43:
Thanks for your reply and my apologies for my slow reply.
You're correct - changing the email format to html displays the images in-line.
Thanks again!

PR #3774, opened 2023-10-01
fix(php8): AllowDynamicProperties

Simpler1:
This fixes a problem that was introduced in PHP 8 which deprecated the use of dynamic properties.

The problem can be seen in the [zoneminder-containers/zoneminder-base](https://github.com/zoneminder-containers/zoneminder-base) after [pull request 56.](https://github.com/zoneminder-containers/zoneminder-base/pull/56)

Symptom:
When you click on any of the monitor names from the Console view, you will get a 502 Bad Gateway message when running with nginx.

nginx log files show the following message many times:
`PHP message: PHP Deprecated:  Creation of dynamic property <property name> is deprecated in <file name>`

### DIFF
--- a/web/includes/Object.php
+++ b/web/includes/Object.php
@@ -4,6 +4,12 @@ require_once('database.php');
 
 $object_cache = array();
 
+/**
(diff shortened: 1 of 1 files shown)


web/includes/Object.php
baudneo, line 1: Yeah boi!

PR #3779, opened 2023-10-29
Updates to Getting Started Doc Section

efranzwa:
Refresh the entire Getting Started (gettingstarted.rst) section to be more consistent with recent ZM version, reST style guide recommendations and more tailored to a first time user.


efranzwa:
Follow up on this PR. I noticed that the Read the Docs builds are failing as they no longer include the sphinx_rtd_theme as part of the build (I think they used to a few years ago) and it needs to be added as a requirement.  According to Read the Docs they recommend using requirements.yaml to call out python requirements under docs/requirements.txt.  This seems fairly straightforward to implement.  See some info here: https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html    I can work on this if you need me to. Cheers.

PR #3785, opened 2023-11-15
Fix Monitor Source Tab Rendering

efranzwa:
Monitor -> Source (for Web Site) was missing the 'Web Site Refresh (Optional)' description label.

Monitor -> Source (for NVSocket) was not rending correctly. Added missing `<table>` tags. Now renders okay but not as clean as other source types due to the use of table formatting. May consider cleaning this up later if NVSocket is to be kept in ZM 1.38.


connortechnology:
nvsocket will be going away. I'll remove it sooner rather than later.

PR #3787, opened 2023-11-21
Prevent allocation on already allocated context, fix #3786

### LINKED ISSUE #3786: Memory leak in video store - audio context allocation
**Describe Your Environment**
- ZoneMinder master branch commit 092f8df
- compiled from source
- own custom OS

**Describe the bug**
In the `VideoStore` class, if audio record is enabled, the `audio_in_ctx` context gets allocated two times additionally:
- https://github.com/d-bum/zoneminder/blob/092f8df115870f90ea203ef380d40a74ddfe53c2/src/zm_videostore.cpp#L463
- https://github.com/d-bum/zoneminder/blob/092f8df115870f90ea203ef380d40a74ddfe53c2/src/zm_videostore.cpp#L781

The context is already allocated at this point by the `FFmpeg_Input` class:
- https://github.com/d-bum/zoneminder/blob/092f8df115870f90ea203ef380d40a74ddfe53c2/src/zm_ffmpeg_input.cpp#L94

All subsequent allocations on the same pointer result in losing the reference to the allocated memory. This memory can not be freed anymore.

**Expected behavior**
Check the `audio_in_ctx` pointer before the first allocation in `VideoStore` class for `nullptr` and do not allocate if pointer is not `NULL`, just modify the context by the `avcodec_parameters_to_context` function.
The second allocation can be removed as it is enough to modify the context by `avcodec_parameters_to_context`.

d-bum:
Fixes issue #3786

### DIFF
--- a/src/zm_videostore.cpp
+++ b/src/zm_videostore.cpp
@@ -452,7 +452,9 @@ bool VideoStore::open() {
       if (!audio_out_codec) {
         Error("Could not find codec for AAC");
       } else {
-        audio_in_ctx = avcodec_alloc_context3(audio_out_codec);
+        if (audio_in_ctx == nullptr) {
+          audio_in_ctx = avcodec_alloc_context3(audio_out_codec);
+        }
         ret = avcodec_parameters_to_context(audio_in_ctx, audio_in_stream->codecpar);
         if (ret < 0)
           Error("Failure from avcodec_parameters_to_context %s",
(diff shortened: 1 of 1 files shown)


connortechnology:
I'm willing to merge this, but I'm thinking I would go further... I'm looking at the comment on 778... I'm not sure we should even be touching a ctx that isn't ours. Perhaps we should just allocate our own.  

Do you want to do more or are content to leave it for another day?

connortechnology:
Merging as the mem leak is important to fix.

d-bum:
Hello Mr. Connor, do you already have an idea or should I propose an implementation?

PR #3804, opened 2024-01-12
Debian error message

schabau:
In Zoneminder 1.37 this error message is thrown on Debian:

zmstats[951].WAR [main:73] [Failed getting user_utilization from ]

The problem is that the Top command on Debian does not output Cpu(s) but CPU(s).
This change would fix the problem.

### DIFF
--- a/scripts/zmstats.pl.in
+++ b/scripts/zmstats.pl.in
@@ -61,7 +61,7 @@ while (!$zm_terminate) {
   }
 
   # Get CPU utilization percentages
-  my $top_output = `top -b -n 1 | grep "^%Cpu(s)" | awk '{print \$2, \$4}'`;
+  my $top_output = `top -b -n 1 | grep -i "^%Cpu(s)" | awk '{print \$2, \$4}'`;
   my ($user_utilization, $system_utilization) = split(/ /, $top_output);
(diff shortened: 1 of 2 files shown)


connortechnology:
I have a branch that parses /proc instead of using top... This effort may be a waste of time.

schabau:
Well, currently it would be faster to add the "-i" than every Debian user having to adapt zmstats.pl with every update from 1.37. There will certainly be a few more updates to 1.37 before the change to /proc is made. ;-)

But fine, if not then not.

connortechnology:
Well, whatever you have done here, there are no changes, so if you want it merged... you need to make it have changes..

schabau:
I have to admit, I don't know anything about Github and I tried to make my suggestion here because it obviously went unnoticed in the Zoneminder forum.

I thought my suggestion for the zmstats.pl file would be listed here. At least that's how I see it here. Or am I not understanding that correctly?
github_Screenshot 2024-01-12 213810
But simply line 64 of the zmstats.pl file:
```
- my $top_output = `top -b -n 1 | grep "^%Cpu(s)" | awk '{print \$2, \$4}'`;
+ my $top_output = `top -b -n 1 | grep -i "^%Cpu(s)" | awk '{print \$2, \$4}'`;
```

PR #3817, opened 2024-02-06
Introducing a new ZM_DB_SSL_VERIFY_SERVER_CERT configuration option t…

jmpalacios:
…o allow

deployments to specify whether they wish to verify their database server's certificate or not, mapped to PHP's PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT attribute.

Resolves 3816.

### DIFF
--- a/web/includes/database.php
+++ b/web/includes/database.php
@@ -55,6 +55,7 @@ function dbConnect() {
         PDO::MYSQL_ATTR_SSL_CA   => ZM_DB_SSL_CA_CERT,
         PDO::MYSQL_ATTR_SSL_KEY  => ZM_DB_SSL_CLIENT_KEY,
         PDO::MYSQL_ATTR_SSL_CERT => ZM_DB_SSL_CLIENT_CERT,
+        PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT => ZM_DB_SSL_VERIFY_SERVER_CERT,
       );
       $dbConn = new PDO($dsn, ZM_DB_USER, ZM_DB_PASS, $dbOptions);
     } else {
--- a/cmakecacheimport.sh
+++ b/cmakecacheimport.sh
@@ -58,6 +58,7 @@ echo "Database password          : Not shown"
 echo "Database SSL CA Cert       : $ZM_DB_SSL_CA_CERT"
 echo "Database SSL Client Key    : $ZM_DB_SSL_CLIENT_KEY"
 echo "Database SSL Client Cert   : $ZM_DB_SSL_CLIENT_CERT"
+echo "Verify database SSL Cert : $ZM_DB_SSL_VERIFY_SERVER_CERT"
 
 
 CMPATH="CACHE PATH \"Imported by cmakecacheimport.sh\" FORCE"
@@ -78,6 +79,7 @@ echo "set(ZM_DB_PASS \"$ZM_DB_PASS\" $CMSTRING)">>zm_conf.cmake
 echo "set(ZM_DB_SSL_CA_CERT \"$ZM_DB_SSL_CA_CERT\" $CMSTRING)">>zm_conf.cmake
 echo "set(ZM_DB_SSL_CLIENT_KEY \"$ZM_DB_SSL_CLIENT_KEY\" $CMSTRING)">>zm_conf.cmake
 echo "set(ZM_DB_SSL_CLIENT_CERT \"$ZM_DB_SSL_CLIENT_CERT\" $CMSTRING)">>zm_conf.cmake
+echo "set(ZM_DB_SSL_VERIFY_SERVER_CERT \"$ZM_DB_SSL_VERIFY_SERVER_CERT\" $CMSTRING)">>zm_conf.cmake
 
 echo ""
 echo "Wrote zm_conf.cmake"
--- a/scripts/ZoneMinder/lib/ZoneMinder/Database.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Database.pm
@@ -102,7 +102,8 @@ sub zmDbConnect {
           'mysql_ssl=1',
           'mysql_ssl_ca_file='.$ZoneMinder::Config::Config{ZM_DB_SSL_CA_CERT},
           'mysql_ssl_client_key='.$ZoneMinder::Config::Config{ZM_DB_SSL_CLIENT_KEY},
-          'mysql_ssl_client_cert='.$ZoneMinder::Config::Config{ZM_DB_SSL_CLIENT_CERT}
+          'mysql_ssl_client_cert='.$ZoneMinder::Config::Config{ZM_DB_SSL_CLIENT_CERT},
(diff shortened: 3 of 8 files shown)


web/includes/database.php
```diff
@@ -55,6 +55,7 @@ function dbConnect() {
         PDO::MYSQL_ATTR_SSL_CA   => ZM_DB_SSL_CA_CERT,
         PDO::MYSQL_ATTR_SSL_KEY  => ZM_DB_SSL_CLIENT_KEY,
         PDO::MYSQL_ATTR_SSL_CERT => ZM_DB_SSL_CLIENT_CERT,
+        PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT => ZM_DB_SSL_VERIFY_SERVER_CERT,
```
connortechnology, line 58: My concern here is upgrades, people's existing files in /etc/zm.conf won't have the define, so this will break, so we need to ensure that ZM_DB_SSL_VERIFY_SERVER_CERT gets defined no matter what.

connortechnology:
Looks good, but we need to add it to config files in /etc/zm/ so that it gets defined, and looks like you've missed the perl side of things, in scripts/ZoneMinder/lib/ZoneMinder/Database.pm.  Good work tho! Thanks!

jmpalacios:
> Looks good, but we need to add it to config files in /etc/zm/ so that it gets defined, and looks like you've missed the perl side of things, in scripts/ZoneMinder/lib/ZoneMinder/Database.pm. Good work tho! Thanks!

Awesome, thanks for the initial review!

How would I add it to the main config file? Wasn't that covered in my patch to `zm.conf.in`?

And as for Perl, which I'm a complete neophyte on, I'll be pushing a commit shortly to `scripts/ZoneMinder/lib/ZoneMinder/Database.pm`, assuming the attribute to use would be `'mysql_ssl_verify_server_cert'`, similar to how I used `MYSQL_OPT_SSL_VERIFY_SERVER_CERT` in `src/zm_db.cpp:49`, but please correct me if I'm wrong because this is a complete guess (and, to be honest, so was `MYSQL_OPT_SSL_VERIFY_SERVER_CERT`!).

jmpalacios:
ZM_DB_SSL_VERIFY_SERVER_CERT configuration added to the Perl database module in 51812416bc9bf553378fd20cdf6cd99fcf63c926.

connortechnology:
Looks good to me now.  We can fix what breaks after merge.

connortechnology:
Aha, first problem: 
MYSQL_OPT_SSL_VERIFY_SERVER_CERT (argument type: my_bool *)

Enable or disable verification of the server's Common Name identity in its certificate against the host name used when connecting to the server. The connection is rejected if there is a mismatch. For encrypted connections, this feature can be used to prevent man-in-the-middle attacks. Identity verification is disabled by default.

This option does not work with self-signed certificates, which do not contain the server name as the Common Name value.

This option is deprecated as of MySQL 5.7.11 and is removed in MySQL 8.0. Instead, use MYSQL_OPT_SSL_MODE with a value of SSL_MODE_VERIFY_IDENTITY.

jmpalacios:
OK, so, as I suspected, I had indeed read about these SSL changes in MySQL 5.7, I was just a bit confused in the context of ZoneMinder & PHP support for them.

So, all in all, on the side of MySQL we have, from 5.7.11 onwards, valid [options](https://dev.mysql.com/doc/c-api/5.7/en/mysql-options.html) `SSL_MODE_DISABLED`, `SSL_MODE_PREFERRED `, `SSL_MODE_REQUIRED `, `SSL_MODE_VERIFY_CA `, and `SSL_MODE_VERIFY_IDENTITY ` for the `MYSQL_OPT_SSL_MODE ` setting/enumeration, with `SSL_MODE_VERIFY_IDENTITY` mapping to the old-style `MYSQL_OPT_SSL_VERIFY_SERVER_CERT ` option that I introduced in `src/zm_db.cpp:49`.

However, I don't know if ZoneMinder aims to support RDBMs other than MySQL, and/or if it intends to support MySQL releases older than 5.7.11 (even with 5.7's EOL date fast approaching). Moreover, PHP's PDO, as far as I can tell, seems to still only support the old-style MySQL configuration via its sole `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` attribute; I can't find anything anywhere to the effect of supporting the more fine-grained configuration, not in PHP's documentation nor even in the latest state of its underlying code for the upcoming 8.3.3 release:

- https://www.php.net/manual/en/ref.pdo-mysql.php
- https://github.com/php/php-src/blob/php-8.3.3RC1/ext/pdo_mysql/pdo_mysql.c#L125

And you'll notice that, under-the-hood, PHP [maps](https://github.com/php/php-src/blob/php-8.3.3RC1/ext/pdo_mysql/mysql_driver.c#L891) this sole `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` constant to OpenSSL's `CLIENT_SSL_VERIFY_SERVER_CERT` parameter, so there doesn't seem to be much leeway there to get creative with MySQL's modern SSL verification granularity.

All in all, I'm unsure at this point how this SSL verification configuration should be exposed to the ZoneMinder configuration, and how it should be used across its different layers. The CPP, Perl, and PHP layers do seem to be hardwired to use MySQL, which should allow us to expose the more fine-grained approach to the configuration layer, even if the PHP layer is short-circuited to always use `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` with a true/false value (that'd be trivial); but that'd still leave the question of whether MySQL releases < 5.7.11 should be supported.

And if there is indeed a need to support other RDBM's, then I would be even more unsure of how to approach the problem.

Please advise!

connortechnology:
At the moment we are hard wired to use mysql/mariadb, but there is a PR to convert the c++ side to something that can also do Postgresql and others. 

FOr now we can just worry about mysql/mariadb.  I think we do need to continue to support 5.7 though.  I think the c++ side can be solved just with some #ifdefs.

jmpalacios:
> At the moment we are hard wired to use mysql/mariadb, but there is a PR to convert the c++ side to something that can also do Postgresql and others.
> 
> FOr now we can just worry about mysql/mariadb. I think we do need to continue to support 5.7 though. I think the c++ side can be solved just with some #ifdefs.

I don't know if MariaDB supports any of these constants, but I'd assume it does, given its origins. So, from that point of view, there shouldn't be a need for any conditional support of either approach when considering MariaDB, if I'm not mistaken.

As for still supporting MySQL 5.7, until when would you expect to keep it around? It's end-of-life date was October 2023 (I was mistaken when I said "fast-approaching" some comments above), and Amazon RDS will fully drop support for it on February 29, 2024. Your support timeline might as a result be the biggest factor in deciding on whether we move to the new finer-grained constants, or keep using the older approach for the time being.

connortechnology:
Since I reverted this, it isn't really merged.  Do you mind creating a new PR?

Issue #3821, opened 2024-02-09
Incorrect chronology of joining several videos when downloading (v1.37.50 )


IgorA100:
**Describe Your Environment**
- Version of ZoneMinder - 1.37.50
- How you installed ZoneMinder - PPA (Updated 06 feb 24)
- Full name and version of OS - Ubuntu 23.10 mantic

**Describe the bug**
In the final video file, the first event becomes the last, and the last one becomes the first!

**To Reproduce**
![1_2](https://github.com/ZoneMinder/zoneminder/assets/5006170/78607cab-1f8e-44e8-a8a7-5215bbee14f6)

**Expected behavior**
The first event (chronologically) in the final file should be the first, and the last one should be the last, and not vice versa.

PR #3822, opened 2024-02-09
Sorting an array of events when merging into a file

IgorA100:
When merging events into a file, they must first be sorted by time and start date of the event.

Closing the issue: #3821

### DIFF
--- a/web/includes/download_functions.php
+++ b/web/includes/download_functions.php
@@ -18,10 +18,6 @@
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 // 
 
-function cmpDataTime($a, $b) {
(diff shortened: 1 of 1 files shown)


web/includes/download_functions.php
```diff
@@ -18,6 +18,10 @@
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 // 
 
+function cmpDataTime($a, $b) {
+    return strcmp(strtotime($a->StartDateTime), strtotime($b->StartDateTime));
```
connortechnology: strtotime returns an integer so we shouldn't really use strcmp... can just use <=> I think

IgorA100: Yes, of course you're right.
I optimized the code.

PR #3824, opened 2024-02-10
regexp: drop obsolete pcre3 and port regexp usage to pcre2

davvid:
pcre3 is deprecated and was replaced by pcre2. This fixes the issue that caused zoneminder to be dropped from Debian/testing.

A flavor of this patch has already been submitted to Debian: https://salsa.debian.org/debian/zoneminder/-/merge_requests/2

Related-to: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=999919

### DIFF
--- a/distros/debian/control
+++ b/distros/debian/control
@@ -3,6 +3,7 @@ Section: net
 Priority: optional
 Maintainer: Isaac Connor <iconnor@connortechnology.com>
 Build-Depends: debhelper (>= 9), cmake
+    , libphp-serialization-perl
     , libgnutls28-dev | libgnutls-dev
     , libmysqlclient-dev | libmariadbclient-dev | default-libmysqlclient-dev
     , libjpeg8-dev | libjpeg-dev
--- a/distros/debian/control
+++ b/distros/debian/control
@@ -3,17 +3,16 @@ Section: net
 Priority: optional
 Maintainer: Isaac Connor <iconnor@connortechnology.com>
 Build-Depends: debhelper (>= 9), cmake
-    , libphp-serialization-perl
     , libgnutls28-dev | libgnutls-dev
-    , libmysqlclient-dev | libmariadbclient-dev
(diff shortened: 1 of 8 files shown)


connortechnology:
Are there not actual api differences that have not been addressed?

We actually don't need a pcre library. It is used in the old remote monitor code and without it we do our own parsing of http headers etc.  

So my question is, do we compiled with pcre2?  If not, we will need those commits as well.  Or perhaps we should simply remove the code entirely.

davvid:
Everything compiled with libpcre2 installed but I didn't check whether it's actually being used, and relatedly, whether it's just compiling the same way as it would if pcre were not installed at all and using the fallback code.

I'll take a closer look to see if it's actually getting used and if not get it enabled and port the code. I didn't realize that there's a chance it could be falling back to builtin code.

Thanks for the heads-up that this code could possibly be removed. If you have a preference for doing it that way I can do that instead, but I'll take a stab at porting it first unless I hear otherwise.

davvid:
I've pushed an update that ports the code over to pcre2.

Here are some links that I used as reference:

* https://github.com/PCRE2Project/pcre2/issues/51
* https://github.com/i3/i3/issues/4682#issuecomment-973076704
* https://www.pcre.org/current/doc/html/index.html

Please test this thoroughly before merging as I don't have an easy way to test this out locally. I can confirm that it all compiles and links, though.

distros/debian/control
```diff
@@ -3,17 +3,16 @@ Section: net
 Priority: optional
 Maintainer: Isaac Connor <iconnor@connortechnology.com>
 Build-Depends: debhelper (>= 9), cmake
-    , libphp-serialization-perl
```
connortechnology, line 6: Why are you dropping this dependency?

davvid: Oops, that was a mistake. I rebased and squashed this fixup in.

PR #3825, opened 2024-02-10
ConfigData: default to not checking for updated versions of ZoneMinder

davvid:
Users should get updates through the package.
Also prevents "phoning home" to zoneminder.com.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
+++ b/scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in
@@ -2647,9 +2647,9 @@ our @options = (
       website to determine the most recent release. These checks are
       infrequent, about once per week, and no personal or system
       information is transmitted other than your current version
-      number. If you do not wish these checks to take place or your
-      ZoneMinder system has no internet access you can switch these
-      check off with this configuration variable
+      number. If you wish these checks to take place and your
+      ZoneMinder system has internet access you can switch these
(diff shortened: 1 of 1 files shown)


davvid:
The first patch in this series originated from Debian's patches:
https://salsa.debian.org/debian/zoneminder/-/tree/master/debian/patches?ref_type=heads

SteveGilvarry:
I believe there is an edit config data script in utils designed for distros
to change config for packages. Mobile so can’t grab a link for you.

davvid:
Thanks for the pointer @SteveGilvarry much appreciated. I mostly submitted this patch because I agree with Debian's position that the tool should be privacy-respecting by default (for all users) rather than something that's Debian-only. This isn't really a big deal for Debian since they're already patching it, but it seemed worth sharing in case y'all agree.

We can close this PR if maintainers feel otherwise, though. It's all good. Thanks again for zoneminder.

connortechnology:
The problem here is that people do not manually check for updates.  We still have people using ZoneMinder from 10 years ago, that we know are completely insecure and broken. 

We agree with the general stance of respecting privacy, but we also want people to get updates.  So since distros manage the updates, that's cool that they disable this.  For anyone else, I think we should stick with it enabled.  Now, it probably SHOULD be a bigger, more in your face decision to make upon first install, like the telemetry configuration.  

I would certainly merge a PR that did that in addition to this one.

PR #3826, opened 2024-02-10
zm_user: avoid memory leak when user cannot be authenticated

davvid:


davvid:
This is probably a minor issue in practice, though.

Slightly related ~ are the maintainers amenable to switching over to extensive use of `std::shared_ptr` and `std::unique_ptr` instead of raw pointers? There's a lot of manual memory management that would be simplified by using these but doing so would probably be a pretty major change.

PR #3833, opened 2024-02-20
Add zmDbReconnect, and more!

connortechnology:
Add zmDbReconnect, which closes the connection before trying to reconnect, in an attempt to prevent mem leak.  This adds a 1 sec sleep if we fail to connect. This also removes the disabling of db logging before doing logging... because we are going to assume (possibly incorrectly) that the logging code is correct and generating well-formed sql, and hence the error is external. This might be a bad idea.

### DIFF
--- a/src/zm_db.cpp
+++ b/src/zm_db.cpp
@@ -21,6 +21,7 @@
 #include "zm_logger.h"
 #include "zm_signal.h"
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


webgeek1234 (approved):
From a code review perspective, this looks like an improvement. Cleaning up a stale db connection before trying to open a new one. Have not verified by testing.

PR #3840, opened 2024-02-24
Fix: Add Cpu Usage % to header for standalone server

IgorA100:
After:
After

### DIFF
--- a/web/ajax/status.php
+++ b/web/ajax/status.php
@@ -19,6 +19,7 @@ if ($_REQUEST['entity'] == 'navBar') {
   // Call the functions we want to dynamically update
   $data['getBandwidthHTML'] = getBandwidthHTML($bandwidth_options, $user);
   $data['getSysLoadHTML'] = getSysLoadHTML();
+  $data['getCpuUsageHTML'] = getCpuUsageHTML();
   $data['getDbConHTML'] = getDbConHTML();
   $data['getStorageHTML'] = getStorageHTML();
   //$data['getShmHTML'] = getShmHTML();
--- a/web/includes/Server.php
+++ b/web/includes/Server.php
@@ -199,31 +199,45 @@ class Server extends ZM_Object {
     return $this->CpuLoad;
   }
 
-  public function getServerLoadData() {
-    if (is_readable("/proc/stat")) {
-      $stats = @file_get_contents("/proc/stat");
+  public function getServerLoadData($mode) {
+    $stats = false;
+    $fileName = null;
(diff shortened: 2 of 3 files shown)


connortechnology:
The reason I didn't implement this, although it is nice to have as part of the toolset, is that I WILL NOT add a 1 second pause to loading the UI, so the plan was to make sure the db Server object existed and got updated even for non-multi-server case.

 Also, beware, not all OS's have /proc.  In fact, linux may not have /proc either.

IgorA100:
> I WILL NOT add a 1 second pause to loading the UI,

I've updated the code.
The 1 second pause is no longer used.

> Also, beware, not all OS's have /proc. In fact, linux may not have /proc either.

For Windows, you can receive the download without any gaps. I can add some code. But can Zoneminder be installed on Windows?
What systems can ZM be installed on?

> linux may not have /proc either.

I’m not ready to answer anything yet, I need to think.
Which Linux systems don't have /proc

connortechnology:
OpenBSD is the prime example of not having /proc.  In linux land a container or something might not have it.  It is uncommon, but possible.  Which is why I left the top parsing code in perl-side.  

The top parsing is not reliable though.  I intend to change the parsing to perl regexps instead of bash awk, etc to see what the actual output is that fails. 

In terms of SERVER_ID, we can just use = 0 when not in multi-server mode.  Code just needs a few tweaks to make that work.  

This is all work in progress.  Might be better if you joined slack, or discord or something to discuss with me more to prevent overlap of effort.  Or maybe even the discussion parts of github since you are already here. Or forums...

IgorA100:
Ok Isaac, I get it.
I'm more used to communicating on the forum.
You've done a lot of work on detecting memory leaks. I haven't found any obvious problems with memory usage in the latest versions!

connortechnology:
I'll try communicate better.  Too used to working alone around here.  I'll likely merge this, but just be aware that my plan is for most of this nice work to be not needed as the values will be updated in background in the db.

IgorA100:
I understand your thoughts.
Yes, my solution is temporary.
Maybe I'll look at the Pearl scripts a little later.
You already have more important tasks.

davidcomerford:
I'm confused why we'd want CPU usage if we already have system load average?

IgorA100:
These are slightly different values.

connortechnology:
They are very different values.  On many of my servers, the load can be 20 or more, but cpu usage is about 50%.  There are lots of different reasons why processes might be waiting on service (which is what load is a measure of).

IgorA100:
Closed as a draft because... there is a more recent version #3843

PR #3841, opened 2024-02-24
Makes the "Events" part of the ONVIF URL configurable

ovargasp:
Makes the "Events" part of the ONVIF URL configurable
as required per some ONVIF version 21.06 cameras
by default sets the Events Path equals to "/Events" to maintain functionality


jp-bennett:
Looks good to me!

PR #3842, opened 2024-02-26
Changing uptime return line parsing for CpuLoad() & add CpuNicePercent, CpuIdlePercent, CpuUsagePercent to Server_Stats

IgorA100:
Changing uptime return line parsing for CpuLoad()
Depending on the locale it could be like this:
`16:50:53 up 1 day, 1:10, 3 users, load average: 0.89, 0.94, 1.14` 
or like this:
`16:50:53 up 1 day, 1:10, 3 users, load average: 0.89, 0.94, 1.14`


connortechnology:
Likely just an oversight.  I'll review and add them today.

connortechnology:
SO you write to the servers table instead of writing to a file in /tmp.

IgorA100:
Please try merge and build.
In theory, values should be stored in DB,

PR #3843, opened 2024-02-27
Refactoring: Stats to header (Cpu Usage %)

IgorA100:
Perl script enters the calculated CPU loads into DB.
PHP gets data from DB and displays it


IgorA100:
@connortechnology 

I tried to implement your idea.
For a single server everything works correctly.
I don’t have the opportunity to check for multi-servers, but in theory it should work.

Please merge and build.

IgorA100:
We need to resolve the issue with "$Config{ZM_SERVER_ID}"
When it is NULL it is bad because:
“NULL” is entered into the database, and on the “Options->Servers” page it is designated as “1”. Because of this, the values are not displayed correctly on the “Options->Servers” page.
But I don't know how to correctly change the "$Config{ZM_SERVER_ID}" setting from NULL to 1

IgorA100:
@connortechnology 
Is there something you don't like or don't have time to check the code?

connortechnology:
See the comments.  Remove removed the caching of previous calls and added a sleep.  I don't want to sleep if we don't have to.

IgorA100:
Ok, I'll change the code.

IgorA100:
@connortechnology
I made changes and abandoned sleep().

IgorA100:
Without using a temporary file with the previous state it will not work.
I don't see any point in storing temporary information in the "Server_Stats" table. In addition, the table will have to be enlarged by adding 4 columns.
Do you like it or not?

IgorA100:
Did I do something wrong?

connortechnology:
At no time did I mention the Server_Stats table.  I mean leave it in the in-memory server object.

IgorA100:
@connortechnology 
Forgive me Isaac.
I don't know Perl very well.

I hope you are now satisfied with my code?

connortechnology:
When I find the time, I'll modify it to suit me and merge.  I'm actually on vacation right now... but also working and signing up new clients.  Work load 400%.

IgorA100:
> I'm actually on vacation right now... 

Sorry, Isaac.
It's good for you to rest!

IgorA100:
@connortechnology 
Didn't find time to look at this PR?
It works for me without problems.

connortechnology:
I have merged it, and also reverted the parts I didn't like. ANd fixed the sql for loading of stats in php.

PR #3845, opened 2024-02-29
Added a link to live viewing from monitor event viewer

IgorA100:
1. When viewing events, there is no way to go to a live view of the monitor in real time in one click.
2. But, there are two links to go to the monitor settings.

It was decided to replace one of the links with a transition to live viewing.

P1040680

### DIFF
--- a/web/skins/classic/css/base/skin.css
+++ b/web/skins/classic/css/base/skin.css
@@ -582,6 +582,7 @@ body.sticky #page {
 
 #content table input[type=submit], #content table input[type=button], #content table button, #content table .btn-primary {
     margin-top: 1px;
+    margin-right: 4px;
 }
 
 #contentButtons {
--- a/web/skins/classic/css/base/views/event.css
+++ b/web/skins/classic/css/base/views/event.css
@@ -104,10 +104,10 @@ height: 100%;
 #eventStatsTable {
 }
 #eventVideo {
-  width: 100%; /* flex will make it fill the available space */
+  /* width: 100%; *//* flex will make it fill the available space */
 }
 .eventStats {
-  margin-right: 5px;
+  /* margin-right: 5px; */
 }
 
 #videoFeed {
--- a/web/skins/classic/css/classic/skin.css
+++ b/web/skins/classic/css/classic/skin.css
@@ -416,6 +416,7 @@ th.table-th-sort-rev span.table-th-sort-span {
 #content table button,
 #content table .btn-primary {
     margin-top: 0;
+    margin-right: 4px;
 }
 
 #contentButtons {
--- a/web/skins/classic/css/classic/views/event.css
+++ b/web/skins/classic/css/classic/views/event.css
@@ -68,8 +68,8 @@ height: 100%;
 }
 
 #eventVideo {
-   display: inline-block;
-   position: relative;
+   /* display: inline-block;
+   position: relative; */
 }
 
 #imageFeed {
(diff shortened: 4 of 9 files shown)


connortechnology:
But you removed the link to go to monitor settings.  How about instead of this, we use icons and have on for settings and one for live view?

IgorA100:
I replaced ONE link out of two!
The second one remains. Using it you can easily get into the monitor settings.
12r

I thought about adding a button, but I didn’t like this idea because... it will be a redundant element.
There is already a clickable place. Why have two links to the same action next to each other? And I decided to replace one of the links. This way, unnecessary elements are not added and the functionality is expanded.

connortechnology:
Ok, but it's not very good UI.  I actually think the two lines should be merged.  It isn't intuitive that clicking on one goes one place and the other a different place, which is why I think we should use icons.

IgorA100:
Ok, since you think it would be better to add another button and remove one duplicate line, I can do that.
I suggest making a button for live viewing like this: https://fontawesomeicons.com/television
or this one: https://fontawesomeicons.com/video-camera
Or offer your own option.
And in what place is it best to add it?

connortechnology:
That icon looks fine.  We include both material icons and font awesome... it would be good at some point to standardize on one of them and remove the other.  At this time I don't know which is better, although I tend to gravitate towards material icons, but then I often don't find what I need.  I would just insert the buttons on the line with the Monitor Id/Name.

IgorA100:
I changed the design.
I moved name & Id of the monitor to another place, similar to live viewing. I think it will be better this way.
I would add another button to go to view all events. Currently there is only a button at the top to view all frames for the event.

Yes, I saw that different sets of icons are used.
I agree that ideally you should use one set. But I can’t say anything concrete yet.
I'll see if it's possible to give up one of the sets.

IgorA100:
For example like this:

IgorA100:
@connortechnology 
If you like the picture, I'll make changes to the code.

connortechnology:
It looks great, but I want you to keep the Id and Name before the buttons.

IgorA100:
> It looks great, but I want you to keep the Id and Name before the buttons.

No problem.


But it won’t be pretty, especially if the monitor name is long :(
Why do this? The monitor ID and its name are normally visible under the video.

connortechnology:
THere is enough space there for it to fit.  I think we just need a little css magic to make it layout better.   honestly I would remove the monitor info at the bottom.  
My feelings on the matter are simply that I am used to looking there for the name, I want to know which monitor it is.

IgorA100:
> THere is enough space there for it to fit. I think we just need a little css magic to make it layout better. 

css is not a magician :)
Yes, with the current monitor name and high resolution there is no problem displaying everything in one line.
But, moving icons to a new line will happen in the following case:
- long monitor name
- low screen resolution
There is no way to fix this with css.
Or make two lines at once (one with ID and name, the second with icons).

IgorA100:
In the works, a fix will be coming soon.
A lot of work. time is short.

IgorA100:
I finished work
Now like this:




Now everything scales correctly depending on the screen size.

IgorA100:
If you have any comments, I am ready to discuss.
But I think it turned out not bad.

connortechnology:
Good work, looks great.  I was wanting this feature just the other day.

IgorA100:
@connortechnology 
Thank you!
If I find time I will make changes (using responsive Bootstrap) for live view. There are some things that are wrong.

PR #3846, opened 2024-03-01
Placeholder text for monitor date time format.

Simpler1:
After spending an hour trying to figure out which date time format to enter here, I thought it would be best to just spell it out for users trying to customize this.

### DIFF
--- a/web/skins/classic/views/monitor.php
+++ b/web/skins/classic/views/monitor.php
@@ -1331,7 +1331,7 @@ $codecs = array(
 ?>
             <li>
               <label><?php echo translate('TimestampLabelFormat') ?></label>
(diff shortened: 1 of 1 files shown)


connortechnology:
This is fine, however we really should link to some useful documentation instead.  We have a few extra characters that aren't posix. I thought there was some in readthedocs but I can't find it.

connortechnology:
What docs there are here: https://zoneminder.readthedocs.io/en/latest/userguide/definemonitor/definemonitor_timestamp.html

Simpler1:
This is more of a quick reminder.  We can also add a link to documentation, but that change would be more involved and require a new button/link.

PR #3849, opened 2024-03-04
Fix: Responsive design with use bootstrap for live view

IgorA100:
Changed responsive design.
Now everything is scaled and displayed more correctly.


IgorA100:
Now like this:

PR #3852, opened 2024-03-04
Fix style in options pages

IgorA100:

### DIFF
--- a/web/skins/classic/css/dark/skin.css
+++ b/web/skins/classic/css/dark/skin.css
@@ -42,6 +42,10 @@ h4 {
     color: #117AAd;
 }
 
+a {
+  cursor: pointer;
+}
+
 p {
     color: #dddddd;
 }
--- a/web/skins/classic/css/dark/skin.css
+++ b/web/skins/classic/css/dark/skin.css
@@ -42,10 +42,6 @@ h4 {
     color: #117AAd;
 }
 
-a {
-  cursor: pointer;
-}
-
 p {
     color: #dddddd;
 }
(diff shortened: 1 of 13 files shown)


web/skins/classic/css/dark/skin.css
```diff
@@ -42,6 +42,10 @@ h4 {
     color: #117AAd;
 }
 
+a {
```
connortechnology: This shouldn't be neccessary as you've added it to the base... skin.css.. or base/skin.css.  We always include base.

IgorA100: Of course you're right, sorry.
I changed code.

connortechnology:
This is all fine... but in general we have an inconsistency about where to put the buttons, top or bottom.  I was leaning towards top, and generally having them stay sticky so that they are always available to click.  We can address that in another PR though. Just wanted to bring it up.

IgorA100:
Yes, I should have asked about the button placement. But, on this layout there is no problem moving them up. 



Most of the buttons for working with tables were at the bottom, so I moved the rest down.
The "Save" button is still at the top.
With buttons sticking when scrolling, it is necessary to leave the header motionless.
Perhaps we should think about a new accordion-style left navigation menu design and reconsider the top menu in the header?

IgorA100:
If scroll table with thead fixed, then at high resolution everything will be beautiful, but at low resolution everything will be bad from the fixed filters.
Land01
Land1

connortechnology:
Console needs a lot of work. Filters should not all be visible but show on demand. And collapse.

PR #3856, opened 2024-03-07
Fix design when viewing list of events

IgorA100:
Change layouts & space in code.
`<div id="inner-content">` now not used.


connortechnology:
This broke scroll bars.  Please fix.

IgorA100:
@connortechnology 
Explain to me what's wrong here? What don't you like? What should I change?

PR #3857, opened 2024-03-07
Fix "overflow-y" for "#monitorList" in console

IgorA100:
For "#monitorList" there is no need to set "overflow-y: scroll;" Otherwise, scrolling will always be displayed.


connortechnology:
This broke scroll bars.  I've put them back, you may want to review

IgorA100:
@connortechnology 
I wrote in a comment to your commit

PR #3860, opened 2024-03-10
Feat: Control of collapsing/expanding blocks

IgorA100:
Added functions for controlling the visibility of blocks when clicking on the control button.
Now template is like this:
1. Collapsed/expanded object::
```
    <div class="filterBar hidden" id="fbpanel">
      .....
    </div>
```
Where:
- class "filterBar" - any design class, not required
- class "hidden" necessary so that when the page is refreshed, there is no unnecessary expansion in case of a collapsed display.
- id="fbpanel" - unique object ID

2. Control button:
```
<a href="#" data-flip-сontrol-object="#fbpanel">
     <i id="fbflip" class="material-icons" data-icon-visible="filter_alt_off" data-icon-hidden="filter_alt"></i>
</a>
```
or
```
<a href="#" data-flip-сontrol-object="#fbpanel">
     <i id="fbflip" class="fa-font-avesome" data-icon-visible="filter_alt_off" data-icon-hidden="filter_alt"></i>
</a>
```
Where:
- data-flip-сontrol-object="#fbpanel" - attribute equal to managed object ID
- id="fbflip" - unique object ID, not required
- class="material-icons" OR class="fa-font-avesome" - indicated when using the corresponding icon set.
- data-icon-visible="filter_alt_off" - icon displayed when an object is expanded
- data-icon-hidden="filter_alt"- icon displayed when object is collapsed

You don't need to specify anything else in the template. The script does the rest.
- There can be an unlimited number of objects on one page.
- Support for Font Awesome and Material icons.
- Clean template code.
- Very easy replacement of the icon on the control button
- It is planned to be used to control the visibility of filters, so as not to write your own code for each filter block.


IgorA100:
Initially, I wanted to replace the filter display control icons on the console, but I couldn’t do it quickly. That's why I decided to write this code.
After writing this code, icons can be changed in a few seconds and in only one place in the template.
I think it would be better this way:


Or don't like the new filter icon?

IgorA100:
@connortechnology 
I'm interested to know your opinion.
This code shouldn't break anything, because... These are just additional features that are not yet used.

web/skins/classic/js/skin.js
```diff
@@ -300,6 +301,53 @@ if ( currentView != 'none' && currentView != 'login' ) {
       }
     });
 
+    // Manage visible object & control button (when pressing a button)
+    $j("[data-flip-сontrol-object]").click(function() {
+      let objIconButton = $j(this).find("i");
```
connortechnology: really nitpicking, but why not just do something like const jthis = $j(this); once at the top and use it forever more.

connortechnology:
In general looks good... I definitely like the filter icon instead of a chevron.

web/skins/classic/js/skin.js
IgorA100: I had this idea, but decided not to do it (at a slight cost to performance), because:
1. "$j(this)" doesn't appear that many times, only three times in one function and four times in another.
2. Now it’s more clear.
3. Add two lines of code.

But now I doubt that I did the right thing....
Maybe it's really worth changing the code?

IgorA100:
> I definitely like the filter icon instead of a chevron.

I made changes to the code for "Console" page

web/skins/classic/js/skin.js
IgorA100: I added
const _this_ = $j(this);

connortechnology:
The size of the filter icon is weird on mobile. Smaller than buttons so it just looks a little weird. 
THe animation to hide/show the filters is a little jerky

IgorA100:
> The size of the filter icon is weird on mobile. Smaller than buttons so it just looks a little weird.

Now it displays like this for me:


The filter button can be aligned vertically (by placing it in the button block).


Or design it like all other buttons.


Which one do you prefer?

> THe animation to hide/show the filters is a little jerky

JQ is entirely responsible for this, I did not touch the algorithm and left it as it was before. It used to be jerky too.
I'll see what i can do.

IgorA100:
@connortechnology 
Jerks occur only in "stickey" mode.
So far I haven't been able to find a solution.

connortechnology:
Don't worry about it.

PR #3863, opened 2024-03-13
Code cleaned & added table header background in events

IgorA100:
Fixing sticky navbar problem

Corrected #3856


IgorA100:
@connortechnology 
I saw your hot fix, but I made minor improvements to cleaner and more correct code. Everything should now work correctly with sticky and moving styles.

PR #3864, opened 2024-03-13
Fix page console

IgorA100:
- Removed duplication of scroll bars
- Fix table header when using sticky

### DIFF
--- a/web/skins/classic/views/console.php
+++ b/web/skins/classic/views/console.php
@@ -231,7 +231,8 @@ echo $navbar ?>
 <?php
 ob_start();
 ?>
-	<div id="monitorList" class="container-fluid table-responsive pt-2">
+    <div class="pt-2"></div>
+    <div id="monitorList" class="container-fluid table-responsive">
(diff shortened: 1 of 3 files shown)


web/skins/classic/views/console.php
```diff
@@ -231,7 +231,8 @@
 <?php
 ob_start();
 ?>
-	  <div class="container-fluid table-responsive-sm pt-2" id="monitorList">
+    <div class="pt-2"></div>
```
connortechnology: Is this pt-2 div even necessary?  Just thinking we could achieve the spacing another way.. no biggie though.

IgorA100: Another option is to set the class "pb-2" for "id="contentButtons"", but you need to look at what this will affect.
I did the most painless option (so that nothing would break) for correct operation.
There is still a lot of work on the templates.

IgorA100: I made changes.
Nothing has broken globally.

PR #3865, opened 2024-03-13
Fis style in Options page

IgorA100:
Commit https://github.com/ZoneMinder/zoneminder/commit/95d2e0a27cf54daf69ca2fec5530ced4edf154cd fixed one thing and broke something else

"display: flex;" do not use for "form", because will cause other problems, for example with buttons on the "API" page
margin-right: -15px; It is not allowed to use "#options"; other places in the template break.

I tried to fix the problems as cleanly as possible, but everything is not perfect in terms of writing the code, although visually everything is good.

In the future I will try to find a more optimal solution instead:
```
body.sticky #options {
  height: calc(100% - 50px); /* It's bad, but it's temporary */
  overflow-y: auto;
}
```
Right now I don't know how to change this using pure CSS.

Now it looks like this

### DIFF
--- a/web/skins/classic/css/base/views/options.css
+++ b/web/skins/classic/css/base/views/options.css
@@ -34,6 +34,9 @@ input.large {
   min-width: 160px;
 }
 
+#optionsContainer {
+}
+
 #options {
 /*  margin-right: -15px;*/ /* you shouldn't do this for the sake of shifting the scrool bar, it breaks the style in different places. done differently*/
   width: 100%;
--- a/web/skins/classic/css/base/views/options.css
+++ b/web/skins/classic/css/base/views/options.css
@@ -37,10 +37,14 @@ input.large {
 #options {
 /*  margin-right: -15px;*/ /* you shouldn't do this for the sake of shifting the scrool bar, it breaks the style in different places. done differently*/
   width: 100%;
-  margin-top: 15px;
 }
 
-body.sticky #sidebar, body.sticky #options {
+body.sticky #options {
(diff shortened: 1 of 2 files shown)


connortechnology:
Can we not make the whitespace changes.. it seems to be confusing github and I can't tell if you've accidentally broken the code to display the options or not.  I suspect not, but things like that have certainly snuck in before.

Just FYI, I really don't like the bootstrap way of doing things, like adding pb-2 classes to an otherwise unneeded div.  I'm ok with it if it gets the job done, but I don't like it.  I like a more semantic approach and using css to style things. The main reason being that we actually do many white labels of this UI and being able to drastically change the look and feel all with CSS is really useful.

IgorA100:
> Can we not make the whitespace changes.. 

I added a new DIV block and had to add spaces to align the code. Yes, it makes a mess when committing, but this makes the code more readable and easier to correct in the future.
 
> Just FYI, I really don't like the bootstrap way of doing things, like adding pb-2 classes to an otherwise unneeded div. 

I agree. But, in this case, this is a shorter way to achieve the goal. This is due to stickiness
In the future, I will analyze the code in more detail and try to optimize it.

connortechnology:
I decided years ago to treat html indentation and php indentation separately, and not be a fanatic about it.  If we can can't read the diff, then we havn't achieved anything.

IgorA100:
Yes, you are right.
That's what I did before. And now I'm too carried away...
I removed previously added spaces in PHP code

PR #3871, opened 2024-03-16
Language display according to translation

IgorA100:


IgorA100:
Lots of dependencies on the getLanguages() function. We need to look into it more carefully. I'm closing it for now.

PR #3878, opened 2024-03-19
Fix: Location of PTZ Controls buttons when scaling browser less than 100%

### LINKED ISSUE #3877: PTZ Controls layout could use improvement
See here for a pic example:
https://forums.zoneminder.com/viewtopic.php?t=33148

I've never gotten around to improving them

IgorA100:
Possible fix #3877

### DIFF
--- a/web/skins/classic/css/base/views/control.css
+++ b/web/skins/classic/css/base/views/control.css
@@ -100,9 +100,66 @@
 }
 
 .ptzControls .controlsPanel .pantiltPanel .pantiltControls .arrowBtn {
-    width: calc(32px - 0.1rem); /* ( - 0.1rem) Necessary for correct operation at browser scale less than 100% */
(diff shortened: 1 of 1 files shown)


connortechnology:
This looks fine... just wondering what display: flex would do... probably not get it quite right... just that display: flex seems to work so well 80% of the time.   Just asking for your thoughts, otherwise I'll merge in the morning.

IgorA100:
display: flex - this is a quite useful (albeit a little mysterious) property, but in this case we have no pixels in stock, and when scaling, a recalculation almost always occurs with a shift of several pixels.
This is my subjective opinion based on some experience.

IgorA100:
I changed code.
It will be better this way.
Because when calculating +/- 1px there will be an error and if you look closely, this error is noticeable.
Now the location is fixed and it is almost perfect.

PR #3880, opened 2024-03-20
feat(events): Narrow mode layout (mobile portrait) for events view

Simpler1:
Reworked layout of search terms at the top when then screen width is less than 500px.  This comes into play when viewing in portrait mode on a mobile device.

image


connortechnology:
How does one navigate without navbars? They should become hamburgers.. I think you should tackle navbars in a separate PR.

Do we really need all the extra spans?

@IgorA100 thoughts?

IgorA100:
As far as I understand, in this PR only the filter in “#toolbar” was changed, and “#navbar-container” remained unchanged.
Just a screenshot, unfortunately, not complete.
I had an idea to make "#navbar-container" more adaptive, but time problems :(

> Do we really need all the extra spans?

It would probably be better to completely revise the filter block and implement it using bootstrap (raw, col-xx). https://getbootstrap.com/docs/4.0/layout/grid/
The result would be more readable code, and the design would be responsive to mobile devices with a screen width of <576px

Simpler1:
The additional spans allow for formatting control over the term label and operator as a single entity. (They will be needed even with the bootstrap implementation, but just as div instead of span.)

When viewing in portrait mode on mobile, I don't need to navigate.  I can easily rotate to landscape to get the navigation controls back.

I tried implementing the bootstrap (row/col-xx), but it doesn't look nearly as nice.  It tries to center each group of elements.  Also, you have to manually define the width ratio for each line.

The way that I implemented this, the label and operator take whatever space they need to stay on one line (left justified), then the remaining space is given to the input field.  It looks nice. 

I'm not sure what is meant by "Just a screenshot, unfortunately, not complete."

Simpler1:
I can separate these two commits into two different PRs if that's useful.  That was my original intent.  They are independent, but both make the mobile experience better.

Simpler1:
Primary PR moved to PR #3881.

PR #3881, opened 2024-03-20
feat(events): Narrow mode layout (mobile portrait)

Simpler1:
Separated from PR #3880 

Reworked layout of search terms at the top when then screen width is less than 500px. This comes into play when viewing in portrait mode on a mobile device.

image


IgorA100:
It certainly looks better than it does now.
I like.
I think it won't be difficult to add a hide/show button for this filter after the approval of my PR #3860

Simpler1:
Yea.  For reference, here's the current layout:

image

PR #3882, opened 2024-03-20
feat(narrow): Remove entire header when narrow

Simpler1:
When in viewing on a narrow screen (mobile in portrait mode), remove the navbar and a couple of the event buttons to make it display nicer.  The user can easily rotate to landscape mode to get them back.

### DIFF
--- a/web/skins/classic/css/base/skin.css
+++ b/web/skins/classic/css/base/skin.css
@@ -524,6 +524,19 @@ body.sticky #page {
   }
 }
 
+@media screen and (max-width:500px) {
+  .eventStats,
+  #navbar-one,
+  #navbar-two,
+  #exportBtn,
+  #downloadBtn,
+  #videoBtn,
+  #statsBtn,
+  #fullscreenBtn,
+  #zoomOutBtn {
+    display: none;
+  }
+}
 
 #header h2 {
     left: 0;


IgorA100:
I wouldn't completely hide the navigation bar, especially since its height (red stroke) takes up noticeably less space than the filter itself (blue stroke).
I think that adding a hide filter button will give us noticeably more space than hiding the navigation bar.
Rotating the device is, IMHO, not a good idea.

Simpler1:
My header is much taller than what you show.  

image

How about we collapse it down to a single hamburger menu button when narrow?

Also, I change the filter values frequently (i.e. setting Notes to Person and then removing) when on mobile.  I never change anything in the header when I'm checking on the events.

connortechnology:
Looks like the menu is correctly becoming a hamburger, good.  The second icon is supposed to show/hide the load/etc.  If we can get all that stuff into just one line of header, that'll be an ok compromise.

IgorA100:
> My header is much taller than what you show.

Simpler1:
It looks like it's supposed to be honoring the cookie for zmHeaderFlip such that once you collapse it, it should stay collapsed even on a refresh.  That doesn't seem to be happening.  It keeps coming back expanded and then the first click tries to expand it again before the second click collapses it.

IgorA100:
Yes, there is a similar problem, but I don't know if it's related to cookies.
I haven't researched this yet.

PR #3883, opened 2024-03-21
Fix: Elimination of jerks for slideToggle on console

IgorA100:
@connortechnology 

Fixes jerking when hiding the display of the filter panel on the console when the "stickey" mode is enabled.
Try this PR for yourself.
In theory it should work.

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -292,6 +292,7 @@ if ( currentView != 'none' && currentView != 'login' ) {
       const objIconButton = _this_.find("i");
       const obj = $j(_this_.attr('data-flip-сontrol-object'));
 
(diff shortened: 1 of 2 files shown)


connortechnology:
Honestly this didn't do much for me.  Changing the animation to fast is at least a little less annoying.

IgorA100:
Strange...
I'll look more carefully.

IgorA100:
@connortechnology 

I made more changes.
I checked and now it works fine for me.
Please check again for yourself!

IgorA100:
Now there is no need to worry about the element always having the display=block property.
Isaac, there's no way you couldn't tell the difference. Everything should be smoother now.
I set the mode to fast, as you liked.

PR #3888, opened 2024-03-23
Fix translate "Network Scan"

borozenetsvv:
Fix translate "Network Scan" in console view.


web/lang/en_gb.php
```diff
@@ -421,6 +421,7 @@
     'MustConfirmPassword'   => 'You must confirm the password',
     'MustSupplyPassword'    => 'You must supply a password',
     'MustSupplyUsername'    => 'You must supply a username',
+    'NetworkScan'           => 'Scan Network',
```
connortechnology, line 424: This is uneccessary.  The source text should be left as Scan Network. There is nothing wrong with spaces.

PR #3902, opened 2024-03-25
Fix: css, remove important, add fix style .chosen-choices

IgorA100:
In some places Important no longer needs to be used.
Fix style  for .chosen
Fix style for table border in dark css


IgorA100:
Fix text color in the Input field
Close #3906

PR #3908, opened 2024-03-26
Feat: Change scrollbar style for dark css

IgorA100:
I'm not sure about changing the style of the scrollbar, but I thought the dark theme made the scrollbar look too prominent.
I decided to change it a little.
Tested in Chrome and Firefox.

### DIFF
--- a/web/skins/classic/css/dark/skin.css
+++ b/web/skins/classic/css/dark/skin.css
@@ -21,6 +21,11 @@
  * Primary look and feel styles
  */
 
+:root {
+  --scrollbarBG: #CFD8DC;
+  --sliderBG: #95AFC0;
+}
+
 body {
     color: #dddddd;
     background-color: #222222;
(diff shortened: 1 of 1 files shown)


connortechnology:
@Simpler1 thoughts?

Simpler1:
I've tried in Chrome and Firefox on Linux Mint as well as Android.
I've used Shift-F5 to reload the page.  Still no change in the scroll bars.

Simpler1:
The only place where I see the scrollbar changes is in the list of tags (which is longer than the screen) on the event view.  The main window doesn't seem to be picking up the changes at all.

The narrower scroll bar takes more effort to select with the mouse (just because of the smaller area).

IgorA100:
> (just because of the smaller area).

The scrollbar can be made to a standard width, but in this case the bar will take up a lot of useful space. Although....
I need to install a new Linux virtual machine to test the changes.
I tested on Windows.

Simpler1:
I see the color change on the filters view, but not the size difference. 
image

I only see the size difference for tags.  
I don't see either the color or size change for the overall window.

IgorA100:
> I see the color change on the filters view, but not the size difference.

I haven't published these changes yet. Only I have them.
I showed a screenshot that I like. Do you like him or not?
I'll post the changes now.

> I don't see either the color or size change for the overall window.

I'll try to look at Linux tomorrow

IgorA100:
Ahhhh.....
I remembered.
My changes for the overall window so far only work when the WEB_NAVBAR_STICKY mode is enabled

IgorA100:
I've made changes code.
Now works in both modes.
Please check

Simpler1:
Looks like the additional changes are not in this pull request (the patch-41 branch).

IgorA100:
I checked everything again.
There are changes, and they are precisely in this PR.
Try this PR again (just replace your zoneminder\www\skins\classic\css\dark\skin.css file with my file)

Simpler1:
image

Simpler1:
Ok.  I see your changes now.  

Looks like Firefox does a nice hiding of the main scrollbar when it doesn't have focus.  
Chrome doesn't do that.  Can it be added to Chrome?

I agree it looks nicer than the default scroll bar.
I'm good with putting it in.

IgorA100:
> Chrome doesn't do that. Can it be added to Chrome?

I'm not sure if this is possible. But I'll take a look...

PR #3910, opened 2024-03-27
Feat: For <select> set .chosen & apply chosen

IgorA100:
If more 10 options in <selector>, search is turned on, if less 10 options, the search is not turned on. This has already been implemented and is working.
The `<select>` style must be uniform.

- Changed for the Options page.
- Changed for the Montage page.
- Changed for the Filter page.

On  Montage page more 10 options in select, which is why the search is enabled. If you don’t like this behavior, then the search can be excluded for some specific select elements, but leave the “chosen” style.

### DIFF
--- a/web/skins/classic/includes/functions.php
+++ b/web/skins/classic/includes/functions.php
@@ -1021,8 +1021,17 @@ function xhtmlFooter() {
   $skinJsFile = getSkinFile('js/skin.js');
 ?>
   <script src="<?php echo cache_bust($skinJsFile) ?>"></script>
-  <script nonce="<?php echo $cspNonce; ?>">$j('.chosen').chosen({disable_search_threshold: 10});</script>
-  </body>
+  <script nonce="<?php echo $cspNonce; ?>">
+    function applyChosen() {
+      const limit_search_threshold = 10;
+
+      $j('.chosen').chosen('destroy');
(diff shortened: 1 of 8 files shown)


IgorA100:
My work is finished.
I accept comments if something is not satisfactory.

web/skins/classic/includes/functions.php
```diff
@@ -1021,7 +1021,17 @@ function xhtmlFooter() {
   $skinJsFile = getSkinFile('js/skin.js');
 ?>
   <script src="<?php echo cache_bust($skinJsFile) ?>"></script>
-  <script nonce="<?php echo $cspNonce; ?>">$j('.chosen').chosen({disable_search_threshold: 10});</script>
+  <script nonce="<?php echo $cspNonce; ?>">
```
connortechnology: Instead of having this in php, let's stick the function in skin.js

IgorA100: No problem. I thought about it, but decided not to touch it for now.
I don't know what to do with the $cspNonce token yet, but I'll look at the code

connortechnology: remove it. Remove the entire script section.

IgorA100: I transferred script
The $cspNonce token now applies to entire main JS script.
I have no problems with tests.
I haven’t tested the multi-server mode, but in theory there shouldn’t be any problems.

Issue #3887, opened 2024-03-23
Add support for DBD-MariaDB after DBD-mysql dropped support of the MariaDB


VVD:
**Describe Your Environment**
- Version of ZoneMinder: 1.36.33
- How you installed ZoneMinder: build from ports
- Full name and version of OS: FreeBSD 13.3 amd64
- Browser name and version (if this is an issue with the web interface)

**If the issue concerns a camera**
- Make and Model
- frame rate
- resolution
- ZoneMinder Source Type:

**Describe the bug**
DBD-mysql dropped support of the MySQL 5.7 and MariaDB: https://github.com/perl5-dbi/DBD-mysql/issues/405#issuecomment-1818501464
If use MariaDB then can't use Zoneminder with recent DBD-mysql.
If install https://github.com/perl5-dbi/DBD-MariaDB instead of DBD-mysql then Zoneminder can't run.

**To Reproduce**
1. Replace DBD-mysql with DBD-MariaDB.
2. Start Zoneminder.
3. Zoneminder can't run.

**Expected behavior**
Run and work without errors with DBD-MariaDB.
Article "Migrating from DBD::mysql to DBD::MariaDB":
https://blogs.perl.org/users/grinnz/2023/12/migrating-from-dbdmysql-to-dbdmariadb.html

**Debug Logs**
```
Use of uninitialized value $DBI::errstr in concatenation (.) or string at /usr/local/lib/perl5/site_perl/ZoneMinder/Database.pm line 114.
Use of uninitialized value $DBI::errstr in concatenation (.) or string at /usr/local/lib/perl5/site_perl/ZoneMinder/Database.pm line 114.
03/23/24 03:47:34.518328 zmpkg[25250].ERR [ZoneMinder::Database:114] [Error reconnecting to db: errstr: error val:install_driver(mysql) failed: Can't locate DBD/mysql.pm in @INC (you may need to install the DBD::mysql module) (@INC contains: /usr/local/lib/perl5/site_perl/mach/5.36 /usr/local/lib/perl5/site_perl /usr/local/lib/perl5/5.36/mach /usr/local/lib/perl5/5.36) at (eval 10) line 3.
Perhaps the DBD::mysql perl module hasn't been fully installed,
or perhaps the capitalisation of 'mysql' isn't right.
Available drivers: DBM, ExampleP, File, Gofer, MariaDB, Mem, Proxy, Sponge.
[...]

PR #3914, opened 2024-03-28
Support mariadb

connortechnology:
Couple of small bits to enable use of DBD::MariaDB

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,14 +128,16 @@ set(ZM_CONTENTDIR "/var/lib/zoneminder" CACHE PATH
 set(ZM_FONTDIR "${CMAKE_INSTALL_FULL_DATADIR}/zoneminder/fonts" CACHE PATH
   "Location of the font files used for timestamping, default: <prefix>/${CMAKE_INSTALL_DATADIR}/zoneminder/fonts")
 
+set(ZM_DB_TYPE "mysql" CACHE STRING
+  "Type of database where the ZoneMinder database located, default: mysql")
 set(ZM_DB_HOST "localhost" CACHE STRING
-  "Hostname where ZoneMinder database located, default: localhost")
+  "Hostname where the ZoneMinder database located, default: localhost")
 set(ZM_DB_NAME "zm" CACHE STRING
-  "Name of ZoneMinder database, default: zm")
+  "Name of the ZoneMinder database, default: zm")
 set(ZM_DB_USER "zmuser" CACHE STRING
-  "Name of ZoneMinder database user, default: zmuser")
+  "Name of the ZoneMinder database user, default: zmuser")
 set(ZM_DB_PASS "zmpass" CACHE STRING
-  "Password of ZoneMinder database user, default: zmpass")
+  "Password of the ZoneMinder database user, default: zmpass")
 set(ZM_WEB_USER "" CACHE STRING
   "The user apache or the local web server runs on. Leave empty for automatic detection.
      If that fails, you can use this variable to force")
@@ -745,7 +747,6 @@ set(WEB_PREFIX "${ZM_WEBDIR}")
 set(CGI_PREFIX "${ZM_CGIDIR}")
 set(WEB_USER "${ZM_WEB_USER}")
 set(WEB_GROUP "${ZM_WEB_GROUP}")
-set(ZM_DB_TYPE "mysql")
 if(ZM_PERL_SEARCH_PATH)
   set(EXTRA_PERL_LIB "use lib '${ZM_PERL_SEARCH_PATH}'; # Include custom perl install path")
 else()
--- a/scripts/ZoneMinder/lib/ZoneMinder/Database.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Database.pm
@@ -112,7 +112,7 @@ sub zmDbConnect {
         .$socket . $sslOptions . ($options?join(';', '', map { $_.'='.$$options{$_} } keys %{$options} ) : '')
         , $ZoneMinder::Config::Config{ZM_DB_USER}
         , $ZoneMinder::Config::Config{ZM_DB_PASS}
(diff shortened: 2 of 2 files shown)


connortechnology:
Might fix #3887

gdt:
In pkgsrc, p5-DBD-mysql is at 5, which rejects MariaDB.   Version 4 was working fine with MariaDB 10.6.

I believe this PR's commits have not been merged to 1.36.   So I cherry-picked the three commits onto 1.36.37, and then used that diff as a patch in the package.   The package, when configured to use MariaDB, depends on p5-DBD-MariaDB, mariadb itself, and sets ZM_DB_TYPE as a cmake arg, so the default config is right.

Starting up, I get from the web:
```
ZoneMinder Error

Unable to connect to ZM db using dsn MariaDB:host=localhost;dbname=zm;charset=utf8mb4

could not find driver

ZoneMinder will retry connection in 10 seconds.
```

If I change ZM_DB_TYPE back to mysql, it starts up, but I also have p5-DBD-mysql  4 installed, the last version that accepts MariaDB as a mysql flavor.

My guess is that something in PHP is also keying off ZB_DB_TYPE and looking for a mariadb pdo, but I can't find where that might be happening.

gdt:
Also, with just these three commits cherry-picked, on starting zoneminder (the perl stuff, not php via apache), I see
```
DBD::MariaDB::db STORE failed: Unknown attribute mysql_auto_reconnect at /usr/pkg/lib/perl5/vendor_perl/5.42.0/ZoneMinder/Database.pm line 119.
01/ 4/26 17:52:49.556704 zmpkg[12852].FAT [ZoneMinder::Database:121] [Can't set mysql_auto_reconnect on in database connection]
```
so it seems there are more.

Do people think using p5-DBD-MariaDB works with 1.36.37?

update:

By changing the php database connect:

```
function dbConnect() {
  global $dbConn;

  $dsn = ZM_DB_TYPE;
  if ( $dsn == 'MariaDB' ) {
    $dsn = 'mysql';
  }
```

things are working (with this PR cherry-picked, and ZM_DB_TYPE=MariaDB).  The code looks similar master vs 1.36.37.   Does this work for anyone else?   I find no evidence that there is php-pdo_MariaDB and that I'm missing it.

connortechnology:
THe auto-reconnect thing is kindof a biggy.  It requires us to handle a connection failure throughout the codebase.  Which we did in 1.37.*.  

So I can cherry-pick some more of this stuff... but I think 1.36 is going to be less and less viable as time goes on.  

See e163509d57d5cf0909d4c9c02cd496a53ed92d0a

As to the php side of things... looks like there isn't an equivalent php mariadb driver?! So you are going to have to add a line to pass mysql instead of mariadb there

gdt:
My belief for php-pdo is that you use the mysql driver for mariadb and it works fine.   I did add a mapping and that got me most of the way there.  I also commented out the autoconnect, not the same as your commit but functionally equivalent for me wiht mariadb, just don't do it.  I get it that not autoconnecting is likely bad, although with localhost likely it's sort of ok.

As a packager (and somewhat as a user), I don't really want to run 1.37 which AIUI is labeled unstable/development.   So It seems like I should be labeling the package as working with mysql only, which is really unfortunate as it seems mariadb is the standard approach for mysql now.

All of this leads to "1.38 is overdue", but I know you already knew that and saying there's one more reason doesn't help make it so :-)

gdt:
If you are up for cherry-picking enough to make a 1.36.x that is just ok with mariadb out of the box that would be awesome.

connortechnology:
Well I'll pull in a few more of these for 1.36.38.  

Yeah the idea was to get 1.38 out by end of year...  but I didn't get as much done over the holidays as I wanted.  Hopefully by end of January.

connortechnology:
If you could test our commits to release-1.36 branch, that would be a big help.

gdt:
Thanks - I will test that, either building direct or via patch from the 1.36.37..branch diff.    Your changes aren't quite what I tested - not a complaint, I will try your bits.

gdt:
I have dropped the cherry-picks from the PR, and updated (not yet published) the pkgsrc package to the tip of release-1.36, specifically
```
commit c02f5a8e989ae8cac6b78ceeedabcd023aac880e (HEAD -> release-1.36, origin/release-1.36)
Author: Isaac Connor <isaac@zoneminder.com>
Date:   Sun Jan 4 18:45:18 2026 -0500

    If db_type == MariaDB then use the mysql driver
```

You didn't add this commit:
```
commit 649bceb24083639483d828e1288b3ebef8fc14ba
Author: Isaac Connor <isaac@zoneminder.com>
Date:   Thu Mar 28 09:55:43 2024 -0400

    Improve handling of ZM_DB_TYPE
```

and the only real harm, I think, is that the example config file says mysql, because `ZM_DB_TYPE` is overridden.  (In the package, I addded CMAKE_CONFIGURE_ARGS with the chosen type so the default config aligns with the modules/db that are provided, mysql vs mariadb.)

After manually changing ZM_DB_TYPE in zm.conf, zoneminder starts and appears to be running ok.

I suggest cherry-picking 649bceb too, and I'll update, clear out my configs so I get the sample one (real config is in `conf.d/zmcustom.conf`) and retest.

connortechnology:
I have added 649bceb24083639483d828e1288b3ebef8fc14ba

gdt:
I've locally updated pkgsrc to that commit and am successfully starting up with mariadb (10.6.23) with the installed default config and p5-DBD-MariaDB.  It's easy for pkgsrc to be on a commit along the branch with a faux version number of 1.36.37.7, so it doesn't really matter to me if you snap a micro or you don't, but it would help anyone using mariadb whose p5-DBD-mysql becomes up to date (5).

Thanks very much for the discussion and cherry-picks.   (I have an outstanding issue about printf specifiers, and a few more minor pending patches that I avoided filing until I was caught up and understood better, but those aren't about mariadb at all.)

connortechnology:
Let me know anything you've got and I'll get them in.
People on debian/ubuntu can get point builds from zmrepo or ubuntu ppa, so they are good.  I'm fine to release this all relatively soon to keep the changes small.

PR #3920, opened 2024-03-28
Feat: Ability to download videos on Event view page with a multiport configuration

IgorA100:
If the port is different from 80, the browser will start watching the video instead of downloading.

Close https://forums.zoneminder.com/viewtopic.php?t=33167

### DIFF
--- a/web/includes/Event.php
+++ b/web/includes/Event.php
@@ -257,15 +257,19 @@ class Event extends ZM_Object {
 
   public function getStreamSrc( $args=array(), $querySep='&' ) {
 
-    $streamSrc = '';
     $Server = $this->Server();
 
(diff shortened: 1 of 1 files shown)


connortechnology:
This does nothing for me here.  Yes it removes the port but the behaviour is the same.  Could be because mine is a multi-server setup and the url points to another host entirely.

IgorA100:
> Could be because mine is a multi-server setup and the url points to another host entirely.

Probably yes, I have no way to check the work on a multi-server system. But if there is only one server and there are many servers, then without this change the browser tries to play the video instead of downloading it, which naturally leads to an error.
Since this change did not break your system, I think this PR can be approved.

IgorA100:
@connortechnology 
What confuses you about this PR?

PR #3924, opened 2024-03-28
Fix style for classic skin

IgorA100:
I removed the width limitation, it looks better, but overall the skin is in very poor condition.
I'm not sure anyone uses it.


connortechnology:
I'd be in favour of removing it. Could move it into it's own repo as an add-on if people want to continue using it.   I only feel the need to maintain light and dark.

IgorA100:
I don't know if it should be removed or if there is something unique about it.
I don't really like him yet...

IgorA100:
I see :)
In this case, you probably need to remove it.

PR #3925, opened 2024-03-28
Fix: We do not set chosen container width to 100% for simple_widget() elements

IgorA100:
For now this spoils the design in some places, for example on the Montage review page.


IgorA100:
We do not currently apply this PR in connection with PR #3930.
After applying PR #3930, we'll come back again.

Simpler1:
This PR can be removed.  It is no longer relevant.

PR #3930, opened 2024-03-29
feat(): Term fields spacing for both console and events

Simpler1:
Modified spacing for term fields in both the console and events views:

Console from:
image

to:
image

Events from:
image

to:
image

### DIFF
--- a/web/skins/classic/css/base/skin.css
+++ b/web/skins/classic/css/base/skin.css
@@ -156,7 +156,7 @@ button.btn {
   line-height: 1;
   font-size: 18px;
 }
-input,textarea,select,button,.btn-primary {
+input, textarea, select, button, .btn-primary {
     border: 1px #ccc solid;
     padding: 5px;
     border-radius: 1px;
@@ -167,7 +167,6 @@ input,textarea,select,button,.btn-primary {
     background-color: #f8f8f8;
     text-align: left;
     border-radius:4px;
-    margin: 1px 0;
 }
 
 input.noborder {
@@ -743,7 +742,7 @@ ul.nav.nav-pills.flex-column {
 
 .chosen-container {
 text-align: left;
-/*min-width: 11em;*/ /* Why ???*/
+min-width: 11em; /* Makes a nice uniform display */
 }
 
 .chosen-container-active .chosen-choices {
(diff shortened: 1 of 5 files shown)


web/skins/classic/css/base/skin.css
```diff
@@ -743,7 +742,7 @@ ul.nav.nav-pills.flex-column {
 
 .chosen-container {
 text-align: left;
-/*min-width: 11em;*/ /* Why ???*/
+min-width: 11em; /* Makes a nice uniform display */
```
IgorA100: I would not apply such strict restrictions globally, because... This is not necessary everywhere.
For example, on the Filter page, a block with brackets will not be beautiful.



And in some other places it will also be bad because of this.
Perhaps it’s worth adding another class with a width constraint and applying it as needed?

IgorA100: I think you could write it like this:
```
#fbpanel .chosen-container, #toolbar .chosen-container {
  min-width: 11em;
}
```

IgorA100: @Simpler1 @connortechnology 
What are your thoughts on my proposal?

Simpler1: The fields that you show (brackets, and, or) do not have the class ".chosen-container", so this does not apply to them.

Have you made other changes that are conflicting?

If this min-width did apply to the brackets, and, and or fields, then I agree that the more restrictive property is more appropriate.

Simpler1: @IgorA100 I made the change you suggested.

I also made all of the other views look correct and consistent for both normal and narrow (<500px wide).
Views: Console, Events, Montage, MontageReview, AuditEventsReport.

IgorA100: > The fields that you show (brackets, and, or) do not have the class ".chosen-container"

You're wrong. PR #3910 applies .chosen to almost all `<select>`. And if it is not used somewhere else, then this must be done so that the style is uniform.

> Have you made other changes that are conflicting?

I quickly looked at the code, but did not use it because... I'm currently working on other things. I like your changes, but I've been doing a lot of PR the last few days. Try testing your changes on the latest version of Master Branch.

Simpler1: All my changes are on to of the latest master branch.  Try it when you get time.

IgorA100: @Simpler1 
I checked your PR, good job, but I'm probably being too picky...
It is necessary to add vertical padding for the buttons to make it more beautiful.

Simpler1: I agree.  I didn't touch that code, but will followup with a separate PR.

Simpler1:
@connortechnology This one is ready for merge.

PR #3934, opened 2024-04-01
Feat: Mouse moving an enlarged video frame during live viewing

IgorA100:
Now it is possible to move an enlarged video frame by moving the mouse and not just by clicking while holding down "Shift"
This function works by simultaneously pressing “Ctrl” + “Shift”. Those. you need to hold down “Ctrl” + “Shift”, then by pressing the left mouse button “capture” the video frame and moving the mouse will move the video frame.
I found this very convenient, especially when it is necessary to move a greatly enlarged video frame a short distance.
All the previous functionality when holding “Ctrl” (decrease) or “Shift” (positioning) or simply “click” (increase) remained unchanged.

### DIFF
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -16,6 +16,14 @@ var classMonitorW_SB_L = 'col-sm-9'; /* id="wrapperMonitor" ONLY WITH LEFT */
 var classMonitorW_SB_R = 'col-sm-10'; /* id="wrapperMonitor" ONLY WITH RIGHT */
 var classMonitorWO_SB = 'col-sm-12'; /* id="wrapperMonitor" MAXIMUM width */
 
+var PrevCoordinatFrame = {x: null, y: null};
+var coordinateMouse = {
+  start_x: null, start_y: null,
+  shiftMouse_x: null, shiftMouse_y: null,
+  shiftMouseForTrigger_x: null, shiftMouseForTrigger_y: null
+};
+var leftBtnStatus = {Down: false, UpAfterDown: false};
+
 /*
 This is the format of the json object sent by bootstrap-table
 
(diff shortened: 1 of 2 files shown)


IgorA100:
If this PR is approved, then perhaps it is worth reconsidering the algorithm for enlarging the image by clicking the mouse?
Make it more intuitive?
Those. to enlarge exactly the visual area of the image on which the mouse was clicked, and not the virtual area that is comparable to the click location at a scale of 1:1?
I don't know if you'll understand my point? :

IgorA100:
I now have another thought about changing the behavior algorithm.
I can make an algorithm like this:
If the left mouse button is pressed and released without displacement (displacement less than XX pixels (for example 30) - increase the video frame.
If the left mouse button is pressed and the mouse moves by more than XX pixels (for example, 30), move the video frame.
How do you like this idea?

web/skins/classic/views/js/watch.js
```diff
@@ -593,27 +601,105 @@ function fetchImage(streamImage) {
...
+    leftBtnStatus.Down = true;
+  } else if (event.buttons == 0 && leftBtnStatus.Down == true) { //Up left button after pressed
+    leftBtnStatus.Down = false;
+    leftBtnStatus.UpAfterDown = true;
+  }
+
+  if ((leftBtnStatus.UpAfterDown) || //The left button was raised or the cursor was moved more than 30 pixels relative to the actual size of the image
+    ((coordinateMouse.shiftMouseForTrigger_x > 30) && leftBtnStatus.Down) ||
+    ((coordinateMouse.shiftMouseForTrigger_y > 30) && leftBtnStatus.Down)) {
+    //We perform frame shift
```
IgorA100, line 699: The video frame is shifted at intervals of 30 pixels.
This seemed to me the best option, because... if the value is reduced, the load increases and the shift may not keep up with the position of the mouse; if the value is increased, movement is less intuitive.

IgorA100:
@connortechnology 
If you like this PR, I will do the same for viewing the recorded events.

PR #3938, opened 2024-04-01
Feat: Added a button to hide/show the timeline panel on the "Montage Review" page

IgorA100:
This PR includes two changes:
**1.** Added the ability to execute a function BEFORE or AFTER the hide/show event of a DOM element block. In addition to the implementation in PR #3860

**Usage:** 
Add an attribute to the control button: `data-flip-control-run-before-func="NameFuntion"` or `data-flip-control-run-after-func="NameFuntion"`

**2.** Added a button to hide/show the timeline panel on the "Montage Review" page

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -365,8 +365,9 @@ if ( currentView != 'none' && currentView != 'login' ) {
     // Manage the web console filter bar minimize chevron
     $j("#mfbflip").click(function() {
       $j("#mfbpanel").slideToggle("slow", function() {
-        //changeScale(); //This function disappeared somewhere... It needs to be restored. Function call appeared Aug 10, 2023 https://github.com/ZoneMinder/zoneminder/commit/b8982dce53137d86d33d2d6d79104d4a25673ef7
-        redrawScreen(); //Temporarily use changeScale() instead. You can try using refreshWindow()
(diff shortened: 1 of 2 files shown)


web/skins/classic/js/skin.js
```diff
@@ -346,7 +361,8 @@ if ( currentView != 'none' && currentView != 'login' ) {
     // Manage the web console filter bar minimize chevron
     $j("#mfbflip").click(function() {
       $j("#mfbpanel").slideToggle("slow", function() {
-        changeScale();
+        //changeScale(); //This function disappeared somewhere... It needs to be restored. Function call appeared Aug 10, 2023 https://github.com/ZoneMinder/zoneminder/commit/b8982dce53137d86d33d2d6d79104d4a25673ef7
+        redrawScreen(); //Temporarily use changeScale() instead. You can try using refreshWindow()
```
IgorA100: This is a temporary solution.
@connortechnology , maybe you know what happened with the changeScale() function?

connortechnology: I think it exists in an unfinished branch of mine.  It exists in watch.js and montage.js.  So we should test for existence of function and use what exists.

IgorA100: I added a function exists check

PR #3939, opened 2024-04-01
feat(logs): Consistent spacing for terms and buttons

Simpler1:


web/skins/classic/views/log.php
```diff
@@ -39,46 +39,58 @@
...
       </span>
-      <span class="EndDateTimeFilter">
+      <span class="term EndDateTimeFilter">
         <label><?php echo translate('End Date/Time') ?></label>
-        <input type="text" name="filterEndDateTime" id="filterEndDateTime" value=""/>
+        <span class="term-value-wrapper">
+          <input type="text" name="filterEndDateTime" id="filterEndDateTime" value=""/>
+        </span>
       </span>
+      </div>
```
IgorA100, line 93: Specify which DIV is closing.
If `<div class="controlHeader">` is closed, then you can write it like this:
`</div><!--.controlHeader-->`
If you don’t specify which tag is closing, then it’s difficult to edit the code later.

PR #3949, opened 2024-04-05
Feat: Added the ability to zoom in/out image of each monitor separately on the "Montage" page

IgorA100:
I haven’t checked it 100% yet, but it seems to work and doesn’t break anything...
Please test it.


connortechnology:
That's cool... the actual resizing doesn't work very well here though and maybe we need to put a border on them to make it more noticeable where to click to do the resize....

IgorA100:
> the actual resizing doesn't work very well here though and maybe we need to put a border on them to make it more noticeable where to click to do the resize....

Yes, these problems exist and require solutions, but these problems are long-standing, and not because of my PR.
I'll try to fix them, but a little later.

IgorA100:
@connortechnology 
Try new code.
Not perfect, but better.
For ideal positioning, need to rewrite a lot of code and most likely need to use absolute positioning of monitors in Layout and abandon Flex. All monitor sizes must be managed by JS.

IgorA100:
@connortechnology
I made some more changes to the code.
Try it.

web/js/MonitorStream.js
```diff
@@ -139,7 +139,7 @@ function MonitorStream(monitorData) {
     if (width && (width != '0px') && (img.style.width.search('%') == -1)) {
       monitor_frame.css('width', parseInt(width));
     }
-    if (img.style.width) img.style.width = 'auto';
+    if (img.style.width) img.style.width = '100%';
```
connortechnology, line 142: I might be wrong but I think the reason for auto instead of 100% was to prevent stretching the image. The decision was that we would fix the heights and let the browser leave padding around the width to maintain aspect ratio.

IgorA100: With Height, everything is now confused and 'changeHeight()' did not work at all before.
Now it works, but only changes the height of the container in which the frame is located, but not the height of the frame itself. What's probably wrong? Yes?
I think it would be correct to set the height of the frame and calculate the height of the container.

My change from width = 'auto' to '100%' allows the Frame to scale correctly in edit mode when changing "Width" in the setting. Because Changing the "Width" setting changes the width of the container, not the frame!

I don't know how clearly I explained...

I can't fix all the problems right away.

**I see this algorithm for changing Height and Width:**
1. Setting the height and width is set for Frame. The width and height of the container are calculated by JS.
2. If both Height and Width are set, then we display Frames without recalculation, but this is the worst setting.
3. If the Height is set, then we calculate the Width from the proportions and then calculate the size of the container in JS.
4. If the Width is set, then we calculate the Height from the proportions and then calculate the size of the container in JS.

PR #3956, opened 2024-04-07
To configure detection zones, allow float % input

IgorA100:


connortechnology:
But step=any allowed float input.  You have limited it to .1 increments so now we can't do two decimal places..

IgorA100:
Eeeehhh.....my bad English again....

step - does not prohibit or allow anything and by default for numbers step=any=1
Those. if step=any is written in the tag and enter the value 1.24 and press the up arrow on the input field, then the increment will be +1, i.e. the next value will become 2.24.
After my change, the increment will become = 0.1, not 1, but when you press the up arrow, the value will become = 1.3, then 1.4, etc., but you can write 1.422 by hand

Although, it’s probably not worth making my changes.

PR #3957, opened 2024-04-08
Fix: If Select has no Options, don't try to select 1 Option in Filter page

IgorA100:
Because selecting an option that does not exist results in an error.
For example, when cloning a Tags filter string when there are no tags.

### DIFF
--- a/web/skins/classic/views/js/filter.js
+++ b/web/skins/classic/views/js/filter.js
@@ -383,7 +383,8 @@ function addTerm( element ) {
   row.find('select').chosen('destroy');
   var newRow = row.clone().insertAfter(row);
   newRow.find('select').each( function() { //reset new row to default
-    this[0].selected = 'selected';
+		if ($j(this).find('option').length > 0 )
+      this[0].selected = 'selected';
   });
   newRow.find('input[type="text"]').val('');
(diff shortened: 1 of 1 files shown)


web/skins/classic/views/js/filter.js
```diff
@@ -383,7 +383,8 @@ function addTerm( element ) {
   row.find('select').chosen('destroy');
   var newRow = row.clone().insertAfter(row);
   newRow.find('select').each( function() { //reset new row to default
```
IgorA100: But I’m not at all sure about the need to reset new row to default, because we are cloning a string, which means we want the new string to be similar to the one we are cloning.
@connortechnology ???

connortechnology: You are suggesting that we NOT reset the values to defaults... I can see the argument either way.  I think I am leaning towards agreeing

IgorA100: > You are suggesting that we NOT reset the values to defaults...

Yes. This seems more correct to me.
If you agree, I will temporarily comment out the condition, but will not completely delete it.

connortechnology: I agree

IgorA100: Made

PR #3969, opened 2024-04-17
Global change grid layout on Montage page

IgorA100:
Added support for GridStack (https://github.com/gridstack/gridstack.js) & PanZoom (https://github.com/timmywil/panzoom) on Montage page
Innovations:
- Fully adaptive grid (scaling, moving)
- The image of each monitor separately can be Scale & Pan
- Zoom and other buttons appear automatically when you hover the mouse over a specific monitor.
- Ability to change the location of monitor statistics.

Note:
- No compatibility with saved Layouts in previous versions of ZM. When you try to load such a Layout, a warning is displayed.
- Supports only the following preset Layouts: "1 Wide", "2 Wide", "3 Wide", "4 Wide", "6 Wide", "8 Wide", "12 Wide" and "16 Wide". Custom Layouts are limited only by your imagination.
- Removed the ability to set Width, Height and Scale through presets in "Select".


IgorA100:
@connortechnology 
I need your help.
Will you be able to make changes to the database in the "MontageLayouts" table?
You need to remove the pre-installed Layouts "5 Wide", "7 Wide", "9 Wide" and "10 Wide" 
Add "1 Wide" (not sure what is needed, it's up to you), "12 Wide", "16 Wide"

IgorA100:
@connortechnology
Please take the time to check my PR.
You should like it.

connortechnology:
Amazing work!   Two things:
When the status is inner, the font needs to change colour to something lighter. 
Because you removed the fixed height option, we are stuck with a layout mess if you have cameras with different aspect ratios.  We need the height option to come back, or be replaced with a forced aspect ratio option.

IgorA100:
> Amazing work!

I'm glad you liked it.

> hange colour to something lighter

Unfortunately, I did not test it on a light theme.
Of course I'll fix it.

> or be replaced by a forced aspect ratio parameter.

I've thought about this, but I don't have a ready-made solution in my head yet. But I will think...

PR #3970, opened 2024-04-19
Write jpegs and stream mjpeg using avcodec

webgeek1234:


webgeek1234:
The works for jpegs being written to disk, like saving events. It does not cover jpegs being written to memory, like used to prime streaming. It does not convert metadata that was written by libjpeg. If this is needed, equivalent avformat code needs written.

PR #3971, opened 2024-04-21
Adds floorplan functionality to ZM (based on leaflet and leaflet.draw)

ovargasp:
requires to manually upload floorplan as a jpg to the folder "/usr/share/zoneminder/www/graphics" with the name "floorplan.jpg"

### DIFF
--- a/web/js/panzoom.js
+++ b/web/js/panzoom.js
@@ -251,6 +251,14 @@ var zmPanZoom = {
         }
       }
     }
+
+    if (this.panZoom[id].getScale().toFixed(1) > 1) {
+      this.panZoom[id].setOptions({handleStartEvent: (event) => {
+        event.preventDefault();
+      }});
+    } else {
+      this.panZoom[id].setOptions({handleStartEvent: (event) => {}});
+    }
   },
 
   click: function(id) {


web/fp_delete.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+$mid = $_GET["mid"];
+/*if ($_GET["ismon"] = "true")
+{ $ismon = true}else{$ismon = false};*/
+//$ismon = $_GET["ismon"];
+//$json = $_GET["json"];
+
+
+$file_handle = fopen('/etc/zm.conf', 'r');
```
connortechnology: Why are we doing all this, instead of using the built in ZM functions that automatically do this? So that it can stand alone outside of a ZM install? Seems weird.  If /etc/zm/zm.conf exists, then you should be able to include the database and config.php includes.

ovargasp: I started it as a stand alone function, so it can be used as a simplified user interface with some branding (colors, fonts, etc.) based in the floorplan and just for viewing purposes. Because of that, I tried to use the just the config file to get database parameters and then make it as  autonomous as possible, we can create a separate config file, but is up to what we want to do with it...

PR #3982, opened 2024-05-03
FIx: Montage page, remove stream.src = ''

IgorA100:
I don’t understand why it was necessary to reset the attribute if it is immediately redefined “stream.src = src” ???


connortechnology:
Because some browsers don't  accept the new value until you have cleared it.

PR #3986, opened 2024-05-08
Feat: Used PanZoom on Watch page 

IgorA100:
Similar to #3969
Approve only after #3969


IgorA100:
I had to change the "setScale" function in MonitorStream.js. This resulted in changes to montage.js
I did not do another PR, but decided to combine them in this PR.

IgorA100:
@connortechnology 
Try to find time and check/approve this PR.
I want to create more PR, but I'm afraid of getting confused...

connortechnology:
Looks like we can remove the zoom out button below the stream.  Also, the old code had a shift click option to zoom out all the way instead of having to click multiple times.

connortechnology:
THe new zoom is better of course, but I liked th eold behaviour where clicking on the image would direct where it zoomed into.  ALso ctrl-click anywhere on the image means that no mouse traversal is necessary.  The zoom in/out buttons are actually quite small.

IgorA100:
> but I liked th eold behaviour where clicking on the image would direct where it zoomed into.

I'll think of something.

> The zoom in/out buttons are actually quite small.

Oh, on the contrary, they seemed big to me :)
How big do you want them to be?

PR #3988, opened 2024-05-08
Fix: CSS file loading order (functions.php)

IgorA100:
Load third-party files first

### DIFF
--- a/web/skins/classic/includes/functions.php
+++ b/web/skins/classic/includes/functions.php
@@ -114,6 +114,10 @@ echo output_link_if_exists(array(
   'js/bootstrap-table-1.22.3/bootstrap-table.min.css',
   'js/bootstrap-table-1.22.3/extensions/page-jump-to/bootstrap-table-page-jump-to.min.css',
 ), true);
+
+if ( $basename == 'montage' ) {
+  echo output_link_if_exists(array('/assets/gridstack/dist/gridstack.css', '/assets/gridstack/dist/gridstack-extra.css'));
+}
 ?>
   <link rel="stylesheet" href="skins/classic/js/jquery-ui-1.13.2/jquery-ui.theme.min.css" type="text/css"/>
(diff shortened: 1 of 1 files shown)


connortechnology:
We used to do a lot of this if view = stuff in the past.  I went instead with just sticking it into the view file itself.. Not sure if it matters in terms of performance but it just seemed cleaner.

IgorA100:
I did this because:
- CSS should always be loaded in the <head>, not the <body>, because... The CSS must be loaded first so that there is no jerking on the page. Yes, there may be exceptions when it is difficult to do otherwise.
- When CSS is loaded from one place you can always control the order in which the files are loaded. Third-party files must be downloaded first, your own files last, because... Sometimes you have to override values in your own files. In this case, overriding styles can be done without using "important".
- More correctly, all CSS and JS files need to be compressed before sending to the browser, which means they should be loaded from one place. I'm not sure ZM does this.
- I’m more used to having all CSS and JS files loaded from one place.

IgorA100:
I think it would be more correct if all CSS and JS files are loaded in one place, but not in the “function.php” file, but in a separate one, for example “load_css_js.php”

connortechnology:
css should be first, js should be last, in general these rules only need apply to the bulk of the core UI parts.  Third party libraries if only used on view don't matter so much.  They tend to all end up being cached anyways.  

I would not accept a PR moving loading of CSS/js into another file.  They are fine in function.php.  

The biggest optimisation we can do right now is to deal with the fonts.  These are large and largely unused.

IgorA100:
> The biggest optimisation we can do right now is to deal with the fonts. These are large and largely unused.

I remember about this.
I haven't come up with a good idea yet...
I'm currently working on changing the Ratio

connortechnology:
https://stackoverflow.com/questions/64614572/creating-a-material-icons-subset

Lots of people do it.  I started to get into it one day but ran out of time.

PR #3991, opened 2024-05-13
Enabled support for old ZoomPan algorithm, Added ZoomPan (In & Out) options  & other (watch.js)

IgorA100:
- Added variables: "panZoomMaxScale", "panZoomStep"
- The "setScale" function is no longer used
- Changed the "changeScale" Function (Fix: problem with image scaling when changing browser width)
- Enabled support for the old ZoomPan algorithm (Added a checkbox for switching)
- Added reset zoom option (Ctrl+click on the Zoom Out button)
- Added an option to double the Zoom In step (Ctrl+click on the Zoom In button)
- Fix: border IMG for alarm & alert status

Added:
- "Shift" button + mouse wheel - Zoom In / Zoom Out
- "Shift" button + mouse click anywhere in the image = applies Zoom Out to the clicked location.
- "Ctrl" button + mouse click anywhere in the image = reset Zoom to 1.
- Depending on the available actions, the appearance of the mouse cursor changes.

### DIFF
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -28,6 +28,8 @@ var panZoomEnabled = true; //Add it to settings in the future
 var panZoomMaxScale = 10;
 var panZoomStep = 0.3;
 var panZoom = [];
+var shifted;
+var ctrled;
 
 /*
 This is the format of the json object sent by bootstrap-table
@@ -631,6 +633,19 @@ function handleClick(event) {
         window.location.assign(url);
       }
     }
+
+    if (obj.getAttribute('id').indexOf("liveStream") >= 0) {
+      if (ctrled && shifted) {
+        return;
+      } else if (ctrled) {
+        panZoom[monitorId].zoom(1, {animate: true});
+      } else if (shifted) {
+        const scale = panZoom[monitorId].getScale() * Math.exp(panZoomStep);
+        const point = {clientX: event.clientX, clientY: event.clientY};
+        panZoom[monitorId].zoomToPoint(scale, point, {focal: {x: event.clientX, y: event.clientY}});
+      }
+      monitorsSetScale(monitorId);
+    }
   } else {
     // +++ Old ZoomPan algorithm.
     if (!(event.ctrlKey && (event.shift || event.shiftKey))) {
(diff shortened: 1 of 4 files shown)


IgorA100:
@connortechnology 
I added a commit that you should like with the new ZoomPan algorithm. It looks like the old one, but I found it even more convenient!!!
Please, check it.

web/skins/classic/views/js/watch.js
```diff
@@ -947,10 +975,61 @@ function streamStart() {
...
+    document.getElementById('zoomOutBtn').classList.remove("hidden");
+  } else {
+    document.getElementById('zoomOutBtn').classList.add("hidden");
+  }
+  $j("#use-old-zoom-pan").click(function() {
+    useOldZoomPan = this.checked;
+    setCookie('zmUseOldZoomPan', this.checked);
+    location.reload();
+  });
+  document.getElementById('use-old-zoom-pan').checked = useOldZoomPan;
```
connortechnology, line 1024: I think you are missing a commit because use-old-zoom-pan doesn't exist for me and hence this line fails.
I suggest something like:
const use_old_zoom_pan = $j("#use-old-zoom-pan");
  use_old_zoom_pan.click(function() {
    useOldZoomPan = this.checked;
    setCookie('zmUseOldZoomPan', this.checked);
    location.reload();
  });
  use_old_zoom_pan.prop('checked', useOldZoomPan);

connortechnology:
There is a problem here and with the previous patches.  Your removal of the ability to scale the image larger.  We have users that need to scale the image larger than the browser window and scroll to where they want to view.  You have removed this functionality.

IgorA100:
> Your removal of the ability to scale the image larger.
> ...
> We have users that need to scale the image larger than the browser window and scroll to where they want to view.

At the moment, the image is enlarged up to 20 times, regardless of the width of the browser window. Panning occurs when you click on the image. This way you can view any location in the enlarged image.

web/skins/classic/views/js/watch.js
IgorA100: Sorry, you had an error because I forgot to commit the “watch.php” file to which I added a check box
I've made changes, try again.

IgorA100:
You can always turn on full screen mode for easy viewing of large images.
But, if, for example, the “RTSP2Web Live Stream” setting is enabled, then there is no button to switch to full screen mode. I don't know the reason why the button is missing.

IgorA100:
Yes, now the old "Scale=auto" mode is actually used, but with the possibility of flexible zooming and panning.
I can add an option to make the image the same width as the browser window without enabling full screen mode. But is this necessary?
When an image is larger than the width of the browser window and has a scroll bar, it looks very unattractive and makes no sense with current scaling and panning.

IgorA100:
I added "Scale" setting #3995

PR #3992, opened 2024-05-14
Feat: Managing Ratio on Mintage page (montage.js)

IgorA100:
- Added a list of possible Ratios (auto, real, 1:1, 5:4, 4:3, etc.) for selection.
- Added the ability to specify Ratio both for all displayed monitors and for a specific monitor.
* Option "auto" - the average Ratio is calculated for all displayed monitors, then the closest value is selected from the preset list and applied to all monitors.
* Option "real" - real Ratio for each monitor without recalculation.
- Added "Monitor status position = Show on hover"
- Added a button to open (appears when you hover over the monitor image) of any monitor in full screen mode

### DIFF
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -16,6 +16,34 @@ var changedMonitors = []; //Monitor IDs that were changed in the DOM
 var panZoomEnabled = true; //Add it to settings in the future
 var panZoom = [];
 
+const presetRatio = new Map([
+  ['auto', ''],
+  ['real', ''],
+  ['1:1', '1.000'],
+  ['5:4', '1.250'],
+  ['4:3', '1.333'],
+  ['43:32', '1.344'],
+  ['11:8', '1.375'],
+  ['3:2', '1.500'],
+  ['25:16', '1.563'],
+  ['16:10', '1.600'],
+  ['5:3', '1.667'],
+  ['16:9', '1.778'],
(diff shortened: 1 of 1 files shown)


IgorA100:
@connortechnology 
I hope I did everything well :)

IgorA100:
@connortechnology
I'm interested to know your opinion when displaying monitors with different Ratios and different orientations.
Did you like it?
It seems to me that it was done well and I don’t know what else could be improved related to Ratio

connortechnology:
I think it is very good.  Works well here. I'm not sure of the value of all the different ratio options in the drop down.  Seems to me we should only offer the ratios that are present in the video streams, because the only reason to do this is to make them all the same so that they layout nicely.

IgorA100:
> I'm not sure of the value of all the different ratio options in the drop down.

Maybe yes. In most cases, "Auto" (I really hope that my algorithm will be optimal) or "Real" will be used.
Such a large list of Ratios is necessary to select the optimal average value.
I did not reduce the visible list in Select so that any, the rarest monitors would be displayed most correctly. Monitors with a large spread of Ratio should also be displayed more or less correctly

> because the only reason to do this is to make them all the same so that they layout nicely.

I don’t agree, because... for example, there is a 16:9 and 4:3 monitor, the average Ratio value (from the standard ones) will be 25:16. And if we leave only 16:9 and 4:3 in the list, then 25:16 cannot be selected. This is the beauty of the "Auto" mode

PR #3996, opened 2024-05-14
Moved the "Fullscreen" button to another block (watch.php)

IgorA100:
Moved from block `id="dvrControls"` to block `id="extButton"` Now the button does not disappear if there is no block `dvrControls`


connortechnology:
I'm ok with this, but I think a better aproach would be to ensure that dvr controls is always present.  We do this with event viewing and video.js linking the buttons to video.js commands.  We should always move towards consistency of UI I think regardless of underlying playback technique.

IgorA100:
> but I think a better aproach would be to ensure that dvr controls is always present

I agree. I don't yet understand the reason that dvr controls are not always displayed. I can do this, but a little later.
Now, as a temporary solution, I propose a simple way of full-screen viewing by moving the button to another block.

connortechnology:
We do if($streamMode=='jpeg') to not include the dvr controls.  Because we havn't written the code to hook the buttons up to the webrtc player yet.

PR #3999, opened 2024-05-15
Chore: eslint (monitor.js)

IgorA100:


connortechnology:
I prefer to remove the debugging lines. This is no longer needed.

Issue #4006, opened 2024-05-17
Video.js используется устаревший 4.11.4, на данный момент есть версия 8.12.0


IgorA100:
**Describe Your Environment**
- Version of ZoneMinder 1.37

http://videojs.com
https://github.com/videojs/video.js

IgorA100:
@connortechnology 
I'll see when I have time

connortechnology:
For consistency of UI, video.js may not really be the way to go.  Might be better to integrate something like https://github.com/numberwolf/h265web.js which would give us h265 as well.

IgorA100:
Agree.
I will think.

connortechnology:
See some partial work I did here https://github.com/connortechnology/ZoneMinder/tree/h265web.js

IgorA100:
> partial work

Why didn't you finish your work? What were the difficulties?

connortechnology:
I think it basically worked, just needed to hook up all the UI elements

IgorA100:
https://github.com/ZoneMinder/zoneminder/commit/93248954fe5854e942a1bb45cac115378d0b47b7

PR #4007, opened 2024-05-17
Fix: Autoplay for MP4 (event.php)

IgorA100:
Now there are no problems with autoplay.
Probably you still need to update the player #4006 or you don’t need to install “autoplay” & “muted” in “data-setup”

The error in the browser console has been resolved: "Uncaught (in promise) DOMException: play() failed because the user didn't interact with the document first."


connortechnology:
Doesn't this mean it will always start up muted? That's not behaviour that I want.

IgorA100:
Yes this is bad.
I didn't pay attention to this because... I still haven’t been able to properly broadcast sound from the cameras :(

PR #4008, opened 2024-05-17
Feat: Added support for PanZoom on Event page

IgorA100:


IgorA100:
@connortechnology 
Please check.

connortechnology:
Uncaught TypeError: Cannot read properties of undefined (reading 'getScale')
    at ￼skins_classic_views_…15962920.js:1695:74

connortechnology:
nothing has the zoompan class.  Missing a commit?

IgorA100:
Is this the only error?
Very strange.
If there were no class there would be an error when initializing zoompan at line 1885

IgorA100:
@connortechnology 
File added.
Ready for testing!

connortechnology:
Looks better, doesn't seem to work with video.js player. 

One question, Love the shift to zoom in and the ctrl to zoom all the way out.  How do I just zoom out a little bit? Seems to me that ctrl-click should back out a little, and ctrl-shift-click should zoom all the way out.  ztrl-shift changes the mouse icon, but not sure what it means.  

Mouse zoom in/out is really awesome.

IgorA100:
> doesn't seem to work with video.js player.

Is Zoom not working or are there any problems with the player?
It works for me without any problems with video.js
Are there any errors in the browser console?

> How do I just zoom out a little bit?

I just posted a message on the forum: https://forums.zoneminder.com/viewtopic.php?t=33253

> Seems to me that ctrl-click should back out a little, and ctrl-shift-click should zoom all the way out

No problem I can add this to all Watch, Montage & Event pages

> ztrl-shift changes the mouse icon, but not sure what it means.

This means that the key combination is invalid.

> Mouse zoom in/out is really awesome.

I'm also delighted with how it turned out :)

IgorA100:
video.js add a bunch of its own blocks and therefore the bottom panel of the player also scales along with the image.
I don’t know how to avoid this yet, because...
1. It is necessary to analyze video.js more deeply, but since... We have an ancient version installed, then....
2. You need to update video.js to the latest version, but....
3. Do you want to add H.265

I don't know what to do with the bottom panel of the player yet.

В остальном у меня нет проблем с video.js

IgorA100:
@connortechnology
I added your wishes:
Ctrl + mouse click = zoom out
Ctrl + Shift + mouse click = reset scale to "1"

And indeed Zoom In/Out did not work on **mouse click** when using video.js. It only worked when twisted.
I have corrected.

connortechnology:
Ideally I would like to get rid of video.js.  It doesn't really do all that much for us. I would like to switch the h265web.js as it adds h265 capability.

IgorA100:
> I would like to switch the h265web.js

I already understood that.
But "right now" I won't be able to replace video.js with h265web.js
I need some time.
In addition, now I want to take a break (for 1-2 weeks) from working on ZM and work on my garden.
Those. I will fix some obvious bugs, I will think about improvements, but most likely I will not make global changes.
What to do with this PR?

connortechnology:
It has already taken years, it can wait another two weeks :) Seriously, no pressure from anyone here, you've done amazing work.

IgorA100:
Ok, Isaac.
When I do work, I am responsible for its correctness.
When you approve of my PR, you trust me.
And if there are any mistakes in my work, I am obliged to correct them as soon as possible so that other people do not experience problems.
I've done a lot over the past couple of months and there are probably some bugs somewhere that need to be fixed.

PR #4012, opened 2024-05-18
Fix: Select Layouts on Montage page

IgorA100:
**Preset Layouts should be first in the Layots list**

- Moved list of preset Layouts from montage.js to montage.php
- Changed the definition of the default Layout

### DIFF
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -53,7 +53,7 @@ function stringToNumber(str) {
 }
 
 function isPresetLayout(name) {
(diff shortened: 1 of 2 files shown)


IgorA100:
Please check and approve before use #4014

connortechnology:
Since we may come to the montage page with different selections of monitors selected, maybe by group, perhaps we should have an auto layout option, which tries to guess how many across it should use.  Because once the cookie is set, it won't change. Just a thought.

IgorA100:
> Because once the cookie is set, it won't change. Just a thought.

Thanks for the good idea, I'll think about it.
Maybe you should remove "Freeform" (now there is practically no use for it) and add the "auto" mode you suggest?

IgorA100:
@connortechnology 
Done #4019
I did as you requested :)
I removed the "Freeform" layout because... I find it pointless and added the "Auto" mode

PR #4014, opened 2024-05-19
Feat: Delete Layout on Montage page

IgorA100:


IgorA100:
Please check and approve after use #4012

PR #4015, opened 2024-05-21
Chore: May not need to hide ".chosen" (skin.css)

IgorA100:
Additional analysis is required.


IgorA100:
Everything seems to be fine, but let the line be frozen for now.
I'll delete it later if no one has any problems.

PR #4029, opened 2024-05-22
Fix: placement of monitors when initializing Layout (montage.js)

IgorA100:
This is a temporary dirty code to check if the monitors are positioned correctly on preset layouts


connortechnology:
Yes that fixes it both the weird layout and the scroll jumping

PR #4030, opened 2024-05-22
Fix: placement of monitors when initializing Layout №2 (montage.js)

IgorA100:
This code is clean.
Should fix the situation with incorrect placement of monitors on preset layouts, but should be checked.


connortechnology:
This fixes the layout problem but not the jiggling problem.

IgorA100:
I think this PR should be combined, because... it fixes one of the problems.

PR #4031, opened 2024-05-22
Fix: Jiggling montage layout (montage.js)

### LINKED ISSUE #4028: Jiggling montage layout
On my system, I have 12 cameras, a mix of 1080p, 720p, some webrtc some playing by zms.  If I select a 4wide layout, and hide the filter bar, they almost exactly fill the space.  However it clearly overflows by like 1 pixel and then the scroll bar appears, but then the layout code does it's thing and the scroll bar disappears and then the layout code does it's thing and the scroll bar reappears, etc.  Causing them all to bounce around by a pixel or two.

IgorA100:
Attempt to fix #4028
It seems to work for me, but there may be “side effects”, because... The analysis of monitor size changes is interrupted. Additional testing is required.

### DIFF
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -12,6 +12,7 @@ var objGridStack;
 
 var layoutColumns = 48; //Maximum number of columns (items per row) for GridStack
 var changedMonitors = []; //Monitor IDs that were changed in the DOM
+var scrollBbarExists = false;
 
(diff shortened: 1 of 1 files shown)


connortechnology:
Jiggles for a bit then settles.  Good enough for me.

IgorA100:
Yes, there are small jiggles and they disappear quickly, but sometimes they don't appear at all
I can't catch them :(

PR #4039, opened 2024-05-24
Fix:  changedMonitors.forEach (montage.js)

### LINKED ISSUE #4037: The placement of monitors on the Montage page is broken again
**Describe Your Environment**
- Version of ZoneMinder 1.37

Someone somewhere broke something...... #4032

IgorA100:
Iterating through the array was incorrect

Possible fixes #4037

### DIFF
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -798,7 +798,7 @@ function initPage() {
 
   setInterval(() => { //Updating GridStack resizeToContent, Scale & Ratio
     if (changedMonitors.length > 0) {
-      changedMonitors.forEach(function(item, index, object) {
+      changedMonitors.slice().reverse().forEach(function(item, index, object) {
         const value = getSelected(document.getElementById("ratio"+item));
         const img = document.getElementById('liveStream'+item);
         const currentMonitor = monitors.find((o) => {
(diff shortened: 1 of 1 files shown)


web/skins/classic/views/js/montage.js
```diff
@@ -798,7 +798,7 @@ function initPage() {
 
   setInterval(() => { //Updating GridStack resizeToContent, Scale & Ratio
     if (changedMonitors.length > 0) {
-      changedMonitors.forEach(function(item, index, object) {
+      changedMonitors.slice().reverse().forEach(function(item, index, object) {
```
connortechnology, line 801: It is not clear to the casual observer why the slice and reverse are neccessary. perhaps a comment would help.
Alternatively a simple for loop instead of a forEach is actually faster and you can just clear the array at the end. Would be easier to read as well.

IgorA100: At the time of execution of this code, other data may be filled into the "changedMonitors" array. Therefore, it cannot be cleaned after processing all elements. Step by step cleaning is required. This is why "for" is not used.
slice().reverse() - to iterate over a copy of the array from the end, otherwise it will not be possible to correctly remove the processed monitor from the array

connortechnology: Aha!  Excellent content for a comment.

So the slice makes a copy...  which we could use for on.  Maybe we should just use a while with popping off elements until the array is empty... so that if another item gets added we will process it.

connortechnology:
Id say leave it for now.  It's not a big deal.

web/skins/classic/views/js/montage.js
IgorA100: I'm not ready to say anything yet.

IgorA100:
I made changes
Try again.

> Id say leave it for now. It's not a big deal.

:)

connortechnology:
That's got it. Well done.

PR #4040, opened 2024-05-26
Optimizing code related to GridStack (montage.js)

IgorA100:
And fix monitor positioning when switching Ratio


IgorA100:
@connortechnology 
Please check.
I hope it doesn't happen again #4037

PR #4042, opened 2024-05-27
Chore: Moved PanZoom functions to "web/js/panzoom.js"

IgorA100:
Moved from Montage, Watch & Event page


IgorA100:
I added another commit https://github.com/ZoneMinder/zoneminder/pull/4042/commits/9ec5c1ad90fe72fdf97469459db1d7bb71bbdc83
Although it is not related to the transfer of PanZoom functions, it is related to the monitorsSetScale function

@connortechnology 
If you are against it, then I can delete the commit and create another PR after this PR is approved

connortechnology:
No problem with the scale stuff.
This looks great, I'm just wondering if we shouldn't stick it into it's own .js file instead of skin.js.  And only load that file on pages that need it.

IgorA100:
> It would likely be useful to other skins.

Ok

IgorA100:
Done

PR #4044, opened 2024-05-27
Feat: Added Rate control on Montage page

IgorA100:
So far I have changed the Rate for all monitors.
I did not change the Rate for each monitor separately, because... it's probably not necessary. Or is it necessary? Probably not necessary :)

### DIFF
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -1086,6 +1086,25 @@ function monitorsSetScale(id=null) {
   }
 }
 
+function changeMonitorRate() {
+  const rate = $j('#changeRate').val();
+  monitorsSetRate(rate);
+  setCookie('zmMontageRate', rate);
+}
+
+function monitorsSetRate(fps, id=null) {
+  if (id) {
+    var currentMonitor = monitors.find((o) => {
+      return parseInt(o["id"]) === id;
+    });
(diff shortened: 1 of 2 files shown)


IgorA100:
Significantly saves network traffic.
Ready for testing.

web/skins/classic/views/js/montage.js
```diff
@@ -1086,6 +1086,25 @@ function monitorsSetScale(id=null) {
...
+  monitorsSetRate(rate);
+  setCookie('zmMontageRate', rate);
+}
+
+function monitorsSetRate(fps, id=null) {
+  if (id) {
+    var currentMonitor = monitors.find((o) => {
+      return parseInt(o["id"]) === id;
+    });
+    currentMonitor.streamCommand({command: CMD_MAXFPS, maxfps: fps});
```
IgorA100, line 1100: It would be more correct to write:
`currentMonitor.setMaxFPS(fps);` but due to #4045 this code will not work.

connortechnology:
Not sure why you felt the need to do this when I said I already had it done and was going to merge it.
Oh well.

IgorA100:
> Not sure why you felt the need to do this

I don't know myself :) It just took 30 minutes of work.

> I said I already had it done and was going to merge it.

I forgot about this :(.

PR #4053, opened 2024-05-29
Stop monitors when closing montage page

### LINKED ISSUE #4051: An error appears when switching from Montage page to Console page.
**Describe Your Environment**
- Version of ZoneMinder 1.37

`ERR | Error writing packet: Invalid argument | zm_videostore.cpp`

At the moment the Montge page is closed, something probably needs to be stopped.

IgorA100:
Possibly closed #4051

### DIFF
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -1103,6 +1103,15 @@ function changeMonitorStatusPositon() {
 // Kick everything off
 $j(window).on('load', () => initPage());
 
+//Stop monitors when closing page
+document.onvisibilitychange = () => {
+  if (document.visibilityState === "hidden") {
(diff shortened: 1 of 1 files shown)


connortechnology:
This seems good, but what happens when the tab becomes visible again?

connortechnology:
Hmm also doesn't seem to do anything here...

IgorA100:
> This seems good, but what happens when the tab becomes visible again?

Yes, you need to restart the monitors. I'll correct it, BUT!


> Hmm also doesn't seem to do anything here...

Nothing happens because... the code `monitors[i].kill();` does not stop the monitors!

IgorA100:
I made an addition: #4054
But my monitors still don't stop.

PR #4054, opened 2024-05-29
Start monitors when show page after hidden on Montage page

IgorA100:
Correction to #4053

### DIFF
--- a/web/js/MonitorStream.js
+++ b/web/js/MonitorStream.js
@@ -316,6 +316,7 @@ function MonitorStream(monitorData) {
     }
     stream.onerror = this.img_onerror.bind(this);
     stream.onload = this.img_onload.bind(this);
+    this.started = true;
   }; // this.start
 
   this.stop = function() {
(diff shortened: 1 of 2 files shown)


connortechnology:
check latest master.  Should actually kill the stream.

IgorA100:
Fantastic !

Everything works correctly :)
Now, if the Montage page is not displayed in the browser, then the load on the server CPU is significantly reduced.

IgorA100:
Probably the same should be applied to the Watch page?

connortechnology:
Yes definitely apply to watch.  And zones and zone edit.   Maybe it should be automatic in MonitorStream.js.  
Need to also pause rtsp2web. I think janus is already handled.

IgorA100:
> Yes definitely apply to watch. And zones and zone edit.

Done

> Maybe it should be automatic in MonitorStream.js.

Yes, it worked!

> Need to also pause rtsp2web.

I haven't figured out rtsp2web yet :( 

> I think janus is already handled.

I installed Janus, but I don't like how it works at all and I haven't figured it out yet.

PR #4055, opened 2024-05-29
Added starting/stopping monitor when show/hide Watch page

IgorA100:
But stopping the monitor does not work correctly.
The monitor stops and starts again.
I don't know how to fix this yet.
It's probably not enough to change mode=jpeg to mode=single in MonitorStream.js

### DIFF
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -1414,6 +1414,5 @@ document.onvisibilitychange = () => {
     if (!monitorStream.started) {
       monitorStream.start();
     }
-
   }
 };
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -1412,6 +1412,9 @@ document.onvisibilitychange = () => {
     monitorStream.kill();
   } else {
     //Start monitor when show page
-    monitorStream.start();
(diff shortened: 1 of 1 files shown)


connortechnology:
It's probably a failed ajax causing a reload.  Changing to single will definitely kill the stream, but it can still be restarted by changing it back to jpeg.  Def needs more work.

IgorA100:
My holey head.... :)
For some reason I decided to create a new monitorStream, if it didn’t disappear anywhere....
I've made corrections.

connortechnology:
Hey I love small digestible PRs but maybe we could go a little larger. Some people are complaining about the volume of emails they are getting.

IgorA100:
Ok.
I try to ensure that each PR relates to a specific logical change in behavior. This is how I was taught in other projects. And in this case, the inspector, i.e. It’s easier for you to analyze changes and reject or approve PR.
I can reduce the number of PRs by combining changes into one PR instead of 2-3.

connortechnology:
I often use two tabs and flip between them to get info about the monitor while watching the live view.  I think we should set a timeout on stopping the stream, maybe 5 seconds or so.

IgorA100:
You're probably right.
I will think.

IgorA100:
I created PR #4060
I suggest continuing the discussion there.

PR #4058, opened 2024-05-29
Added Scales = Max 480, 640, 800, 1024, 1280, 1600 px & added settings Stream Quality on Watch, Event & Montage page

IgorA100:
https://forums.zoneminder.com/viewtopic.php?t=33262


IgorA100:
@connortechnology 
I decided to add the $streamQuality parameter and committed https://github.com/ZoneMinder/zoneminder/pull/4058/commits/18628dbc4422944567f6a6cedd35b5931d02d916

Initially, I planned to make the settings in Bandwidth control, but then I didn’t really like this idea.
I hope you don't mind?
In this way, you can very flexibly manage both the quality and load on network traffic.

connortechnology:
I agree with the approach.  We might do a default value in bandwidth settings but yeah it should be selectable at all times.

IgorA100:
> We might do a default value in bandwidth settings

Maybe yes. But this can be done later.

IgorA100:
Done.
Please check.

web/skins/classic/views/event.php
```diff
@@ -63,8 +63,18 @@
   $scale = validInt($monitor->DefaultScale());
   $scaleSelected = $monitor->DefaultScale();
 }
-if (!validInt($scale) and $scale != '0') {
-  $scale = '0';
+//if (!validInt($scale) and $scale != '0') {
+//  $scale = '0';
+//}
+$scale = '10'; // temporarily to adapt the new algorithm, not used in new calculations!
```
IgorA100, line 69: Doesn't matter. Next (in JS) is the recalculation of Scale

IgorA100:
Done/
Please check.

IgorA100:
I've added support for $streamQuality on Montage page.
Now it works on Watch, Event & Montage page.
Isaac, you can check.

IgorA100:
@connortechnology 
Is there something you don't like or don't have time to check the code?

PR #4062, opened 2024-05-31
When viewing in a cycle, move to the next stream without reloading Watch page

### LINKED ISSUE #4059: UI Improvement on watch view
Initial video load makes the UI jump around too much.  We should just figure out the scaling and start the stream all in javascript.

The UI jumps again on cycle change.  It seems to jump stopping the existing stream, then do a full page reload. Ideally we would just change the stream without a full reload.

IgorA100:
Just rough sketches for now.
Moving to the next stream works.
Otherwise, there are still many nuances that need to be corrected.

Partial fix #4059

### DIFF
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -302,7 +302,7 @@ function onPlay() {
   setButtonStateWatch('stopBtn', 'inactive');
   setButtonStateWatch('playBtn', 'unavail');
   if (monitorStream.status.delayed == true) {
-    setButtonState('stopBtn', 'inactive');
+    //setButtonState('stopBtn', 'inactive');
     if (monitorStreamReplayBuffer) {
       setButtonState('fastFwdBtn', 'inactive');
       setButtonState('slowFwdBtn', 'inactive');
@@ -310,7 +310,7 @@ function onPlay() {
       setButtonState('fastRevBtn', 'inactive');
     }
   } else {
-    setButtonState('stopBtn', 'unavail');
+    //setButtonState('stopBtn', 'unavail');
     if (monitorStreamReplayBuffer) {
       setButtonState('fastFwdBtn', 'unavail');
       setButtonState('slowFwdBtn', 'unavail');
@@ -967,7 +967,7 @@ function streamPrepareStart(monitor=null) {
       monitorsSetScale(monitorId);
     }
   }, 100);
-  setButtonState('stopBtn', 'active');
+  setButtonStateWatch('stopBtn', 'active');
 }
 
 function handleMouseEnter(event) {
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -295,7 +295,7 @@ function streamCmdPause(action) {
 }
 
 function onPlay() {
-  //monitorStream.setup_onplay(onPlay); //IgorA100 Added for testing, but probably not required 
+  //monitorStream.setup_onplay(onPlay); //IgorA100 Added for testing, but probably not required
   //setButtonState('pauseBtn', 'inactive');
   //setButtonState('playBtn', 'active');
   setButtonStateWatch('pauseBtn', 'inactive');
@@ -959,7 +959,7 @@ function streamPrepareStart(monitor=null) {
   el.addEventListener('mouseenter', handleMouseEnter);
   el.addEventListener('mouseleave', handleMouseLeave);
 
-  let i = setInterval(function() {
+  const i = setInterval(function() {
     if (document.querySelector('[id ^= "liveStream"]').offsetHeight > 20) {
       //You need to wait until the image appears.
       clearInterval(i);
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -1002,7 +1002,7 @@ function streamStart(monitor = null) {
   } else {
     monitorStream = new MonitorStream(monitorData[monIdx]);
   }
-   monitorStream.setBottomElement(document.getElementById('dvrControls'));
+  monitorStream.setBottomElement(document.getElementById('dvrControls'));
   // Start the fps and status updates. give a random delay so that we don't assault the server
   //monitorStream.setScale($j('#scale').val(), $j('#width').val(), $j('#height').val());
   monitorsSetScale(monitorId);
@@ -1046,7 +1046,7 @@ function streamReStart(oldId, newId) {
 
   //Change main monitor block
   document.getElementById('monitor').innerHTML = currentMonitor.streamHTML;
- 
+
   //Change active element of the navigation menu
   document.getElementById('nav-item-cycle'+oldId).querySelector('a').classList.remove("active");
   document.getElementById('nav-item-cycle'+newId).querySelector('a').classList.add("active");
@@ -1258,13 +1258,13 @@ function cycleNext() {
   monitorStream.kill();
 
   // +++ Start next monitor
-  let oldId, newId;
+  let oldId;
   if (monIdx == 0) {
     oldId = monitorData[monitorData.length-1].id;
   } else {
     oldId = monitorData[monIdx-1].id;
   }
-  newId = monitorData[monIdx].id;
+  const newId = monitorData[monIdx].id;
   streamReStart(oldId, newId);
   cycleStart();
   // --- Start next monitor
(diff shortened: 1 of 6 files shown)


IgorA100:
@connortechnology 
You can try my changes to understand whether I’m working in the right direction.

IgorA100:
I need to know what tags (and their attributes) can be for Stream.
I currently know two options, for example:
`<video id="liveStream14" autoplay="" muted="" controls="" playsinline="" style="height: 558px;"></video>`
`Camera_8`

connortechnology:
The video tag is for when we are using janus/rtsp2web.  So the code in js/MonitorStream should handle altering it.  
For the img stream, you just need to change the monitor=6

web/skins/classic/views/js/watch.js
```diff
@@ -1190,7 +1190,33 @@ function cycleNext() {
...
+
+  // +++ Старт следующего монитора
+  monitorStream = new MonitorStream(monitorData[monIdx]);
+  const img = document.getElementById('liveStream'+monitorData[monIdx-1].id);
+  const src = img.src;
+  if (src) {
+    const url = new URL(src);
+    url.searchParams.set('monitor', monitorData[monIdx].id);
+    url.searchParams.delete('connkey');
+    url.searchParams.set('mode', 'single');
```
connortechnology: why start with single?  Let's go straight to streaming

connortechnology: doesn't really matter I suppose

IgorA100: Today the decision was made to do everything completely differently.

IgorA100:
@connortechnology 
Please test it.
I believe that 99% of the work is done, but perhaps you will find some problems.
I had to do a lot of work rather than just changing the SRC because... switching from one camera to another requires many checks.
This change requires careful and complete testing with different cameras and settings.

web/skins/classic/views/js/watch.js
IgorA100: It was “such a small thing” compared to the current code changes :)

IgorA100:
Currently known issues:
- The monitor starts with Scale=100, then scale takes on the correct value. Affects the jump in network traffic at the initial moment.
- Small problem with image size when switching to the next or previous monitor.
- Display DVR control buttons

IgorA100:
> * Display DVR control buttons

I think watch.php should display all DVR control buttons and watch.js should display only those that can be used with the current monitor.
Isaac, do you agree?

connortechnology:
Yes, I think for consistency, buttons should all be there and should be in an expected location and be enabled/disabled appropriately.

IgorA100:
Ok.
I'll deal with this later.

connortechnology:
Looks good to me on my mix of zms/rtsp2web/janus etc.

IgorA100:
> Looks good to me

Great news !

IgorA100:
With the DVR control buttons, almost everything is incomprehensible to me :(
Everything is confusing here:
```
function setButtonState(element_id, btnClass) {
  var element = document.getElementById(element_id);
  if ( element ) {
    element.className = btnClass;
    if (btnClass == 'unavail' || (btnClass == 'active' && (element.id == 'pauseBtn' || element.id == 'playBtn'))) {
      element.disabled = true;
    } else {
      element.disabled = false;
    }
  } else {
    console.log('Element was null or not found in setButtonState. id:'+element_id);
  }
}
```
Here "element.disabled" (in skin.js) is used, and the classes 'unavail', 'active', 'inactive' are assigned via setButtonState() (in watch.js). At the same time, these classes are described in CSS styles only for the “input” tag and not for the “button”.

I'm confused.....
What should I do ?

IgorA100:
I partially fixed some minor issues: https://github.com/ZoneMinder/zoneminder/pull/4062#issuecomment-2147479279
Please test again.
Pause for Janus doesn't work. But most likely it was like this before.
DVR control buttons require special attention in another PR

If you don't have any problems, I think this PR should be merged.

IgorA100:
If, when viewing on the Watch page, you go to the Event view (table below the image), then when going to the next/previous event it does not work correctly. Those. the transition is made to the camera that was originally opened on the Watch page, and not to the currently selected one in the left navigation menu.
Strange behavior.

IgorA100:
I fixed the issue https://github.com/ZoneMinder/zoneminder/pull/4062#issuecomment-2150165728

IgorA100:
Is support for the old Pan & Zoom option necessary?
Or can the old code be removed?

IgorA100:
It is necessary to fix the link for the "All events" button

IgorA100:
I fixed the issue https://github.com/ZoneMinder/zoneminder/pull/4062#issuecomment-2150554173

web/skins/classic/views/js/watch.js
```diff
@@ -335,18 +295,22 @@ function streamCmdPause(action) {
...
-  setButtonState('playBtn', 'active');
+  //monitorStream.setup_onplay(onPlay); //IgorA100 Added for testing, but probably not required
+  //setButtonState('pauseBtn', 'inactive');
+  //setButtonState('playBtn', 'active');
+  setButtonStateWatch('pauseBtn', 'inactive');
+  setButtonStateWatch('stopBtn', 'inactive');
+  setButtonStateWatch('playBtn', 'unavail');
   if (monitorStream.status.delayed == true) {
-    setButtonState('stopBtn', 'inactive');
+    //setButtonState('stopBtn', 'inactive');
```
IgorA100, line 305: I still don’t understand why this was necessary?

web/skins/classic/views/js/watch.js
```diff
@@ -335,18 +295,22 @@ function streamCmdPause(action) {
...
     if (monitorStreamReplayBuffer) {
       setButtonState('fastFwdBtn', 'inactive');
       setButtonState('slowFwdBtn', 'inactive');
       setButtonState('slowRevBtn', 'inactive');
       setButtonState('fastRevBtn', 'inactive');
     }
   } else {
-    setButtonState('stopBtn', 'unavail');
+    //setButtonState('stopBtn', 'unavail');
```
IgorA100, line 313: I still don’t understand why this was necessary?

web/skins/classic/views/js/watch.js
connortechnology: A lot of this stuff is old.  The fact that the styles were only defined for inputs is because back in the day we used <input type="button">.  At some point, I decided to do away with all that because using <button> Was better.  Clearly I didn't update the css.  I don't think we need three states.  Just enabled/disabled will do.  I do think that if something is disabled, it is useful to make sure the title tooltip says why.  I also hate, but have never gotten around to doing... is remove the stop button.  The play/pause/stop concept was for cassette tapes, where pausing would actually damage/wear out the tape/mechanisms.  In a digital world, the difference is that stop goes back to the beginning of the file and pause doesn't.  The I don't feel that is useful.  Others disagree with me.

connortechnology:
Another idea, if I pause/stop or otherwise interact with the video, perhaps we should pause the cycle.

web/skins/classic/views/js/watch.js
IgorA100: I have a video cassette recorder and an audio recorder and a vinyl player with records (left over from the 90s) and I sometimes watch and listen to them :) I know what it is. Yes, the digital world has changed many things.
Regarding the Stop/Pause buttons - Pause can only be useful when watching streaming video that is cached. Those. after releasing the Pause button, viewing will continue from the point where it was paused (i.e. not in real time, but with a delay). I don't know if this behavior is provided in ZM or not

> I don't think we need three states. Just enabled/disabled will do

Agree. But now I'm not ready for the "great feat of global redesign" of the DVR control buttons. Maybe later...

IgorA100:
> perhaps we should pause the cycle.

Ok, good idea.

IgorA100:
Done.

IgorA100:
I added one more logical improvement.
https://github.com/ZoneMinder/zoneminder/pull/4062/commits/cbf3812a30fd55f7c899dcf5ab4d441c10275983

web/skins/classic/views/js/watch.js
connortechnology: Here is an example of inactive vs disabled, etc.  The ptzToggle button: should always be visible, unless ptz is disabled globally.  So always visible, but if the monitor doesn't support ptz, then it is disabled.  If the monitor DOES support ptz, then it should be either inactive or active depending on whether we are showing the ptz controls.

connortechnology:
I just fixed an issue with ptz not showing up.
I think currentMonitor should be a global and use it instead of a bunch of global vars for with and height etc.

IgorA100:
> I just fixed an issue with ptz not showing up.

Thank you!

> I think currentMonitor should be a global and use it instead of a bunch of global vars for with and height etc.

Agree.

PR #4065, opened 2024-06-04
Fix: in function "scaleToFit" do not take padding into account if the…

IgorA100:
… width is enough (skin.js)


IgorA100:
@connortechnology 
Please check on different pages with Scale=Auto
In theory, it should get better, not worse.

connortechnology:
Looks fine to me...

PR #4069, opened 2024-06-06
Apply Tooltip to objects with [data-toggle="tooltip"] & [title] attributes (skin.js)

IgorA100:


IgorA100:
It appears unsuccessfully in the image (at the very bottom and a scroll bar appears), and requires improvement.

connortechnology:
What is the point of supporting data-tooltip in addition to title?

IgorA100:
I wanted to make the title more beautiful

connortechnology:
I get making titles more beautiful, do we have anything with data-toggle=tooltip?  Why not just use title?

IgorA100:
> I get making titles more beautiful,

Great.

> do we have anything with data-toggle=tooltip?

No.

> Why not just use title?

Yes, you are right.
Not the best idea came to my mind with Tooltip

PR #4075, opened 2024-06-11
Fix: PTZ control error on Watch page

IgorA100:
+ Fix: Limit scope for "handleClick"

https://forums.zoneminder.com/viewtopic.php?p=134573#p134573

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -60,49 +60,6 @@ function checkSize() {
   }
 }
 
-// Polyfill for NodeList.prototype.forEach on IE.
-if (window.NodeList && !NodeList.prototype.forEach) {
-  NodeList.prototype.forEach = Array.prototype.forEach;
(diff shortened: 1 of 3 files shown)


IgorA100:
There is also a problem in watch.php
I'm working on solving a problem.

IgorA100:
Please check.

IgorA100:
As far as I understand, the main problem is that dataOnClick() fires before the PTZ control block is rendered to the DOM.

IgorA100:
> but nothing happens.

Are you sure you applied the changes to skin.js? This is what corrects the situation.

IgorA100:
I temporarily added a dataOnClick() call from the Watch page, this will definitely fix the problem.
Tomorrow I will delve deeper.

IgorA100:
I made another commit.

IgorA100:
Please test the current version.

PR #4085, opened 2024-07-04
Feat: Added the ability to save objects and arrays in cookies using JSON (skin.js)

IgorA100:


web/skins/classic/js/skin.js
```diff
@@ -771,18 +783,28 @@ function setCookie(name, value, seconds) {
     // 2147483647 is 2^31 - 1 which is January of 2038 to avoid the 32bit integer overflow bug.
     expires = "; max-age=2147483647";
   }
-  document.cookie = name + "=" + (value || "") + expires + "; path=/; samesite=strict";
+  document.cookie = name + "=" + (newValue || "") + expires + "; path=/; samesite=strict";
 }
 
+/**
+* If JSON is stored in cookies, the function will return an array of values.
```
connortechnology: Why an array?  Shouldn't it return a json object?

IgorA100: Will return either an array or an object (depending on what was encoded), because We cut off everything else in “isJSON” so that there are no problems.
I have corrected the code comment.

PR #4091, opened 2024-07-15
Feat: Moved part of the code to function buildMonitorsFilters. Don't use ob_get_contents (_monitor_filters.php)

IgorA100:
In fact, the code is practically unchanged!
All code logic is not affected.
Some of the code is output to the buildMonitorsFilters function, which returns an array. There are no other changes.
Firstly, it is necessary to abandon the use of ob_get_contents for the new Montage page
Secondly, this makes the code more understandable.


IgorA100:
I ask for verification and, if possible, approval.

connortechnology:
I definitely like getting rid of ob_start etc.

PR #4097, opened 2024-07-17
Do not create a Warning if you could not get the current monitor status from the DB (Monitor.php)

IgorA100:
The only reason why the status could not be obtained is the absence of a monitor in the 'Monitor_Status' table. For example - Montage page is open and
- at this moment the connection with the camera is lost
- go to the camera settings and resave
- disabled Capturing
- and so on. In all of the above cases, a useless warning will be generated that only fills the log.

Or am I wrong?


connortechnology:
If there isn't a status update in over a minute, the records will get deleted.  SO this could actually happen, although I've never seen it logged.. 

So I'm tempted to agree with you,  Maybe it could be debug level.

PR #4114, opened 2024-08-17
Update debian.rst

beltphed:
Failing to install `apache2` results in subsequent calls to `a2enconf` and `a2enmod` failing because they might not have been installed. (in my case, I was using a minimal Debian 12 install which doesn't install `apache2` by default)

### DIFF
--- a/docs/installationguide/debian.rst
+++ b/docs/installationguide/debian.rst
@@ -39,7 +39,7 @@ Run the following commands.
 
 ::
 
-    sudo apt install mariadb-server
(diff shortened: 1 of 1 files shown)


connortechnology:
apache is a hard dependency of zoneminder... it will get installed when zoneminder is installed. This change doesn't hurt anything but should be  unnecessary.  I'd be curious to know how you got into this situtation.

connortechnology:
apache2 is a hard dependency in our packages, but apparently debian has removed it in theirs.  Sigh.

Issue #4115, opened 2024-08-17
MariaDB Support?


robertsilen:
Hello, I happend to notice that [MySQL](https://zoneminder.readthedocs.io/en/stable/search.html?q=mysql&check_keywords=yes&area=default) is mentioned more than [MariaDB](https://zoneminder.readthedocs.io/en/stable/search.html?q=mariadb&check_keywords=yes&area=default#) in your documentation - any particular reason for that? Can I make a PR to add more MariaDB? Where are the corresponding files in github in that case?

connortechnology:
Because mysql predates mariadb?

files are in the docs directory.

robertsilen:
Thanks, I made a suggestion: https://github.com/ZoneMinder/zoneminder/pull/4117

PR #4117, opened 2024-08-19
add MariaDB to docs

robertsilen:
Suggesting to add MariaDB to docs, as originally mentioned in https://github.com/ZoneMinder/zoneminder/issues/4115 - hope it is ok?


robertsilen:
Thank you! Any idea when it will turn up on the website, e.g. here: https://zoneminder.readthedocs.io/en/stable/faq.html

connortechnology:
It is already in https://zoneminder.readthedocs.io/en/latest/faq.html. stable will require a release of 1.36.35.

ottok:
Sorry for posting here, I should probably open a new issue, but just a quick note that the page https://zoneminder.readthedocs.io/en/latest/installationguide/debian.html still mentions MySQL, even though Debian hasn't shipped MySQL at all since Debian 9 in 2017.

PR #4122, opened 2024-08-23
Feat: Added support for full screen when double clicking on Stream on Watch page

IgorA100:
Similar to #4120


IgorA100:
Please check.

PR #4124, opened 2024-08-23
Minor styling improvements for mobile devices. ".sticky #mfbpanel" height no more than 40% of the screen (skin.css)

IgorA100:
Relevant for mobile devices.


connortechnology:
I'm not seeing any change in behaviour here...

IgorA100:
Perhaps the estimated height of your mobile device screen is greater than 600px.
For example, when viewing the list of recorded events (view=events), the table with events was not visible, because #toolbar occupied all the vertical space.

IgorA100:
@connortechnology 
Is there something you don't like about this PR?
Should I close this PR or make some corrections to it?

connortechnology:
I have nothing against it, it just does nothing for me.  But I can see that if you had a really short phone, limiting the height of the toolbar would be a good idea.  

If I put my phone into landscape mode, then even with the limited toolbar there is no space for events.  The main header navbar lays out terribly, so there is lots to do.

PR #4128, opened 2024-09-06
Added IP2Location.io API as failover for geolocation

ip2location:


abdullahdevrel:
Just for context:

- IPinfo.io provides 1,000 requests per day without a token.
- IPinfo provides 50,000 requests per month with a free tier token.
- Universal access to a globally distributed load-balancing network regardless of access tier.
- Higher accuracy driven by latency-based IP geolocation methods. ([Blog post](https://ipinfo.io/blog/probe-network-how-we-make-sure-our-data-is-accurate/))

In addition to offering generous no-tier and free-tier access, we also have incredible uptime hours. If access or limits to our service is an issue, we are happy to look into it. Thanks.

Issue #4127, opened 2024-09-03
GridStack slow to init when a lot of monitors on montage


connortechnology:
With over 70 monitors, it can take over 40 seconds before first render.  firefox will complain about gridstack init slowing down the browser, etc.

I notice that we are calling init twice, which could perhaps be improved.  Not sure what else can be done.

IgorA100:
I'll definitely watch it, but later.
I don't have time right now :(

IgorA100:
> With over 70 monitors, it can take over 40 seconds before first render

1. Can you make a video of the whole rendering process?
2. Does the rendering speed depend on the selected "Layout"? I think "8Wide" will load faster than "3Wide"

connortechnology:
https://github.com/user-attachments/assets/4ecf6c90-9580-4ede-af1c-857b46ea22aa

connortechnology:
Yeah 8 Wide is faster.

IgorA100:
> I notice that we are calling init twice

GridStack initialization consumes almost no resources. Yes, when opening a page, initialization occurs twice, this is normal for now.

Thanks for the video.

As far as I understand, the long page rendering occurs due to loading a large amount of information.
Reducing "Stream quality" to -50 should speed up page rendering.

IgorA100:
Maybe I found the reason.

connortechnology:
It is massively improved, while faster would be even better, I think we can close this.

PR #4133, opened 2024-09-14
Fix: Don't specify "sizeToContent: true" when initializing GridStack on Montage page

IgorA100:
Additional verification is required, including on Custom layouts 
I haven't tested everything. Maybe something will break somewhere....
Possible closure #4127, but I'm not sure.

### DIFF
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -872,7 +872,7 @@ function initGridStack(grid=null) {
   const opts = {
     margin: 0,
     cellHeight: '1px',
-    sizeToContent: true, // default to make them all fit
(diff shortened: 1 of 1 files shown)


IgorA100:
This PR slightly violates the sort order different from the console

IgorA100:
Sorting should now work correctly.

connortechnology:
Some missing pieces in there, I've fixed them up but you may  want to review my fixes.

connortechnology:
Hasn't made a big difference for #4127

IgorA100:
> to review my fixes

I looked at it, everything is fine.
Thanks for the help.
My montage.js file is very heavily modified now and due to this there were some stupid mistakes when transferring the code.

PR #4136, opened 2024-09-17
Feat: Progress bar now works when viewing MP4 via video.js on Event page

### LINKED ISSUE #4134: video.js progress bar broken by panzoom
5e76dfe273691e0d4d521e3ed643430ab7169a23 caused clicking on the progress bar provided by video.j.s to stop working.

IgorA100:
I'm not sure if this PR fixes #4134
But now the Progress bar correctly displays the timeline..
I don't know if this ever worked before....
12345r

### DIFF
--- a/web/skins/classic/views/js/event.js
+++ b/web/skins/classic/views/js/event.js
@@ -960,16 +960,22 @@ function drawProgressBar() {
 
 // Shows current stream progress.
 function updateProgressBar() {
+  if (vid) {
+    if (!eventData) return;
+    var currentTime = vid.currentTime();
+    var progressDate = new Date(currentTime);
+  } else {
   if (!(eventData && streamStatus)) {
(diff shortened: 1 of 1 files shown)


connortechnology:
This might allow use to simply hide the video.js controls, which may be a good idea.  The ZM native progress bar is ... kinda ugly in comparison, but is at least consistent when using zms or h265web.js.

IgorA100:
> This might allow use to simply hide the video.js controls,

But this panel has volume control. I don't know how relevant it is. You can hide the panel without any problems.

> The ZM native progress bar is ... kinda ugly in comparison

The design can be changed, but I haven't thought about it yet.

connortechnology:
in my h265web.js branch, we have added a volume control, so eventually we won't need video.js controls.

Issue #4132, opened 2024-09-14
Fullscreen mode not persistent (replay/Gapless Events)


BlueMax:
**Describe Your Environment**
zoneminder/jammy,now 1.37.63~20240906093428-jammy amd64
ppa.launchpadcontent.net/iconnor/zoneminder-master/ubuntu/ jammy main
Firefox 130

**Describe the bug**
Event view:
When enabling 'replay/Gapless Events' mode and switching to 'fullscreen' (button) it jumps back to non-fullscreen on next event (after event ends). It should keep fullscreen mode until user exits it on purpose.

connortechnology:
Unfrotunately there is no way to maintain fullscreen between page reloads.  So what we have to do here is NOT do a page reload.

IgorA100:
If you hide the Stats display to the left of the video, then the full-screen mode works correctly, because the page does not reload.
Of course, you need to do it so that the page does NOT reload when Stats is displayed

BlueMax:
Stats are always hidden here. Page reloads are still active.

BlueMax:
Mjpeg has bad quality/efficiency. I wouldn't trade that for the problem.

connortechnology:
Have we finished addressing this?  Can we close it?

BlueMax:
The problem is still there. Same page reloads as before.

IgorA100:
This problem requires deeper analysis and probably AJAX requests should be used.

PR #4143, opened 2024-09-20
Don't execute "location.replace" when moving to next event (event.js)

IgorA100:
This works for Gapless.
I haven't tested the rest.
The information block certainly isn't updated. Maybe something else isn't updated either.
I don't know who did this and why they didn't finish it....

Related to #4132

### DIFF
--- a/web/skins/classic/views/js/event.js
+++ b/web/skins/classic/views/js/event.js
@@ -690,8 +690,8 @@ function streamNext(action) {
   // How about we just reload the page?
   //
   if (vid==null) streamReq({command: CMD_QUIT});
-  location.replace(thisUrl + '?view=event&eid=' + nextEventId + filterQuery + sortQuery);
-  return;
+  //location.replace(thisUrl + '?view=event&eid=' + nextEventId + filterQuery + sortQuery);
+  //return;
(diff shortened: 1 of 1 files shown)


web/skins/classic/views/js/event.js
```diff
@@ -690,8 +690,8 @@ function streamNext(action) {
   // How about we just reload the page?
```
IgorA100, line 690: Why did you decide to do it this way?

connortechnology: because I was lazy. Also much has changed since then.

IgorA100: For now I'm transferring it to draft.
When I have time, I'll look at the code in more detail.

connortechnology: I reviewed the code and added some comments, applied use strict etc.  

THe problem is that we might be coming from an event playing with zms to an event with video.js.  We need to be able to update the actual player in js.  At the time (the video.js stuff was new) it was easier to just reload. We didn't have fullscreen back then.  ANyways, it shouldn't be a big deal now.  

I want to transition the streaming code to an EventSTream.js module like MonitorStream.  So it can be reused for montage etc. I have a need for that in the short term, so I will be working on that.

PR #4159, opened 2024-10-01
Fix: Always use only #wrapperEventVideo to enable fullscreen mode on Event page

IgorA100:
addEventListener should be used only for one block on the page, otherwise there will be problems. Closed: https://forums.zoneminder.com/viewtopic.php?p=135677#p135677


IgorA100:
I will delete the commented code later if everything is ok

PR #4163, opened 2024-10-02
Chore: Removed the use of the "delay" constant as it is no longer used when starting monitors on the Montage page

IgorA100:
Probably it broke at some point. Or maybe it never worked, but in montage.js monitors start with a delay: "monitor.start(delay);"
Essentially, the changes consist only of wrapping the function code in
```
    setTimeout(function() {
      ......
    }.bind(this), delay);
```
and adding spaces

### DIFF
--- a/web/js/MonitorStream.js
+++ b/web/js/MonitorStream.js
@@ -263,7 +263,7 @@ function MonitorStream(monitorData) {
           if (useSSL) {
             hlsUrl = "https://" + rtsp2webModUrl + "/stream/" + this.id + "/channel/0/hls/live/index.m3u8";
           } else {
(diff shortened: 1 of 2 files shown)


IgorA100:
> that makes for bad UI

In principle I agree.


> I switched from a self-refreshing setTimeout to an Interval.

Yes, I saw that.
So the code in **montage.js** should be changed from

```
      // Start the fps and status updates. give a random delay so that we don't assault the server
      const delay = Math.round( (Math.random()+0.5)*statusRefreshTimeout );
      monitors[i].start(delay);
```
on
`      monitors[i].start();`

And in **MonitorStream.js** remove the "delay" argument from function. Otherwise there will be some confusion.

connortechnology:
yes

IgorA100:
Done

PR #4198, opened 2024-12-16
Strip build time from compressed manpages

sertonix:
The manpage compression metadata include the true build time (and not SOURCE_DATE_EPOCH like the manpages themselves). Excluding the build time allows the manpage to be build reproducibly.

Ref https://reproducible-builds.org/


sertonix:
The build failure seems unrelated :thinking:

PR #4214, opened 2025-01-29
Fix: Stop RTSP2Web Type HLS when stream stops

IgorA100:
Also, we should not try to restore playback of the stream if we have forcibly stopped it.


IgorA100:
Moved to #4218

PR #4217, opened 2025-02-02
The most correct transmission of RTSP stream with audio support when using RTSPToWeb

IgorA100:
Now it will be possible to turn on the audio track when viewing online using RTSPToWeb Type WebRTC


connortechnology:
Is this ready to go ?  Um, my streams here work fine with audio...

IgorA100:
> Is this ready to go ?

I think the PR is ready.
There will be another big change in the way Type MSE works, but that will be another PR, which I will most likely publish within a week.

> Um, my streams here work fine with audio...

With or without this PR?

PR #4227, opened 2025-02-23
Fix: Call the "on_scroll" function for browsers that do not support the "scrollend" event for Montage page

IgorA100:


connortechnology:
Googling says only old IE don't support scrollend... is this really necessary? Is there a modern browser that doesn't support it?

IgorA100:
> Is there a modern browser that doesn't support it?

Oddly enough, but I was very surprised that Chromiun 109.0.5414.120 does not support scrollend
This is very strange..
Perhaps some other browser does not support "scrollend", which greatly affects the loading of the client machine.

connortechnology:
My own testing doesn't agree, but it shouldn't hurt anyone else.  Just concerned that it's a red herring.

IgorA100:
> My own testing doesn't agree

Have you tried it in Chromiun and did everything work for you?

IgorA100:
I just checked on another computer with Windows10 & Chromiun 109.0.5414.120 - and indeed there is support for "scrollend"
Earlier I tried on a computer with Windows8.1 & Chromiun 109.0.5414.120 and there was no support for "scrollend"
Only riddles.....

PR #4235, opened 2025-02-27
Feat: Ability to switch Primary/Secondary stream for RTSP2Web on Watch page

IgorA100:
Also, this PR implements more correct operation of RTSP2Web type MSE.
I did not split it into two PRs, because many changes in MonitorStream.js relate to both moments.

### DIFF
--- a/web/js/MonitorStream.js
+++ b/web/js/MonitorStream.js
@@ -291,12 +291,6 @@ function MonitorStream(monitorData) {
             videoEl.src = hlsUrl.href;
           }
         } else if (this.RTSP2WebType == 'MSE') {
-          videoEl.addEventListener('pause', () => {
-            if (videoEl.currentTime > videoEl.buffered.end(videoEl.buffered.length - 1)) {
-              videoEl.currentTime = videoEl.buffered.end(videoEl.buffered.length - 1) - 0.1;
-              videoEl.play();
-            }
-          });
           const mseUrl = rtsp2webModUrl;
           mseUrl.protocol = useSSL ? 'wss' : 'ws';
           mseUrl.pathname = "/stream/" + this.id + "/channel/" + this.currentChannelStream + "/mse";
@@ -373,6 +367,7 @@ function MonitorStream(monitorData) {
 
   this.stopMse = function() {
     this.RTSP2WebMSEBufferCleared = false;
+    this.streamStartTime = 0;
     return new Promise((resolve, reject) => {
       if (this.mseSourceBuffer && this.mseSourceBuffer.updating) {
         this.mseSourceBuffer.abort();
@@ -452,10 +447,10 @@ function MonitorStream(monitorData) {
 
   this.restart = function(channelStream = "default", delay = 200) {
     this.stop();
-    setTimeout(function() {
-      this.start(channelStream);
+    setTimeout(function() {// During the downtime, the monitor may have already started to work.
+      if (!this.started) this.start(channelStream);
     }, delay);
-  };
+  }
 
   this.pause = function() {
     if (this.RTSP2WebEnabled) {
@@ -886,7 +881,7 @@ function MonitorStream(monitorData) {
     // We correct the lag from real time. Relevant for long viewing and network problems.
     if (this.RTSP2WebType == 'MSE') {
       const videoEl = document.getElementById("liveStream" + this.id);
-      if (this.wsMSE && videoEl.buffered != undefined && videoEl.buffered.length > 0) {
+      if (this.wsMSE && videoEl.buffered != undefined && videoEl.buffered.length > 0 && this.streamStartTime !==0) {
         const videoElCurrentTime = videoEl.currentTime; // Current time of playback
         const currentTime = (Date.now() / 1000);
         const deltaRealTime = (currentTime - this.streamStartTime).toFixed(2); // How much real time has passed since playback started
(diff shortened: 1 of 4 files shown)


web/js/MonitorStream.js
```diff
@@ -12,10 +12,16 @@ function MonitorStream(monitorData) {
...
   this.RTSP2WebStream = monitorData.RTSP2WebStream;
+  this.currentChannelStream = null;
+  this.RTSP2WebMSEBufferCleared = true;
   this.webrtc = null;
   this.hls = null;
   this.mse = null;
   this.wsMSE = null;
+  this.streamStartTime = 0; // Initial point of flow start time. Used for flow lag time analysis.
+  this.waitingStart;
+  this.mseListenerSourceopenBind = null;
```
connortechnology, line 23: Nitpicky but why isn't the o in open and the e in end captitalized?

web/js/MonitorStream.js
```diff
@@ -351,21 +354,65 @@ function MonitorStream(monitorData) {
     this.statusCmdTimer = clearInterval(this.statusCmdTimer);
     this.streamCmdTimer = clearInterval(this.streamCmdTimer);
     this.started = false;
-    if (this.webrtc) {
+    if (this.RTSP2WebType == 'WebRTC' && this.webrtc) {
       this.webrtc.close();
       this.webrtc = null;
-    } else if (this.hls) {
+    } else if (this.RTSP2WebType == 'HLS' && this.hls) {
```
connortechnology, line 360: The test for RTSP2WebType seems superfluous here, is it? Same about with webrtc

IgorA100: Yes, you are right. I forgot to remove it.
How good that you checked it with a fresh look...
Only I should leave the "this.RTSP2WebType" check, and this.hls and this.webrtc should be removed. This will be more correct, because at the moment of start there is not always an object this.hls || this.webrtc || this.mse.
I will fix it.

web/js/MonitorStream.js
IgorA100: I specified "mseListenerSourceopenBind" instead of "mseListenerSourceOpenBind" because "Sourceopen" is the name of the Event Listener and it is written everywhere as "sourceopen" and not "sourceOpen"
I thought about how to write it correctly. but in the end I chose the option with a small "o".
Should I change it to "mseListenerSourceOpenBind"?

> and the e in end captitalized?

What word is this?

connortechnology: Next line.  That sounds like a reasonable reason.  Good enough for me.

web/js/MonitorStream.js
connortechnology: Is it possible for type == HLS but for some other reason this.hls is null?  If it were me, I'd probably go with if(this.hls), but either should work.

web/js/MonitorStream.js
IgorA100: > Next line.

"Updateend" is also the name of the Event Listener, as in the previous line.
Leave it as is?

connortechnology: sure

web/js/MonitorStream.js
IgorA100: > Is it possible for type == HLS but for some other reason this.hls is null?

I haven't checked with type == HLS, but I think it's possible.
With type == MSE - it's possible.

IgorA100: Oh, I thought that this is a piece of code at start, but this is a piece of code at stop.
Everything is different here..
Regarding type == MSE - at the moment of "STOP", this.mse may not be there because the socket has not yet had time to open, and the stream stops. This situation is possible at the moment when you scroll up/down on the Montage page very quickly and for a long time or in other situations.
I analyze "this.mse" in a separate function "stopMse"
I decided to do this to unify the code and avoid possible errors.
Yes, you can probably make the code like this:

```
    if (this.webrtc) {
      this.webrtc.close();
      this.webrtc = null;
    } else if (this.hls) {
      this.hls.destroy();
      this.hls = null;
    } else if (this.RTSP2WebType == 'MSE') {
      this.stopMse();
    }
```

But this is NOT possible

```
    if (this.webrtc) {
      this.webrtc.close();
      this.webrtc = null;
    } else if (this.hls) {
      this.hls.destroy();
      this.hls = null;
    } else if (this.mse) {
      this.stopMse();
    }
```

IgorA100: What should I do?
Leave the code as is or change it to:
```
    if (this.webrtc) {
      this.webrtc.close();
      this.webrtc = null;
    } else if (this.hls) {
      this.hls.destroy();
      this.hls = null;
    } else if (this.RTSP2WebType == 'MSE') {
      this.stopMse();
    }
```
I would leave it as is, but the final word is yours, Isaac :)

connortechnology: Leave it as is.

IgorA100: So you can do a merge.
I am very happy now.
In any case, everything works fine for me with the second stream. And you can always watch the first stream on the Watch page. This is especially relevant for 4K cameras.
Now you don't need to have a bunch of duplicate cameras for the second stream.
Do you use RTSP2Web?

IgorA100: Debugging MSE took a very, very long time. :(
But now the average stream delay does not exceed 1-2 seconds, which is good.
I do not observe any buffer memory leaks either.
And it should work even with a CPU load of 99%, yes there will be frequent restarts, but there should be no errors.
WebRTC also seems to work fine.
I have not dealt with HLS....
Phew...
I'll go have a rest....

PR #4236, opened 2025-02-27
Fix: Always send AJAX requests to the standard port 80 (MonitorStreamjs)

IgorA100:
If you send an AJAX request taking into account the monitor port, then in some situations there may be CORS security errors, because we are ~~always on a page with a standard port~~.
While writing, I realized that ZM can work on a non-standard port.
So I need to remove the monitor port and add the port of the page we are on...
Right?

### DIFF
--- a/web/js/MonitorStream.js
+++ b/web/js/MonitorStream.js
@@ -6,6 +6,7 @@ function MonitorStream(monitorData) {
   this.id = monitorData.id;
   this.connKey = monitorData.connKey;
   this.url = monitorData.url;
+  this.urlWOPort = deletePortFromUrl(monitorData.url);
   this.url_to_zms = monitorData.url_to_zms;
   this.width = monitorData.width;
   this.height = monitorData.height;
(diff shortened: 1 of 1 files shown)


IgorA100:
Add a commit https://github.com/ZoneMinder/zoneminder/pull/4236/commits/5bd6f683f71a07bb494a57a25f5b38771aa2d6e9
"Send AJAX requests through the port on which the page being viewed is located"

connortechnology:
No.... none of this should be necessary and I think it is ugly.  If there is a CORS problem, we can fix that at the CORS level 
The only thing that REALLY matters is that we are talking to the right server (so that zms can talk to zmc), but this multi-port stuff is to get around the 6 connection limit.  If you start doing all your ajax on the same port they will get backed up even though they really shouldn't take more than a fraction of a second each..
I use multi-port and multi-server extensively and when I have seen a CORS issue it has been due to something else stupid...

IgorA100:
> If you start doing all your ajax on the same port they will get backed up even though they really shouldn't take more than a fraction of a second each..

Probably yes.
I don't remember exactly the reason for the CORS problem, why I decided to do it this way. I did it a long time ago, and now I decided to create a PR.
Ok, I will watch the code work without this PR.

PR #4237, opened 2025-02-28
Fix: Start auto Play when using videojs on Event page

IgorA100:
It is necessary to add the "muted" attribute to the "video" tag
Otherwise, the browser policy prohibits starting auto Play.
But the browser policy allows auto Play to start if the user has performed some actions on the page.
Therefore, when clicking on an event from the list, you can allow Audio to be turned on, which is what we do in event.js
That is, the final algorithm is as follows:
1. When clicking **links** on events, playback will start with sound (as it was before)
2. When refreshing the page (Ctrl+F5 || Ctrl+R || pasting a link into the browser line), playback will start without sound (previously, playback did not start at all and there was an error in the browser console)
3. When the "onplay" event is triggered for the "vid" object, you should not execute "vid.play()", since we use the "autoplay" attribute for the "video" tag, which could lead to unpredictable consequences.
This algorithm is fully compliant with browser policies.

Also, this PR actually hides the standard "Play" button of the "videojs" player, since playback always starts automatically.
Closed https://forums.zoneminder.com/viewtopic.php?t=33839&sid=f06098a69ceed3de64de4a8fe5b873b5

I tried to implement something similar in #4007 but then I abandoned it. Now it seems to have worked out.

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -1253,8 +1253,18 @@ var doubleClickOnStream = function(event, touchEvent) {
 
   if (target) {
     if (document.fullscreenElement) {
+      if (getCookie('zmEventStats') && eventStats) {
+        eventStats.toggle(true);
+        wrapperEventVideo.removeClass('col-sm-12').addClass('col-sm-8');
+        changeScale();
+      }
       closeFullscreen();
     } else {
+      if (getCookie('zmEventStats') && eventStats) {
+        eventStats.toggle(false);
+        wrapperEventVideo.removeClass('col-sm-8').addClass('col-sm-12');
+        changeScale();
+      }
       openFullscreen(target);
     }
     if (isMobile()) {
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -1253,18 +1253,8 @@ var doubleClickOnStream = function(event, touchEvent) {
 
   if (target) {
     if (document.fullscreenElement) {
-      if (getCookie('zmEventStats') && eventStats) {
-        eventStats.toggle(true);
-        wrapperEventVideo.removeClass('col-sm-12').addClass('col-sm-8');
-        changeScale();
-      }
       closeFullscreen();
     } else {
-      if (getCookie('zmEventStats') && eventStats) {
-        eventStats.toggle(false);
-        wrapperEventVideo.removeClass('col-sm-8').addClass('col-sm-12');
-        changeScale();
-      }
       openFullscreen(target);
     }
     if (isMobile()) {
--- a/web/skins/classic/views/event.php
+++ b/web/skins/classic/views/event.php
@@ -331,7 +331,7 @@ if (file_exists($Event->Path().'/objdetect.jpg')) {
 <?php
 if ($video_tag) {
 ?>
-                  <video autoplay id="videoobj" class="video-js vjs-default-skin"
+                  <video muted autoplay id="videoobj" class="video-js vjs-default-skin"
                     style="transform: matrix(1, 0, 0, 1, 0, 0);"
                    <?php echo $scale ? 'width="'.reScale($Event->Width(), $scale).'"' : '' ?>
                    <?php echo $scale ? 'height="'.reScale($Event->Height(), $scale).'"' : '' ?>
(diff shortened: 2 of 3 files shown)


IgorA100:
@connortechnology 
I mistakenly included the commit https://github.com/ZoneMinder/zoneminder/pull/4237/commits/27dd13fe746ecb527d00e81b0392838ac182c8cd in this PR
And I don't know how to undo it :(
I wanted to make a separate PR for this commit.
What should I do?

IgorA100:
I just made another commit that reverts https://github.com/ZoneMinder/zoneminder/commit/27dd13fe746ecb527d00e81b0392838ac182c8cd
And created a new PR

connortechnology:
My only thought here is that if someone actually HAS gone and clicked in their settings to always allow playing of video with sound, this will make them have to unmute the audio in order to hear it.  

I wonder if we can detect if we have permission to auto play with sound..

IgorA100:
> and clicked in their settings to always allow playing of video with sound

Where is this setting?

connortechnology:
In firefox Settings->Privacy & Security -> Permissions -> Autoplay

IgorA100:
> In firefox

Ah, these are browser settings... I thought we were talking about ZM.
Probably possible. Although, I think 99% of users don't change anything there :)
But I think it's better to have separate settings in ZM. Only I haven't decided whether the settings should be global on the Options page or on the Event page. And where to save the settings, in the DB, cookies or sessions :)
I thought and thought about it and decided to do it without settings for now. In any case, I didn't make it worse, but now autoplay will work.

IgorA100:
If the browser is set to prohibit playback, then even now the sound will not play!
That is, "vid.muted(false);" will not turn on the sound. Therefore, there is no point in even trying to get the sound settings in the browser.

IgorA100:
Alternatively, I can add a global setting.
Options->Web->WEB_SOUND_ON_START_PLAY_EVENT
By default, it will be enabled. When enabled, everything will work as it does now.

When disabled, the event will start playing without sound. You can always enable the sound by clicking on the icon in the player.
But you, Isaac, will have to write another code to update the database :)

IgorA100:
I forgot a little how to add/remove settings :(
I thought that I needed to make an entry in the DB.
But now it seems to me that it is enough to add only an entry in "scripts/ZoneMinder/lib/ZoneMinder/ConfigData.pm.in"
Or is there something else that needs to be done?

connortechnology:
My thought is that no one will use that config entry.  It would really be better to detect if we can autoplay with audio and put up a friendly popup.

IgorA100:
> My thought is that no one will use that config entry.

Maybe. Although on the other hand - if the speakers or headphones are on and you listen to music, and simultaneously watch recorded events (especially not one, but many), then you need to press the mute button on each event. But, on the other hand, the option to turn off in the settings also requires pressing the buttons many times :)
Okay, let's not make any additional settings for now.

> if we can autoplay with audio and put up a friendly popup.

If the sound is turned off in the browser, then there will be no sound in either automatic or manual mode!
I will read if it is possible to get the state of the sound on or off in the browser..

IgorA100:
I have no more thoughts on how to improve this PR.

IgorA100:
@connortechnology 
Is there something you don't like about this PR?

connortechnology:
I don't like it because  it is not obvious why the audio isn't playing.  At least when when video doesn't play it is obvious what is going on, if you click play you get audio.  With this you don't and it isn't clear why. THe volume control in the player controls doesn't even indicate muted, which may be a bug in video.js, I don't know.  

So I think we just need to make it obvious if we can. Maybe a text overlay or just down near the buttons or something indicating what is going on. Maybe even direct them to the settings.

Further, I am someone who has set those settings to allow sound, and this disables sound for me on a refresh.  

The only thing we really need to do is somehow make it easier for them to change their settings to allow sound.

IgorA100:
You need to decide:
1. Do you need autoplay, but without sound, or
2. Do you not need autoplay.
If p.1, then where and how to place the text or icon that the sound is off?
If p.2, then this PR can be closed. Because it is not possible to combine autoplay with sound.

> THe volume control in the player controls doesn't even indicate muted, which may be a bug in video.js, I don't know.

Yes, this is a problem in video.js. The latest version doesn't seem to have this problem.
In the version of video.js used in ZM, you just click on the sound icon in the player control panel and the sound will appear.

IgorA100:
If you choose p.1, then I can do it like this:



I'm waiting for your decision, Isaac.

connortechnology:
Long term I think we actually want to move away from video.js.  We don't really need it.  So I'd say p2.

_I_ don't need autoplay, because I went in the settings next to the URL and clicked allow sound and so it just autoplays.  It sucks but thats what everyone has to do. This isn't unique to ZoneMinder. They should only ever have to do this once and it will work fine without us jumping through any hoops.

IgorA100:
> Long term I think we actually want to move away from video.js

I agree.
But the behavior of any other player will be similar, because it depends not on the player, but on the browser.

PR #4238, opened 2025-02-28
feat(tags): Filter option for No Tag

Simpler1:
Add an option to filter for events that have "No Tag".

If users have a tag with ID 0, they will need to delete that tag since ID 0 is now used to identify "No Tag".

### DIFF
--- a/web/CMakeLists.txt
+++ b/web/CMakeLists.txt
@@ -6,7 +6,7 @@ add_subdirectory(api)
 configure_file(includes/config.php.in "${CMAKE_CURRENT_BINARY_DIR}/includes/config.php" @ONLY)
 
 # Install the web files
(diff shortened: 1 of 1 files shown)


IgorA100:
Good idea!

> If users have a tag with ID 0, they will need to delete that tag since ID 0 is now used to identify "No Tag".

Isn't it possible "No Tag" = "" instead of "No Tag" = 0 ?

connortechnology:
Yeah there shouldn't be any tags == 0, but in general and for consistency, We tend to use "" in the html input value for this...

Simpler1:
Filtering for tags is not a freeform text field.  It only allows selection of existing tags.  This is the reason for needing a value for "No Tag" to appear at the top of the dropdown.

Simpler1:
Sorry, but I think we're talking past each other.  Replacing the "'0'" with "", simply doesn't find the events with no tag.  Each entry in the dropdown has an ID and a Name.  I'm just using 0 as the ID.

I've been using this in my system for about a year now and it works great.

Simpler1:
"No Tag" doesn't mean that the user explicitly entered "0" for the tag name.  It means that it has no tag assigned.  Therefore a value of "".  That's what this filter looks for.

PR #4243, opened 2025-03-03
Fix: Changed RTSP2WebStream => RTSP2WebStreamOptions on Watch page

IgorA100:
In addition to https://github.com/ZoneMinder/zoneminder/commit/5cff2f0904a01234425c5c77dad5b38a76175905


IgorA100:
@connortechnology 
This is also necessary, otherwise the Watch page will not work!
Sorry for the collision

PR #4247, opened 2025-03-05
When scaling the browser window, save the browser size in cookies (skin.js)

IgorA100:
Saved in JSON
innerWidth: window.innerWidth
innerHeight: window.innerHeight
outerWidth: window.outerWidth
outerHeight: window.outerHeight

Request: #4228

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -701,7 +701,7 @@ function endOfResize(e) {
     }
   }, 250);
 }
-$j(window).on('resize', endOfResize);
+window.onresize = endOfResize;
 
 /* scaleToFit
  *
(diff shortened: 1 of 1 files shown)


web/skins/classic/js/skin.js
```diff
@@ -700,7 +710,7 @@ function endOfResize(e) {
  * scaleEl is the thing to be scaled, should be a jquery object and should have height
  * */
 function scaleToFit(baseWidth, baseHeight, scaleEl, bottomEl, container, panZoomScale = 1) {
-  $j(window).on('resize', endOfResize); //set delayed scaling when Scale to Fit is selected
```
connortechnology, line 703: Does this not mean that we never call endOfResize?  This stuff should happen regardless of scaleToFit functions...

IgorA100: > Does this not mean that we never call endOfResize?

No

Previously, the assignment of "onresize" to the "endOfResize" function occurred after the call to "scaleToFit".
Now the assignment occurs regardless of "scaleToFit".

PR #4249, opened 2025-03-06
fix(filters): Remove the ID from the filter names in the dropdown

Simpler1:
Simplify the UI slightly by removing the ID from the filter name.

### DIFF
--- a/web/skins/classic/views/filter.php
+++ b/web/skins/classic/views/filter.php
@@ -43,7 +43,7 @@ if ( isset($_REQUEST['Id']) and $_REQUEST['Id'] ) {
 }
 $filter = null;
 foreach ( ZM\Filter::find(null,array('order'=>'lower(Name)')) as $Filter ) {
-  $filterNames[$Filter->Id()] = $Filter->Id() . ' ' . $Filter->Name();
+  $filterNames[$Filter->Id()] = $Filter->Name();
   if ( $Filter->Background() )
     $filterNames[$Filter->Id()] .= '*';
   if ( $Filter->Concurrent() )


IgorA100:
Why delete useful information?

Simpler1:
Why would an ID be useful information for a named filter?  This is internal data that is not useful to a user.  It's just clutter.

IgorA100:
> This is internal data that is not useful to a user

For example, you pressed the "Save as" button, but did not enter a new name. There will be two filters with the same name. How can you tell which was the last one?
Or, for example, to debug filters, it is useful to have a visual comparison of the number and name. Especially when there are many filters.
Surely someone once needed this, since there is code that is not erroneous.

I just expressed my opinion.
In any case, only Isaac makes the decision.

connortechnology:
I find the ids to be useful, and indeed I am the one that put the id in the dropdown.  When looking at processes and logs, the name is not present and not useful.  Hence the id.  I agree it isn't great UI for non power users, but for this power user it is very useful.  If we had a filter list view then maybe it wouldn't be so important.

Simpler1:
For debugging, it makes sense to show the filter ID.  For general usage, it causes confusion.  Consider the following two images:
Screenshot from 2025-03-10 09-15-48
Screenshot from 2025-03-10 09-13-24

If you don't want to add this, you can just close this PR.

IgorA100:
> it causes confusion. 

1. You yourself decided to use the ">" sign in the filter name.
2. Alternatively, the ID can be framed with quotation marks or brackets or something else... If Isaac doesn't mind.

connortechnology:
I'm not here to be s tyrant any more than i have to be. We could make it configurable or maybe with further discussion we can come with something better

connortechnology:
How about we do either (id) or id: (should make this consistent with monitor ids shown elsewhere).  Also can put it in a span or something so that we can hide it with css.

connortechnology:
Closing this as I implemented a config entry for it. See 6d3b7ddebcdc8392aad3c9dcc2b4c67306a2ac9b

PR #4253, opened 2025-03-10
feat(tags): Filter option for No Tag

Simpler1:
I moved this to a separate branch so that I could continue other work while we discuss any concerns.

Being able to filter for events that don't have a tag is very useful functionality.

The "0" is not the name of the tag, it's just the first tag in the dropdown list.


Simpler1:
This is an indexed array (not an associative array).  With the change that you describe, it now has an index of [] instead of [0].  There is no way check for an index of "blank".

How do you propose to do the check?

Here's what the indexed array looks like with your proposed change:
(
    [] => No Tag
    [86] => Delete
    [2] => Car
)

PR #4259, opened 2025-03-15
Feat: Added the ability to save monitor data without closing the page when editing a monitor

IgorA100:
And also added the appearance of a fashionable window with a request to save data when switching to zone management pages.


IgorA100:
@connortechnology 
Closed https://zoneminder-chat.slack.com/archives/CC2T6P1KQ/p1741275291089969

IgorA100:
As far as I understand, when switching from Monitor page to Zone page using Java, the referer attribute is not lost. Although it should be lost and the "<-" button should not work, but it does.
Please check this point.

connortechnology:
Works well. One thought is that when going to Zones, if nothing was changed, we don't need the warning. 

On another project I had input elements track their initial value so it was easy to check for changes..  Could also set a changed flag on the form element.  Makes it easy to check.

Eventually perhaps zones shouldn't actually leave the monitor edit. Something to think about long term.

IgorA100:
Maybe it is necessary to implement this for all forms?
I think the idea with seralization is very interesting...

IgorA100:
PR #4277

PR #4264, opened 2025-03-16
Fix: Call "getBodyTopHTML" instead of the <body> tag (video, stats, controlcap, export, error, login, optionhelp, files, privacy, control, monitorpreset, frames, timeline, postlogin, cycle, bandwidth, plugin)

IgorA100:


IgorA100:
@connortechnology 
I checked all the changes, everything works correctly.
Except for "plugin" & "timeline", because I don't know how to call them yet :( But in these two files the changes should work correctly.

PR #4267, opened 2025-03-17
Fix: If "eid" is not passed, then you need to assign an empty value to the variable (frames.js.php)

IgorA100:
Otherwise there will be an error in JS and the script will stop.


web/skins/classic/views/js/frames.js.php
```diff
@@ -1,3 +1,3 @@
-var eid = <?php echo validInt($_REQUEST['eid']) ?>;
```
connortechnology, line 1: empty string isn't sufficient, needs to be '0'; Or "''"

IgorA100: This PR inserts exactly "''" :)

connortechnology: No it doesn't.  You would have to write it as " '' " Or` '\'\''`

IgorA100: Sorry Isaac, the quotation marks got lost somewhere and I didn't notice...
I fixed it.

Issue #4272, opened 2025-03-25
Stream Quality option should only show if a secondary stream is available


connortechnology:
If nothing has been entered for the Secondary stream url, then we shouldn't give it the option.

IgorA100:
Agreed.
I'll get to it after I finish the sidebar menu..

PR #4275, opened 2025-03-25
Fix: Stream Quality option should only show if a secondary stream is available

IgorA100:
Closed https://github.com/ZoneMinder/zoneminder/issues/4272


IgorA100:
But there is one inconvenience.
When editing the monitor, "Viewing->Stream source" is not updated in real time when changing "Source->SourceSecondPath"
Maybe leave the settings as they were before? Allow selecting any stream regardless of the filled in SourceSecondPath value?

IgorA100:
I will make the change to make the choice available in real time, but a little later....

IgorA100:
Done.
You can check.
The "Select" tag always has two options: Primary & Secondary, but if the monitor's "SecondPath" is not filled in, the selection line will be inactive. I think this is a more correct option.
Now, when changing the "SecondPath" value on the monitor edit page, the Primary/Secondary stream selection item will automatically become active or inactive.

Be sure to check this PR before approving, because I have a lot of changes related to the new menu in the left Sidebar and many of my PRs have not yet been approved. Due to this, some of the code may not have been transferred or, on the contrary, something unnecessary may have gotten in.

IgorA100:
@connortechnology 
Please, if possible, when you have free time, consider this PR,
Since the function initPageGeneral() is added here, I cannot make other PRs that will use this function.

PR #4279, opened 2025-03-31
Moved panZoom button control code to global scope from watch.js to skin.js with touch support

IgorA100:
New function managePanZoomButton(event) in skin.js

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -1422,6 +1422,82 @@ function handleChangeInputTag(evt) {
   manageRTSP2WebChannelStream();
 }
 
+/* Handle any action on the touch screen */
+function handleTouchActionGeneral(action, evt) {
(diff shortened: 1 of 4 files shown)


web/skins/classic/js/skin.js
```diff
@@ -1513,7 +1504,7 @@ function initPageGeneral() {
   });
 
   // Support for touch devices.
-  "touchstart touchend touchcancel touchmove".split(" ").forEach(function(action){
+  "touchstart touchend touchcancel touchmove".split(" ").forEach(function(action) {
```
connortechnology: Why not just ['touchstart','touchend' etc].forEach?

IgorA100: I used to use it like that and got used to it, I guess...
It was a long time ago, I don't remember why I did it like that...
Edited by:https://github.com/ZoneMinder/zoneminder/pull/4279/commits/b47292a4c3ebb6b9075f108b08c5cbc58f80ef70

PR #4282, opened 2025-04-01
Feat: Smoothly collapse main content when leaving the page (skin.js)

### LINKED ISSUE #4273: boken images look bad
Say when we go from montage to console by click on the word console.  We kill all the streams, which is good because otherwise it can take a while for the browser to actually stop streaming itself and go to the console.

However for a brief second we see a bunch of broken images before the page changes.

Maybe when stopping these streams we should replace them with a 1 pixel black image or something instead.

IgorA100:
Closed #4273


IgorA100:
Conflict will be resolved after approval #4279

PR #4284, opened 2025-04-02
Fix: When connection is lost using RTSP2Web type==WebRTC, restart the stream (MonitorStream.js)

IgorA100:


web/js/MonitorStream.js
```diff
@@ -1134,6 +1134,10 @@ const waitUntil = (condition) => {
 
 function startRTSP2WebPlay(videoEl, url, stream) {
   const mediaStream = new MediaStream();
+  if (stream.webrtc) {
```
IgorA100, line 1137: This change is not directly related to the created PR, but I decided not to replicate the PR, but to make a commit. Because this fix is ​​related to one object RTSP2Web type WebRTC

PR #4292, opened 2025-04-04
Fix: Instead of "streamPlay" call "streamPause" when initializing the page (Event.js)

IgorA100:
Because playback may not start automatically!
Further, the player buttons are controlled by video.js events


IgorA100:
> have the settings to allow sound, the video plays, with audio, but the play button indicates that it is paused.

Got it. I'll fix it soon.

IgorA100:
Ready #4298

PR #4293, opened 2025-04-05
Fix: Avoid double monitor start when initializing a Montage page

IgorA100:
1. Do not execute "on_scroll" until all monitors are positioned on the screen. Otherwise, the monitors will be started twice when the page is initialized. The first time is when on_scroll() is called from selectLayout(), the second time is started from startMonitors() after all monitors are positioned.

2. When executing startMonitors(), you cannot change the SRC attribute of the IMG tag before executing "monitor.start()", because in monitor.start() we have an assignment of this.img_onload. That is, the "onload" event can occur earlier (due to the SRC change) than we assign a function to it!

Scaling occurs in monitorStream

### DIFF
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -537,16 +537,6 @@ function handleClick(evt) {
 function startMonitors() {
   for (let i = 0, length = monitors.length; i < length; i++) {
     const monitor = monitors[i];
-    // Why are we scaling here instead of in monitorstream?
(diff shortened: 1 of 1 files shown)


connortechnology:
I think for img.src, we need to always maintain a var with what it will be and then assign it to the img.  There is an issue with timeouts as well that may be addressed by this.

IgorA100:
> I think for img.src, we need to always maintain a var with what it will be and then assign it to the img. There is an issue with timeouts as well that may be addressed by this.

Probably yes.

PR #4299, opened 2025-04-08
Chore: Minor code optimization on Montage page

IgorA100:


connortechnology:
In case you havn't seen it: https://forums.zoneminder.com/viewtopic.php?p=137057&sid=49d15ef308517f4c6bcac5eac6c73278#p137057

IgorA100:
@connortechnology 
I saw this topic.
Indeed, there is a problem only with Firefox
I will fix it soon.

PR #4301, opened 2025-04-09
Fix: Problem with long page opening in Firefox Montage page

IgorA100:
Returned previously deleted code.
Closing the issue: https://forums.zoneminder.com/viewtopic.php?p=137059#p137059 We need to figure out what's wrong with scaling!!!

### DIFF
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -553,7 +553,7 @@ function startMonitors() {
       console.log(`startMonitors NOT FOUND ${'liveStream'+monitor.id}`);
     }
     /* --- */
-		const isOut = isOutOfViewport(monitor.getElement());
+    const isOut = isOutOfViewport(monitor.getElement());
     if (!isOut.all) {
       monitor.start();
     }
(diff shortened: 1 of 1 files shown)


web/skins/classic/views/js/montage.js
```diff
@@ -538,6 +538,21 @@ function handleClick(evt) {
 function startMonitors() {
   for (let i = 0, length = monitors.length; i < length; i++) {
     const monitor = monitors[i];
+    // Why are we scaling here instead of in monitorstream?
```
connortechnology, line 541: I think the issue is, on first load, the img tags are not scaled, so it will be streaming full res images.  This does an initial scale so that they start up somewhat scaled. 

Now that we have window sizes in the cookies, we should be able to pre-scale in php side.  

Also, I am changing my thoughts about initially loading a single-shot image and then switching to streaming.  I am leaning more towards just starting up streaming.

IgorA100: Ok, Isaac, I got you.
In addition to the browser window size, you also need to know the grid. When the grid is standard (2, 4, 8... Wide) everything is relatively simple. When the grid is custom, one image can be 900px wide, and another 300px. You will probably need to get the grid from the database. In general, you need to think...
I'll probably get into this, but later.
I have a lot of work to do right now and I want to finish the left sidebar. It is almost ready, it requires a little optimization of the code and CSS.
Let it remain as it was for now. Because in the coming days I will be busy.

PR #4302, opened 2025-04-10
Feat: Material icons styles in a separate file from skin.css to material-icons.css

IgorA100:
This will allow us to load this style first.
That is, on a slow Internet channel or a slow browser, icons may load with a delay, which does not look nice.

### DIFF
--- a/web/skins/classic/css/base/material-icons.css
+++ b/web/skins/classic/css/base/material-icons.css
@@ -1,0 +1,48 @@
+@font-face {
+  font-family: "Material Icons";
+  font-style: normal;
+  font-weight: 400;
+  font-display: block;
+  src: url("../../../../fonts/material-icons.woff2") format("woff2"), url("../../../../fonts/material-icons.woff") format("woff");
+}
+.material-icons {
+  font-family: "Material Icons";
+  font-weight: normal;
+  font-style: normal;
+  font-size: 24px;
+  line-height: 1;
+  letter-spacing: normal;
+  text-transform: none;
+  display: inline-block;
+  white-space: nowrap;
+  word-wrap: normal;
+  direction: ltr;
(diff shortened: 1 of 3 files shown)


connortechnology:
I look forward to testing this, I am surprised that it makes that much of a difference

IgorA100:
> I look forward to testing this,

Yes, definitely test it.
I liked it.

IgorA100:
> On my test server (pi on a very slow connection) it takes the same amount of time to load, regardless of this. 16-17s. Very painful.

Can you make a video from the beginning load to the end?
I'll think about what can be done.

connortechnology:
Eventually I can make a video. Busy now.  Best thing to do is shrink the material icons file by only including the ones we use.

web/skins/classic/css/base/material-icons.css
```diff
@@ -0,0 +1,48 @@
...
+  display: inline-block;
+  white-space: nowrap;
+  word-wrap: normal;
+  direction: ltr;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  text-rendering: optimizeLegibility;
+  font-feature-settings: "liga";
+  vertical-align: middle;
+  display: none;
```
IgorA100, line 25: @connortechnology 
I probably know how to solve your problem....
Try deleting the line "display: none;"
Now our CSS is loaded almost at the beginning. Perhaps we will either abandon loadFontFaceObserver or reconsider how to optimally call it

IgorA100:
> Best thing to do is shrink the material icons file by only including the ones we use.

This is one of the points on loading optimization. I remember you told me about it. I thought a little and tried something, but since I didn't succeed, I put it off :(
Nevertheless, icon fonts should be loaded as early as possible.

PR #4303, opened 2025-04-10
Fix: "Pale moon" browser does not understand overflow: hidden auto; (skin.css)

IgorA100:
I'm shocked by this......

In "sticky" mode, some pages were missing vertical scrollbar
I understand that "Pale moon" browser is quite rare and exotic, but nevertheless. Maybe someone uses it or similar ones.


connortechnology:
That is shocking, thankfully the fix is simple.

PR #4305, opened 2025-04-10
Feat: Font "Material Icons" load and display as soon as possible

IgorA100:
Now "FontFaceObserver" will be called first
In addition to #4302

### DIFF
--- a/web/skins/classic/includes/functions.php
+++ b/web/skins/classic/includes/functions.php
@@ -65,6 +65,24 @@ if (defined('ZM_WEB_FAVICON')) {
 ';
 }
 echo output_link_if_exists(array('css/base/material-icons.css'));
+echo output_script_if_exists(array(
+  'js/fontfaceobserver.standalone.js',
+));
+
+?>
+  <script nonce="<?php echo $cspNonce; ?>">
+    const fontMaterialIcons = new FontFaceObserver("Material Icons", {weight: 400});
+    fontMaterialIcons.load(null, 30000).then(function() {
+      console.log("Material Icons is loaded");
+      var _style_ = document.createElement('style');
+      _style_.innerHTML = `.material-icons {display: inline-block !important;}`;
+       document.querySelector('head').prepend(_style_);
+    }, function() {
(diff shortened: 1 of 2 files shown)


IgorA100:
@connortechnology 
Try it, you might like it.
On a slow channel, the font used to appear after 30 seconds, now after 20 seconds.

connortechnology:
https://stackoverflow.com/questions/36259752/fastest-way-to-load-a-web-font-for-a-website

connortechnology:
tried putting a <style> in the header had no effect.

IgorA100:
@connortechnology 
Try again after my commit https://github.com/ZoneMinder/zoneminder/pull/4305/commits/e3acc2a2d6725ce0ff1f68efda45be3fa8c17f03
There is nothing else you can speed up, because most of the time is spent loading the file "material-icons.woff2" It is loaded at the very beginning, then there is a small delay for the server to respond, and then as soon as the font file is loaded, it is immediately displayed on the screen.
This is the fastest and most correct loading option.

IgorA100:
> I wonder if we used an inline style to load it... instead of a .css

This will speed up loading by only 10-15% (this is not much and will not really be noticeable. But I don’t know yet how to synchronize with FontFaceObserver in this case

IgorA100:
@connortechnology 
I also sped up the font loading to the maximum possible.
It won't be possible to speed it up any more, and I'm 100% sure of that. Unless you reduce the file size.
Try it.

connortechnology:
That's pretty good, it now loads the font first.  Still takes 13s total but that's not our fault.  I'm actually pretty impressed that our payload is ony 594kb.  We could maybe pre-compress some of the js and css like bootstrap etc.. maybe merged what we use into 1 file.. not sure it's worth it.

IgorA100:
> We could maybe pre-compress some of the js and css like bootstrap etc.. maybe merged what we use into 1 file.. not sure it's worth it.

Yes, I had such a thought. Maybe I will do it, but I don't know when...

PR #4306, opened 2025-04-11
Fix: Broken style on Timeline page

IgorA100:
The problem appeared after #4268
Closed https://forums.zoneminder.com/viewtopic.php?p=137074#p137074

### DIFF
--- a/web/skins/classic/views/timeline.php
+++ b/web/skins/classic/views/timeline.php
@@ -676,7 +676,7 @@ xhtmlHeaders(__FILE__, translate('Timeline'));
 getBodyTopHTML();
 echo getNavBarHTML();
 ?>
-  <div id="page p-0">
+  <div id="page" class="p-0">
     <div class="d-flex p-1">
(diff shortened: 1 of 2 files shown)


web/skins/classic/views/timeline.php
```diff
@@ -676,7 +676,7 @@ function drawYGrid( $chart, $scale, $labelClass, $tickClass, $gridClass ) {
 getBodyTopHTML();
 echo getNavBarHTML();
 ?>
-  <div id="page p-0">
```
IgorA100, line 679: Oooo... It was broken back on August 24, 2020 :))))))
https://github.com/ZoneMinder/zoneminder/commit/0879ecf9a5f3118597264cf68a3b4133956e58da

connortechnology: Not many people evenknow timeline view exists, let alone use it!

IgorA100: Yes, I agree.
I myself went to Timeline only for the third time in a year..

PR #4309, opened 2025-04-12
Feat: Added "preload" for "fontawesome-webfont.woff2" (functions.php)

IgorA100:
When updating the file "css/font-awesome.min.css" you MUST check what path to "fontawesome-webfont.woff2" is specified, as well as the version after the "?" If necessary, make changes to this file. Otherwise, "preload" will not work.

Similar to "Material Icons" https://github.com/ZoneMinder/zoneminder/pull/4305


connortechnology:
I don't know what the current state of affairs is, but in the past, browsers would not cache urls with a ? in them.  So it would be best to not have ? in it.

Also, we really should sync on just one icon set instead of two.

IgorA100:
> I don't know what the current state of affairs is, but in the past, browsers would not cache urls with a ? in them.

I don't know how things are now either. :(

> So it would be best to not have ? in it.

It doesn't depend on me, it's specified in the standard font-awesome.min.css. I didn't change anything.

> Also, we really should sync on just one icon set instead of two.

Sure. But honestly, I don't know what's better Font Awesome or Material Icon :(

IgorA100:
FF, Chrome and Chromium cached. I haven't tested it on other browsers.

PR #4310, opened 2025-04-14
Fix: Firefox return to previous page (skin.js)

IgorA100:
Firefox has a feature of caching and returning to the previous page. Moreover, in different versions of FF the behavior is different. Added by listener pageshow
Closed https://forums.zoneminder.com/viewtopic.php?p=137084#p137084

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -41,6 +41,7 @@ var icons = {
 var panZoomEnabled = true; //Add it to settings in the future
 var expiredTap; //Time between touch screen clicks. Used to analyze double clicks
 var shifted = ctrled = alted = false;
+var mainContent = document.getElementById('content');
 
 function checkSize() {
   if ( 0 ) {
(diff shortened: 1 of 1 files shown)


web/skins/classic/js/skin.js
```diff
@@ -1527,6 +1528,13 @@ function initPageGeneral() {
     alted = e.altKey;
   });
 
+  if (['montage', 'watch', 'devices', 'reports', 'monitorpreset', 'monitorprobe', 'onvifprobe', 'timeline'].includes(currentView)) {
+    mainContent = document.getElementById('page');
+  } else if (currentView == 'options') {
+    mainContent = document.getElementById('optionsContainer');
+  }
+  var mainContentJ = $j(mainContent);
```
IgorA100, line 1536: We use two mainContent objects. The first "mainContent" - globally (using vanilla JS)
The second "mainContentJ" - locally as a JQ object, because it is a little easier to work with.
If all goes well, then I will probably rewrite the code without using JQ, only in vanilla JS.
I don't want to do this now, so as not to break anything else.
Ideally, we need to move to a single standard, and not use different main content identifiers `#page, #options, #content `for different pages. But this is quite a long and painstaking work for the future...

PR #4311, opened 2025-04-15
Feat: Added left side menu

### LINKED ISSUE #4077: Replace the top menu with a side menu
I propose to discuss the option of adding a side menu like here: https://www.bootstrap.gallery/demos/milestone-admin-dashboard/index.html#

Sources: https://www.bootstrap.gallery/details/milestone---bootstrap-admin-panel

IgorA100:
Closed #4077

### DIFF
--- a/web/skins/classic/includes/functions.php
+++ b/web/skins/classic/includes/functions.php
@@ -251,6 +251,14 @@ function buildSidebarMenu() {
               </a>
             </li>
             <li class="menu-item">
+              <a href="https://zmninja.zoneminder.com/" target="_blank">
+                <span class="menu-icon">
+                  <i class="material-icons">face_5</i>
+                </span>
+                <span class="menu-title">' . translate("zmNinja") . '</span>
+              </a>
+            </li>
+						<li class="menu-item">
               <a href="https://wiki.zoneminder.com/" target="_blank">
                 <span class="menu-icon">
                   <i class="material-icons">article</i>


IgorA100:
@connortechnology 
Haven't you tested this PR yet or didn't you like it?

connortechnology:
I do not have time to look at it.  It will not go in 1.38. It will get merged likely immediately after 1.38 is released, as 1.39.0

IgorA100:
@connortechnology 
I consider this PR complete.

connortechnology:
I'm not doing so well at getting 1.38 out, so I might take a look at this over the next couple days with an eye to merging it before 1.38.  Stay tuned.

IgorA100:
Requires addition from https://github.com/ZoneMinder/zoneminder/commit/762ee7a2a3bb607236a8ace6a3856d512c07806c
But, it will be implemented after approval (if any) of this PR.

IgorA100:
@connortechnology Haven't you looked at this PR?

connortechnology:
I did, it didn't apply cleanly to the branch I was working on.  I do think I will try to get it in soon before 1.38 release.

IgorA100:
> I did,

Did you like it or do you have any suggestions?

connortechnology:
What are the X buttons for in the filter menu?  Are they for when not using chosen? I guess if you had multiple things chosen it would be nice to be able to clear them all...  which explains why they don't appear for the text inputs... Maybe give them a title like "Clear all" so that it is clear on mouse over.

I see no reason to have a use old/new menu toggle in the main UI.  Move it to options -> Web. 

Also, there is a thing about the words Old/New.  They connotate old=busted,bad vs new=good. There is an implied value statement.  I would prefer a more accurate less judgey term like Top Menu / Left Menu.  Also, some people might want both.  It shouldn't be an either/or.

connortechnology:
I'm not sure why things are going in assets.  WHy another dir to contain third party stuff.  We now have 3.  I'd kinda like to have that cleaned up.  

I like all the link to docs, zmNinja etc, but some people may want to remove that to save space.  We have an option for that in th eold header to remove content from home logo button.  

ALso support for my new NAVBAR_LINKS stuff.

IgorA100:
> What are the X buttons for in the filter menu?

For "Clear all"

> Maybe give them a title like "Clear all" so that it is clear on mouse over.

Yes. I will do that.

> I see no reason to have a use old/new menu toggle in the main UI. Move it to options -> Web.

I did it for quick switching. Okay, I'll move it to "options -> Web"

> like Top Menu / Left Menu

Okay, I'll rename it.

> Also, some people might want both. It shouldn't be an either/or.

So I need to make such a choice in the options?
- Top Menu
- Left Menu
- Top and Left menu
Did I understand you correctly?

IgorA100:
> I'm not sure why things are going in assets. WHy another dir to contain third party stuff. We now have 3. I'd kinda like to have that cleaned up.

These are three third-party plugins used to create the menu
/assets/mb.extruder/
/assets/pro-sidebar-template/
/assets/swiped-events/
I don't understand yet what I need to do :(

> I like all the link to docs, zmNinja etc, but some people may want to remove that to save space. We have an option for that in the old header to remove content from home logo button.

All this is located almost at the very bottom of the menu. The menu has correct scrolling, including on mobile devices.
It seems to me that these items do not take up extra space and do not interfere at all. I have used this menu for a long time on various devices and I was comfortable. But if you insist, then I can make an option to disable these menu items or use the existing HOME_ABOUT option
What should I do?

> ALso support for my new NAVBAR_LINKS stuff.

Ok.

connortechnology:
I do not insist, you asked for comments.  I gave comments.  We can wait for people to complain about the links, but for my pro users I will have to hide them.  

I have merged this, it looks good enough as is, we can now get comments from users.

IgorA100:
> I gave comments.

Thank you! I am really interested in this.

> but for my pro users I will have to hide them.

Why do you need to do this? Since it is critical for you, I can analyze the "HOME_ABOUT" setting and hide or show additional links.
Do you agree?

PR #4312, opened 2025-04-16
Fix: Instead of switch "( $max_aspect_ratio )" use a formula for recalculation aspect ratio on Timeline page

IgorA100:
In general, we need to somehow calculate the values ​​through a formula.
But I still haven't figured out which formula to use....

### DIFF
--- a/web/skins/classic/css/base/views/timeline.css.php
+++ b/web/skins/classic/css/base/views/timeline.css.php
@@ -30,6 +30,9 @@
 .imageHeight {
 <?php
 switch ( $max_aspect_ratio ) {
+  case 0.56: // 1080x1920, 1520x2688
+  case 0.57:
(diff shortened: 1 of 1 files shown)


connortechnology:
Maybe there is a better way.  This was done to center the video I think... it was a long time ago. Maybe a fresh approach to layout could do away with it.

IgorA100:
Can be done like on Montage page using gridstack.js

IgorA100:
I looked at the Timeline page code.
It needs a lot of changes to integrate correctly with gridstack.js
I don't have time to do it yet :(

IgorA100:
@connortechnology 
The conversion formula was very simple :)
It was just a matter of doing (100/$max_aspect_ratio) and rounding, rather than specifying specific values ​​for each aspect ratio.

PR #4318, opened 2025-04-29
feat(header): Chevron right for collapsed - chevron down for expanded

Simpler1:
Simple change to make the chevrons easier to understand when the main-header-nav is collapsed or expanded.

The right chevron indicates collapsed.
The down chevron indicates expanded.

(Reference: https://stackoverflow.com/questions/18325779/bootstrap-3-collapse-show-state-with-chevron-icon)


connortechnology:
@IgorA100 opinion on this?

IgorA100:
> opinion on this?

To be honest, after this PR it becomes even more unclear :)
Perhaps it really is worth thinking about changing the icons, but I don't really like the option with "keyboard_arrow_right".
Perhaps I'm too picky.

As an option - maybe replace with collapse_all + expand_all
In general, the entire top horizontal menu needs to be redesigned. Make it more convenient for narrow screens and mobile devices.

PR #4320, opened 2025-04-30
Dark mode

Simpler1:
A simple css change to make the borders not so bright when in dark mode.


Simpler1:
I need to isolate this change and then resubmit this PR

PR #4322, opened 2025-05-01
Replace old style layouts with new style layouts + added "24 Wide" & "48 Wide" Layouts (montage page)

### LINKED ISSUE #4315: MontageLayouts need cleanup
On a fresh install we are still including the default old-style layouts.  That needs to be cleaned up, and the invalid (not a divisor of 48) removed.

Update scripts should do this job as well.

IgorA100:
New style layouts were implemented in:
https://github.com/ZoneMinder/zoneminder/commit/99e87a9c1c705042e4abc02dbbb035428e88c70e https://github.com/ZoneMinder/zoneminder/pull/3969

I haven't tested it, but it should work.
Please check it.

Closed #4315

### DIFF
--- a/db/zm_create.sql.in
+++ b/db/zm_create.sql.in
@@ -1194,16 +1194,14 @@ CREATE TABLE MontageLayouts (
   PRIMARY KEY (`Id`)
 );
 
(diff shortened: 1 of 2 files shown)


IgorA100:
Done.
But I haven't checked!!!
Although everything should work correctly.

PR #4327, opened 2025-05-20
Fix deprecated AVCodecContext::channel_layout to use ch_layout

faginbagin:
Moving from ubuntu jammy with libavcodec 58.134.100 to ubuntu noble with libavcodec 60.31.102 broke the ability to capture events using some cameras whose audio streams did not use the AAC codec. This fixes the problem by using the newer AvCodecContext::ch_layout.

For more details, see forum post: https://forums.zoneminder.com/viewtopic.php?t=33900


faginbagin:
This pull request fixes issue: https://github.com/ZoneMinder/zoneminder/issues/4328

PR #4342, opened 2025-06-27
Fix: Don't limit stream height and round parent height to integer Montage page

### LINKED ISSUE #4340: Ever increasing image size on montage when image stream fails to load
I think a network issue caused the fail to load, will need to look into how to simulate it.

While it was failing... the image height/width constantly increased by small amounts.

IgorA100:
Closed #4340


IgorA100:
I hope everything is okay now.

IgorA100:
@connortechnology 
Please test it longer before approving.
But it seems to me that there are no side effects or problems now.

IgorA100:
I had no problems with this PR. Everything works correctly.

connortechnology:
This definitely improves the behaviour.  CUrrently when streams go off-view, they can vertically expand forever after the stream has been stopped.  With this they do a little expand/contract then settle.   Ideally stopping a stream will have no effect on the size but hey, this is better.

IgorA100:
> With this they do a little expand/contract then settle.

Hmm, that's weird, I don't see this happening on my device. Maybe it's related to some settings?
I'd like to fix this, but I don't know how to reproduce the problem.

PR #4347, opened 2025-07-02
Now we build the Options submenu in only one place and using the "ZM_Menu" class

IgorA100:
@connortechnology 
I decided to make a separate file for greater clarity.
But then I thought, maybe it would be better to place the function in the file '\skins\classic\includes\functions.php'?
I'm waiting for your decision.


connortechnology:
Most of these could be generated from the db Config table.  THere are only a couple that aren't present there.  I am tempted to go with functions.php, and also look at generating them from the already loaded in ram zm_config variable.

IgorA100:
> I am tempted to go with functions.php, and also look at generating them from the already loaded in ram zm_config variable.

OK.

IgorA100:
> zm_config variable

I looked and did not find the necessary information in "$zm_config" to build a submenu for "Options". There are only constant values.
Probably it is stored in some other variable.
The information I need seems to be in "ConfigData.pm.in", and it is also somewhere in the PHP code?
But there is little information here.
For example, I can use the value in "category" to build a submenu, but there is no data here on what name of the 'system' section should be displayed in the UI
```
{
name => 'ZM_STATS_UPDATE_INTERVAL',
default => '60',
description => 'How often to update the database statistics',
help => q`
The zmstats daemon performs various db queries that may take
a long time in the background.
`,
type => $types{integer},
category => 'system',
},
```

IgorA100:
It would be possible to take the category names from "ConfigData.pm.in" and display them in the UI almost without changes (replace only the first letter with a capital letter), but it's not that simple...
For example, the category "medband" should be displayed as 'MediumBW'
`$tabs['medband'] = translate('MediumBW');`
I don't know what to do yet....

connortechnology:
In the options view, we get all that from the db.  You can just query the db.  It is already loaded in order to show the options.

connortechnology:
See line 160.

IgorA100:
Ok, but how can I get the category names for the Options section so that these names can be displayed in the UI?
I didn't find such information in the DB

IgorA100:
Maybe we are talking about different things?
I need to get a list to build a submenu for Options
This list should be in one place. Now it is duplicated in two different PHP files.
```
      $tabs = array();
      if (!defined('ZM_FORCE_CSS_DEFAULT') or !defined('ZM_FORCE_SKIN_DEFAULT'))
      $tabs['skins'] = translate('Display');
      $tabs['system'] = translate('System');
      $tabs['auth'] = translate('Authentication');
      $tabs['config'] = translate('Config');
      if (defined('ZM_PATH_DNSMASQ_CONF') and ZM_PATH_DNSMASQ_CONF) {
        $tabs['dnsmasq'] = translate('DHCP');
      }
      $tabs['API'] = translate('API');
      $tabs['servers'] = translate('Servers');
      $tabs['storage'] = translate('Storage');
      $tabs['web'] = translate('Web');
      $tabs['images'] = translate('Images');
      $tabs['logging'] = translate('Logging');
      $tabs['network'] = translate('Network');
      $tabs['mail'] = translate('Email');
      $tabs['upload'] = translate('Upload');
      $tabs['x10'] = translate('X10');
      $tabs['highband'] = translate('HighBW');
      $tabs['medband'] = translate('MediumBW');
      $tabs['lowband'] = translate('LowBW');
      $tabs['users'] = translate('Users');
      $tabs['groups'] = translate('Groups');
      $tabs['control'] = translate('Control');
      $tabs['privacy'] = translate('Privacy');
      $tabs['MQTT'] = translate('MQTT');
      $tabs['telemetry'] = translate('Telemetry');
      $tabs['version'] = translate('Versions');
```
12q

connortechnology:
It is in the Category column.  Not all tabs are represented there, but they are the special cases.

IgorA100:
> It is in the Category column

Yes, but!
The category says 'medband', but it needs to be displayed in the UI as 'MediumBW'
I couldn't find any mappings 'medband' <=> 'MediumBW' anywhere except in PHP files
`$tabs['medband'] = translate('MediumBW');`

Also, how can I determine the order of categories in the menu?
How do I know that, for example, "Config" should be after "Authentication"?

connortechnology:
Well, we could put the order in the db or just go alphabetically.

 I think the medband=>"medium BW" should be done in translations.

IgorA100:
> Well, we could put the order in the db

That would be great!
Can you do it?

> I think the medband=>"medium BW" should be done in translations.

I guess I agree, it makes sense!

IgorA100:
> or just go alphabetically

It is possible, but probably many users have already gotten used to the order in which the categories are currently arranged and will not be happy. Or maybe not...

connortechnology:
That is a good point.  I think we should just hard code the order in functions.php.  Putting it in the db is over complicating things.

IgorA100:
Okay, I'll think about it...

IgorA100:
Isaac, look what I did.
I added the ability to safely call the "mb_ucfirst" function. It should work on any version of PHP, if 'mbstring' is not loaded, then "ucfirst" will be called
In the future, I think it will be more correct to move all the menu building functions to the "ZM_Menu" class
I liked this solution. For example, if some new item is added to "ConfigData.pm.in", but for some reason it is not specified in "$categoryDisplayOrder", then it will still be present in the Options submenu, but only at the very end.
Waiting for your comments.

IgorA100:
@connortechnology 
Do I need to finish this PR or are you not satisfied with something?
Basically, all that remains is to add the translation and everything is ready.

connortechnology:
Looks fine to me.

IgorA100:
I think this PR is finished.
Please check the correctness of the display for Top & Left menu.

connortechnology:
quick testing here, the Display (skins) tab doesn't work so I can't enable left menu...

IgorA100:
> the Display (skins) tab doesn't work

How exactly does not work?
The inscription `"There are no config entries for Category skins."` or ... ?

IgorA100:
@connortechnology 
I made a small correction. Replacing "skins" with "display".
Now everything should work.
I haven't taken into account your commit
https://github.com/ZoneMinder/zoneminder/commit/225b428592a4fba860a5897445174603f5a31bbb

IgorA100:
@connortechnology Have you checked? Everything should work correctly.

PR #4356, opened 2025-07-13
Fix unnecessary zmc exit caused by occasional frame errors.

Marine-Mike:
The error counter in zm_jpeg is never reduced so set a threshold of acceptable frame loss.


connortechnology:
I'm not sure I see the value in dying at all due to jpeg errors...  oh well, I've never actually seen this code get triggered.

PR #4359, opened 2025-07-15
Use variables to specify color (CSS)

IgorA100:


IgorA100:
For now, let it be so, but in general, there is a lot of work to do to put the CSS files in order, so that it would be possible to quickly change the color scheme without rewriting many colors in different places.

PR #4364, opened 2025-07-17
Clarify GO2RTC_PATH use

rabsym:


connortechnology:
I think using a better default value will suffice.

PR #4370, opened 2025-07-19
Fix: Correct switching of players on Watch page

IgorA100:
Always load hls.js library (watch.php)
Because now we switch players "on the fly"

### DIFF
--- a/web/skins/classic/views/watch.php
+++ b/web/skins/classic/views/watch.php
@@ -103,16 +103,12 @@ if (!$cycle and isset($_COOKIE['zmCycleShow'])) {
 }
 #Whether to show the controls button
 $hasPtzControls = false;
-$hasHLS = false;
 foreach ($monitors as $m) {
   if (( ZM_OPT_CONTROL && $m->Controllable() && canView('Control') && $m->Type() != 'WebSite' )) {
     //If there is control for at least one camera, then we display the block.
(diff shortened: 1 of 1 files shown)


web/js/MonitorStream.js
```diff
@@ -257,7 +270,7 @@ function MonitorStream(monitorData) {
 
     $j('#volumeControls').hide();
 
-    if (this.Go2RTCEnabled && ((!this.player) || (-1 != this.player.indexOf('go2rtc')))) {
+    if ((this.Go2RTCEnabled && !this.player) || (-1 != this.player.indexOf('go2rtc'))) {
```
connortechnology: This is incorrect as it allows go2rtc for monitors that do not have it enabled.

IgorA100: No. My option works correctly
My condition will work:
- If Go2RTC is enabled in the monitor and the player is not selected.
- The go2rtc player is selected. BUT if go2rtc is disabled in the monitor, then it will not be in the list of available players!

connortechnology: Ugh, agreed, but why?  What does this change achieve?

IgorA100: Absolutely not critical.
Initially I wanted to do it a little differently, but I haven't done it yet
I returned the conditions as you had

PR #4376, opened 2025-07-22
Run ResizeObserver for Go2RTC on Montage page

IgorA100:


IgorA100:
@connortechnology 
Please check, it should work.

connortechnology:
Totally breaks scaling here. I don't even have any go2rtc streams viewable. Can only first first stream, full width no matter what.  No js errors.

IgorA100:
It's very strange. I don't understand yet...
This PR can't break everything so globally.
Everything works fine for me.
There are almost no changes here...

IgorA100:
I checked everything several times, everything should work.
I can't figure out the reason.
Is the problem definitely related to this PR???

IgorA100:
I just updated from https://launchpad.net/~iconnor/+archive/ubuntu/zoneminder-master
And applied this PR.
No problems, everything works correctly.

connortechnology:
yeah seems ok now.

PR #4378, opened 2025-07-22
We do not use the document hiding/showing analysis from "video-rtc.js", because we have our own analysis (MonitorStream.js)

### LINKED ISSUE #4367: AYSW not affecting go2rtc
Could be that coming back to the tab is restarting the video, but AYSW is up so it shouldn't.

IgorA100:
Closed #4367


connortechnology:
Nice.

Issue #4381, opened 2025-07-23
When using go2rtc, only Primary stream is always used and switching to Secondary does not work


IgorA100:
I think in the go2rtc config you can do it like this:
```
streams:
  "id_0":
    PATH for primary stream
  "id_1":
     PATH for secondary stream
```

Isaac, if you do this, I'll do the rest in JS.
If you agree, I will sketch out the draft code.

connortechnology:
Sure. We also need the ability to stream from ZM directly intead of hitting the camera.  So there might be 4 streams.

IgorA100:
Isaac, when you make changes in the go2rtc config as I wrote, you can check with #4382 streams should switch.

IgorA100:
> We also need the ability to stream from ZM directly intead of hitting the camera.

Do you mean viewing recorded events from a file using go2rtc or am I missing something????

connortechnology:
No, using the rtsp_server functionality of ZM to re-stream.  We may have to apply privacy zones or other changes to the video. Also many cameras can't handle too many streams. So re-streaming is a plus.

IgorA100:
> using the rtsp_server functionality of ZM to re-stream.

Yes, I understand.
By the way, you don't want to use go2trc to view recorded events, it can do it, right?

connortechnology:
As far as I know it can.

connortechnology:
I implemented the C++ side parts.

The stream choice doesn't seem to stick on a reload.

IgorA100:
> The stream choice doesn't seem to stick on a reload.

Yes, you are right, we do not save the selected stream.
We will not be able to save the selected stream in cookies, because the cookie size is limited. In case of using hundreds of cameras, there will be problems.
Perhaps we can save in PHP sessions, the memory limit is higher there, but I do not know how this will affect performance. The entire session is loaded into memory.
You can also save in the DB.

At the moment, it works like this:
For RTSP2Web, there is a "Stream source" setting. When opening the Watch page, the "Stream quality" selector is set in accordance with the "Stream source" value from the monitor settings.
Since we do not use the "Stream source" setting for go2rtc (and by default we always select the "Primary" stream), a collapse occurs.
We can also start using the "Stream source" setting from the monitor for go2rtc. In this case there will be no collapse, but the selected "Stream quality" selector will not be saved (as for RTSP2Web)
Isaac, I'm waiting for your decision.

IgorA100:
[...]

PR #4383, opened 2025-07-24
Fix: More correct switching of stream channels for RTSP2Web and Go2RTC with saving the selection in cookies on Watch page

### LINKED ISSUE #4373: When cyclically viewing, available players are not updated when switching to the next or previous camera.

IgorA100:
Now the selected channel is saved in cookies.
But the saved channel is applied to any monitor, not to a specific one. Let it be like this for now. It's better than what's happening now with the collapse....
Later, a new solution will need to be developed.
You can also pass the channel in the browser line via `streamChannel=0 или streamChannel=secondary`

The "RTSP2WebStream" setting is now visible for both RTSP2Web and Go2RTC
Probably, in the future it is necessary to rename "RTSP2WebStream" for example to "StreamSource"

Closed #4373 #4381

### DIFF
--- a/web/js/MonitorStream.js
+++ b/web/js/MonitorStream.js
@@ -100,6 +100,46 @@ function MonitorStream(monitorData) {
     return this.player = p;
   };
 
+  this.manageAvailablePlayersOptions = function(action, opt) {
+    const titleOption = "The player is disabled in the monitor settings."
+    if (action == 'disable') {
+      opt.setAttribute('disabled', '');
+      opt.setAttribute('title', titleOption);
+    } else if (action == 'enable') {
+      opt.removeAttribute('disabled');
+      opt.removeAttribute('title');
+    }
+  }
+
+  this.manageAvailablePlayers = function() {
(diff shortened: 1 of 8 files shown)


IgorA100:
For Montage page everything is a bit more complicated...
Now the channel is also taken from cookies saved on Watch page. I think this is not correct and it should be done as it was before, i.e. the channel should be taken from monitor settings.
Do you agree?
I am not ready to implement the channel switch on Montage page yet, there are many nuances there...

IgorA100:
Now the Montage page uses the channel specified in the monitor, including for go2rtc

IgorA100:
I added a channel stream check when switching to another monitor.
If the channel is not available at the monitor, the primary channel will be used

There is still a lot of work to do related to switching players and channels, but I think it is worth considering this PR and then systematically making other changes.

IgorA100:
I made the players switch when selecting the next/previous camera, as you wanted.
Closed #4373

IgorA100:
Now I have definitely completed this PR :)

IgorA100:
It seems like all switching now works correctly.

PR #4387, opened 2025-07-25
Fix: Don't apply "chosen" to hidden objects ('.hidden, .hidden-shift') (skin.js)

IgorA100:


connortechnology:
co-pilot  says this has issues.  You might try playing around with it.

My only thought on the matter is wow this is getting nasty.  

Why is it needed?

IgorA100:
> Why is it needed?

At the moment we have <select #streamQuality> used for ZMS and <select #streamChannel> used for Go2RTC and RTSP2Web. They are both in the document. Depending on which player is enabled, one or the other <select> is displayed. Chosen will apply to both in Yoga, both <select> will be displayed, which is incorrect.
You can use one <select> and rebuild it with JS, but this will complicate the code a little and increase the load on the browser a little.
And in general, it seems logical to me not to apply "chosen" to any hidden <select>, because it will break the UI.

IgorA100:
> co-pilot says this has issues.

Resolved

IgorA100:
> is getting nasty

Why?
It seems logical to me not to display hidden <select> to which "chosen" is applied
Of course, you can analyze the object before applying "chosen" and if it has a hiding class, then do not apply "chosen" to it, but this will again complicate the code and you will always need to remember this....

PR #4400, opened 2025-07-30
Fix: When changing the player and after starting the stream (hiding - displaying the page), you also need to call monitorsSetScale() on Watch page

IgorA100:
Otherwise, there may be problems with the image if we were viewing using "ZMS MJPEG" and then switched, for example, to the Go2RTC secondary channel and we have different image proportions for "ZMS MJPEG" always using the primary channel and Go2RTC using the secondary channel


IgorA100:
I also added this to this PR:

You also need to reset the previous state prevStateStarted, otherwise after the first triggering (document.visibilityState != "hidden") after hiding and then displaying the page, the stream will be restarted within 15 seconds, and this should not happen.

PR #4414, opened 2025-08-07
Fix: To stop ZMS, call CMD_QUIT instead of CMD_STOP (MonitorStream.js)

IgorA100:


web/js/MonitorStream.js
```diff
@@ -541,7 +541,8 @@ function MonitorStream(monitorData) {
           stream.src = src;
         }
       }
-      this.streamCommand(CMD_STOP);
+      //this.streamCommand(CMD_STOP);
+      this.streamCommand(CMD_QUIT);
```
connortechnology, line 545: We don't want to do this.  stop means stop.  QUIT goes in kill() or anywhere we actually change src.   You'll notice I did if (0) above.  We shouldn't be changing src here. (I think).

connortechnology: The problem with play after stop is that our this.started variable has been set to false, so we end up calling start().  Our use of started is no good.  start is the inverse of kill.  play/pause/stop should keep zms around.

IgorA100: > You'll notice I did if (0) above

Yes, I just noticed...
Okay, I won't bother you.

connortechnology: OR... we could kill zms.  Just adds a tiny tiny bit more latency if we then hit play.  So.... maybe we get rid of kill() and just do stop().

IgorA100: You can kill ZMS with CMD_QUIT and it should probably be used in MonitorStream.kill()
It is better to use CMD_STOP in MonitorStream.stop() Now, when executing CMD_STOP, the ZMS process will not send anything and will not consume resources?
It is probably better to use .kill() on the Watch page, because when switching a large number of monitors, there will be many ZMS processes.
The question with conkey remains open, now after CMD_STOP it is necessary to change conkey for the stream to start, after CMD_QUIT you can reuse the old conkey.

Issue #4415, opened 2025-08-09
When clicking the "DEBUG" button on the Filter page, the error "form is not defined" will appear if "ZM_ENABLE_CSRF_MAGIC" is disabled


IgorA100:
For a long time I couldn't understand where the GLOBAL variable "form" comes from. I couldn't find....
After a long debugging I realized that the global variable "form" appears due to the function "CsrfMagic.end" in "includes\csrf\csrf-magic.js"
At the same time, "CsrfMagic.end" will not be called if "ZM_ENABLE_CSRF_MAGIC" is disabled.

When you click the "DEBUG" button, the function "debugFilter" will be called
```
function debugFilter() {
getModal('filterdebug', $j(form).serialize());
}
```
Which tries to use the global variable "form", but it doesn't exist....

I think everything is wrong here....
You need to use "use strict"; for csrf-magic.js and clean things up. Also, in the debugFilter() function you should NOT use the global variable "form"

connortechnology:
I agree.

IgorA100:
I'm surprised that for so many years no one has pressed the "DEBUG" button with ENABLE_CSRF_MAGIC disabled :)))

connortechnology:
It really isn't recomended to disable csrf... should probably be hidden under and "Advanced" section

IgorA100:
> It really isn't recomended to disable csrf... 

Yes, of course. But I had some problem (I don't remember what it was) during the tests, I disabled CSRF and then forgot to enable it. Eventually I found the problem...

> should probably be hidden under and "Advanced" section

It should probably be in the "Security" section, but there doesn't seem to be such a section.

PR #4417, opened 2025-08-09
Fix: Do not generate global variables "form" and "forms" (csrf-magic.js)

IgorA100:
It is necessary to analyze the consequences of such a change. Related to #4415

Perhaps somewhere else there was a use of the global variable "form", which is essentially very incorrect!

### DIFF
--- a/web/includes/csrf/csrf-magic.js
+++ b/web/includes/csrf/csrf-magic.js
@@ -8,7 +8,8 @@
 // Here are the basic overloaded method definitions
 // The wrapper must be set BEFORE onreadystatechange is written to, since
 // a bug in ActiveXObject prevents us from properly testing for it.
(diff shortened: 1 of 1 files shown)


connortechnology:
Looks good to me.

connortechnology:
I went through the UI and had to no js issues.  This was with use strict enabled.

IgorA100:
For now, we won't use "use strict"
Because of it, when saving changes in the monitor, an extra request appears in the browser with confirmation of the action. I haven't figured out why this happens yet. The code needs to be analyzed. Perhaps problems will pop up somewhere.

IgorA100:
I also reviewed the code and for me not everything is obvious in the csrf-magic.js code
I think for now we should not use "use strict"

PR #4430, opened 2025-08-29
Fix: After the cursor leaves Thumbs, ZMS must always be stopped, even in case of network problems.

### LINKED ISSUE #4418: ZMS process is not killed on Console page after cursor leaves monitor image
On the Console page, hover the cursor over the camera image.
The image will enlarge and be viewed in real time, and the ZMS process will appear.
After the cursor is removed from the image, there is no video animation, but the ZMS process continues to exist and consume resources.

IgorA100:
For all Thumbs we will assign the attribute "data-monitor-url" so that we can send ZMS the CMD_QUIT command without using MonitorStream.
We should not rely only on SIGPIPE, we must terminate ZMS correctly in any case.

Closed #4418


connortechnology:
That isn't going to do it.

We need a keepalive.   The zms cgi needs a keepalive parameter, so that if it doesn't get a status update request after so many seconds, it dies.  We can't depend on CMD_QUIT.

IgorA100:
> That isn't going to do it.

Why? It works fine for me! I'm debugging JS now.

> The zms cgi needs a keepalive parameter

Yeah, I agree. That would be a great option. But I don't know how to do it :(
If you did this in C++, I would arrange for a command to be sent to maintain the connection.

> We can't depend on CMD_QUIT.

Yeah, but it's better than relying on SIGPIPE alone

IgorA100:
This PR works!
But of course it is better to use keepalive for ZMS

IgorA100:
@connortechnology 
In this PR we add a attribute to the IMG so that we can send a command to the server where the monitor is served. This attribute will also be needed if you implement keepalive for ZMS. If keepalive is implemented for ZMS, then a very small change will need to be made to skin.js.
This PR works, but I certainly like the implementation of keepalive for ZMS more, but I can't do it in C++

IgorA100:
@connortechnology 
Should I continue working on this PR and resolve the conflicts or should I close this PR?

connortechnology:
Is it still needed? There have been a few changes that may have made it obsolete.
I like the placeholder image stuff and having a connkey generator. 
I don't like the putting urltoindex into each thumbnail... seems like not the best way to do it.  Maybe a better way to go would be mid, and then we can just look up in our js data structures the monitor object and get the url from there.

IgorA100:
> Is it still needed?

I'm not sure, but it's likely necessary. I have one server that hasn't been updated in a long time and it uses this PR. Some other servers don't use a domain name, just an IP address, and there are no problems. I'll need to retest this PR in more detail with the current code.

> I don't like the putting urltoindex into each thumbnail... 

This was the fastest solution with minimal code changes. I think it's possible to skip urltoindex and get the data using JS.

IgorA100:
No problems were found in the current version, so I am closing this PR.

PR #4466, opened 2025-11-15
[release-1.36] Fix building against FFmpeg 8.0

peat-psuwit:
`avcodec_close()` is now removed in FFmpeg 8.0, so #ifdef its calls everywhere. This commit is a minimal amount of changes which allow 1.36 series to build against FFmpeg 8.0, as opposed to cherry-picking commits on the master branch which remove older supports for older FFmpeg versions.

Most call sites have ifdef'ed call to `avcodec_free_context()`. In such case, hitch-hike it and put `avcodec_close()` in the else case as `avcodec_free_context()` calls (the equivalent of) `avcodec_close()` internally.

zm_videostore.cpp has un-ifdef'ed `avcodec_free_context()` together with `avcodec_close()` for some reason, so add the corresponding ifdef.

Calls at zm_mpeg.cpp and zm_remote_camera_rtsp.cpp has never been ported to `avcodec_free_context()`, so they're ported. The comment in zm_remote_camera_rtsp.cpp makes me unsure if this porting is correct, but I essentially followed commit 8822710 [^1] on the master branch, so either both are correct, or both are incorrect. 😅

[^1] Commit 8822710 ("Removing avcodec_close, avcodec_free_context is
     replacement and been around since 2014, so should be no reason to
     version test.")

### DIFF
--- a/src/zm_ffmpeg.cpp
+++ b/src/zm_ffmpeg.cpp
@@ -605,10 +605,11 @@ int zm_send_frame_receive_packet(AVCodecContext *ctx, AVFrame *frame, AVPacket *
 
 void zm_free_codec(AVCodecContext **ctx) {
   if (*ctx) {
-    avcodec_close(*ctx);
 #if LIBAVCODEC_VERSION_CHECK(57, 64, 0, 64, 0)
     // We allocate and copy in newer ffmpeg, so need to free it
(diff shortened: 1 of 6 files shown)


connortechnology:
Thanks!

Issue #4319, opened 2025-04-30
REGRESSION:  Back to an event list from an event requires a manual refresh now.


Simpler1:
**Describe Your Environment**
- Version of ZoneMinder: 1.37.66 (latest commit: b357ce52aa6397818ff1d8c362122fbc8f6e3df5)
- How you installed ZoneMinder: Docker: zoneminder-containers/zoneminder-base with eventserver-base
- Full name and version of OS: Linux Mint 20.3 Cinnamon
- Browser name and version: Chromium: Version 130.0.6690.0 (Official Build) stable (64-bit)

**Describe the bug**
Some change was made between the following two commits: (It's been a month and a half since I've updated)
cd914f885dda2dee5675401e065a6cb791d3267e on March 3, 2025
b357ce52aa6397818ff1d8c362122fbc8f6e3df5 on April 22, 2025
When clicking on an event from a list of events and then clicking the back button (in the top left of the UI), the list of events would always display.  Now, it just shows the header with no list.  I have to manually refresh the page (F5) to get the list to reappear.

**To Reproduce**
1. Go to any list of events
2. Click on an event
3. Click on the left arrow in the top left of the UI
4. Notice that the list is not displayed
5. Manually refresh the page by pressing F5
6. Notice that the list appears.

**Expected behavior**
A manual refresh should not be required after backing out of an event.

@IgorA100

IgorA100:
Chromium 109.0.5414.120 & Win 8.1 I can't reproduce the problem.
As far as I remember, I didn't have any problems with the newer Chromium on Win 10 either.
Maybe the problem is only in Linux? or in some ZM settings.
You don't have a chance to check in Windows?
I don't have any computers with Linux at the moment :(

Simpler1:
I've verified the problem on Android 14 with Chrome also.   The problem
doesn't occur on Android 14 with Firefox.

IgorA100:
Android 10 & Chrome 135.0.07049 no problem found :(
Probably the problem is related to caching. But the strange thing is that everything is read from the cache except DataTable.
I need to somehow reproduce this problem on my own....

IgorA100:
Can I get a screenshot including console?

IgorA100:
@Simpler1 
Try to remove the function "window.addEventListener('beforeunload', function(event) {" in the skin.js file
Will your problem remain or will everything be fine?

Simpler1:
Removing that function seems to have fixed the problem.

[...]

PR #4470, opened 2025-11-25
Fix Issue #4319

Simpler1:
Removing these lines fixes issue #4319.

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -2108,12 +2108,6 @@ function initPageGeneral() {
         // If flex-grow is set to a value > 0 then "height" will be ignored!
         mainContentJ.css({flex: "0 1 auto"});
       }
-
-      mainContentJ.animate({height: 0}, 300, function rollupBeforeunloadPage() {
-        const btnCollapse = $j('body').find('#btn-collapse');
-        if (btnCollapse) btnCollapse.css({display: "none"});
-        mainContentJ.css({display: "none"});
-      });
     }
     //event.returnValue = '';
   });


IgorA100:
Removing the code is a very basic solution.
Probably the best option is to find the problem and fix it.
But I can't reproduce the problem to fix it.
This is just my personal opinion, and it may be wrong.

Simpler1:
Until you can fix the code, it's better to remove the problem.

IgorA100:
> Until you can fix the code

I have no idea how to fix the problem if I can't reproduce it.

Simpler1:
That's why we would remove the non-essential code until it works.

connortechnology:
The "non-essential" code exists to solve a very real problem, that we actually can reproduce.

It might be interesting to see if just removing the animate fixes the problem, while leaving the display:none lines.  The animate is not essential.

I can't imagine why those lines would affect content on another page.  And since it isn't happening to us, I don't think we need to do any knee-jerk actions.  

Can you test just without the animate but leaving the display: none ?

Simpler1:
> The "non-essential" code exists to solve a very real problem, that we actually can reproduce.
> 
> It might be interesting to see if just removing the animate fixes the problem, while leaving the display:none lines. The animate is not essential.
> 
> I can't imagine why those lines would affect content on another page. And since it isn't happening to us, I don't think we need to do any knee-jerk actions.
> 
> Can you test just without the animate but leaving the display: none ?

I wasn't aware that even some of this function was fixing another problem.  I've been running with this function removed for months and it's been working fine.  (I'm just tired of removing this function every time I update.)  What original problem should I be seeing when this function is removed?

Also, I'm not sure how to remove the animation and keep the display:none lines.  If you can provide some example lines to replace this function, I'll test.

Simpler1:
The animate seems to be the problem.  Here's an updated function (with the animate commented out) that works.  I'm not sure how to test the original problem since I don't know what it is/was.  

The "if" statement that was checking for btnCollapse always returns true, so it's not necessary.

Can you test this?
```
      mainContentJ
          // .animate({height: 0}, 300)
          .promise()
          .done(() => {
            $j('#btn-collapse').hide();
            mainContentJ.hide();
          });
```

Simpler1:
After some more testing (both desktop and mobile), I believe it behaves better with this function completely removed.  On both desktop and mobile, I use the buttons to apply tags and move to the previous/next event many times per day (on the order of 2000 times per day).  When this function is removed, the tag that gets applied is displayed for a fraction of a second so that I can confirm that it applied what I expected.  When this function is in place, the screen goes blank before the tag is displayed.
image

IgorA100:
> The "if" statement that was checking for btnCollapse always returns true, so it's not necessary.

This condition is mandatory! If you use the top menu, btnCollapse will be set to false!!!

IgorA100:
> Can you test this?

Why are you using "promise" in your code?
If the problem is only with the animation, you can use this code (there won't be any animation, of course):
`mainContentJ.hide();`
Without animation, there is no need to use "btnCollapse.css({display: "none"});" and it is not needed.

But I still can't understand why "beforeunload" is triggered twice and how to avoid it.

IgorA100:
Does the problem only occur when navigating from the Event page to the Events page?
Does the problem occur with other navigations (from the Event page to a page other than the Events page or to an Events page other than the Event page)?

connortechnology:
I think animation should probably be configurable. Not everyone wants animation effects.  I'll add a config flag for that.  

We do need to figure out the beforeunload running twice though.

IgorA100:
> We do need to figure out the beforeunload running twice though.

It might be related to "bfcache", or maybe something else. If I could reproduce the problem, I could probably solve it.

I'm curious to see if the code posted here: https://github.com/ZoneMinder/zoneminder/issues/4319#issuecomment-3616059147
solves the problem or not.

Simpler1:
> > The "if" statement that was checking for btnCollapse always returns true, so it's not necessary.
> 
> This condition is mandatory! If you use the top menu, btnCollapse will be set to false!!!

jQuery objects are always truthy, even if empty, so this condition doesn’t actually check existence. The code will run css() even if no element is found — but jQuery will safely no-op in that case.

Simpler1:
I found some very odd behavior testing...  I opened two tabs in the Chromium browser and ran the same filter in each tab.  I then selected the same event in each tab.  From each event, I pressed the back button in the UI.  One of the tabs displayed the table without a refresh.  The other required a refresh to display the table.   Not sure how to explain this.  I even started from a "Shift F5" on the events page to ensure that it forced a reload in both tabs.

IgorA100:
> I found some very odd behavior testing

I'm confused....

Does the problem only occur when navigating from the Event page to the Events page?
Does the problem occur with other navigations (from the Event page to a page other than the Events page or to an Events page other than the Event page)?

Simpler1:
> > I found some very odd behavior testing
> 
> I'm confused....
> 
> Does the problem only occur when navigating from the Event page to the Events page? Does the problem occur with other navigations (from the Event page to a page other than the Events page or to an Events page other than the Event page)?

Once the problem occurs in a browser tab, it continues to occur in that browser tab.
The problem occurs when using the "back" button (either in the UI or in the browser).
The problem has been seen going back from an event to events.
The problem has been seen going back from events to console.
The problem has been seen going back from events to filters.
The problem has been seen going forward from events to event.
The problem has been seen going forward from events to filters.

The problem never occurs when clicking the "Console" text button at the top of the screen.
The problem never occurs when clicking the "LIST MATCHES" button at the bottom of the Filters page.

Simpler1:
No longer relevant.

PR #4475, opened 2025-12-08
Display the "No Tag" filter label

Simpler1:
When "LIST MATCHES" is pressed from the filters page from a filter that has the "No Tag" filter applied, show "No Tag" in the filter area on the events page.


Simpler1:
Not ready to merge yet

PR #4476, opened 2025-12-08
Fix(tags):  "No Tag" filter now works properly.

Simpler1:
Multiple fixes for tags.
Filtering for "No Tag" is not the same as not filtering for any tag.
The "No Tag" label is properly displayed in the filter area of the events screen when a named filter is used with "LIST MATCHES".

### DIFF
--- a/web/includes/Filter.php
+++ b/web/includes/Filter.php
@@ -979,7 +979,7 @@ class Filter extends ZM_Object {
         }
       }
     }
-    $availableTags = array(translate('No Tag'));
+    $availableTags = array('0'=>translate('No Tag'));
     foreach ( dbFetchAll('SELECT Id, Name FROM Tags ORDER BY lower(`Name`) ASC') AS $tag ) {
       $availableTags[$tag['Id']] = validHtmlStr($tag['Name']);
     }
@@ -1115,7 +1115,7 @@ class Filter extends ZM_Object {
     for ( $i = 0; $i < 7; $i++ ) {
       $weekdays[$i] = date('D', mktime(12, 0, 0, 1, $i+1, 2001));
     }
-    $availableTags = array(translate('No Tag'));
+    $availableTags = array('0'=>translate('No Tag'));
     foreach ( dbFetchAll('SELECT Id, Name FROM Tags ORDER BY lower(`Name`) ASC') AS $tag ) {
       $availableTags[$tag['Id']] = validHtmlStr($tag['Name']);
     }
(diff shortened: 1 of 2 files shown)


web/skins/classic/views/filter.php
```diff
@@ -132,7 +132,7 @@
   }
 }
 
-$availableTags = array(''=>translate('No Tag'));
+$availableTags = array(translate('No Tag'));
```
connortechnology: I realise the effect is the same... but I'd really prefer if we would write '0'=>translate('No Tag').  It is also not entirely clear to me why '0' is better than ''.

web/includes/Filter.php
```diff
@@ -1146,22 +1149,18 @@ public function simple_widget() {
           $html .= htmlSelect("filter[Query][terms][$i][val]", $archiveTypes, $term['val'],['id'=>'filterArchived', 'class'=>'chosen chosen-auto-width']).PHP_EOL;
           $html .= '</span>';
         } else if ( $term['attr'] == 'Tags' ) {
-          $selected = empty($term['val']) ? [] : json_decode($term['val']);
-          // echo '<pre>selected: '; print_r($selected); echo '</pre>';
-          if (count($selected) == 1 and !$selected[0]) {
-            $selected = null;
-          }
+          $selected = explode(',', $term['val']);
```
connortechnology, line 1152: Actually think the json_decode is the correct thing. When we store things in cookies, we JSON.stringify .  From what I understand, using json to serialize values to cookies is the preferred thing to do these days.

web/skins/classic/views/filter.php
Simpler1: I'll make this change to '0'.  It doesn't work when using ''.

web/includes/Filter.php
Simpler1: It doesn't work with the json_decode.  Try it on your machine.  
Create and save a filter named "No Tags" with a term of "Tag" "equals" "No Tag".
Click "LIST MATCHES".
Notice that the "Tags =" field on the events page is blank.  It should be showing "No Tag".

Simpler1: Also, you can uncomment one of the debug echo lines that I added to see what $selected is set to when the events page loads:
`// echo '<div style="background-color:lightblue"><pre>selected: '; print_r($selected); echo '</pre></div>';`

connortechnology: Ok, I have come around to the use of 0 for No Tag.  Also.. our use of empty($term[val]) is no good because empty('0') = true.  

So I'm going to merge this but I've got some commits to add as well, so be sure to update.

Issue #4464, opened 2025-11-03
Problem generating downloads of large event files


IgorA100:
For example, we need to generate a 60GB file, but we only have 40GB of disk space.
In this case, Apache will take up all the disk space, causing a problem. But we won't know about it because the Frontend will timeout (unless, of course, the timeout is set to 10 minutes or more).
The temporary file (or rather, TWO huge ones!) will remain in the system until you manually delete them!

The second problem: We have 40GB of disk space, and we want to generate a 30GB file. But this will also fail because the file will be archived and will eventually take up ~60GB, which we don't have...

Yes, with infinite disk space and the Frontend timeout set to infinity, there won't be any problems, but in reality, things are different...

I don't yet know how to quickly fix this problem.
I see something like this:
- Analyze disk space before file generation and notify the user if space is low. - Offer a download option without archiving (archiving videos is pointless, and you can always sacrifice the text file included in the archive for space)
- Run the generation and download process in the background to avoid Frontend timeouts

connortechnology:
We already estimate the size of the download based on a sum of event disk space.   So it should be trivial to provide a warning.

IgorA100:
> We already estimate the size of the download based on a sum of event disk space.

Yes, that's right.
But all this needs to be done in the background so it doesn't depend on Frontend timeouts and add the ability to download unzipped files.
When I have time, I'll probably come up with something.

IgorA100:
After downloading the generated archive, all files (generated mp4, txt, zip) remain on my server until I restart the server.
Tested in Chrome & Chromium
I don't think this is a good idea.

connortechnology:
Hard to know when to delete, which is why we let tmpfiles.d auto-delete them at the end of the day.  Still a potential DOS though.  

I don't actually see why this was done in a two-step way, step 1 create the file return the file name, step 2 browser requests the file.  If we did it all in one request, we could delete it at the end of the request.

IgorA100:
Yes, I also have questions about generating and downloading files.
I'll think about the best way to do it.

IgorA100:
I'm working on this problem. A lot of things need to be redone.
Firstly, files will be added to the archive one by one. That is, an MP4 file is generated and immediately added to either a ZIP or TAR file. This significantly saves disk space for intermediate files.
Second, it will be possible to download files without archiving. Links will be generated and displayed to the user.
Third, all intermediate temporary files will be deleted as the archive is created.
[...]

PR #4488, opened 2025-12-17
Feat: Optimizing disk space when generating and downloading events on the Events page

IgorA100:
- Added the ability to download files in MP4 format. This will allow you to download large files.
For example, let's say we want to download 20 event files, each 1GB in size. This will generate a 20GB file and need to be enclosed in a ZIP or TAR file. At some point, we'll need 20GB * 2 = 40GB of space, but we only have 30GB available. In this case, we need to download in MP4 format.
- Added a display of free disk space information.
- Deleting generated temporary files after downloading

There are still some unresolved issues:
- ~~If the download fails (canceled in the browser, network problems, browser blocking multiple downloads, etc.), the generated files are not deleted. I have some ideas, but that's for later... WOW! I know how to solve this problem!!!~~ **Problem solved.**
- When generating multiple MP4 download links, after each successful download and deletion of the temporary file, I need to deactivate the download link in the modal window.
- If the files were generated but for some reason didn't start downloading, the files will still remain on the disk. I'm addressing this issue in another PR.
- Disable generating files if there isn't enough disk space. I think it's necessary to leave at least 500 MB of free space.. I'm addressing this issue in another PR.
- ~~When the "GENERATE DOWNLOAD" button is clicked **again** in the modal window, I need to force the deletion of previously generated files (especially relevant for multiple MP4 files).~~ **Problem solved.**

### DIFF
--- a/web/skins/classic/js/export.js
+++ b/web/skins/classic/js/export.js
@@ -10,10 +10,7 @@ function startDownload(exportFile) {
 
 function getFileNameFromURL(url) {
   const matches = url.match(/file=(.*?)&/) || url.match(/file=(.*)/);
-  if (matches)
-    return matches[1];
-  else
-    return '';
+  return (matches) ? matches[1] : '';
 }
 
 function exportResponse(data, responseText) {
@@ -46,7 +43,6 @@ function exportResponse(data, responseText) {
           $j(nodeCopy).attr("href", thisUrl + file);
         }
       }
-
     } else {
       $j('#downloadLink').text('Download ' + '"' + getFileNameFromURL(exportFile) + '"');
       $j('#downloadLink').attr("href", thisUrl + exportFile);
@@ -77,7 +73,7 @@ function exportEvent() {
       downloadLink.remove();
       i++;
     } else {
-       i = 1000;
+      i = 1000;
     }
   }
 


IgorA100:
@connortechnology 
I've completed this PR.
I'd really like you to take a look at it and give me your opinion.

Closed #4464

connortechnology:
After the auto-download, the listed .zip is still there, and clicking on it silently fails because the download has been deleted.
Console->events->click all->Click Download button->Click generate button.  download is generated and auto-downloads.  modals stays up, with .zip still listed.

Perhaps the modal should close.... or .zip should get greyed out or something.  I missed the actual auto-download happen visually so maybe just need some visual indicator that it was successful.

IgorA100:
> or .zip should get greyed out or something.

Yes, I plan to do this in another PR. I plan to make the link inactive after the download starts. It's not exactly fast; you'll need to use AJAX.

> I missed the actual auto-download happen visually so maybe just need some visual indicator that it was successful.

I haven't thought about that yet...
It would be possible to somehow indicate that the auto-download has started. Perhaps it would also be possible to somehow indicate when the download is complete. I'll have to think about that.

But these are all additional improvements; there could be many more such improvements in the loading module. I think it would be better to organize all of this into separate PRs. Probably.....

PR #4495, opened 2025-12-19
Fix(event): Avoid error applying tag before tags are loaded

Simpler1:
Avoids an error condition that can occur when first loading an event and the available tags have not yet been loaded and one of the buttons in the tag field is pressed to apply the last tag and move to the next/previous event.


connortechnology:
I find this code style bad... going to just do if(tag&& etc)

PR #4507, opened 2025-12-28
Perl5 ExtUtils/Command/MM.pm expects Makefile

nabbi:
"Makefile" is hardcoded into Perl's MakeMaker, it does not hornor renaming. i.e. MakefilePerl

The patch suppresses 3077 uninitialized occurrences when building ZoneMinder

Use of uninitialized value in numeric gt (>) at
/usr/lib64/perl5/5.42/ExtUtils/Command/MM.pm line 147.

A seperate feature enchancement ticket has also been logged with upstream: 
https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/479


onvif/modules/CMakeLists.txt
```diff
@@ -9,10 +9,10 @@ endif()
 # MAKEMAKER_NOECHO_COMMAND previously defined in /scripts/zoneminder/CMakeLists.txt
 
 # Add build target for the perl modules
-add_custom_target(zmonvifmodules ALL perl Makefile.PL ${ZM_PERL_MM_PARMS_FULL} FIRST_MAKEFILE=MakefilePerl DESTDIR="${CMAKE_CURRENT_BINARY_DIR}/output" ${MAKEMAKER_NOECHO_COMMAND} COMMAND make -f MakefilePerl pure_install COMMENT "Building ZoneMinder perl ONVIF proxy module")
+add_custom_target(zmonvifmodules ALL perl Makefile.PL ${ZM_PERL_MM_PARMS_FULL} FIRST_MAKEFILE=Makefile DESTDIR="${CMAKE_CURRENT_BINARY_DIR}/output" ${MAKEMAKER_NOECHO_COMMAND} COMMAND make -f Makefile pure_install COMMENT "Building ZoneMinder perl ONVIF proxy module")
```
Leont: The default value is `Makefile`, so you can also just leave out the `FIRST_MAKEFILE` argument entirely

connortechnology:
This cannot be merged because there is a conflict on the Makefile.  cmake creates a Makefile, so teling MakeMaker to use Makefile does not work. it is literally why we use Makefile.PL instead.

nabbi:
I flipped this to draft to ponder it further.

I can't say I am seeing the cmake Makefile conflict as it's not generated within the gentoo build sandbox; that might be an artifact of how "ebuild cmake" behaves, we don't need it.

ExtUtils/Command/MM.pm has been hard coded with "Makefile" 24 years ago, nearly since its inception.
So what that process is likely being feed is the mtime of your cmake "Makefile", not that perl override of MakefilePerl.
It "works" but perl's makemaker isn't properly sourcing the correct perl makefile

https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/blame/c91610afcaff3fc1cbed12deecacfd28f492ef1e/lib/ExtUtils/Command/MM.pm#L114

connortechnology:
That is a good point.  I am in the habit of building/make installing directly from the src dir, which is bad habit, but very ... easy and convenient.  

I wonder if we could rename the cmake generated Makefile... 

I do think it is ridiculous that we can't tell MakeMaker to use a different file... but yeah 24years later.

I've spent some time asking google if we can do without MakeMaker and it's not trivial.. 

Perhaps the best thing would be to fix our package builds and my own workflow to build in a separate dir.

nabbi:
@connortechnology Suspect we both have higher priorities to focus on :)

I explored some other options last night but haven't found functional workaround yet. In the mean time I can ship a local patch, ie this PR, in the ebuild package to suppress how this issue presented itself. Was hoping we could fix it at the source but you confirmed it's more complicated; needs other tooling / workflow changes to accommodate

connortechnology:
I actually Claude burn through my end of day credits on removing MakeMaker and going with a pure cmake option.  It actually wasn't that bad.  We don't actually use MakeMaker for the real purposes it is useful for (I think). I wasn't aroudn when it was implemented in ZM :).  We really just need to copy all the .pm to the right place in fs.  I'll create a PR, maybe you can take a look and test.

nabbi:
#4617 supersede

PR #4508, opened 2025-12-30
Prevent packaging runs outside of Zoneminder\zoneminder repo. I noted…

SteveGilvarry:
… that on my fork these workflows are running and failing, well maybe not aarch64 it is just waiting for a runner.

Can't really test this outside the repo, but it did stop them running on my repo when I created this so that is 50% right.


connortechnology:
Looks good to me... I did something similar on other branches recently...

PR #4510, opened 2025-12-31
Videojs update

SteveGilvarry:
Ok well that took a while and found some old stuff along the way.


connortechnology:
So I tested, and it works.  Still not clear on what video.js is actually doing for us though...
I gave it a try with video.js commented out and there is a lot of js errors... but it looks and plays the same...
I'm not wild about the script tags being automatically included in functions.php.

connortechnology:
I fixed up the js errors... and now the only thing missing here is a volume control, which I need to copy over from live view.
So in short... what is video.js getting us? A quick google gives us:

connortechnology:
Cross-Browser Consistency: Different browsers render the default HTML5 video player with varied appearances and controls. Video.js provides a unified, CSS-based "skin" that ensures a consistent look and feel across all modern desktop and mobile browsers, which can also be easily customized.
Adaptive Streaming Support: It allows playback of advanced streaming formats like HLS (HTTP Live Streaming) and DASH (Dynamic Adaptive Streaming over HTTP) everywhere, even where those formats are not natively supported by the browser, by using its built-in playback engine (VHS).

IgorA100:
> A quick google gives us:

That's the correct answer. These days, I think every browser has its own video player. But they all look different. Video.js brings a more or less unified look and feel to all players.
I remember you, Isaac, once said we needed to implement a player with H265 support. I read a bit, thought about it, looked at your code, but I never found the time to look into it in more detail.

connortechnology:
Thankfully we no longer need an h265 player.  All my browsers now play h265.  I did implement a javascript based one but the performance was never acceptable.

PR #4519, opened 2026-01-05
Add Tags to event search and return tag data with events

SteveGilvarry:
Isaac said in chat that Tags in API doesn't work, figured must have been related to Askers question about filtering event view by tags, so anyway I did this, can filter on Tag (single only) and TagID (single tag and multiple)


connortechnology:
Things to note:  For performance reasons, we should not join the other tables unless we are filtering by tags.  And I don't think we should include the tags in the results.  People can query the tags on a per-event basis.

PR #4527, opened 2026-01-07
Fix ONVIF PropertyOperation handling according to spec

SteveGilvarry:
Properly implement ONVIF PropertyOperation handling according to the ONVIF Profile T Specification. This fixes issues where alarms were incorrectly triggered or cleared due to misinterpretation of event messages.

Changes:
- Add interpret_alarm_value() helper to handle various value formats including "000", "0", "false", "true", etc.
- Implement proper PropertyOperation handling:
  * "Deleted" - alarm has ended, clear it
  * "Initialized" - current state at subscription (sync, don't trigger)
  * "Changed" - actual state transition (trigger alarm on/off)
- Remove premature alarm clearing on empty PullMessages responses
- Add comprehensive comments explaining ONVIF spec compliance

The root cause was that the parser was extracting different values from the same XML after commit 3f6495644. The new parser prioritizes SimpleItem Value attributes which may contain camera-specific values like "000". By properly interpreting PropertyOperation, we now correctly handle all ONVIF-compliant cameras regardless of their value format.

Fixes issues where events incorrectly went to ALERT state.

### DIFF
--- a/src/zm_monitor_onvif.cpp
+++ b/src/zm_monitor_onvif.cpp
@@ -393,12 +393,13 @@ void ONVIF::WaitForMessage() {
       {  // Scope for lock
         std::unique_lock<std::mutex> lck(alarms_mutex);
 
-        bool has_messages = tev__PullMessagesResponse.wsnt__NotificationMessage.size() > 0;
-        if (!has_messages and !parent->Event_Poller_Closes_Event and alarmed) {
-          alarmed = false;
-          alarms.clear();
-        }
-        
+        // Note: We do NOT clear alarms on empty PullMessages response.
+        // According to ONVIF spec, alarms should only be cleared based on explicit
+        // PropertyOperation="Deleted" or PropertyOperation="Changed" with inactive value.
+        // The old code cleared on empty messages because it wasn't properly handling
+        // PropertyOperation - cameras DO send proper "Deleted" or value changes, we just
+        // weren't interpreting them correctly.
+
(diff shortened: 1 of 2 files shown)


SteveGilvarry:
I need to test, but need to get a camera up, so have a go also. Would be good to collect a selection of cameras that we can test, or at least the onvif messages from a selection.

src/zm_monitor_onvif.cpp
```diff
@@ -393,12 +393,13 @@ void ONVIF::WaitForMessage() {
...
 
-        bool has_messages = tev__PullMessagesResponse.wsnt__NotificationMessage.size() > 0;
-        if (!has_messages and !parent->Event_Poller_Closes_Event and alarmed) {
-          alarmed = false;
-          alarms.clear();
-        }
-        
+        // Note: We do NOT clear alarms on empty PullMessages response.
+        // According to ONVIF spec, alarms should only be cleared based on explicit
+        // PropertyOperation="Deleted" or PropertyOperation="Changed" with inactive value.
```
connortechnology, line 398: This remains to be proven.  Our understanding is that not all cameras report end of event.

connortechnology:
I have a Tapo C520WS, Annke 918 (Hikvision), Amcrest, and many LTS (HikVisions) of various models to test with.

src/zm_monitor_onvif.cpp
SteveGilvarry: Lol yes I agree, but then also not Onvif compliant.

PR #4528, opened 2026-01-08
Add gSoap logging plugin

SteveGilvarry:
First this will log headers so that comes with security risks that people should know about.
Controlled via soap_log=/var/log/zm/onvif_soap.log, and it will dump a lot of data pretty quick as polling often.
image
I will do some more playing with camera I have setup at the moment to see what I can find, but init message was spamming a lot, so added something to log and ignore it.

### DIFF
--- a/src/zm_monitor_onvif.cpp
+++ b/src/zm_monitor_onvif.cpp
@@ -409,7 +409,9 @@ void ONVIF::WaitForMessage() {
         // PropertyOperation - cameras DO send proper "Deleted" or value changes, we just
         // weren't interpreting them correctly.
 
+        int msg_index = 0;
         for (auto msg : tev__PullMessagesResponse.wsnt__NotificationMessage) {
+          msg_index++;
           std::string topic, value, operation;
           
           // Use improved parsing that handles different message structures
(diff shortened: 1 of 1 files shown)


SteveGilvarry:
Also did some more fixes to the parser for data I was seeing from my camera which had 
 tns1:VideoSource/MotionAlarm (ONVIF Core Spec)
  - Data: Name="State" Value="true|false"

  tns1:RuleEngine/CellMotionDetector/Motion (ONVIF Analytics)
  - Data: Name="IsMotion" Value="true|false"

  tns1:RuleEngine/TamperDetector/Tamper (ONVIF Analytics)
  - Data: Name="IsTamper" Value="true|false"

  tns1:RuleEngine/FieldDetector/ObjectsInside (ONVIF Analytics)
  - Data: Name="IsInside" Value="true|false"

  tns1:VideoSource/GlobalSceneChange/ImagingService (ONVIF Imaging)
  - Data: Name="State" Value="true|false"

SteveGilvarry:
First go round so feel free to suggest changes to how it works. Some more time with the ONVIF specs and I could probably do better.

PR #4530, opened 2026-01-08
Add missing ONVIF field translations to en_gb.php

SteveGilvarry:
Add human-readable translations for ONVIF configuration page labels that were displaying raw field names with underscores instead of proper labels.

Added translations for:
- ONVIF (tab name)
- ONVIF_URL, ONVIF_EVENTS_PATH, ONVIF_Options
- ONVIF_Alarm_Text, ONVIF_Event_Listener
- Username, Password (generic fields used in ONVIF section)
- SOAP WSA COMPLIANCE


web/lang/en_gb.php
```diff
@@ -478,6 +484,7 @@
     'PanTilt'               => 'Pan/Tilt',
     'ParentGroup'           => 'Parent Group',
     'PasswordsDifferent'    => 'The new and confirm passwords are different',
+    'Password'              => 'Password',
```
connortechnology, line 487: Is this really necessary?  We don't actually need to translate things that don't need translation

PR #4550, opened 2026-01-12
I returned the old "this.controlMute" function in MonitorStream.js

IgorA100:
It's not entirely clear why it was necessary to change the "this.controlMute" function in the commit.https://github.com/ZoneMinder/zoneminder/commit/f796fa913fe3bedbd13c707dfacc8c649ed1018f

After this change, the "volumeSlider" and "iconMute" styling sometimes doesn't work correctly.

The old code was 99.9% debugged and worked without any issues.

It's also unclear why "this.muted" was added? After all, we already had "audioStream.muted." Now we're probably getting duplicates.

@connortechnology I propose to discuss here an issue that occurs when using my code. I will try to solve it (if it actually occurs...). With this PR, all my monitors start with the sound muted on the Montage page.

### DIFF
--- a/web/js/MonitorStream.js
+++ b/web/js/MonitorStream.js
@@ -914,7 +914,7 @@ function MonitorStream(monitorData) {
       iconMute.innerHTML = (volume == 'on')? 'volume_up' : 'volume_off';
     }
     return iconMute;
-  }
+  };
 
   /*
   * volume: on || off
@@ -929,7 +929,7 @@ function MonitorStream(monitorData) {
       }
     }
     return volumeSlider;
-  }
+  };
 
   /*
   * mode: switch, on, off
@@ -938,7 +938,7 @@ function MonitorStream(monitorData) {
     const mid = this.id;
     let volumeSlider;
     const audioStream = this.getAudioStream(mid);
-    if (!audioStream){
+    if (!audioStream) {
       console.log(`No audiostream! in controlMute for monitor ID=${mid}`);
       return;
     }


connortechnology:
I forget exactly what the issues were, but the effort was to deal with montage.  Having volume on for many streams was a problem.

IgorA100:
Could you try reproducing the issue with the current code?
I've tested it with various settings (except Janus, which doesn't work properly for me), and all cameras start with the sound muted on the Montage page.

connortechnology:
The purpose of a lot of the changes is that the old code returned if the mute icon didn't exist, which is doesn't on montage.

I fail to see how the change could have affected the styling... but hey maybe it does.  Current code works great on montage.

One thing that I am seeing is that the volume controls aren't visible until play happens... In general I think showing/hiding is less than enabled/disabled.

IgorA100:
> if the mute icon didn't exist

Oh, my god... I used the mute icon and volumeSlider everywhere, since it will be present in the audio visualization...
I'll fix the code soon. We also won't need to create the `#volumeControls` slider as a JS object, nor will we need to set a listener for it, etc.


> I am seeing is that the volume controls aren't visible until play happens...

Yes, because there may be no audio in the stream. And until playback starts, it's impossible to determine whether there is an audio stream. If there is no audio stream, there's no point in displaying the volume control. But this is happening either because of the browser or because of video.js.

IgorA100:
> but hey maybe it does.

Yes. Under certain conditions (using a loop and periodically turning the sound on and off on the Watch page), there are issues.

IgorA100:
@connortechnology 
I've completed this PR.
Everything works fine on Montage and Watch pages, and the code is clearer.
Please check it out..

PR #4562, opened 2026-01-21
ONVIF Polling Optimization

BHMSD-bdailey:
Problem

  ONVIF event polling shared a 10-second loop with other managers (Amcrest, RTSP2Web, Go2RTC, Janus). With a 2-second
   pull_timeout, there was an 8-second "deaf" window per cycle where events couldn't be received.

  Solution

  Move ONVIF to a dedicated thread with optimized timing:

  1. Dedicated thread: ONVIF now runs independently, not blocking or blocked by other managers
  2. Longer pull_timeout (PT10S): Maximizes the "listening" window when connected to the camera
  3. Short sleep_for (500ms): Minimizes gap between poll cycles
  4. Thread safety: Added std::atomic<bool> for alarmed_ and healthy_ flags

  Thought Process

  The ONVIF PullMessages API is a long-poll pattern—the camera holds the connection open and returns immediately when
   an event occurs. The optimal strategy is:

  - Maximize listening time: Longer pull_timeout = more time connected, waiting for events
  - Minimize dead time: Shorter sleep between polls = smaller window where events are missed
  - Stay under renewal threshold: pull_timeout must be < 20 seconds to allow subscription renewals

  Latency Comparison

  ┌─────────────────────┬─────────────┬─────────────┐
  │ Metric              │ Before      │ After       │
  ├─────────────────────┼─────────────┼─────────────┤
  │ pull_timeout        │ 5 seconds   │ 10 seconds  │
  │ Loop interval       │ 10 seconds  │ ~10.5 sec   │
  │ Listening time      │ 5s (50%)    │ 10s (95%)   │
  │ Dead time           │ 5s (50%)    │ 0.5s (5%)   │
  │ Worst-case latency  │ ~5 seconds  │ ~500ms      │
  │ Typical latency     │ ~2.5 sec    │ Near-instant│
  │ Other managers      │ Blocked     │ Unaffected  │
  └─────────────────────┴─────────────┴─────────────┘

  Visual Timeline (10-second window)

  BEFORE (shared Poll() loop):
  [==LISTEN==][--------DEAF--------][==LISTEN==]
     0-5s            5-10s              10-15s

  AFTER (dedicated thread):
  [======LISTEN======][.][======LISTEN======][.]
         0-10s        10s      10.5-20.5s    21s
                     500ms                  500ms

  Files Changed

  - src/zm_monitor_onvif.h - Added thread members, atomic flags
  - src/zm_monitor_onvif.cpp - Added Run() loop, thread lifecycle management
  - src/zm_monitor.cpp - Removed ONVIF from shared Poll(), call start() at creation

### DIFF
--- a/src/zm_monitor_onvif.cpp
+++ b/src/zm_monitor_onvif.cpp
@@ -337,9 +337,9 @@ void ONVIF::Run() {
     } else {
       // Attempt to re-establish connection
       Debug(1, "ONVIF: Unhealthy, attempting to restart subscription");
+      std::this_thread::sleep_for(std::chrono::milliseconds(500));
       start();
     }
-    std::this_thread::sleep_for(std::chrono::milliseconds(500));
   }
   Debug(1, "ONVIF: Polling thread exiting");
 }


connortechnology:
I had set the pull timeout to 5s because when it is time to shutdown zmc, it might have to wait 5s on that thread before zmc can shutdown.  I definitely agree with this being it's own thread.  Good work.  Being it's own thread I would imagine there would only be nanoseconds before it made another pull... so I'm not sure how much it matters how long the timeout is in terms of missing events..

connortechnology:
Since the Pull is blocking, do we even need to sleep at all?

BHMSD-bdailey:
OK. The benefit of the longer timeout is less connects and reconnects (network traffic) when there is no ONVIF activity.  So if there was a 1 hour period with no ONVIF messages and pull_timeout set to 10 there would be ~343 (3600/10.5) HTTP polling post, with pull_timeout set to 5 there would be ~655 (3600/5.5) HTTP polling post. I am fine reverting the default back to 5. Just let me know and I can update the PR and put the default back to 5. It is configurable via ONVIF_Options.

BHMSD-bdailey:
I think sleeping is good in case there is an error condition and to not overrun a weaker camera? I think the value could definitely be debated. Initially claude wrote it as 100ms. I was thinking 250ms and when we did the final analysis and code change it landed at 500ms. I am good with changing the value or eliminating it but I think my preference would be to have some at least small amount of breathing room.

connortechnology:
When in doubt, make it configurable!  I actually REALLY want zmc to shutdown and restart quickly... because while it is doing it, we aren't recording... so there's that goal.  I actually think a request once per second is still ok.  We are doing WAY MORE than that with zms and getting status updates.  Definitely need to sleep on error.  Maybe we could measure how long the pull request took and sleep however long it would take to achieve the desired frequency, but if we can miss events if we aren't polling, then we should just be polling.

PR #4563, opened 2026-01-21
Fix: More correct handling of "muted" and copying of all attributes when cloning a DOM node (MonitorStream.js)

IgorA100:


web/js/MonitorStream.js
```diff
@@ -338,6 +338,12 @@ function MonitorStream(monitorData) {
     if (statusEl) statusEl.innerText = '';
   };
 
+  this.copyAllAttributes = function(fromEl, toEl) {
```
IgorA100, line 341: When cloning a DOM node, we need to copy all attributes, not just "ID" and "style", since there may be other attributes that we will need later.

web/js/MonitorStream.js
```diff
@@ -358,10 +364,9 @@ function MonitorStream(monitorData) {
 
         const old_stream = this.getElement();
         const stream = this.element = document.createElement('video-stream');
-        stream.id = old_stream.id; // should be liveStream+id
-        stream.style = old_stream.style; // Copy any applied styles
+        this.copyAllAttributes(old_stream, stream);
         stream.background = true; // We do not use the document hiding/showing analysis from "video-rtc.js", because we have our own analysis
-        stream.muted = this.muted;
+        //stream.muted = this.muted;
```
IgorA100, line 369: This is not required with a probability of 99.9%

web/js/MonitorStream.js
```diff
@@ -378,8 +383,9 @@ function MonitorStream(monitorData) {
         if (-1 != this.player.indexOf('_')) {
           stream.mode = this.player.substring(this.player.indexOf('_')+1);
         }
-        const video_el = document.querySelector('video');
+        const video_el = document.querySelector('#liveStream'+this.id+' video');
```
IgorA100, line 386: It's important to get the correct element, not the first one encountered. This is especially important for the Montage page.
Otherwise, on line 390, "this.createVolumeSlider();" will trigger for an unknown element (the very first one encountered), but the "mute" will be correct. Naturally, the sound for other elements might turn on for a fraction of a second on the Montage page.

web/js/MonitorStream.js
```diff
@@ -378,8 +383,9 @@ function MonitorStream(monitorData) {
...
         if (video_el) {
+          video_el.muted = this.muted;
```
IgorA100, line 388: And here, we will preemptively enable "muted"; we will not rely on further work (even if it is correct and now without delays) of "controlMute()". What if we decide to use a different slider to control the sound? :)

PR #4568, opened 2026-01-25
API: Support multiple Event IDs in TagsController index

### LINKED ISSUE #4567: API: TagsController index does not support filtering by multiple Event IDs
## Summary

The Tags API endpoint `/api/tags/index/Events.Id:<id>.json` works correctly for a single Event ID, but does not support querying tags for multiple events in one API call.

## Current Behavior

- Single event works: `/api/tags/index/Events.Id:40434571.json` ✓
- Multiple events fails: `/api/tags/index/Events.Id:40434571,40434611.json` ✗

The comma-separated value is treated as a single string rather than being parsed into an array for a SQL `IN` clause.

## Expected Behavior

`/api/tags/index/Events.Id:40434571,40434611.json` should return tags for both events, similar to how `EventsController` handles multiple Tag IDs:

```php
$tag_ids = is_array($tag_filter_value) ? $tag_filter_value : explode(',', $tag_filter_value);
$tag_ids = array_map('intval', $tag_ids);
```

## Proposed Solution

In `TagsController::index()`, after the FilterComponent builds conditions, check if `Events.Id` contains comma-separated values and convert them to an integer array for proper SQL `IN` clause handling.

## Use Case

This would allow clients to efficiently retrieve tags for multiple events in a single API call, reducing the number of requests needed when displaying tag information for event lists.

pliablepixels:
## Summary

- Add support for comma-separated Event IDs in the Tags API endpoint
- Enables efficient retrieval of tags for multiple events in a single API call

**Before:** `/api/tags/index/Events.Id:123,456.json` treated `123,456` as a single string value (failed)

**After:** `/api/tags/index/Events.Id:123,456.json` properly splits into `[123, 456]` for SQL `IN` clause

## Changes

In `TagsController::index()`, when `Events.Id` is set, check if it contains comma-separated values and convert to an integer array:

```php
$event_id_value = $conditions['Events.Id'];
if (!is_array($event_id_value) && strpos($event_id_value, ',') !== false) {
  $conditions['Events.Id'] = array_map('intval', explode(',', $event_id_value));
}
```

This approach mirrors how `EventsController` handles multiple Tag IDs.

## Test plan

- [ ] Verify single Event ID still works: `/api/tags/index/Events.Id:123.json`
- [ ] Verify multiple Event IDs now work: `/api/tags/index/Events.Id:123,456,789.json`
- [ ] Verify array values are sanitized to integers (injection prevention)

Fixes #4567

### DIFF
--- a/web/api/app/Controller/TagsController.php
+++ b/web/api/app/Controller/TagsController.php
@@ -48,12 +48,18 @@ class TagsController extends AppController {
     );
 
     if (isset($conditions['Events.Id'])) {
-      $find_array['joins'][]   = [ 
+      // Support comma-separated Event IDs for querying tags across multiple events
+      $event_id_value = $conditions['Events.Id'];
+      if (!is_array($event_id_value) && strpos($event_id_value, ',') !== false) {
+        $conditions['Events.Id'] = array_map('intval', explode(',', $event_id_value));
+      }
+
+      $find_array['joins'][]   = [
           'table' => 'Events_Tags',
           'type'  => 'inner',
           'conditions' => ['Events_Tags.TagId = Tag.Id'],
       ];
-      $find_array['joins'][]   = [ 
+      $find_array['joins'][]   = [
           'table' => 'Events',
           'type'  => 'inner',
           'conditions' => ['Events.Id = Events_Tags.EventId'],


connortechnology:
Looks good, note to self, also implement for other controllers.  FYI this was possible before, you just have to specify multiple EventId:= in the url... which takes up a lot of space. This is better.

PR #4571, opened 2026-01-26
Remove triggers views instead

BHMSD-bdailey:


connortechnology:
Just a heads up, I'm finally testing this.

connortechnology:
I don't believe views can do the job in a performant way.  I think the current state of master is acceptable.  Closing.

Issue #4561, opened 2026-01-19
ONVIF with Hikvision intrusion detection lead to infinite recording


ssymo84:
**Describe Your Environment**
- Version of ZoneMinder: v1.37.76
- How you installed ZoneMinder: https://zmrepo.zoneminder.com/debian/master/
- Full name and version of OS: Debian 12.12

**If the issue concerns a camera**
- Make and Model: Hikvision DS-2CD2347G2H

**Describe the bug**
I'm using Hikvision cameras DS-2CD2347G2H with Intrusion Detection enabled.
On zm I set `ONVIF Alarm Text: ObjectsInside` and it seems working correctly, the recording start but never stop.
ZM log:
```
zmc_m1	16527	INF	Porta Ingresso: 960 - Opening event timestamp 1768667725 % 325	zm_monitor.cpp	2407
zmc_m1	16527	INF	Porta Ingresso: 960 - Gone into alarm state PreAlarmCount: 0 > AlarmFrameCount:1 Cause:ONVIF	zm_monitor.cpp	2224
zmc_m1	16525	INF	ONVIF [before_renewal_check]: Subscription terminates at 2026-01-17 17:45:10 (in 584s), renewal at 2026-01-17 17:45:00 (in 574s)	zm_monitor_onvif.cpp	874
zmc_m1	16525	INF	Setting ClosesEvent (camera supports PropertyOperation)	zm_monitor_onvif.cpp	524
zmc_m1	16525	INF	ONVIF Alarm already active on subscription (Initialized): tns1:RuleEngine/FieldDetector/ObjectsInside	zm_monitor_onvif.cpp	510
zmc_m1	16525	INF	ONVIF Got Event [msg 1/1]! topic:tns1:RuleEngine/FieldDetector/ObjectsInside value:26080 operation:Initialized	zm_monitor_onvif.cpp	465
```
On ODM:

<img width="2226" height="179" alt="Image" src="https://github.com/user-attachments/assets/9fceddd6-f0a1-4cef-8979-692cd6b9550e" />

This type of ONVIF events do not have a closing event, so recording should last for a maximum time and then stop if no other similar events are received.

Another difference is that the Operation is different, on ODM is `changed ` while in ZM is `Initialized`, looking at the source code of zm_monitor_onvif.cpp I see that when the operation is `initialized` closes_event is set which I imagine waits for a closing event.

Thanks for your support.

BHMSD-bdailey:
[...]

PR #4573, opened 2026-01-27
feat: add per-topic alarm expiry using PullMessagesResponse Terminati…

BHMSD-bdailey:
…onTime

Cameras like Reolink send alarm=true but never send the corresponding false, causing alarms to stick indefinitely. Use the TerminationTime from PullMessagesResponse to auto-expire stale per-topic alarms.

- Add AlarmEntry struct with value and termination_time fields
- Extract TerminationTime from each PullMessagesResponse and attach it to alarm entries; refresh on re-trigger so active alarms persist
- Sweep expired alarms after processing messages and on poll timeout
- Add expire_alarms option (default: true) to disable via onvif_options
- Fix TOCTOU race: remove unsynchronized alarms.empty() check before acquiring mutex in the timeout sweep path
- Simplify SetNoteSet with C++17 structured bindings
- Add Catch2 tests for alarm expiry logic (mirrored struct to avoid gSOAP header dependency)

### DIFF
--- a/src/zm_monitor_onvif.cpp
+++ b/src/zm_monitor_onvif.cpp
@@ -77,6 +77,7 @@ ONVIF::ONVIF(Monitor *parent_) :
   ,use_absolute_time_for_renewal(false)
   ,renewal_enabled(true)
 #endif
+  ,expire_alarms_enabled(true)
   ,terminate_(false)
 {
   parse_onvif_options();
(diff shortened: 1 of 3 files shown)


BHMSD-bdailey:
This is a first pass at resolving Issue #4561

PR #4574, opened 2026-01-28
fix: Storage.DiskSpace drifts to negative values over time

lukolszewski:
## Problem

  Storage.DiskSpace in Options->Storage shows incorrect negative values (e.g., -588 GB)
  instead of the actual disk usage.

  ## Root Cause

  There's an asymmetry in how Storage.DiskSpace is maintained:
  - **Event creation** (C++): Does NOT add to Storage.DiskSpace
  - **Event deletion** (Perl): Subtracts from Storage.DiskSpace

  This causes Storage.DiskSpace to drift negative over time as events are deleted
  without corresponding additions when created.

  ## Fix

  Remove the subtraction on event deletion in `Event.pm`. Storage.DiskSpace is now
  maintained solely by zmaudit, which periodically recalculates it from
  `SUM(Events.DiskSpace)`.

  ## Workaround for existing installations

  Run this SQL to fix corrupted values:
  ```sql
  UPDATE Storage SET DiskSpace = (
    SELECT SUM(DiskSpace) FROM Events
    WHERE Events.StorageId = Storage.Id AND DiskSpace IS NOT NULL
  );

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/Event.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Event.pm
@@ -418,11 +418,10 @@ sub delete {
     ZoneMinder::Database::zmDbDo('DELETE FROM Events WHERE Id=?', $$event{Id});
     $ZoneMinder::Database::dbh->commit() if ! $in_transaction;
 
-    my $storage = $event->Storage();
-    if ($event->DiskSpace() and $storage->Id()) {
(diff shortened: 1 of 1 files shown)


connortechnology:
zmaudit is not meant to be run all the time.  It is resource intensive.
The correct thing to do is update Storage in C++.

See 2033b5eb7

PR #4576, opened 2026-01-29
Fix(): Find server in database instead of defaulting to port 80

Simpler1:
An empty Server object has no hostname/protocol/port, so JavaScript was filling in the blanks incorrectly as port 80 (even with https).


connortechnology:
The problem here is that some people like to delete the default server entry.  We need to fall back to nothing in the db.

connortechnology:
I did a possibly quick fix for getting the port under nginx... maybe it will solve the problem? Maybe.  Also, the path to api/zms when there is no server record is set in /etc/zm/conf.d/01-system-paths.conf so those entries need to be set to match whatever you are actually using.

Simpler1:
I tried your fix, but it didn't change the problem.  Without my proposed fix (from Claude), when I load the Montage Review page, I get this in the console:
```
error 
{readyState: 0, getResponseHeader: ƒ, getAllResponseHeaders: ƒ, setRequestHeader: ƒ, overrideMimeType: ƒ, …}
￼csrf-magic.js:42 ￼
 GET ￼https://zoneminder.my_site.com:80/api/events/index/DateTime%20%3E=:2026…3%3A18/MonitorId:10.json?username=my_user&password=my_password net::ERR_SSL_PROTOCOL_ERROR
￼skins_classic_views_…-1769781521.js:1257
```

It's adding in the ":80" when it shouldn't.

connortechnology:
That fix might work for you but it won't work for anyone else. We need to understand why it is doing this.  at the very least it should have defaulted to 443 since it is https. 

Perhaps I can try to recreate your setup.  nginx, ssl, different port.

connortechnology:
So I setup nginx on port 8080... and it correctly uses 8080 for the default server entry and  hence is able to load the data from the api.

So the servers entry is populated from web/skins/classic/js/skin.js.php line 51.  It does a new ZM\Server() and then a Server->to_json().

So it should be calling ->Port() which looks in $_SERVER['HTTP_X_FORWARDED_PORT'], 
then $_SERVER['HTTP_HOST'] 
then $_SERVER['SERVER_PORT']
then finally looks at Protocol for https and defaults to 443 or 80.  

So one of those first 3 must be 80.  Can you insert a debug line to print them out and see what we get?

Simpler1:
Here are the debug output lines:
```
2026-01-31 08:34:50 PM.476212 web_php[440].DBG [45.55.35.48] [$_SERVER['HTTP_X_FORWARDED_PORT'] is ] at skins/classic/js/skin.js.php line 50
2026-01-31 08:34:50 PM.476290 web_php[440].DBG [45.55.35.48] [$_SERVER['HTTP_HOST'] is zoneminder.mysite.com] at skins/classic/js/skin.js.php line 51
2026-01-31 08:34:50 PM.476320 web_php[440].DBG [45.55.35.48] [$_SERVER['SERVER_PORT'] is 80] at skins/classic/js/skin.js.php line 52
```

Simpler1:
I tweaked this PR (with the help of Claude).  I think it satisfies your concern.  I tested and it works for me.

PR #4591, opened 2026-02-04
Fix: More correct logic for #volumeSlider processing & Improved functionality for RTSP2web, Janus, and ZMS

IgorA100:
- Added search for audio and video tracks in the stream.
- If there is no audio track, #volumeSlider becomes unavailable for control.
- Improved functionality for RTSP2web, Janus, and ZMS.

### DIFF
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -896,6 +896,13 @@ function streamReStart(oldId, newId) {
   //Change main monitor block
   monitor_div.innerHTML = currentMonitor.streamHTML;
 
+  const volumeControls = document.getElementById('volumeControls'+oldId);
+  if (volumeControls) volumeControls.id = 'volumeControls'+newId;
+  const volumeSlider = document.getElementById('volumeSlider'+oldId);
+  if (volumeSlider) volumeSlider.id = 'volumeSlider'+newId;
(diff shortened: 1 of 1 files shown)


IgorA100:
@connortechnology 
view=cycle - is it used now or not???

connortechnology:
not. Kept ir around in case someone wanted it.. or something... but no one has. I think we can delete it.

IgorA100:
Ok.

IgorA100:
@connortechnology 
Is there something about this PR that you don't like, or haven't checked it yet?

connortechnology:
Sorry didn't notice the most recent marked it ready for review. Thought it was still a draft

IgorA100:
No problem.

web/js/MonitorStream.js
```diff
@@ -344,6 +350,18 @@ function MonitorStream(monitorData) {
     }
   };
 
+  this.replaceDOMElement = function(fromEl, toTypeEl) {
```
connortechnology, line 353: This seems like a generic utility function that could go in skin.js instead of here

IgorA100: Yes, that's right.
I'll do it.

IgorA100: Ready. #4614

PR #4605, opened 2026-02-08
perf: skip Frames table scan during event playback unless debugging

nabbi:
I was exploring event playback delays and decided to pull on this thread; beyond the theoretical gains I have no measurable benchmarks
Also have back port https://github.com/nabbi/zoneminder/commit/901e62e2a179095f9785bd5be5107f5d5f4ca3ae for 1.38.0 

Thank you

Remove the FramesDuration subquery from the event metadata query in EventStream::loadEventData(). Previously every playback initiation ran:

  SELECT max(Delta)-min(Delta) FROM Frames WHERE EventId=Events.Id

as an embedded subquery. Without a covering index on (EventId, Delta), MySQL walks the EventId_idx to find matching rows then fetches each table row to read the Delta column. For a 10-minute event in Record/Mocord mode at 30fps with bulk_frame_interval=100, that is ~180 index lookups + row fetches. For alarm-heavy events where every frame gets a DB row, it can reach thousands of row fetches. This blocks playback start on every event view.

The FramesDuration value was only consumed by a Debug level 2 log comparing it against Event Length. It was never used in any playback computation, frame timing, seek logic, or client-facing output. The frames_duration field has been removed from the EventData struct entirely. The diagnostic query and its log are now colocated inside a logLevel() >= DEBUG2 guard using a local variable.

Production impact: none. Default ZoneMinder log level is INFO (0). DEBUG2 is only reached via explicit operator configuration (Options > System > LOG_LEVEL_FILE or ZM_DBG_LEVEL env variable). No production deployment runs at DEBUG2 as it generates thousands of log lines per second per daemon. The subquery code path is unreachable under default configuration. Operators who enable DEBUG2 get the same diagnostic output as before.

Theoretical gains on playback initiation per event:
- Eliminates 1 SQL subquery performing N row fetches from the critical path (N = Frames rows for the event)
- Record mode, 10min event: ~180 fewer row fetches
- Alarm-heavy 10min event: ~3,000-18,000 fewer row fetches
- Reduces MySQL buffer pool pressure from Frames page reads
- Reduces InnoDB row lock contention with concurrent frame INSERTs from active recording daemons hitting the same EventId range

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/General.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/General.pm
@@ -20,6 +20,7 @@ our @ISA = qw(Exporter ZoneMinder::Base);
 our %EXPORT_TAGS = (
     functions => [ qw(
       executeShellCommand
+      findDbCommand
       getCmdFormat
       runCommand
       setFileOwner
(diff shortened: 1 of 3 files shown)


nabbi:
This shouldn't conflict with the gcc15-vector feature branch https://github.com/ZoneMinder/zoneminder/pull/4607

PR #4606, opened 2026-02-09
perf: replace O(n) linear frame seeks with O(log n) binary search

nabbi:
I subjectively see a noticeably smoother event playback experience when seeking across the timeline bar after applying these changes.
back port 1.38.0 exists at https://github.com/nabbi/zoneminder/commit/9e88f69e7bdee6ee0011ca8f635c38a10f0fe9f4

Thank you

Two frame search sites in EventStream used linear scans over the frames vector:

1. seek() scanned every frame comparing timestamps until finding the first match. Used during initial event load and time-based navigation.

2. processCommand(CMD_SEEK) used a linear interpolation estimate then walked forward/backward frame-by-frame to refine the position. Used when the user scrubs the timeline during playback. The existing code had a "TODO Replace this with a binary search" comment marking this.

Both are replaced with std::lower_bound / std::upper_bound over the already-sorted frames vector. For a 10-minute event at 30fps (18,000 frames), worst-case comparisons drop from 18,000 to ~14.

seek() uses lower_bound on timestamp to find the first frame at or after the target, then checks one frame back for delta window coverage (preserving the original's "timestamp + delta >= event_time" heuristic). An explicit end() check handles the case where the target equals end_time, since the trailing frame generator produces frames with timestamps strictly less than end_time.

CMD_SEEK uses upper_bound on offset to find the first frame past the target, then steps back one to get the last frame at or before it. This replaces the estimate-then-walk approach and resolves a minor inconsistency in the original where an overestimate walk-back landed one frame after the target while an underestimate walk-forward landed one frame before.

No changes to frame loading, frame sending, streaming, event transitions, or any behavior after the target frame is located. The Debug(1) per-step log spam during CMD_SEEK walks is removed as the iterative walk no longer exists.

### DIFF
--- a/src/zm_eventstream.cpp
+++ b/src/zm_eventstream.cpp
@@ -28,6 +28,7 @@
 #include <arpa/inet.h>
 #include <sys/stat.h>
 
+#include <algorithm>
 #include <filesystem>
 
 #ifdef __FreeBSD__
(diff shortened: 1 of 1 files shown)


nabbi:
I need to incorporate this one into a larger gcc 15 vector feature branch fix

PR #4607, opened 2026-02-09
Fix event playback stability and optimize seek performance

nabbi:
## Fix event playback stability and optimize seek performance

Event stream playback (`zms`) has several latent safety issues in how it manages its frame vector, plus runtime crashes that surface under real-world conditions (broken connections, event boundaries, concurrent seeks).

This PR fixes those issues and then builds on the corrected foundations to replace the `O(n)` seek scans with `O(log n)` binary search.

---

## Why These Ship Together

The six commits form a dependency chain:

1. **Vector safety fixes**
   Correct three related bugs in how `event_data->frames` is indexed:

   * `frame_count` / `last_frame_id` do not account for synthetic interpolated frames
   * `last_frame` becomes a dangling pointer after `emplace_back` reallocation
   * Backward seek has an off-by-one into `frames[-1]`

   These must land first because every subsequent commit (including the binary search) depends on `frames.size()` being the authoritative bounds.

2. **Runtime crash fixes**
   (mutex scope, `SIGPIPE`, keepalive timer) were found during playback testing.
   Without them, `zms` dies silently on broken HTTP connections or races between the command thread and the streaming thread, making the seek improvements untestable in practice.

3. **JavaScript crash recovery**
   Allows the browser-side player to restart the `zms` stream and replay the queued seek when the server process exits (e.g., at event boundaries with the 5-second keepalive gap).
   Without this, a dead `zms` leaves the UI stuck with no way to resume.

4. **Binary search optimization**
   Replaces the two linear scans in `seek()` and `processCommand(CMD_SEEK)` with `std::lower_bound` / `std::upper_bound`.

   For a 10-minute event at 30fps (~18,000 frames), this drops worst-case comparisons from ~18,000 to ~14.

   This commit depends on all the safety fixes above:

   * Uses `frames.size()` for bounds
   * Trusts that the vector is not accessed out of range
   * Assumes `zms` survives long enough for seeks to execute

---

## Commits

* `fix: harden eventstream against vector bounds, dangling pointers, and off-by-one`
* `fix: move sendFrame inside mutex scope to prevent race condition`
* `fix: ignore SIGPIPE in zms to survive broken HTTP connections`
* `fix: reset keepalive timer on event-end pause to prevent connection drop`
* `fix: recover from zms crash by restarting stream on seek`
* `perf: replace O(n) linear frame seeks with O(log n) binary search`

---

## Files Changed

* `src/zm_eventstream.cpp` — vector safety, mutex scope, keepalive, binary search
* `src/zms.cpp` — SIGPIPE handling
* `web/skins/classic/views/js/event.js` — stream crash recovery

---

## Toolchain Note: GCC 15 / _GLIBCXX_ASSERTIONS

The previous out-of-bounds frames[] accesses (using frame_count that does not account for synthetic frames, and the frames[-1] off-by-one) are silent undefined behavior under normal builds.

However, GCC 15 and newer distributions increasingly enable _GLIBCXX_ASSERTIONS by default. This adds runtime bounds checking to std::vector::operator[].

On those platforms, the old code hard-aborts instead of silently corrupting memory. These safety fixes are therefore not only correctness improvements, but also crash fixes for anyone building with a modern toolchain.

### DIFF
--- a/src/zm_eventstream.cpp
+++ b/src/zm_eventstream.cpp
@@ -89,18 +89,42 @@ bool EventStream::seek(SystemTimePoint event_time) {
     return false;
   }
 
-  for (int i=0; i < (int)event_data->frames.size(); i++) {
-    //Info( "eft %d > et %d", event_data->frames[i].timestamp, event_time );
-    if (event_data->frames[i].timestamp >= event_time ||
-        event_data->frames[i].timestamp + event_data->frames[i].delta >=event_time) {
-      curr_frame_id = i + 1;
-      curr_stream_time = event_time;
-      Debug(3, "Set curr_stream_time: %.2f, curr_frame_id: %d",
-          FPSeconds(curr_stream_time.time_since_epoch()).count(),
-          curr_frame_id);
-      return true;
+  // Binary search: frames are sorted by timestamp (ascending).
+  // Find the first frame whose timestamp >= event_time.
+  auto it = std::lower_bound(
+    event_data->frames.begin(),
+    event_data->frames.end(),
+    event_time,
+    [](const FrameData &frame, const SystemTimePoint &t) {
+      return frame.timestamp < t;
+    });
+
+  // If event_time is past all frame timestamps (e.g. at event end_time),
+  // step back to the last frame so the delta check below can evaluate it.
+  if (it == event_data->frames.end() && it != event_data->frames.begin()) {
+    --it;
+  }
+
+  // Check if the previous frame's time window covers event_time.
+  // Original logic: timestamp + delta >= event_time allows a frame whose
+  // timestamp is before the target to match if its display window extends
+  // past it.
+  if (it != event_data->frames.begin()) {
+    auto prev = std::prev(it);
+    if (prev->timestamp + prev->delta >= event_time) {
+      it = prev;
     }
-  } // end foreach frame
+  }
+
+  if (it != event_data->frames.end() &&
+      (it->timestamp >= event_time || it->timestamp + it->delta >= event_time)) {
+    curr_frame_id = static_cast<int>(std::distance(event_data->frames.begin(), it)) + 1;
+    curr_stream_time = event_time;
+    Debug(3, "Set curr_stream_time: %.2f, curr_frame_id: %d",
+        FPSeconds(curr_stream_time.time_since_epoch()).count(),
+        curr_frame_id);
+    return true;
+  }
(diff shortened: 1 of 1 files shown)


src/zms.cpp
```diff
@@ -276,6 +277,10 @@ int main(int argc, const char *argv[], char **envp) {
   Image::Initialise();
   zmSetDefaultTermHandler();
   zmSetDefaultDieHandler();
+  // Ignore SIGPIPE so broken HTTP connections return EPIPE from write()
+  // instead of killing the process. Without this, the 5-second keepalive
+  // gap after event end causes zms to die on the next stdout write.
+  std::signal(SIGPIPE, SIG_IGN);
```
connortechnology: Um... we kinda depend on SIGPIPE to tell use when to quit.  Can't we make this it's own PR instead of including it with the rest here? Iw ould want to test that separately.

web/skins/classic/views/js/event.js
```diff
@@ -827,7 +835,18 @@ function streamSeek(offset) {
...
-    streamReq({command: CMD_SEEK, offset: offset});
+    if (zmsBroke) {
+      // zms has exited; restart stream and queue seek for when it's ready
+      pendingSeek = offset;
+      const img = document.getElementById('evtStream');
+      const url = new URL(img.src);
+      url.searchParams.set('scale', currentScale);
+      img.src = '';
+      img.src = url.href;
+      zmsBroke = false;
```
connortechnology: I'd be tempted to set zmsBroke in the img.onload instead of here... but maybe it doesn't matter.

src/zms.cpp
nabbi: Sure. I was questioning this too if SIGPIPE got muddled with troubleshooting the vector crashing. I can strip that commit

nabbi: We don't need it. I stashed the commit with a [SIGPIPE-ANALYSIS.md](https://github.com/nabbi/zoneminder/blob/research_sigpipe-zms-analysis/SIGPIPE-ANALYSIS.md) under branch [research_sigpipe-zms-analysis](https://github.com/nabbi/zoneminder/tree/research_sigpipe-zms-analysis)

web/skins/classic/views/js/event.js
nabbi: I appreciate your insights to the code base. TL;DR is both locations [research_zms-crash-recovery-onload.md](https://github.com/nabbi/zoneminder/blob/research_zms-crash-recovery-onload/research_zms-crash-recovery-onload.md)

PR #4618, opened 2026-02-11
feat: support token substitution in EventStartCommand/EventEndCommand

pliablepixels:
## Summary

- Replace `execlp()` with `execl("/bin/sh", "sh", "-c", cmd, nullptr)` in `zm_monitor.cpp` for both `EventStartCommand` and `EventEndCommand`, so commands can include arguments
- Add `%EID%`, `%MID%`, `%EC%` token substitution (using existing `ReplaceAll()` from `zm_utils.h`) matching the convention used by `zmfilter.pl`
- `%EC%` (event cause) is only available in `EventStartCommand`; `EventEndCommand` substitutes `%EID%` and `%MID%`

## Motivation

Previously, `execlp()` treated the entire command string as an executable path, making it impossible to pass arguments. This meant users needed wrapper scripts for `EventStartCommand`/`EventEndCommand`. With this change, commands like:

```
/usr/bin/zm_detect.py -c /etc/zm/objectconfig.yml -e %EID% -m %MID% -r "%EC%" -n
```

work directly without a wrapper.

## Test plan

- [ ] Verify `EventStartCommand` fires with correct `%EID%`, `%MID%`, `%EC%` substitution
- [ ] Verify `EventEndCommand` fires with correct `%EID%`, `%MID%` substitution
- [ ] Verify commands with arguments execute correctly (no "Bad file descriptor" errors)

### DIFF
--- a/src/zm_monitor.cpp
+++ b/src/zm_monitor.cpp
@@ -3236,13 +3236,23 @@ Event * Monitor::openEvent(
       logTerm();
       int fdlimit = (int)sysconf(_SC_OPEN_MAX);
       for (int i = 0; i < fdlimit; i++) close(i);
-      std::string cmd = ReplaceAll(ReplaceAll(
-          ReplaceAll(event_start_command, "%EID%", std::to_string(event->Id())),
-          "%MID%", std::to_string(event->MonitorId())),
-          "%EC%", ShellEscape(cause));
-      execl("/bin/sh", "sh", "-c", cmd.c_str(), nullptr);
-      logInit(log_id.c_str());
-      Error("Error execing %s: %s", cmd.c_str(), strerror(errno));
+      if (event_start_command.find('%') != std::string::npos) {
+        std::string cmd = ReplaceAll(ReplaceAll(
+            ReplaceAll(event_start_command, "%EID%", std::to_string(event->Id())),
+            "%MID%", std::to_string(event->MonitorId())),
+            "%EC%", ShellEscape(cause));
+        execl("/bin/sh", "sh", "-c", cmd.c_str(), nullptr);
+        logInit(log_id.c_str());
+        Error("Error execing %s: %s", cmd.c_str(), strerror(errno));
+      } else {
+        execlp(event_start_command.c_str(),
+               event_start_command.c_str(),
+               std::to_string(event->Id()).c_str(),
+               std::to_string(event->MonitorId()).c_str(),
+               nullptr);
+        logInit(log_id.c_str());
+        Error("Error execing %s: %s", event_start_command.c_str(), strerror(errno));
+      }
       std::quick_exit(0);
     }
   }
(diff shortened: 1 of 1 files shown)


pliablepixels:
## Response to Copilot Review Comments

### 1. Shell injection via `%EC%` (cause) — Valid concern

Good catch. The `cause` string can include `trigger_data->trigger_cause` written via shared memory by `zmtrigger.pl` (network-accessible) and `zmu`. While it's limited to 32 bytes (`char trigger_cause[32]`), that's enough for shell injection (e.g., `'; rm -rf / #`). Zone labels from `zone.Label()` are also included and are user-configured via the web UI.

**Fix**: Will shell-escape `cause` before substituting into `%EC%`. `%EID%` and `%MID%` are safe since they're always numeric (`std::to_string` on integer IDs).

### 2. Backward compatibility — Valid concern for both commands

Existing scripts using `$1` (event_id) and `$2` (monitor_id) would silently stop receiving those values with the current change.

**Fix**: Pass the IDs as positional parameters through `sh -c`, so both the new `%EID%`/`%MID%`/`%EC%` token substitution AND the old `$1`/`$2` positional args work:

```cpp
execl("/bin/sh", "sh", "-c", cmd.c_str(), "sh",
      std::to_string(event->Id()).c_str(),
      std::to_string(event->MonitorId()).c_str(),
      nullptr);
```

This gives `$0="sh"`, `$1=event_id`, `$2=monitor_id` — matching the old argv layout.

### 3. Note on Copilot's EventEndCommand code suggestion — Has a bug

The backward-compat concern is valid, but the suggested code is incorrect. With `sh -c command_string [command_name [argument...]]`, the first argument after the command string becomes **`$0`** (the command name), NOT `$1`. The suggestion:

```cpp
execl("/bin/sh", "sh", "-c", cmd.c_str(),
      event_id_str.c_str(), monitor_id_str.c_str(), (char *)nullptr);
```

would give `$0 = event_id`, `$1 = monitor_id` — not matching the old behavior where `$1 = event_id` and `$2 = monitor_id`. The correct form needs a `$0` placeholder (`"sh"`) before the positional args.

pliablepixels:
Pushed fix for the shell injection concern: `cause` is now wrapped via `ShellEscape()` (single-quote wrapping with proper escaping of embedded single quotes) before `%EC%` substitution. `%EID%` and `%MID%` are always numeric (`std::to_string` on integer IDs) so they don't need escaping.

On the backward-compatibility concern (old `execlp()` passed event_id/monitor_id as `$1`/`$2`): intentionally not preserving that. The old `execlp()` treated the entire command string as an executable path, so it was impossible to pass arguments — any working setup was already a bare path like `/usr/bin/myscript.sh` with no args. The `%EID%`/`%MID%` tokens are a better interface. No backward-compat shim to avoid unnecessary complexity.

connortechnology:
Soas not to break existing installs, let's search for a % character and fall back to old behaviour of not found...

connortechnology:
Oh wait I see the comment about intentionally not supporting old functionality... well... That's not smurfy.

pliablepixels:
@connortechnology Done — pushed a fix to preserve backward compatibility.

If the command contains a `%` character, it uses the new token substitution (`%EID%`, `%MID%`, `%EC%`) with `sh -c` execution. If it doesn't, it falls back to the original `execlp()` behavior that passes event_id and monitor_id as `$1` and `$2`.

So existing installs with a bare path like `/usr/bin/myscript.sh` (no tokens) will continue to work exactly as before, while new-style commands with `%EID%` etc. get the token substitution + shell execution.

PR #4619, opened 2026-02-11
fix(build): remove git pull from do_debian_package.sh

pliablepixels:
## Summary

- Remove two `git pull` calls from `utils/do_debian_package.sh`

## Problem

The script has a local-clone code path: when `${GITHUB_FORK}_ZoneMinder.git` exists locally, it clones from that directory instead of fetching from GitHub. However, it then runs `git pull` twice — once in the local `.git` bare repo and once in the release checkout. This overwrites any local commits with whatever is on the remote, silently discarding local changes.

This defeats the purpose of the local-clone path. If you have local commits you want to build (e.g. testing a patch before submitting), the build script throws them away.

## Fix

Remove both `git pull` calls. The script already clones from the local repo — it should build whatever is there. Users who want the latest remote can `git pull` manually before running the build script.


connortechnology:
That's not what git pull does.  It should merge in remote changes but not overwrite local changes.

ZoneMinder_ZoneMinder.git should be a pristine clone.  It was just supposed to prevent always pulling from remote. 

I think we need a different command line parameters to use some other repo to  use and not do a pull from.

PR #4625, opened 2026-02-13
Feat: Merging and moving two identical functions monitorsSetScale from montage.js and watch.js to skin.js

IgorA100:
Also, fixing and optimizing the monitorsSetScale function itself


IgorA100:
@connortechnology 
Please check this PR.
I hope I've migrated all my changes here and there won't be any issues.

PR #4632, opened 2026-02-17
Fix: Event handling for VID (event.js)

IgorA100:
When the "vid.on play" event fires, we don't need to restart playback with "playClicked()"; we just need to redraw the "streamPlay()" buttons. 
Similarly, with "vid.on pause." We use streamPause() instead of pauseClicked().

I'm not 100% sure this is correct, but I think it should be.
Currently, when pressing the buttons in "#dvrControls," Play is called twice and Pause is called twice!


IgorA100:
@connortechnology 
This PR works great for me.

web/skins/classic/views/js/event.js
```diff
@@ -517,8 +517,8 @@ function pauseClicked() {
     vid.pause();
   } else {
     streamReq({command: CMD_PAUSE});
+    streamPause();
```
connortechnology, line 520: Are you sure about this? streamPause updates the buttons state. By moving it up... then it wouldn't do that for video based players... but maybe the events on the video object take care of it? We hould at least put a comment here about that.

IgorA100: > Are you sure about this?

Yes.
For the video player, we call `"vid.pause()"`
Next is the code `"vid.on('pause', streamPause);"`, which calls "streamPause" when the video player pauses.
So everything is fine.
Currently, the button statuses are changing twice.

Issue #4648, opened 2026-02-24
REGRESSION: Show Zones does not work.


Simpler1:
**Describe Your Environment**
ZoneMinder 1.39.0

**Describe the bug**
Show Zones does not work on both the Event page and the Montage page.

**To Reproduce**
1. Go to the Montage page
2. Click on Show Zones
3. The zones are not displayed

**Expected behavior**
Zones should be displayed

IgorA100:
For the Event page, try this PR #4649

connortechnology:
Looks like @IgorA100 is able to reproduce it... all works well for me here though.  Weird.

IgorA100:
Everything worked fine for me too, but the listener assignment was in the wrong place, and the issue could have been spontaneous and depended on many factors (operating system, computer speed, etc.).
I placed the PR in the correct location, and in theory, the problem should have gone away.
In any case, the listener assignment is now called at the right time.
I haven't yet figured out why there might have been a problem with the Montage page. @Simpler1, you need to check the browser console.

Simpler1:
PR #4649 doesn't fix it.

No errors in the browser console on either page.

On the event page, the screen doesn't even refresh when the Show/Hide Zones button is pressed.
On the montage page, the screen refreshes, but no zones are displayed.

IgorA100:
> On the event page, the screen doesn't even refresh
> On the montage page, the screen refreshes, 

This is fine.
I couldn't reproduce the problem.

IgorA100:
@Simpler1 
Check it out #4656

Simpler1:
> [@Simpler1](https://github.com/Simpler1) Check it out [#4656](https://github.com/ZoneMinder/zoneminder/pull/4656)

No change.  Zones are still not displayed in either the event or montage page.  No errors in the console for either.

Simpler1:
This environment to reproduce is easy to set up:
https://github.com/zoneminder-containers/zoneminder-base

IgorA100:
> No change.

This is very, very strange.
I can't reproduce the problem after my 2 PRs yet.
Is your site on a local network or on the internet?
If I had access to your site, I would try to solve the problem.

connortechnology:
The zones are svg placed after the img in the svg and positioned over the img.  It should be trivial to right click on the image, inspect, and find the svg after it and see what it's attributes are and it should highlight in the rendering so you should be able to see where it is.  

Finding out of it is literally not in the html might help.

[...]

PR #4649, opened 2026-02-24
Fix: Assign toggleZones listener only after Event page loads

IgorA100:
Possible partial fix for #4648


IgorA100:
Although this PR does not solve the problem https://github.com/ZoneMinder/zoneminder/issues/4648, it is still more correct and will prevent new problems from appearing in the future....

PR #4662, opened 2026-02-28
Fix: Saving "navbar_type" in cookies

IgorA100:
Also, _REQUEST is more important than _COOKIE
Therefore, we check _REQUEST after checking _COOKIE

### DIFF
--- a/web/index.php
+++ b/web/index.php
@@ -112,20 +112,20 @@ global $navbar_type;
 $navbar_type = isset($_SESSION['navbar_type']) ? $_SESSION['navbar_type'] : '';
 $valid_navbar_types = ['normal'=>1, 'collapsed'=>1, 'left'=>1];
 
+# Cookie overrides session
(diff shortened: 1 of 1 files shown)


web/index.php
```diff
@@ -177,7 +177,16 @@
   $_SESSION['css'] = $css;
   zm_setcookie('zmCSS', $css);
 }
-$_SESSION['navbar_type'] = $navbar_type;
+
+if (
+  !isset($_SESSION['navbar_type']) ||
+  isset($_REQUEST['navbar_type']) ||
```
connortechnology, line 183: I don't think we need these issets here.  We already checked them above, and if there were value above, then navbar_type would be set from either request or cookie.  
We should be able to just test if $vanbar_type is different than session

IgorA100: I did it similarly to "zmCSS"
Okay, I'll study the algorithm in more detail.
At first glance, you might be right and no additional checks are needed. But maybe there are some nuances...

PR #4664, opened 2026-03-01
Fix: For RTSP2Web type=HLS, receive tracks from the stream only at the start of playback, not MEDIA_ATTACHED (MonitorStream.js)

IgorA100:
Otherwise the stream will not be active yet.


IgorA100:
Not relevant due to #4700

PR #4669, opened 2026-03-01
feat: implement Config Requires field for dependent option disable

### LINKED ISSUE #4668: Options UI: implement Config.Requires field for dependent option disable
## Summary

The `Config` table has had a `Requires` column since ZoneMinder v0.9.13 (2003), but it has never been evaluated in the Options web UI. Approximately 50 config entries already have `Requires` values set (across auth, images, logging, mail, upload, web, x10, and system categories), but dependent fields are never disabled when their parent toggle is off.

## Current Behavior

- Unchecking `OPT_USE_AUTH` leaves `AUTH_TYPE`, `AUTH_RELAY`, `AUTH_HASH_*` fields fully editable
- Unchecking `OPT_FFMPEG` leaves `PATH_FFMPEG`, `FFMPEG_FORMATS`, `FFMPEG_*_OPTIONS` fully editable
- Unchecking `OPT_UPLOAD` leaves all `UPLOAD_*` fields fully editable
- Same for `OPT_EMAIL`/`OPT_MESSAGE`, `LOG_DEBUG`, `WEB_LIST_THUMBS`, `WEB_SOUND_ON_ALARM`, `OPT_X10`, etc.

## Expected Behavior

When a parent config is not set to the required value, dependent fields should be disabled (greyed out). Toggling the parent should immediately enable/disable dependents without a page reload.

## Compound Requires

Some entries use compound conditions separated by semicolons, e.g.:
- `ZM_AUTH_HASH_IPS` requires `ZM_OPT_USE_AUTH=1;ZM_AUTH_RELAY=hashed`
- `ZM_SSMTP_MAIL` requires `ZM_OPT_EMAIL=1;ZM_OPT_MESSAGE=1;ZM_NEW_MAIL_MODULES=1`

All conditions must be met (AND logic) for the field to be editable.

## Proposed Fix

- **PHP** (`options.php`): Parse `Requires` on render, evaluate each semicolon-separated condition against current config values, disable field if any condition fails
- **JavaScript** (inline with CSP nonce): On any input change within the options form, re-evaluate all `data-requires` conditions and toggle disabled state live

pliablepixels:
## Summary

- Wire up the long-dormant `Config.Requires` column (added in v0.9.13) so dependent options are disabled when their parent toggle is off
- PHP evaluates `Requires` on page render to set initial disabled state
- Inline JS (with CSP nonce) re-evaluates on any input change for live toggling without page reload
- Supports compound semicolon-separated AND conditions (e.g. `ZM_OPT_USE_AUTH=1;ZM_AUTH_RELAY=hashed`)
- Affects ~50 existing config entries across auth, images, logging, mail, upload, web, x10, and system categories

Fixes #4668

## Test plan

- [ ] Go to Options → Config tab, uncheck OPT_TRAINING → dependent fields disable immediately
- [ ] Options → Auth tab, uncheck OPT_USE_AUTH → AUTH_TYPE, AUTH_RELAY, AUTH_HASH_* all disable
- [ ] Options → Auth tab, check OPT_USE_AUTH but change AUTH_RELAY away from "hashed" → compound deps (AUTH_HASH_IPS, AUTH_HASH_SECRET, etc.) disable
- [ ] Options → Images tab, uncheck OPT_FFMPEG → all FFMPEG fields disable
- [ ] Options → Upload tab, uncheck OPT_UPLOAD → all UPLOAD fields disable
- [ ] Save with parent off, reload page → dependent fields still disabled on fresh render
- [ ] Verify no JS console errors, CSP not blocking the script

### DIFF
--- a/web/skins/classic/views/options.php
+++ b/web/skins/classic/views/options.php
@@ -388,6 +388,9 @@ document.addEventListener('DOMContentLoaded', function() {
   }
 
   $opts.on('change', refresh);
+
+  // Apply initial state so dependents match current values on first render.
+  refresh();
 });
 </script>
 <?php


connortechnology:
We should tell Claude to always use modern js, with const/let instead of var.  Also, things like getVal function in here, we have jquery... $j(thing).val().  Let's have claude clean this up a bit first.

connortechnology:
@copilot code review[agent] instead of making our own getVal() function, this should use $j().val().  Please make a PR against this implementing that

PR #4670, opened 2026-03-01
Chore: Optimized saving of "navbar_type" in session and cookies (index.php)

IgorA100:


web/index.php
```diff
@@ -178,14 +178,9 @@
...
 
-if (
-  !isset($_SESSION['navbar_type']) ||
-  isset($_REQUEST['navbar_type']) ||
-  !isset($_COOKIE['zmNavbar_type']) ||
-  ($_COOKIE['zmNavbar_type'] != $navbar_type)
-) {
-  $_SESSION['navbar_type'] = $navbar_type;
-  zm_setcookie('zmNavbar_type', $navbar_type);
+if (!defined('ZM_FORCE_NAVBAR_TYPE')) {
```
IgorA100, line 181: If we don't check for the presence of ZM_FORCE_NAVBAR_TYPE, the constant's value will always be stored in cookies and the session. But this shouldn't be done, because if a constant is defined, it will be used unconditionally (regardless of the session or cookies). This follows from the code above. `if (defined('ZM_FORCE_NAVBAR_TYPE')) {
`

PR #4701, opened 2026-03-10
Minor change on Zone settings page

Simpler1:
Not a big deal and may only affect me or users who run ZoneMinder on a 1280x1024 monitor with the side menu, but it reduces the width of the settings from 50% to 45 percent so that the graphics and settings appear next to each other instead of on top of each other.

I'm OK if you reject this PR.  It just makes it easier to work with until we can do a more proper layout.

### DIFF
--- a/web/skins/classic/css/base/views/zone.css
+++ b/web/skins/classic/css/base/views/zone.css
@@ -2,8 +2,8 @@
   margin: 2px;
 }
 #settingsPanel {
-	max-width: 600px;
-  width: 50%;
+  max-width: 600px;
+  width: 45%;
   float: right;
 }
 


connortechnology:
I feel like we can do better.  The settings has a fixed width, the rest can be fluid...I'm sure a modern css flexbox layout should work.

connortechnology:
I made changes to auto-shrink the stream.  Lots of other changes too. I think this is no longer needed. let me know what you think.

Simpler1:
I just loaded this up.  This is a much better solution.  

My only concern is the positions of the percent and pixel fields.  I think it would be easier to read if the first column is all percents and the second column is all pixels.  Then you can easily read down the percent or pixel column to see the min/max numbers.

Issue #4703, opened 2026-03-10
Stats on frame of event show way over 100% (zones - Pixels=>Percent)


Simpler1:
**Describe Your Environment**
- Latest commit on github.

**Describe the bug**
Stats on a frame of an event are shown way over 100%

<img width="1207" height="131" alt="Image" src="https://github.com/user-attachments/assets/f8c8a62a-4e40-41ac-9d5d-33cd247424d8" />

**To Reproduce**
1. Open an event
2. Open the frames of the event
3. Click on the Stats button at the top
4. Notice that the percentage calculations are way off

**Expected behavior**
Consistent areas and percents for Alarm Px, Filter Px, Blob Px, Blob Sizes

---

This is still a repercussion of changing zones from pixels to percent.  There are some places that are storing/calculating based on pixels and others based on percent.   Here's a breakdown from Claude:

Bug 1: Threshold comparison (lines 299-304) — min/max_alarm_pixels are stored as percents but compared as raw pixel counts
cpp
`if (min_alarm_pixels && (stats.alarm_pixels_ < (unsigned int)min_alarm_pixels)) {`
`min_alarm_pixels` is loaded from the DB as e.g. 14 (meaning 14%) but `stats.alarm_pixels_` is a real pixel count like 5000. So `5000 < 14` is always false — zones almost never get filtered out.

Bug 2: Score calculation (line 312) — polygon.Area() is correct (pixel-space, since `ParsePercentagePolygon` scales coords properly), but `max_alarm_pixels` used as divisor is still in percent units
`cppstats.score_ = (100*stats.alarm_pixels_)/(max_alarm_pixels ? max_alarm_pixels : polygon.Area());`
When `max_alarm_pixels` is e.g. 38 (percent) and `alarm_pixels_` is 5000, `score = (100*5000)/38 = 13157`. When it falls through to `polygon.Area()` the polygon area IS correct in pixels, so that path would be fine — but max_alarm_pixels being nonzero (almost always true) means it almost always uses the wrong divisor.

The Fix
The DB threshold values are percentages (0–100). They need to be converted to pixel counts at load time, right after the polygon is built. In Zone::Load(), after the polygon is constructed and before zones.emplace_back(...), add the conversion:
cpp
```
// Convert percent-based thresholds to pixel counts now that we have the pixel-space polygon
if (strchr(Coords, '.')) {
  int zone_area = polygon.Area();
[...]

PR #4704, opened 2026-03-11
fix: score calculation always uses polygon area as divisor

Simpler1:
Previously, the score for alarmed/filtered/blob pixels used max_alarm_pixels (or equivalent) as the divisor when nonzero. Since these thresholds are now correctly converted from percent to pixels in Load(), using them as the divisor produced scores relative to the threshold rather than the zone area, and could exceed 100% whenever alarm pixels exceeded the max threshold.

Fix all three score calculations to always divide by polygon.Area(), making score a consistent "percent of zone area alarmed" value (0-100).

Fixes https://github.com/ZoneMinder/zoneminder/issues/4703

### DIFF
--- a/src/zm_zone.cpp
+++ b/src/zm_zone.cpp
@@ -309,7 +309,7 @@ bool Zone::CheckAlarms(const Image *delta_image) {
     return false;
   }
 
-  stats.score_ = (100*stats.alarm_pixels_)/(max_alarm_pixels ? max_alarm_pixels : polygon.Area());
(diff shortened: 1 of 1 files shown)


Simpler1:
This may not be needed with commit 2874751.    That was a different way to fix this.  I'll test with that and see.

Simpler1:
The problem that this PR was solving is now fixed.  Probably with https://github.com/ZoneMinder/zoneminder/commit/2874751c72d1fee0623e512842a73892338f9de5.

PR #4706, opened 2026-03-12
Fix orphaned zms processes accumulating on Montage page tab hide/show

Simpler1:
When the Montage page tab is hidden for more than 15 seconds, the `onvisibilitychange` handler calls `monitors[i].stop()` on all active streams. When the tab becomes visible again, `startVisibleMonitors()` detects `monitor.started === false` and calls `start()`, spawning new zms processes with new connKey values.

The problem is that `stop()` sends `CMD_STOP` to the ZoneMinder API, which does not terminate the zms process. The old zms process continues running with no client consuming its output, no timeout to kill it, and no mechanism to ever clean it up. Each hide/show cycle leaves 3 more orphaned zms processes (one per monitor).

Over time these accumulate until the server exhausts resources, at which point new zms requests begin returning 504 Gateway Timeout errors and the Montage page stops streaming entirely.

This is particularly impactful for users who run the Montage page continuously (e.g. for alarm sound notifications), where the browser's tab visibility API will hide/show the tab frequently due to normal computer use.

### DIFF
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -1402,7 +1402,10 @@ document.onvisibilitychange = () => {
             prevStateStarted = 'played';
             //Stop only if playing (not paused).
             // We might want to continue status updates so that alarm sounds etc still happen
-            monitorStream.stop();
+            // Use kill() instead of stop() to send CMD_QUIT and terminate the zms
+            // process. stop() only sends CMD_STOP which leaves zms running, causing
+            // orphaned processes to accumulate each time the tab is hidden/shown.
+            monitorStream.kill();
           }
         } else {
           prevStateStarted = 'stopped';
--- a/web/skins/classic/views/js/montage.js
+++ b/web/skins/classic/views/js/montage.js
@@ -1046,7 +1046,7 @@ document.onvisibilitychange = () => {
       //closing should kill, hiding should stop/pause
       for (let i = 0, length = monitors.length; i < length; i++) {
         // Stop instead of pause because we don't want buffering in zms
-        monitors[i].stop();
+        monitors[i].kill();
       }
     }, 15*1000);
   } else {
--- a/web/skins/classic/views/js/zone.js
+++ b/web/skins/classic/views/js/zone.js
@@ -793,7 +793,10 @@ document.onvisibilitychange = () => {
     TimerHideShow = setTimeout(function() {
       //Stop monitors when closing or hiding page
       for (let i = 0, length = monitorData.length; i < length; i++) {
-        monitors[i].stop();
+        // Use kill() instead of stop() to send CMD_QUIT and terminate the zms
+        // process. stop() only sends CMD_STOP which leaves zms running, causing
+        // orphaned processes to accumulate each time the tab is hidden/shown.
+        monitors[i].kill();
       }
     }, 15*1000);
   } else {


web/skins/classic/views/js/watch.js
```diff
@@ -1402,7 +1402,10 @@ document.onvisibilitychange = () => {
             prevStateStarted = 'played';
             //Stop only if playing (not paused).
             // We might want to continue status updates so that alarm sounds etc still happen
-            monitorStream.stop();
+            // Use kill() instead of stop() to send CMD_QUIT and terminate the zms
+            // process. stop() only sends CMD_STOP which leaves zms running, causing
+            // orphaned processes to accumulate each time the tab is hidden/shown.
+            monitorStream.kill();
```
connortechnology, line 1408: You didn't read my comment above.  hide/show shouldn't kill zms.  If you kill it then everything stops. No audio alerts, etc.  We just want to stop/pause video streaming.

Alternatively we could switch to monitor status polling instead of working through zms..

connortechnology:
The issue lies elsewhere.  stop() should not alter connKey so that we can later send a play and can continue to poll for monitor status updates.

IgorA100:
> stop() should not alter connKey

stop() doesn't change connKey
In MonitorStream.js, when starting a stream, there's a condition called `if (this.activePlayer == 'zms') {`, but at that point, activePlayer will be empty because we haven't specified it anywhere. We only end up in this place where the above condition is located if we couldn't previously determine which player will be used.
I'm still not entirely clear on these conditions:
```
    if (this.activePlayer == 'zms') {
...
    } else if (-1 != stream.src.indexOf('mode=paused')) {
...
    } else {
...
    }
```

IgorA100:
If you replace
`if (this.started) this.streamCommand(CMD_STOP);`
with
`stream.src = stream.src.replace(/mode=jpeg/i, 'mode=single');`
in MonitorStream.js, the ZMS process is killed and then, in start(), it is started with a new ConKey.
But I don't know why Isaac decided to use `streamCommand(CMD_STOP)` instead of `mode=single`. Perhaps this is to prevent the ZMS process from being killed?

P.S. The check for `if (this.started)` is unnecessary, since we already have a similar check earlier in the code.
P.S. 2 I tried to execute CMD_PLAY after CMD_STOP without changing ConKey, but the video stream does not start and a new ZMS process is created.

connortechnology:
The idea is: pause stops play, but buffers in ram so you can step backwards, etc. It is meant to be short term pause.  STOP stops play, but does not buffer.  It also does not exit, because it is assumed that there will be a subsequent. command. It should also continue to send keep alive frames to keep the stream open.  QUIT means quit outright.  The reason I don't like editing the url, is that it kills the image stream (hopefully).  If zms doesn't die immediately, and we don't change the connkey, the new stream will hang.  So we need a new connKey.  The issue is really just latency.  At the end of the day, we are polishing a turd.  We need to move to a websocket based player.

IgorA100:
>  If zms doesn't die immediately, and we don't change the connkey, the new stream will hang. So we need a new connKey. 

But with the new connKey, the old ZMS process will still exist.
So, after executing CMD_STOP, the old process can't be killed?

connortechnology:
@Simpler1 the diagnosis here was right — processes really do accumulate — but the mechanism turned out to be in three other places, and killing on tab hide takes the status polling and alarm sounds down with it, which is what @connortechnology objected to.

Opened #5120 with fixes for all three:

1. **A stopped `zms` could never exit at all** (`zm_monitorstream.cpp:670`). It slept without ever writing, so a departed client produced no `EPIPE`, and the `continue` also skipped the `ttl` check at the bottom of the loop. Only an explicit `CMD_QUIT` could end it. That is why they accumulate rather than merely appear — this one is the root cause, independent of anything the browser does.
2. **`select_zms()` replaced the connkey without quitting the process it addressed**, so nothing could reach the old `zms` afterwards. The reload path in `getStreamCmdResponse()` already did this correctly; the ordinary start path did not.
3. **The resume branch was unreachable.** `select_zms()` can resume an existing stream with `CMD_PLAY`, but `stop()` clears the `activePlayer` that branch tests, so every re-show built a second `zms` instead of resuming the first.

With those, hide/show resumes the existing process — no kill, no restart latency, alarm sounds uninterrupted — and any `zms` that does get abandoned now cleans itself up.

Thanks for the report and the digging; it pointed straight at the right area.

PR #4707, opened 2026-03-12
Zone min/max layout

Simpler1:
Change min/max layout on zone page for alarmed/filtered/blob areas so that pixel values are lined up in the left column and percent values are lined up in the right column.

Before:
image

After:
image

Also converted all tabs to 2 spaces to be consistent.


IgorA100:
I prefer the current layout using pixel and percentage values; I think it makes more sense.

connortechnology:
THe alignment is visually nice, but I agree with @IgorA100 .  Also I would like them to be on a single line, so we need to adjust spacing/input sizing a bit.   I think overall the zone editing page needs more work but I'm out of ideas and time at the moment.

PR #4722, opened 2026-03-27
Fix: Minor HTML layout issues

IgorA100:
- Removed extra space and extra slash 
- Added missing UL tag
- SPAN tag replaced with DIV tag


jayharv19761-wq:
> - Removed extra space and extra slash 
> 
> - Added missing UL tag
> 
- [ ] > - SPAN tag replaced with DIV tag

web/skins/classic/includes/functions.php
```diff
@@ -99,7 +99,7 @@ function xhtmlHeadersStart($file, $title) {
   <?php #Chosen can't be cache-busted because it loads sprites by relative path ?>
   <link rel="stylesheet" href="skins/classic/js/chosen/chosen.min.css" type="text/css"/>
 <?php
-  echo output_link_if_exists(array('js/noUiSlider-15.8.1/dist/nouislider.min.css?'), false, $param = ['global', 'stylesheet', '  type="text/css"/']);
+  echo output_link_if_exists(array('js/noUiSlider-15.8.1/dist/nouislider.min.css?'), false, $param = ['global', 'stylesheet', ' type="text/css"']);
```
connortechnology, line 102: This may no longer be true, but there was a time where putting the ? at the end meant it wouldn't be cached.  We actually would prefer this to be cached, and should use cache_busting where appropriate. SInce this lib isn't likely to change without a version change, we can just leave it, but maybe remove the ?

web/skins/classic/includes/functions.php
```diff
@@ -427,14 +427,16 @@ function getSidebarTopHTML() {
       <div class="sidebar-footer hidden-for-collapsed">
         <div class="footer-box">
           <div>
+            <ul>
 ' . getAccountCircleHTML($skin, $user, $forLeftBar = true) . '
+            </ul>
           </div>
           <div style="padding: 0 10px">
-            <span style="display: block; margin-bottom: 10px">
+            <div style="margin-bottom: 10px">
```
connortechnology, line 435: I feel like we should remove these hard coded styles and do it in css.  I really prefer semantic styling.

IgorA100: @connortechnology 
Done: #4732
I made the display more correct.

web/skins/classic/includes/functions.php
IgorA100: Done: #4733
In fact, the "?" sign had no effect at all in this case.
Typically, a unique identifier, such as a file modification timestamp, would follow the "?" sign. In this case, when the file was changed, the timestamp would also change, meaning the browser would recognize the file as different and reload it from the server.
In other words, caching was always enabled before.
Now, when the "nouislider.min.css" file is modified, the browser will reload it

PR #4729, opened 2026-04-11
Check for /proc/meminfo before reading its content.

f1rmb:
There is no such proc file system on FreeBSD, hence http server logs a lot of errors.


connortechnology:
In ubuntu questing, systemd hides it by default now too, so this is good.

PR #4730, opened 2026-04-11
Conditionally calls imagedestroy() as it was deprecated since PHP 8.0 doing  nothing except logging errors.

f1rmb:


IgorA100:
@f1rmb
@connortechnology 

Hmm.
I was ignored...
Yes, one bug was fixed, which is great, but we could also fix the memory leak when using PHP>=8.
The garbage collector will of course eventually free up memory when it's running low, but we could speed up this process.
Oh well...

f1rmb:
Well, there is a trick for this (to immediately release the memory, not waiting for PHP good will).
I will raise a PR for this.

PR #4734, opened 2026-04-11
Fix string interpolation and platform where 'ip' command (from iproute2) is unavailable.

f1rmb:
- as from PHP 8.2, ${var} string interpolation is deprecated. the fix looks a bit hackish, but it should parse the correct syntax only.
 - on FreeBSD, the 'ip' command is not available. Also, the old code that parse the 'ifconfig' output looks a bit weird (probably really outdated). The new code works on Linux (6.19.11 here) and FreeBSD. Also, to get the default route interface, I've added a FreeBSD only code section. Later I will add a ZM_PATH_ROUTE in cmake / conf, for now the command is hardcoded, I will also add Linux support for default route detection when 'ip' is not available (unlikely). now get_network() works on FreeBSD too, interfaces array is no more empty.

### DIFF
--- a/web/includes/functions.php
+++ b/web/includes/functions.php
@@ -2259,6 +2259,7 @@ function get_networks() {
             }
         }
         $routes = array();
+        $output = '';
         exec(ZM_PATH_IP.' route', $output, $status);
         if ($status) {
             $html_output = implode('<br/>', $output);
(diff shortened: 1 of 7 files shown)


web/includes/functions.php
```diff
@@ -2249,13 +2249,13 @@ function get_networks() {
...
 	    foreach ( $output as $line ) {
-	      if ( preg_match('/^\d+: ([[:alnum:]]+):/', $line, $matches ) ) {
-          if ( $matches[1] != 'lo' ) {
-            $interfaces[$matches[1]] = $matches[1];
-          } else {
-            ZM\Debug("No match for $line");
-          }
-        }
+              if ( preg_match('/^\d+: ([[:alnum:]]+):/', $line, $matches ) ) {
+                if ( strcmp($matches[1], 'lo') != 0 ) {
```
connortechnology: why is strcmp better here?

f1rmb: Ah, a relic from various changes, which I forgot to revert.
Reverted now.

IgorA100:
@f1rmb
Please format the code (use spaces instead of tabs).

f1rmb:
> @f1rmb Please format the code (use spaces instead of tabs).

Okay. Done. I had to reindent the whole function, as the code is (and still) a mix of tab/space indentation. 
get_network() is now space indented.

IgorA100:
Yes, there is such a problem (spaces/tabs).
We should try to replace all tabs with spaces, including in existing code.

IgorA100:
> get_network() is now space indented.

Your formatting uses 4 spaces, but 99% of the code uses 2 spaces.
It's probably not critical (Isaac has the final say), but it's ugly.

f1rmb:
Well, tell me, I could reformat de whole code. But later, I don't have much time today.

---
73 de Daniel, F1RMB

-------- Original Message --------

connortechnology:
Yes we use 2space formatting. In general we use the google style guidelines, with a few exceptions.

That being said, I don't feel a need to be pedantic about this.  As long as the code you change is formatted correctly and doesn't undo existing good formatting I don't care.

All that being said, I think the entire issue is stupid and I wish the world would move on.  Your editor should display whatever formatting you prefer and the stored code should have almost no unnecessary white space at all.  But I still use vim so I'm part of the problem, and as I age I find that consistent styling, helps my brain parse the code.  So here we are, 2spaces please.

connortechnology:
You may find it useful to setup git to auto-lint the code.  We have an eslint ci setup to that checks js files, but you can set that up in git to do it before you even commit.  Same for c++ code using cpplint etc.  I find that very helpful.

web/skins/classic/views/onvifprobe.php
```diff
@@ -238,10 +238,20 @@ function probeProfiles($device_ep, $soapversion, $username, $password) {
 
   $detprofiles = probeProfiles($monitor['Host'], $monitor['SOAP'], $_REQUEST['Username'], $_REQUEST['Password']);
   foreach ($detprofiles as $profile) {
-    $monitor = $camera['monitor'];
+     $monitor = $camera['monitor'];
+
+    if (PHP_VERSION_ID >= 80200) { // strings ${xxx} are depreated, use {$xxx} instead
```
connortechnology: Instead of all this, how about we just don't use string interpolation?
$sourceString = $profile['Name'].' : '.$profile['Encoding'] ... etc..

f1rmb:
Old school here too, I use Emacs ;-)

---
73 de Daniel, F1RMB

-------- Original Message --------

web/skins/classic/views/onvifprobe.php
f1rmb: Question is, will it work with all PHP version ? (read: I don't want to break anything)

connortechnology: I don't think php will change string concatenation.

f1rmb: Okay.
I will try that later today, and update the PR as well (+ full reindent)

f1rmb: Done.

web/skins/classic/views/onvifprobe.php
```diff
@@ -40,7 +40,7 @@ function execONVIF($cmd) {
         $html_output
         Please run the following command from a command line for more information:
         $shell_command"
-        );
+    );
```
connortechnology: All these whitespace changes should not be in this PR.

web/includes/functions.php
```diff
@@ -2240,59 +2240,82 @@ function i18n() {
...
+                    // Get default route iface
+                    exec("route get default | grep interface | awk '{print $2}'", $defaultIface, $status);
+
+                    if ($status) {
+                        $html_output = implode("\n", $output);
+                        ZM\Error("Unable to get default ip route, status is '$status'. Output was:$html_output");
+                    } else {
+                        if (isset($defaultIface) && isset($defaultIface[0])) {
+	                    $interfaces['default'] = $defaultIface[0];
+                        }
```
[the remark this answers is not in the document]
connortechnology: There are actually security issues around just running things and expecting PATH to find them.  Sometimes the ZM admin is NOT the system admin.  So system admin sets ZM_PATH_IFCONFIG and ZM admin just has to live with it.

f1rmb:
ESLint seems too pedantic, but I won't reindent the whole code (not mine), as I've been told to not do so.

f1rmb:
FYI, I don't get any warning/error on my FreeBSD install now (my codebase is 1.38.1, with a set of patches made from my PRs (merged or awaiting).

PR #4738, opened 2026-04-14
Don't pass null as first parameter to strtotime().

f1rmb:


web/ajax/reports.php
```diff
@@ -190,7 +190,7 @@ function queryRequest($search, $advsearch, $sort, $offset, $order, $limit) {
     $report = new ZM\Report($row);
 
     $row['Name'] = validHtmlStr($row['Name']);
-    $row['StartDateTime'] = $dateTimeFormatter->format(strtotime($row['StartDateTime']));
+    $row['StartDateTime'] = $row['StartDateTime'] ? $dateTimeFormatter->format(strtotime($row['StartDateTime'])) : null;
```
connortechnology, line 193: FYI StartDateTime cannot be null. It is a NOT NULL column.

f1rmb: Here, it's null.
Can you give me some clue of what's wrong on my system ?

connortechnology:
I didn't realise we were looking at Reports.  Yup, can be NULL.  Good check. I merged anyways because it is good practice.

PR #4743, opened 2026-04-19
Feat: Using audio motion analyzer (attempt 2)

IgorA100:
Similar to #4393.
- The HLS player is not yet supported.
- Removed the file "skins/classic/assets/audioMotion-analyzer/src/audioMotion-analyzer.js" because it is licensed under the AGPL-3.0-or-later license.
- Added the ability to work correctly without the file being removed.
- To restore full audioMotion functionality, the user only needs to download and save the file skins/classic/assets/audioMotion-analyzer/src/audioMotion-analyzer.js.
- Added links for manual download of audioMotion-analyzer.js. I'll be adding an automated download process from the UI later.
- In addition to this PR, it is essential to add code to update the database. The code looks something like this:
```ALTER TABLE `Monitors` ADD `WhatDisplay` enum('OnlyVideo','OnlyAudioVisualization','VideoAudioVisualization') NOT NULL DEFAULT 'OnlyVideo' AFTER `Decoding`;```

### CODE REVIEW

**Comment by IgorA100** [comment]
@connortechnology 
Done!
Please check.
Everything should work correctly, whether the audioMotion-analyzer.js file is present or absent.
It is MANDATORY to add the code to update the database
```ALTER TABLE `Monitors` ADD `WhatDisplay` enum('OnlyVideo','OnlyAudioVisualization','VideoAudioVisualization') NOT NULL DEFAULT 'OnlyVideo' AFTER `Decoding`;```
The default monitor settings are stored here.

"What Display" was replaced with "Show."

Regarding moving the display to a side column, I'll consider it, but that would require rotating the analyzer 90 degrees, and I don't know how to do that yet.

**Comment by IgorA100** [comment]
@connortechnology 
Done!
Please check.

It is MANDATORY to add the code to update the database
```ALTER TABLE `Monitors` ADD `WhatDisplay` enum('OnlyVideo','OnlyAudioVisualization','VideoAudioVisualization') NOT NULL DEFAULT 'OnlyVideo' AFTER `Decoding`;```

**Comment by IgorA100** [comment]
I've completed this PR.
Once it's approved (I really hope so), I'll work on the Audio Motion settings.

**Comment by IgorA100** [comment]
- You can position the analyzer vertically to the side of the image, but the low frequencies will be at the top and the high frequencies at the bottom. I don't know yet how to do the opposite (low frequencies at the bottom and high frequencies at the top); I probably need to change the analyzer code.
- If the browser is stretched to almost the entire screen width or we're viewing a video in portrait mode at any browser width, it's better to position the analyzer to the side. If the browser width is half the screen width or less, it's better to position the analyzer at the bottom. Image aspect ratio is also important.
- On smartphones, you also need to analyze portrait or landscape mode.
- Basically, we need to come up with some kind of smart calculation algorithm, as there are many questions and few answers. We need to somehow calculate available space and dynamically insert the analyzer there. We should also consider static placement options at the bottom or side.

**Comment by IgorA100** [comment]
Isaac, have you checked this PR?
Theoretically, everything should be fine, unless I forgot to backport some of my changes.

**Comment by connortechnology** [comment]
Without audiomotion.js we are still showing the Show: options and audio only is an option and none of it works of course. 
Video doesn't play... says paused, may be unrelated.

File: web/skins/classic/assets/version [hunk 1]
```diff
@@ -14,3 +14,6 @@ https://github.com/john-doherty/swiped-events
 ***********************************
 mb.extruder Oct 24, 2018 with magor changes IgorA100
 https://github.com/pupunzi/jquery.mb.extruder
+***********************************
```
connortechnology: Don't need this as we aren't including nit.

connortechnology:
If there is a required db update, then add a zm_update-1.39.whatever.sql file to do it.  There are lots of examples in there.

IgorA100:
> Without audiomotion.js we are still showing the Show:

On which page is this happening (Watch, Event, Montage)?
This shouldn't happen...

web/skins/classic/assets/version
IgorA100: Yes, sure.
Done.

IgorA100:
> then add a zm_update-1.39.whatever.sql file

I created the file "zm_update-1.39.5.sql"

IgorA100:
> Video doesn't play...

Are there any errors in the browser console?
It's strange, but I can't reproduce any of the issues you described yet.
I checked all the files in this PR and they are identical to those on my server.
I'll keep trying to find the problem.

IgorA100:
@connortechnology 
Perhaps you accidentally deleted the file "`skins/classic/js/audioMotionAnalyzer.js`"?
File licensed under AGPL-3: "`skins/classic/assets/audioMotion-analyzer/src/audioMotion-analyzer.js`"

PR #4758, opened 2026-04-27
When using ONVIF, the console page should not display simply "Use ONVIF Events", but should also indicate which event the trigger is configured for. 

IgorA100:

### DIFF
--- a/web/ajax/console.php
+++ b/web/ajax/console.php
@@ -390,7 +390,7 @@ function queryRequest() {
     }
     $row['Analysing'] = isset($monitor['Analysing']) ? $monitor['Analysing'] : 'None';
     $row['Recording'] = isset($monitor['Recording']) ? $monitor['Recording'] : 'None';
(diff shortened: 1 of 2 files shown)


connortechnology:
This is getting to be a lot of info on the console.  Do we really need this here?  I feel like console should list current status more than saved settings... Maybe it should be it's own column and then the user can choose to show/hide it instead.

IgorA100:
> This is getting to be a lot of info on the console.

Yes, you're right. I was actually thinking about creating a toggle setting for "Standard Console" and "Compact Console." The compact console should only display the most important information in a more compact layout.

> Do we really need this here?

When there are many monitors and more than two ONVIF triggers applied to them, it's difficult to understand which monitors are affected by which event triggers. This would be clearer in the PR data. It would be even better to add this to the filter.

> Maybe it should be it's own column

Perhaps, but what exactly should be displayed in this column? In this PR, I'm not adding any new data; I've simply replaced the word "Events" with a word from ONVIF_Alarm_Text (e.g., MotionAlarm). Yes, the word has become a few (5-8) letters longer.

PR #4762, opened 2026-04-28
Now we will pause audioMotion in MonitorStream.js

IgorA100:
This will allow you to pause audioMotion on the Montage page as well.


IgorA100:
I've made all the necessary corrections.

PR #4785, opened 2026-05-02
Feat: Allow national Unicode characters in monitor name

IgorA100:


IgorA100:
@connortechnology 
Are you unhappy with something?

web/includes/actions/monitor.php
```diff
@@ -292,7 +299,7 @@
 
     $saferName = basename($newMonitor['Name']);
     $link_path = $Storage->Path().'/'.$saferName;
-    if (($saferName != $newMonitor['Name']) and !@symlink($mid, $link_path)) {
+    if (($saferName != $newMonitor['Id']) and !@symlink($mid, $link_path)) {
```
connortechnology: this looks wrong

web/skins/classic/views/js/monitor.js.php
```diff
@@ -56,7 +56,7 @@ function validateForm(form) {
     }
   }
 
-  if ( elements['newMonitor[Name]'].value.search( /[^\w\-\.\(\)\:\/ ]/ ) >= 0 )
+  if ( elements['newMonitor[Name]'].value.search( /[^\w\p{L}\-\.\(\):\/ ]/u) >= 0 )
     errors[errors.length] = "<?php echo translate('BadNameChars') ?>";
```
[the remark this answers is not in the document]
connortechnology: The reason for the BadNameChars is specifically for creating a symlink out of the name.  We could strip out badchars when creating the symlink and leave them otherwise. Or we could get rid of that symlink.

web/includes/actions/monitor.php
IgorA100: > this looks wrong

Why?
Previously, we compared the same value before and after processing `basename`. We only created a symlink if any invalid characters were corrected in `$newMonitor['Name']`. If `basename` didn't find any corrections, the symlink wasn't created. And that's not correct.
Now, the symlink won't be created unless the monitor name equals its ID, since a folder with the same ID already exists.

IgorA100: @connortechnology 
So, since the FIRST condition (($saferName != $newMonitor['Name']) wasn't met, we never got around to checking the second condition (!@symlink($mid, $link_path)), which is what actually tries to create the symlink...
Now everything is fixed.

IgorA100:
@connortechnology 
I look forward to your comments and suggestions.

IgorA100:
@connortechnology 
It seems to me that if the monitor's name has changed, then the symlink in all available storage locations should also be changed to the new name.
What do you think, Isaac?

connortechnology:
Yes it should.

IgorA100:
> Yes it should.

Great!
Done: https://github.com/ZoneMinder/zoneminder/pull/4785/changes/9561ed894738a707814f0b25a11420241b89b29d

IgorA100:
@connortechnology 
I made another commit with a description: https://github.com/ZoneMinder/zoneminder/pull/4785/changes/a95efdbd425d20a6091d27d6f0f09b98d3f56da3

Request Copilot review

web/includes/actions/monitor.php
```diff
@@ -175,14 +175,41 @@
...
-          $link_path = $oldMonitor->Storage()->Path().'/'.basename($oldMonitor->Name());
-          if (file_exists($link_path)) {
-            ZM\Debug("Deleting old link  ".$link_path);
-            unlink($link_path);
-          } else {
-            ZM\Debug("Old link didn't exist at ".$link_path);
+          $saferName = basename($newMonitor['Name']);
+          $oldLinkPath = $oldMonitor->Storage()->Path().'/'.basename($oldMonitor->Name());
+          $oldLinkPathFound = false;
+          require_once('includes/Storage.php');
```
connortechnology, line 183: I like requires to all be at the top

web/includes/actions/monitor.php
```diff
@@ -292,7 +319,7 @@
 
     $saferName = basename($newMonitor['Name']);
     $link_path = $Storage->Path().'/'.$saferName;
-    if (($saferName != $newMonitor['Name']) and !@symlink($mid, $link_path)) {
+    if (($saferName != $mid) and !@symlink($mid, $link_path)) {
```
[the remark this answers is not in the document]
IgorA100: Done: https://github.com/ZoneMinder/zoneminder/pull/4785/changes/3de3dc7d2134e757221f4f079fcc2f59b37296dd

IgorA100:
I've followed all the recommendations and can re-request the Copilot check.

IgorA100:
@connortechnology 
I've made the necessary changes to the code.

IgorA100:
Please recheck.

IgorA100:
Please recheck.

IgorA100:
Please recheck.

web/includes/Monitor.php
```diff
@@ -211,7 +211,7 @@ public static function getStreamChannelOptions() {
 
   protected $defaults = array(
     'Id' => null,
-    'Name' => array('type'=>'text','filter_regexp'=>'/[^\w\-\.\(\)\:\/ ]/', 'default'=>'Monitor'),
+    'Name' => array('type'=>'text','filter_regexp'=>'/[^\w\p{L}\-\.\(\)\:\/ ]/u', 'default'=>'Monitor'),
```
connortechnology: The new pattern allows ASCII word chars (`\w`) + Unicode letters (`\p{L}`), but not Unicode digits (`\p{N}`) or combining marks (`\p{M}`). For a feature specifically about national characters, names in scripts that use combining diacritics, or non-ASCII digits (e.g. Arabic-Indic), will still be rejected. Consider `/[^\w\p{L}\p{M}\p{N}\-.()\:\/ ]/u` — `\w` already covers ASCII digits/underscore, so the only real additions are marks + non-Latin digits.

web/ajax/monitor.php
```diff
@@ -0,0 +1,42 @@
...
+      }
+      ajaxResponse(array('response'=>$result, 'monitorName'=>$monitorName, 'cleanedMonitorName'=>$cleanedMonitorName, 'badChars'=>$badChars, 'messageBadNameChars'=>$message));
+    } else {
+      ajaxError(translate('ErrorVerifyingMonitorName'));
+    }
+    break;
+  } // end switch action
+} // end if canView('Monitors')
+
+ajaxError(translate('UnrecognisedAction').' "'.validHtmlStr($_REQUEST['action'] ?? '').'" '.translate('ConjOr').' '.translate('insufficientPermissionsUser').' "'.validHtmlStr($user->Username()).'"');
```
connortechnology: `translate('insufficientPermissionsUser')` (lowercase `i`) doesn't match the lang key `InsufficientPermissionsUser` you added in en_gb/ru_ru. It only resolves via `translate()`'s case-fallback, which returns the value with its first letter lower-cased — a subtle surprise. Use the exact key `InsufficientPermissionsUser`.

web/ajax/monitor.php
```diff
@@ -0,0 +1,42 @@
...
+  switch ( $action ) {
+  case 'validateName' :
+    require_once('includes/Monitor.php');
+    $monitor = new ZM\Monitor($mid);
+    $filterRegexp = $monitor->getDefaults()['Name']['filter_regexp'];
+    $result = true;
+    $badChars = [];
+    $message = '';
+
+    if (isset($_REQUEST['monitorName']) && $_REQUEST['monitorName'] !== '') {
```
connortechnology: `$_REQUEST['monitorName']` can arrive as an array (`?monitorName[]=x`). `trim()` / `preg_replace()` on an array return null and emit warnings (suppressed here by `display_errors=0`), giving confusing results. Add an `is_string($_REQUEST['monitorName'])` guard alongside the `!== ''` check.

web/includes/actions/monitor.php
```diff
@@ -175,14 +175,45 @@
...
+            $dirStoragePath = opendir($storagePath);
+            if($dirStoragePath === false) {
+              ZM\Warning("Unable to check and delete old symlinks to monitor with ID=".$monitor->Id()." in the '".$Storage->Name()."' storage because the path to the storage is missing or unreadable.");
+            } else {
+              while (($file = readdir($dirStoragePath)) !== false) {
+                $linkPath = $storagePath . $file;
+                if (is_link($linkPath)) {
+                  $absolutePath = realpath($linkPath);
+                  if ($oldLinkPath == $linkPath) $oldLinkPathFound = true;
+                  if ($absolutePath == $Storage->Path().'/'.$mid) {
```
connortechnology, line 196: Stale-link detection relies on `realpath($linkPath) == $Storage->Path().'/'.$mid`. Two cases silently skip cleanup: (a) a dangling symlink makes `realpath()` return `false`, so it's never removed; (b) if `$Storage->Path()` has a symlinked parent component it won't match `realpath`'s fully-resolved output. Consider comparing `basename(readlink($linkPath))` against `$mid` (the links are created with a relative `$mid` target), or canonicalize the storage path once before the loop. Not data-destroying — just leaves orphaned links.

web/lang/ru_ru.php
```diff
@@ -607,6 +610,7 @@
     'OptionalEncoderParam' => 'Необязательные параметры кодировщика', // Added - 2019-03-24
     'Options'              => 'Опции',
     'OrEnterNewName'       => 'или введите новое имя',
+    'Or'                   => 'Или',
```
connortechnology: `'Or'` is unused — `ConjOr` (the key the new ajax endpoint actually calls) already exists at line 264. This plus the unrelated `TZUnset` addition in this file look like scope creep; suggest dropping them to keep the translation diff focused on what the PR needs.

connortechnology (commented):
Reviewed the change. Functionally sound and lint-clean (php -l on all PHP incl. the new ajax file, eslint on monitor.js). The JS refactor through `saveMonitorDataPrepare()` and the false "unsaved changes" fix are good improvements, and the symlink rewrite correctly drops the old self-comparison and guards `name == id`. No security regression (output is validHtmlStr/JSON-escaped, alerts are non-HTML, the request path is read-only and the default user keeps `$user->Username()` safe when auth is off).

A few inline points below. The main one worth addressing before merge is the regex coverage (#1), since it's central to the feature. Note also that name validation remains advisory-only: a direct POST to `action=save` still bypasses the regex because `Monitor->save()` never sanitizes Name (pre-existing, not a regression, but this PR could have closed it).

IgorA100:
Isaac, thank you so much for your detailed review and comments.
I'll definitely check everything and make any changes.

IgorA100:
I've made some changes to the code and left comments.
Ready for review.

IgorA100:
Isaac. It would be great if you could fix the code in zm_monitor.cpp so that it would also log the monitor name, taking into account the changes in this PR. Currently, all non-Latin characters are cut off.

`13.06.2026 AD 18:24:50 | zmc_m4 | 1368 | INF | "There should be a name here..."_172 : 3000 - Capturing at 25.80 fps, capturing bandwidth 1296948bytes/sec Analysing at 0.00 fps | zm_monitor.cpp | 2032`

connortechnology:
@IgorA100 see a04cd7a43 .  Switched from utf8 to utf8mb4.  Should fix c++ side.

IgorA100:
> see https://github.com/ZoneMinder/zoneminder/commit/a04cd7a4327989593d05bd123616023a9e015df9

Great!
Thanks Isaac.

PR #4792, opened 2026-05-04
Feat: Consistency in the modal window code for confirming the delete action.

IgorA100:
Similar to #4749

- Moved the getDelConfirmModal() function from options.js to skin.js
- Removed the getDelConfirmModal() function (and its call) from report.js, options.js, reports.js, devices.js, controlcaps.js(?view=options&tab=control), and snapshots.js. This is because it's now a single function in skin.js, which is called when the "Delete" button is pressed, rather than during page initialization.
- Slightly modified the local functions of manageDelConfirmModalBtns(). Removed the listener, so the code is executed immediately.
- You can now call local functions from the global getDelConfirmModal() function (instead of just executing submitThisForm() ). Applies to report.js, reports.js, devices.js, controlcaps.js, snapshots.js
- Changed the style of the "Delete" button on the Groups page
- Added deletion confirmation to the Storage, Roles, and Groups pages
- On the Event page, deletion permissions were broken. Deletes were allowed for those who did NOT have edit permissions, but deletions were blocked for those who did. !
- On the Report page, deletion didn't work and still doesn't work. This needs to be addressed, or maybe it's possible to delete reports only on the Reports page. (FIXED...)
- On the Devices page, the "canEdit.Device" permissions check was being performed, but it should be performed like this: "canEdit.Devices"
- The $action variable was not defined in \ajax\devices.php
- Added translations.
- Added a style for the disabled "Delete" button
- Minor fixes

PR #4793, opened 2026-05-04
Fix broken HTML markup if !$report->FilterId() on Report page

IgorA100:


IgorA100:
The `function deleteRequest($id)` in `\ajax\reports.php` has a `!$report->canEdit()` check, but I couldn't find any access rights settings for Reports.
Is Reports/Report page working correctly?

connortechnology:
Probably not. It was a pretty quick rough in.

IgorA100:
That's what I thought...
Ok.
Some of the code checks permissions for "Events," and others for "Reports."
For now, we'll check for "Events" when deleting Reports. It's probably not entirely correct, but it's better than what we have now.

IgorA100:
Will be implemented here: #4792

PR #4797, opened 2026-05-05
Fix: Display snapshot for incomplete event file (Update image.php)

IgorA100:


web/views/image.php
```diff
@@ -336,7 +336,11 @@ function imagescale($image, $new_width, $new_height = -1, $mode = 0) {
           $path = $Event->Path().'/'.sprintf('%0'.ZM_EVENT_IMAGE_DIGITS.'d', $Frame->FrameId()).'-'.$show.'.jpg';
         } else {
           if ( $Event->DefaultVideo() ) {
-            $file_path = $Event->Path().'/'.$Event->DefaultVideo();
+            if($Event->DefaultVideo() !== 'index.m3u8') {
+              $file_path = $Event->Path().'/'.$Event->DefaultVideo();
+            } else {
+              $file_path = $Event->Path().'/'.$file.find_video($Event->Path());
```
connortechnology, line 342: This is great for now, but if there were multiple /mp4 files in the dir it could be a problem.  I think ideally we would parse the m3u8.  Something to put on todo list.

IgorA100: > but if there were multiple /mp4 files in the dir

I thought about it, but then decided that there can't be multiple MP4 files in a folder.
Okay, so I was wrong.
So I have some questions:
- When can there be multiple MP4 files in one folder?
- Which file should I use to create a snapshot? Right now, it's created from the first MP4 file it finds in the folder.

Issue #4774, opened 2026-04-30
Can no longer see or export current event's video


no1unorightnow:
**Describe Your Environment**
- Version of ZoneMinder [Version: 1.39.5~20260427175441-noble]
- How you installed ZoneMinder [PPA]
- Full name and version of OS: Ubuntu 24.04 (noble)
- Browser name and version (if this is an issue with the web interface):
This issue is browser-independent. Tested Chrome Version 147.0.7727.116 (Official Build) (64-bit), 150.0 (64-bit) (canonical-002 - 1.0) snap

**If the issue concerns a camera**
all cameras / Ffmpeg source with rtsp TCP connection, passthrough video writer with audio, "auto" container

**Describe the bug**
This appears to be two parts, both applicable to "New Event", that is, the most recent event, that has not been completed, but is currently recording per the rolling recording settings (10 minutes)
1. the web view no longer allows scrubbing into the past. So say you are 5 minutes into the 10-minute event, you cannot see what happened 4 minutes ago - only 5 or more minutes ago, or "right now"/live view. In Chrome, it forces the play marker to the end of the video, effectively the same as the "live" view; in Firefox, it just spins a loading symbol over a black screen. No video plays at all.
2. Exporting is broken. Previously, I could export "incomplete.mp4", which would be the current event, up to the point I clicked the Download button. The new behavior is to export a file named "index.m3u8" (which I have found other issues referencing, but not this issue), which is a text file containing partial references to video files.
VLC does not play the file, likely in part due to the fact that no path to the server is exported.
A sample of one of the files I exported is as follows:
```
#EXTINF:1.997,
#EXT-X-BYTERANGE:1065292@1255
index.php?view=view_video&eid=442058&file=incomplete.h264.mp4
#EXTINF:0.083,
#EXT-X-BYTERANGE:1064613@1066547
index.php?view=view_video&eid=442058&file=incomplete.h264.mp4
#EXTINF:1.997,
```
If I replace all instances of "index.php" with "http://fqdn-or-ip-to-server/zm/index.php", then it does play in VLC, but extremely choppily, stuttering audio and video.

**To Reproduce**
1. Go to /zm/?view=console
2. Click on one of the hyperlink numbers under the "Events" column
3. click on the last page number listed
4. click on the hyperlink for "New Event"
5. click on the "download" icon
6. observe the file is index.m3u8 rather than incomplete.mp4, and contains no actual video data, just incomplete references to partial files.

**Debug Logs**
[...]

PR #4800, opened 2026-05-05
Fix: Regexp for replacing index.php with $base_url (view_hls.php)

IgorA100:
With the previous expressions, this shouldn't work at all.
Partial fix for issue #4757
And a possible solution for issue №1 #4774


IgorA100:
There is an error :(

IgorA100:
I corrected my mistake.
I was a bit inattentive.

IgorA100:
I'm closing this PR because another, more global PR #4806 has appeared.

PR #4809, opened 2026-05-06
Feat: Added codec="MP4HLS" on Event page

IgorA100:
Let's try playing any recorded events, not just incomplete ones, using HLS.

### DIFF
--- a/web/skins/classic/views/event.php
+++ b/web/skins/classic/views/event.php
@@ -90,7 +90,7 @@ if (isset($_REQUEST['showZones'])) {
 $codecs = array(
   'auto'  => translate('Auto'),
   'MP4'   => translate('MP4'),
-  'MP4HLS'=> translate('MP4 HLS'),
+  'MP4HLS'=> ['Name'=> 'MP4 HLS', 'disabled'=> (!file_exists($Event->Path() . '/index.m3u8'))],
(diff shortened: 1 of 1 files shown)


IgorA100:
I would completely remove the `str_ends_with($Event->DefaultVideo(), '.m3u8')` condition that was previously in place and allow the user to choose how to view the incomplete event. Currently, users can't natively view the MP4 file of an incomplete event; they always view the incomplete event using the HLS player.

IgorA100:
@connortechnology 
I look forward to your comments and feedback.

web/skins/classic/views/event.php
```diff
@@ -101,8 +102,8 @@
 } else {
   $codec = $monitor->DefaultCodec();
 }
-if (!isset($codecs[$codec])) {
-  ZM\Warning("Invalid value for Codec: $codec, reverting to auto");
+if (!isset($codecs[$codec]) || (is_array(($codecs[$codec])) && $codecs[$codec]['disabled'])) {
+  if (!isset($codecs[$codec])) ZM\Warning("Invalid value for Codec: $codec, reverting to auto");
```
connortechnology, line 106: I don't see the need test isset again.  I would split the two cases with a different warning for is disabled.

PR #4810, opened 2026-05-07
Eliminate Event_Summaries / bucket-table deadlocks (zmstats, zmaudit, Event::delete, zmDbDo retry)

connortechnology:
## Summary

Eliminates the recurring InnoDB deadlock cycle observed under load between `zma`'s `event_update_trigger` path, `zmstats.pl`'s Event_Summaries maintenance, `zmaudit.pl`'s aggregations, and `zmfilter.pl` (`Event::delete`).

The root cause was a multi-table `UPDATE Event_Summaries es JOIN Events_Hour ...` pattern: MariaDB takes S-locks on the joined bucket rows and holds them to TX commit *regardless of isolation level*, which inverts the canonical lock order

```
Events[Id] -> buckets[EventId] -> Event_Summaries[MonitorId]
```

against the trigger writers that hold X-locks on those same bucket rows. The InnoDB deadlock log captured the exact `zma` vs `zmstats` cycle.

### What changes

* **`zmDbDo` auto-retries on errno 1213** (5 attempts, exponential backoff with jitter) when called outside an explicit transaction. Most call sites are autocommit-style and already idempotent on retry.
* **`Event::delete`** sets `READ COMMITTED` for the per-event delete TX, runs Stats/Event_Data/Frames/Events deletes, and retries the whole TX on 1213. Also fixes a latent `commit()` -> `rollback()` bug on the error path.
* **`zmstats.pl` Event_Summaries maintenance** is rewritten to:
  1. Prune aged rows from `Events_Hour/Day/Week/Month` (RC, no gap locks).
  2. Snapshot per-monitor counts/disk-space via plain `SELECT` (consistent read at RC, no locks).
  3. Apply one `UPDATE Event_Summaries` per `MonitorId` using the snapshotted values.
  
  This keeps the lock prefix (`buckets -> Event_Summaries`) consistent with the trigger path and never holds bucket S-locks while writing ES. Wrapped in 5-retry deadlock loop.
* **`zmaudit.pl`** receives the same snapshot-then-per-row treatment for its Event_Summaries and Storage `DiskSpace` recomputations, replacing five correlated/JOINed UPDATEs that had the same lock-ordering hazard.
* **`db/triggers.sql`** picks up a comment block above `event_update_trigger` documenting the canonical lock acquisition order so future writers don't reintroduce the cycle.

### Why isolation level alone wasn't enough

`SET TRANSACTION ISOLATION LEVEL READ COMMITTED` drops the next-key/gap locks on plain DELETE/UPDATE range scans, but a multi-table UPDATE *always* takes S-locks on the JOINed rows for as long as the transaction lives. The structural fix had to be removing the JOIN, not lowering isolation.

### Notes for reviewers

* No schema changes. Pure script + Perl module + retry-loop changes plus one comment block in `triggers.sql`.
* The `zmDbDo` retry is conditional on `AutoCommit` — it will not silently re-run statements that are part of an explicit caller transaction (those callers are responsible for their own retry, e.g. the `Event::delete` loop).
* `Event::delete`'s isolation set + retry path is bypassed when the caller is already inside a transaction (`$in_transaction` short-circuit), so callers that wrap multiple deletes in their own TX keep their semantics.

## Test plan

- [ ] Build clean (`cmake --build build`) — Perl scripts go through cmake `configure_file`.
- [ ] `perl -c` on the touched modules with a real install (placeholders substituted).
- [ ] Run `zmstats.pl`, `zmaudit.pl` against a busy install and confirm no `Deadlock found` errors in `/var/log/zm/`.
- [ ] Force contention by deleting a hot event via filter while `zma` is updating the same monitor; confirm `Event::delete` retries succeed and the row is removed.
- [ ] Verify Event_Summaries totals match recomputed `SUM(DiskSpace)/COUNT(*)` from the bucket tables after a full `zmstats` cycle.
- [ ] Confirm `zmaudit.pl --interactive` Storage DiskSpace UPDATE still produces correct totals.

### DIFF
--- a/db/triggers.sql
+++ b/db/triggers.sql
@@ -116,9 +116,21 @@ FOR EACH ROW
 
 drop procedure if exists update_storage_stats//
 
+/* ============================================================================
+ * Lock-acquisition order for the Events update/delete triggers (and for the
+ * scripts that touch the same tables). InnoDB X-locks the matched Events row
+ * during WHERE evaluation, before either BEFORE or AFTER trigger bodies fire,
+ * so the order is the same regardless of trigger timing:
+ *   Events[Id] -> Events_Hour/Day/Week/Month[EventId] -> Event_Summaries[MonitorId]
+ * zmstats.pl prune+resync follows the matching prefix (bucket DELETEs then
+ * UPDATE Event_Summaries) and crucially does NOT pre-lock Event_Summaries —
+ * pre-locking ES would invert against the trigger body order and reintroduce
+ * deadlocks against filter / zma writers. The bucket update/delete triggers
+ * also propagate into Event_Summaries[MonitorId] in the same direction.
+ * ============================================================================ */
 drop trigger if exists event_update_trigger//
 
-CREATE TRIGGER event_update_trigger AFTER UPDATE ON Events 
+CREATE TRIGGER event_update_trigger AFTER UPDATE ON Events
 FOR EACH ROW
 BEGIN
   declare diff BIGINT default 0;
--- a/db/triggers.sql
+++ b/db/triggers.sql
@@ -127,7 +127,7 @@ drop procedure if exists update_storage_stats//
  * Writers that follow this order (and must continue to):
  *   - event_update_trigger (AFTER UPDATE on Events)
  *   - event_delete_trigger (BEFORE DELETE on Events)
- *   - src/zm_event.cpp Event::createNotification path: INSERT Events, then
+ *   - The Event::Event constructor in src/zm_event.cpp: INSERT Events, then
  *     INSERT Events_Hour/Day/Week/Month, then INSERT/UPDATE Event_Summaries
  *     (event_insert_trigger is commented out below; zmc does it directly)
  *   - The bucket update/delete triggers cascade into Event_Summaries in the
--- a/db/triggers.sql
+++ b/db/triggers.sql
@@ -117,16 +117,27 @@ FOR EACH ROW
 drop procedure if exists update_storage_stats//
 
 /* ============================================================================
- * Lock-acquisition order for the Events update/delete triggers (and for the
- * scripts that touch the same tables). InnoDB X-locks the matched Events row
- * during WHERE evaluation, before either BEFORE or AFTER trigger bodies fire,
- * so the order is the same regardless of trigger timing:
+ * Canonical lock-acquisition order for every writer that touches Events,
+ * the bucket tables, and Event_Summaries. InnoDB X-locks the matched Events
+ * row during WHERE evaluation, before either BEFORE or AFTER trigger bodies
+ * fire, so the order is the same regardless of trigger timing:
+ *
  *   Events[Id] -> Events_Hour/Day/Week/Month[EventId] -> Event_Summaries[MonitorId]
- * zmstats.pl prune+resync follows the matching prefix (bucket DELETEs then
- * UPDATE Event_Summaries) and crucially does NOT pre-lock Event_Summaries —
- * pre-locking ES would invert against the trigger body order and reintroduce
- * deadlocks against filter / zma writers. The bucket update/delete triggers
- * also propagate into Event_Summaries[MonitorId] in the same direction.
+ *
+ * Writers that follow this order (and must continue to):
+ *   - event_update_trigger (AFTER UPDATE on Events)
+ *   - event_delete_trigger (BEFORE DELETE on Events)
+ *   - src/zm_event.cpp Event::createNotification path: INSERT Events, then
+ *     INSERT Events_Hour/Day/Week/Month, then INSERT/UPDATE Event_Summaries
+ *     (event_insert_trigger is commented out below; zmc does it directly)
+ *   - The bucket update/delete triggers cascade into Event_Summaries in the
+ *     same direction
+ *   - zmstats.pl prune+resync (bucket DELETEs then UPDATE Event_Summaries)
+ *   - zmaudit.pl resync (bucket SELECTs then UPDATE Event_Summaries)
+ *
+ * Crucially: do NOT pre-lock Event_Summaries before touching the bucket
+ * tables — that inverts the order and reintroduces the deadlock cycle
+ * against zma/filter/zmc writers.
  * ============================================================================ */
 drop trigger if exists event_update_trigger//
 
--- a/scripts/ZoneMinder/lib/ZoneMinder/Database.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/Database.pm
@@ -263,15 +263,28 @@ sub _sql_with_bind_values {
 }
 
 # Basic execution of $dbh->do but with some pretty logging of the sql on error.
+# Auto-retries on deadlock (errno 1213) only when AutoCommit is on, since
+# inside a caller-managed transaction the caller has to rebuild the whole TX.
 sub zmDbDo {
-	my $sql = shift;
-  my $rows = $dbh->do($sql, undef, @_);
-	if ( ! defined $rows ) {
-		Error('Failed '._sql_with_bind_values($sql, @_).' : '.$dbh->errstr());
-  } elsif ( ZoneMinder::Logger::logLevel() > INFO ) {
+  my $sql = shift;
+  my @params = @_;
+  my $max_attempts = $dbh->{AutoCommit} ? 5 : 1;
+  my $rows;
+  for ( my $attempt = 1; $attempt <= $max_attempts; $attempt++ ) {
+    $rows = $dbh->do($sql, undef, @params);
+    last if defined $rows;
+    if ( ($dbh->err() // 0) == 1213 and $attempt < $max_attempts ) {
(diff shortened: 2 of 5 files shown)


connortechnology:
Pushed two follow-up commits addressing the review feedback:

- **`19c5bcbbd` docs: correct Event::delete lock-order comment** — fixes both review comments about the inconsistent comment vs. trigger definition. The block in `Event.pm` previously claimed `event_update_trigger` was changed to `BEFORE UPDATE`; that change was considered then reverted (lock order doesn't depend on trigger timing — InnoDB X-locks the matched row during WHERE evaluation before either trigger fires), but the comment never got rewritten. Now matches the chain documented in `triggers.sql`: `Events[Id] -> buckets[EventId] -> Event_Summaries[MonitorId]`.
- **`943114da9` fix: don't zero Event_Summaries counters on aggregate SELECT failure in zmaudit** — addresses the third review comment. Now tracks per-aggregate success and only updates the column groups whose SELECT succeeded; if every aggregate fails the resync is skipped entirely with an error logged, and the next zmaudit pass corrects it.

connortechnology:
Pushed `0dcd500b7` addressing the new review comment: the bucket-prune SELECT now checks `$dbh->err()` and bails the same way the DELETE error path does. (Same hazard wasn't present in the snapshot/aggregate sections — those already check explicitly.)

connortechnology:
Addressed the four new review comments:

- **`628bcb3e8` fix: don't log Error on deadlock inside a caller-managed transaction** — `zmDbDo` no longer emits the misleading `Error('Failed ... Deadlock found')` when called from `Event::delete` / `zmstats` prune+resync TXs; those have their own outer retry loops. Now a `Debug` line in that path; `Error` still fires for non-1213 failures and for autocommit calls that exhaust retries.
- **`7b4617b1a` docs: document atomicity tradeoff in zmstats/zmaudit ES resync** — agreed the snapshot-then-update split is non-atomic; that's intentional. The trigger writers maintain ES incrementally between resyncs, drift is bounded by the resync interval and corrected on the next pass, and locking ES before reading aggregates would invert the canonical order and re-introduce the deadlock we just removed. Comment block on both scripts now spells this out.
- **`e254ed41d` docs: expand triggers.sql lock-order block with full writer list** — added `src/zm_event.cpp`'s create path (`event_insert_trigger` is commented out below, so zmc inserts Events then bucket rows then Event_Summaries directly), the bucket triggers cascading into ES, and `zmaudit.pl` alongside `zmstats.pl`. The doc now enumerates every writer obligated to follow the canonical order, not just the Events-level triggers.

PR is now at 11 commits.

connortechnology:
Five new comments addressed in four follow-up commits:

- **\`d01300832\` fix: log Error in Event::delete when deadlock retries are exhausted** — addresses the "silent return after 5 failed attempts" hazard. zmDbDo intentionally suppresses Error on 1213 inside a caller-managed TX, so Event::delete now logs its own Error (with errstr captured before rollback) on the exhausted-retries path.
- **\`1323e9f02\` docs: reference the actual Event::Event constructor in lock-order block** — fixed the bogus \`Event::createNotification\` symbol; the bucket+ES insert sequence lives in \`Event::Event\` at \`src/zm_event.cpp:46\`.
- **\`52d5f80e0\` docs: clarify per-monitor UPDATE lock state in zmstats resync** — agreed the previous comment conflated statement-level locks with TX-level locks. Rewrote it to distinguish the TX's already-held bucket and ES X-locks (acquired in canonical order by the earlier DELETEs and their cascading triggers) from the new statement-level UPDATE.
- **\`ccc0bcf82\` fix: track per-row UPDATE failures in zmaudit ES and Storage resync** — both \`aud_print('Finished ...')\` messages now reflect actual completion: skipped aggregate column groups and per-row \`zmDbDo\` failures are counted and reported instead of being masked. An Error is logged if Storage resync didn't complete.

connortechnology:
Five new comments addressed in three follow-up commits (the zmstats commit folds K + L together because the touched-monitors restriction fixes both at once):

- **\`411b6916b\` fix: avoid DB-backed logging inside zmDbDo when caller manages the transaction** — the most serious of the batch. \`ZoneMinder::Logger->logPrint\` does \`INSERT INTO Logs\` on the same \`\$dbh\`, so calling \`Debug()\`/\`Error()\` from zmDbDo's failure path inside a caller-managed TX would execute another statement, clearing the err/errstr the caller needs for rollback/retry. A caller could then read err=0 after a deadlock-victim TX and \`commit()\` what looks like success but is actually a rolled-back no-op. Now zmDbDo bails silently when AutoCommit is off; the caller's own retry loop handles logging.
- **\`8fb4ad79e\` perf: restrict zmstats ES resync to monitors with pruned buckets** — addresses both the "full Events_Month scan every 60s" workload concern *and* the "X-lock every ES row every pass" contention concern. Capture MonitorIds during the prune SELECT, skip the resync entirely if nothing was pruned, and restrict the aggregate SELECTs (\`WHERE MonitorId IN (...)\`) and per-monitor UPDATEs to the touched set. zmaudit remains the deep-resync safety net for drift in untouched monitors. Also captures errstr before rollback so the gave-up Error reports the actual reason on drivers that clear errstr on rollback.
- **\`d098a055b\` fix: track Event_Summaries enumerate failure as partial resync in zmaudit** — without it, the enumerate failure left \`@skipped\` empty and the audit log still claimed a full resync while monitors with no current bucket rows never got zeroed. Now tracked alongside the per-aggregate flags and surfaced in the partial-resync message.

connortechnology:
Four new comments addressed in three follow-up commits:

- **\`01b142531\` fix: skip zmDbDo success Debug when inside a caller-managed transaction** — covers comment Q. The success-path Debug had the same hazard as the failure-path Debug already fixed: \`Logger->logPrint\` INSERTs into \`Logs\` on the same \`\$dbh\`, so inside a caller-managed TX it added a write and could clobber err/errstr. Now gated on \`AutoCommit\`.
- **\`5183fb647\` fix: use \$dbh->do for SET TRANSACTION to preserve the next-TX isolation level** — covers comments O and P. \`SET TRANSACTION ISOLATION LEVEL\` applies to the very next transaction. With DB debug logging enabled, \`zmDbDo\`'s success Debug INSERT INTO Logs became that "next transaction" and silently consumed the directive — so the actual delete/prune TX ran at default isolation, not READ COMMITTED. Both \`Event::delete\` and \`zmstats\` now call \`\$dbh->do\` directly for \`SET TRANSACTION\` (which can't deadlock, so \`zmDbDo\`'s retry was no benefit anyway).
- **\`07d50a8fe\` fix: CAS-style Storage DiskSpace update in zmaudit to preserve concurrent adjustments** — covers comment R. Unlike Event_Summaries (trigger-maintained), Storage.DiskSpace has no incremental DB-level maintenance — Event::delete and event-finalize do their own +/- in application code. The previous snapshot+overwrite could undo a concurrent \`Event::delete\` adjustment, making accounting transiently wrong. Now reads current DiskSpace alongside the aggregate and uses MariaDB null-safe equality (\`<=>\`) in the WHERE so a concurrent writer's newer value blocks the overwrite. Deferred and failed counts are reported separately in the audit log.

connortechnology:
Four new comments — two were obsolete and two needed fixes:

- **T** (Event.pm SET TRANSACTION via zmDbDo) and **U** (zmstats.pl.in SET TRANSACTION via zmDbDo) — already fixed in **\`5183fb647\`** before this review batch landed. Both files now call \`\$dbh->do\` directly for \`SET TRANSACTION\` so zmDbDo's success Debug INSERT can't become the "next transaction" that consumes the isolation directive.

- **\`673d5a933\` fix: chunk zmstats bucket-prune DELETE IN-list to avoid packet-size limits** — addresses V. The PK-based \`DELETE WHERE EventId IN (?, ?, ...)\` is deliberate (precise per-row PK locking is what keeps the canonical lock order intact; a predicate DELETE would range-lock the bucket index and undo the deadlock work). So instead of switching pattern, chunk the EventId list into 1000-row batches.

- **\`25da6cce9\` fix: CAS-guard Event_Summaries UPDATE in zmaudit to preserve concurrent writes** — addresses S. zmaudit now reads current ES values alongside the aggregate snapshot, skips rows that already match (no X-lock for no-op writes), and guards each UPDATE with null-safe equality (\`<=>\`) on every column it writes. A concurrent trigger writer between our snapshot and UPDATE causes the WHERE to miss, leaving their adjustment in place. CAS-deferred / no-op / failed counts are reported separately in the audit log.

  zmstats does *not* need the same treatment because its UPDATE runs inside a TX that already holds the bucket X-locks gating the trigger writers that update its column set (Hour/Day/Week/Month). zmaudit doesn't hold those locks, hence the CAS.

connortechnology:
Pushed `8a58971be` addressing the BIGINT precision comment. Both the Event_Summaries and Storage skip-if-already-correct checks now compare scalars as strings (`"$a" eq "$b"`) instead of numerically. DiskSpace columns above 2^53 (~9 PB) could otherwise collapse to the same double under Perl's `!=` and cause a real skip-forever drift bug — distinct BIGINTs comparing equal as doubles, deterministically, so the next pass would do the same wrong thing.

PR #4835, opened 2026-05-16
Make HLS event playback actually work end-to-end (fixes #4757)

### LINKED ISSUE #4757: Viewing recorded events using HLS player?
@connortechnology 
I think you wrote code for viewing recorded events using the HLS player, but I couldn't find how to set it up anywhere.
Or am I wrong?

connortechnology:
Eleven commits that fix everything between recording an event and a browser scrubbing it back via HLS. Supersedes #4803 and #4806; closes #4757.

### C++ — m3u8 generation produced duplicate / mis-pointed fragments
`zm_videostore.cpp` snapshotted `avio_tell()` *before* `av_interleaved_write_frame()`, but with `movflags=frag_keyframe` the mov muxer flushes fragment N *inside* that call when keyframe N+1 arrives. So `last_fragment_offset_` described the next fragment while the bytes at that offset hadn't been written yet. `writeM3U8` then push_back'd a tentative entry per live update, and the next keyframe push duplicated it with the correct duration. Result: every fragment past the first appeared twice in the manifest (once with bogus `0.040s`, once correctly), and the final fragment was either missing or pointed at the previous fragment's bytes. Players seek-jerked backwards. Fixed in `0e532cd77` by snapshotting after the write; a new `VideoStore::finalize()` runs the trailer logic and parses the trailing `mfro` box so the last fragment's actual byte range can be recorded.

### PHP — `view_hls.php` produced double-prefixed URLs
The regex captured `index.php?…` in `$1` and prepended `$base_url . '?'`, emitting `/zm/index.php?index.php?view=…` which drops the `view=` route param. Players fetched the manifest as if it were the mp4 fragment. Fixed in `275b675ac` by capturing only the query string.

### PHP — `view_video.php` advertised m3u8 filename when serving mp4 bytes
The fallback resolves `$path` to the actual mp4, but `Content-Disposition` still used `DefaultVideo`, so the download button saved a manifest. Fixed in `20da5e5f1` by deriving the filename from `$path`.

### PHP — closed-event detection bypassed HLS
The original `$has_hls` check required `DefaultVideo` to end in `.m3u8`, but closed events get their `DefaultVideo` rewritten to the renamed mp4. Fixed in `10a4f9af1` / `375e82fd8` / `5c0532fe3` by keying HLS selection off on-disk `index.m3u8` plus a (MP4HLS / auto) codec choice. Explicit MP4 stays native; MJPEG short-circuits earlier via `$video_tag`.

### PHP — `liveui:true` blocked scrubbing
video.js's liveui replaces the seekbar with a live-edge-only control for in-progress events, making it impossible to seek back through the already-recorded portion. Fixed in `e378da3d1` by keeping the standard seekbar.

### C++ — VideoStore lifecycle bugs surfaced by `finalize()`
- `4182a829d` drains `reorder_queues` inside `finalize()` ahead of the trailer write (the destructor's drain would otherwise push packets into a closed AVFormatContext).
- `4aa62a4d8` makes `VideoStore::filename` a `std::string` (the caller renames + reassigns the source `video_incomplete_path`, dangling any `const char*` finalize() held into it for `fopen()`); also adds `oc` null guards in the destructor.
- `2c…` (latest) refactors the destructor to just call `finalize()` so the queue-drain / trailer / close logic lives in one place.

## Supersedes

- #4803 — view_hls.php regex fix
- #4806 — same plus view_video.php and liveui

## Test plan

- [ ] Record a fresh event, open it before it closes — m3u8 loads, scrubbing works
- [ ] Let the event close, reopen — HLS player still picked, m3u8 still loads, byte-range mp4 segments fetch with correct `Content-Type` / `Content-Disposition`
- [ ] Download button saves the mp4 file (not the manifest)
- [ ] Inspect a generated `index.m3u8` — each fragment listed exactly once, byte ranges are contiguous, durations match keyframe intervals (not `0.04s`)
- [ ] Open an older closed event (DefaultVideo is the mp4, not index.m3u8) — PHP fallback picks HLS when manifest exists
- [ ] Pick MP4 explicitly from the codec dropdown on an event whose monitor records H.264 — native mp4 playback, no HLS fetch
- [ ] Pick MJPEG explicitly — MJPEG stream, no HLS fetch

### DIFF
--- a/src/zm_event.cpp
+++ b/src/zm_event.cpp
@@ -202,9 +202,12 @@ Event::~Event() {
   /* Close the video file */
   // We close the videowriter first, because if we finish the event, we might try to view the file, but we aren't done writing it yet.
   if (videoStore != nullptr) {
-    // Finalize last fragment before closing the video store
+    // Flush the trailer + record the final fragment before writing the m3u8.
+    // finalize() must run before writeM3U8 so the manifest contains every
+    // fragment (including the one no later keyframe was around to close).
+    videoStore->finalize();
+
     std::string m3u8_path = path + "/index.m3u8";
-    // Write temporary m3u8 with incomplete filename (writeM3U8 finalizes last fragment)
     std::string video_url_tmp = "index.php?view=view_video&eid=" + std::to_string(id)
       + "&file=" + video_incomplete_file;
     videoStore->writeM3U8(m3u8_path, video_url_tmp, true);
--- a/src/zm_event.cpp
+++ b/src/zm_event.cpp
@@ -258,26 +258,17 @@ Event::~Event() {
   if (frame_data.size()) WriteDbFrames();
 
   uint64_t video_size = 0;
-  bool has_m3u8 = false;
   DIR *video_dir;
   if ((video_dir = opendir(path.c_str())) != NULL) {
     struct dirent *dir_entry;
     while ((dir_entry = readdir(video_dir)) != NULL) {
       struct stat vf_stat;
       if (stat((path + "/" + dir_entry->d_name).c_str(), &vf_stat) == 0 &&
-          S_ISREG(vf_stat.st_mode)) {
+          S_ISREG(vf_stat.st_mode))
         video_size += vf_stat.st_size;
-        if (!strcmp(dir_entry->d_name, "index.m3u8")) has_m3u8 = true;
-      }
     }
     closedir(video_dir);
   }
-
-  // Prefer index.m3u8 as DefaultVideo when an HLS manifest was written, so
-  // event.php picks the HLS player on closed events too — otherwise the player
-  // selection logic falls back to direct mp4 playback for everything except a
-  // manual MP4HLS codec choice.
-  std::string default_video = has_m3u8 ? "index.m3u8" : video_file;
 
   // Use async dbQueue instead of synchronous zmDbDoUpdate to avoid blocking
   // the close_event_thread (which blocks the analysis thread on the next closeEvent).
@@ -295,7 +286,7 @@ Event::~Event() {
       frames, alarm_frames,
       tot_score, static_cast<uint32>(alarm_frames ? (tot_score / alarm_frames) : 0),
       max_score, max_score_frame_id,
-      default_video.c_str(),
+      video_file.c_str(), // defaults to ""
       video_size,
       id);
   dbQueue.push(std::move(sql));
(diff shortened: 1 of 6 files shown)


connortechnology:
Thanks @Copilot — both real bugs.

**reorder_queues** (4182a829d): moved the drain into `VideoStore::finalize()` ahead of the trailer write, and gated the destructor's drain on `!finalized_` so it still runs when nobody calls finalize().

**DefaultVideo='index.m3u8'** (2e83b8b07): reverted. You're right that `getImageSrc()`, `Length()`, `Filesize()`, and `Video()` in `web/includes/Event.php` all treat DefaultVideo as a local file passed to ffmpeg/filesize, and our manifest's relative `index.php?…` segment URLs aren't resolvable from there. The PHP fallback in event.php (codec=auto / codec=MP4HLS keyed on on-disk `index.m3u8`) covers closed-event HLS selection without needing to rewrite DefaultVideo.

connortechnology:
Thanks @Copilot — all three real bugs.

**filename use-after-free** (4aa62a4d8): `VideoStore::filename` is now a `std::string` so it owns its storage. `Event::AddPacket_()` does rename the incomplete file and reassign `video_incomplete_path`, which would have dangled the previous `const char*` pointer the moment finalize() reached its `fopen(filename, …)` for mfro parsing.

**Null oc in destructor** (same commit): added `oc` null check before the trailer/avio block and around the reorder_queue drain, since the queue drain ultimately calls into `av_interleaved_write_frame(oc, …)` which would crash if open() bailed before allocating oc.

**$has_hls overrides explicit MP4** (5c0532fe3): dropped the `str_ends_with(DefaultVideo, '.m3u8')` clause. In-progress events get DefaultVideo='index.m3u8' from the constructor, so that clause silently overrode the user's MP4 choice. $has_hls is now purely (manifest on disk) AND (codec is MP4HLS or auto). MJPEG short-circuits earlier via $video_tag.

connortechnology:
Thanks @Copilot.

**Code duplication** (3b03ec5f1): refactored the destructor to just call `finalize()`. finalize() is already idempotent (returns early when finalized_ is set or oc was never allocated), so the legacy "caller didn't call finalize" path and the open()-failed-early path both work through the same shutdown logic now. Destructor only handles resource deallocation.

**DefaultVideo confusion**: the PR description was stale — the C++ `DefaultVideo='index.m3u8'` rewrite you flagged in your earlier review was reverted in 2e83b8b07. Updated the description to reflect that. The destructor's UPDATE still writes `video_file` (the renamed mp4); HLS routing for closed events is handled entirely by the PHP fallback in event.php.

PR #4847, opened 2026-05-28
If we're trying to view a remote event, we shouldn't pollute the logs

IgorA100:
For example, we were viewing an event, but the page in the browser wasn't closed, and in the meantime, the recorded event was deleted. In this case, there will be a persistent error in the logs.
The code on the Event page probably also needs to be changed.

- Added the "eventNotFound" JS constant for the Event page.
- If a recorded event isn't found (eventNotFound == true), then print a message to the browser console and stop further script processing, as this is pointless, but it will help avoid unnecessary warnings and errors.

### DIFF
--- a/web/skins/classic/views/js/event.js.php
+++ b/web/skins/classic/views/js/event.js.php
@@ -17,6 +17,7 @@
 // PHP variables to JS
 //
 var connKey = '<?php echo $connkey ?>';
+const eventNotFound = <?php echo ((!$Event->Id() || ($Event->Id() and !file_exists($Event->Path()))) ? "true" : "false") ?>;
 
 var eventData = {
 <?php if ( $Event->Id() ) { ?>
--- a/web/skins/classic/views/js/event.js
+++ b/web/skins/classic/views/js/event.js
@@ -1359,6 +1359,10 @@ function onStatsResize(vidWidth) {
 }
 
 function initPage() {
+  if (eventNotFound) {
+    console.log("Event was not found.");
+    return;
+  }
   getAvailableTags();
   getSelectedTags();
 
--- a/web/views/view_video.php
+++ b/web/views/view_video.php
@@ -83,7 +83,7 @@ if ( $errorText ) {
 } 
 
 if ( ! ($fh = @fopen($path, 'rb') ) ) {
-  ZM\Error('Can\'t open video at '.$path);
+  ZM\Debug('Can\'t open video at '.$path);
   header('HTTP/1.0 404 Not Found');
   die();
 }


web/skins/classic/views/js/event.js.php
```diff
@@ -17,6 +17,7 @@
 // PHP variables to JS
 //
 var connKey = '<?php echo $connkey ?>';
+const eventNotFound = <?php echo ((!$Event->Id() || ($Event->Id() and !file_exists($Event->Path()))) ? "true" : "false") ?>;
```
connortechnology, line 20: The only problem here is that the event might exist on a different server.  I would want to look for the file on the server that owns the storage area that the event is stored on.  So..
If Event->Storage>ServerId() != currentServerId then ajax->checkforExistence().

IgorA100: > checkforExistence().

But how can we check a file on another server?
- We don't have access to the other server's file system, right?
- We haven't authenticated on the other server, and we can't pass the authorization token and name in the query string?

The multi-server configuration is still a mystery to me...

IgorA100: @connortechnology 
Moreover, I don't understand how a reference to a recorded event on another BB server could have ended up in the AA server database. I don't think that's possible...

connortechnology: They share the same db

IgorA100: > They share the same db

Oh, even so... I see, interesting, I didn't know that...
But I still have a question: what query can I use to check for a recorded event on another server? Isaac, have you done this before? If so, how?

IgorA100: And again, I don't understand anything...
We have one storage for all servers, which means $Event->Path() will return the correct address.
For example:
there is `Storage_A` and `Storage_B`
there is `Server_A` and `Server_B`
`Storage_A` is connected to `Server_A` at `/mnt/A/S_A`
`Storage_A` is connected to `Server_B` at `/mnt/B/S_A`
`Storage_B` is connected to `Server_B` at `/mnt/B/S_B`
- When requesting an event stored on `Storage_A` from` Server_A`, `$Event->Path()` should return `/mnt/A/S_A`
- When requesting an event stored on `Storage_A` from `Server_B`, `$Event->Path()` should return `/mnt/B/S_A`
- When requesting an event stored on `Storage_B` from `Server_A`, `$Event->Path()` should return nothing, because `Storage_B` isn't connected to `Server_A`. And we have no way to verify the presence of the recorded event.
Isn't that right?

connortechnology: You have to use the API on the assigned server.  Ideally all servers would have access to all storage, but sometimes that isn't true.  Some people want to use a server just for the UI, multiple recording servers, etc.  Sometimes located very differently on the network with slow unreliable network.  So each recording server has it's own local storage which is reliable for recording, but may not be reliable (or performant) for playback.  Which is why we can assign a ServerId to Storage to tell us where the best place to access it is.  You have to ask the API on the server assigned to the Storage.

IgorA100: Thanks for the clarification.

PR #4853, opened 2026-05-28
The "Id" in the "Monitor" object is now public instead of protected. This allows us to retrieve Monitor.Id.

IgorA100:
I also replaced tabs with spaces.

### DIFF
--- a/web/includes/monitor_probe.php
+++ b/web/includes/monitor_probe.php
@@ -1029,7 +1029,7 @@ function probeNetwork() {
       'IP'          => $camera['ip'],
       'camera'      => $camera,
       'Monitor'     => $monitor,
-      'MonitorId'   => ($monitor) ? $monitor->Id() : '',
+      'MonitorId'   => $monitor ? $monitor->Id() : '',
     ];
   } # end foreach stream
   } # end foreach mac
--- a/web/includes/monitor_probe.php
+++ b/web/includes/monitor_probe.php
@@ -1029,7 +1029,6 @@ function probeNetwork() {
       'IP'          => $camera['ip'],
       'camera'      => $camera,
       'Monitor'     => $monitor,
-      'MonitorId'   => $monitor ? $monitor->Id() : '',
     ];
   } # end foreach stream
   } # end foreach mac
--- a/web/includes/monitor_probe.php
+++ b/web/includes/monitor_probe.php
@@ -126,7 +126,7 @@ function probeV4L() {
 } # end function probeV4L
 
 function probeAxisCommunicationsAB($ip, $username, $password) {
-	return probeAxis($ip, $username, $password);
+  return probeAxis($ip, $username, $password);
 }
 
 // Probe Network Cameras
@@ -167,7 +167,7 @@ function probeAxis($ip, $username, $password) {
       }
     }
   } else {
-	  ZM\Debug("No content from $url");
+    ZM\Debug("No content from $url");
   }
   $cameras[] = $camera;
   return $cameras;
@@ -610,12 +610,12 @@ function probeFoscam($ip, $username, $password) {
   $rtsp_port = 0;
   $http_port = 0;
   if (port_open($ip, 554)) {
-	  $rtsp_port = 554;
+    $rtsp_port = 554;
   }
   if (port_open($ip, 88)) {
-	  $http_port = 88;
+    $http_port = 88;
   } else if (port_open($ip, 80)) {
-	  $http_port = 80;
+    $http_port = 80;
   }
   if (!$rtsp_port) $rtsp_port = $http_port;
   $cameras = [];
@@ -921,14 +921,14 @@ function probeNetwork() {
       ZM\Debug("Have match for $ip $mac $macRoot ".$macBases[$macRoot]['type']);
       $macBase = $macBases[$macRoot];
       if ($filter_manufacturer and ($filter_manufacturer != $macBase['type'])) {
-	      ZM\Debug("Continuing because offilter $filter_manufacturer == ".$macBase['type']);
-	      continue;
+        ZM\Debug("Continuing because offilter $filter_manufacturer == ".$macBase['type']);
+        continue;
       } else {
-	      ZM\Debug("Not Continuing because offilter $filter_manufacturer != ".$macBase['type']);
+        ZM\Debug("Not Continuing because offilter $filter_manufacturer != ".$macBase['type']);
       }
       if (function_exists('probe'.$macBase['type'])) {
         if (!$username and isset($monitors[$ip])) {
-		$monitor = $monitors[$ip];
+          $monitor = $monitors[$ip];
           ZM\Debug("Using auth from monitor $ip ".$monitor->User().' '. $monitor->Pass());
           $new_cameras = call_user_func('probe'.$macBase['type'], $ip, $monitors[$ip]->User(), $monitors[$ip]->Pass());
           if (!$new_cameras) {
@@ -1023,12 +1023,13 @@ function probeNetwork() {
     }
 
     $results[] = [
-	    'mac' => $mac,
+      'mac' => $mac,
       'description' => $sourceString,
       'url'         => (isset($camera['monitor']) ? $camera['monitor']['Path'] : ''),
       'IP'          => $camera['ip'],
       'camera'      => $camera,
       'Monitor'     => $monitor,
+      'MonitorId'   => ($monitor) ? $monitor->Id() : '',
     ];
   } # end foreach stream
   } # end foreach mac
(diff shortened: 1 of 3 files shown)


web/includes/monitor_probe.php
```diff
@@ -1023,12 +1023,13 @@ function probeNetwork() {
...
 
     $results[] = [
-	    'mac' => $mac,
+      'mac' => $mac,
       'description' => $sourceString,
       'url'         => (isset($camera['monitor']) ? $camera['monitor']['Path'] : ''),
       'IP'          => $camera['ip'],
       'camera'      => $camera,
       'Monitor'     => $monitor,
+      'MonitorId'   => ($monitor) ? $monitor->Id() : '',
```
connortechnology: Don't need the parenthesis

connortechnology:
@copilot in what circumstance would stream.Monitor.Id not be populated, and hence require this change?

connortechnology:
What the code does today

  probeNetwork() only sets 'Monitor' => $monitor when:
  $monitor = null;
  if (isset($monitors[$ip])) {
    $monitor = $monitors[$ip];  // line 1019
  } 
  
  $monitors[$ip] is populated at lines 867–884 from ZM\Monitor::find(['Type'=>'Remote']) and find(['Type'=>'Ffmpeg']). _find (Object.php:140-141) does new $class($row) where $row is a SELECT * row from the Monitors table. The
  constructor (Object.php:35-37) iterates the row and sets each column as a dynamic public property:
  foreach ($row as $k => $v) {
    $this->$k = $v;
  } 
  
  The Monitors table has an Id column (capital I — matches PHP's case-sensitive property access). ZM_Object carries #[\AllowDynamicProperties] so PHP 8.2+ doesn't strip them. json_encode($monitor) includes dynamic public properties. The
   JS sees Stream.Monitor.Id as the DB row's Id.

  There's no codepath I can find where $monitor->Id would be unset but $monitor is truthy.

  Curious internal inconsistency
  
  add_monitors.js:99 (in parseStreams, unchanged by the PR) already does:
  html += ' is already entered into the system by Monitor ' + stream.Monitor.Id + ' ' + stream.Monitor.Name + '<br/>';
  
  If Monitor.Id were truly missing, the displayed HTML would read "Monitor undefined undefined" — which the maintainer would probably have noticed by now. So either both work and the PR is solving a non-bug, or both are broken and the
  existing HTML rendering displays "undefined" silently.

  Reasonable hypotheses if the bug is real
  
  1. PHP version / AllowDynamicProperties not picked up — if running on a PHP version where the attribute is being ignored, dynamic properties might not serialize. But the rendered HTML at line 99 should still be broken too, so this
  seems unlikely if only the URL path fails.
  2. bootstrap-table mutates the row object between data.Streams[i] assignment to ProbeResults (line 49) and addMonitor reading it back (line 124). The table init could shallow-clone or restructure rows. Possible but not common.
  3. A formatter or extender registered on the table strips object properties to JSON-stringify them as cells.

  Recommendation

  Either:
  - Ask the author for console.log(Stream) output in the failing case. That'll show exactly what's missing and either confirm a real bug or reveal it's not what they think.
  - If you want to merge it anyway — the change is harmless (adds a flat field, slightly belt-and-braces in the if check, no functional regression), and the whitespace cleanup makes monitor_probe.php a bit more consistent (it had mixed
  tabs/spaces). Just note in the merge that the root cause isn't pinpointed.
  
  I'd push back gently on the PR description though — the title says "Correct operation" implying a known bug, but the diff doesn't actually explain what was broken or why Stream.Monitor.Id was missing. A console-log proof would close
  the loop.

IgorA100:
> add_monitors.js:99 (in parseStreams, unchanged by the PR) already does:

Yes, I saw that. I didn't change it because this code is called either when importing from a CSV file or from the `getProbeResponse()` function. But `getProbeResponse() `isn't called anywhere.
Most likely, `stream.Monitor.Id` isn't working, and in the line` html += ' is already entered into the system by Monitor ' + stream.Monitor.Id + ' ' + stream.Monitor.Name + '<br/>';` but since I can't verify this, I didn't make the change.

IgorA100:
> The Monitors table has an Id column

Yes, that's right. But we're not passing the "Monitor" table, but a PHP "Monitor" object, in which the Id is protected.
Here's an example:
```
    [Id:protected] => 8
    [Name] => HeroSpeed_2Mp
```

IgorA100:
Basically, if we don't use `Monitor.Id` as `protected` in \includes\Monitor.php (i.e. remove the line `protected $Id;`), then we'll most likely be able to access `Monitor.Id` in the future. I didn't do this because I don't know what consequences this might lead to and why someone previously made `Monitor.Id` protected.

IgorA100:
I don't use AI to write code or text, and I have trouble with English. 
But I hope you understand me, Isaac? :)

web/includes/monitor_probe.php
IgorA100: My habit is to always wrap conditions in brackets.
Done: https://github.com/ZoneMinder/zoneminder/pull/4853/changes/212b743a14780c605209b98c063737a84cd473fa

IgorA100:
> add_monitors.js:99 (in parseStreams, unchanged by the PR) already does:
html += ' is already entered into the system by Monitor ' + stream.Monitor.Id + ' ' + stream.Monitor.Name + '
';

I've made some changes to the code. https://github.com/ZoneMinder/zoneminder/pull/4853/changes/0df258be3a6e60b6300b0250f1b995fc2f32e7ed
I can't test this change, but it's highly likely the right decision.

connortechnology:
Claude Says: Igor is right — there is a real bug. Root cause confirmed:

  The bug

  web/includes/Monitor.php:377 declares:
  protected $Id;
  
  PHP json_encode skips protected properties (only public + dynamic-public are serialized). So when the probe response is built, every Monitor field that came in via the constructor's foreach ($row as $k => $v) { $this->$k = $v; }
  becomes a dynamic public property (because AllowDynamicProperties is set on ZM_Object) — except for Id, which gets assigned to the explicit protected declaration that already exists. Visibility stays protected, and Id is silently
  dropped from the JSON.

  Igor's print_r output [Id:protected] => 8 is exactly that — PHP's notation for a protected property holding the value 8.

  Why he didn't update parseStreams (his comment)

  He noted that parseStreams() at line 99 (the other Stream.Monitor.Id reference) is only called from getProbeResponse(), which is dead code (no callers). I confirmed:
  - grep getProbeResponse web/ → only definitions / unused.
  
  So stream.Monitor.Id at line 99 is broken too, just hidden because that codepath only runs for the import-CSV flow (line 156) and the unused getProbeResponse(). The visible bug is in addMonitor() which fires from the always-displayed
  Edit button.

  History
  
  The protected $Id; declaration was added in e11b700fb (2023-09-13, "Add Id as a protected member, only lookup monitor status if Id has a value") by Charlie Root. The commit's actual purpose was to add an if ($this->Id()) guard around
  Monitor_Status lookups — for that, the property needed to exist so Id() could return a defined value cleanly. The protected visibility looks incidental and not load-bearing.

  What I'd recommend over PR #4853's approach

  The one-line fix is just protected $Id; → public $Id; (or delete the declaration entirely, letting the constructor's assignment create a public dynamic property like every other column).

  That fixes Monitor.Id everywhere — not just in addMonitor() — including the dead-but-not-deleted parseStreams() HTML at line 99, plus any future caller that legitimately reads monitor->Id from a JSON-encoded Monitor.

  The PR's approach (adding flat MonitorId) works for the addMonitor case but leaves the root-cause invisible everywhere else. Including:

  grep -rn "\.Monitor\.Id\b\|Monitor\['Id'\]" web/skins/classic/ | grep -v "newMonitor" | head

  Probably surfaces more callers that have been silently getting undefined.

  Reply suggestion for the PR:
  
  ▎ Igor — nice catch on the root cause, that print_r output is the smoking gun. Rather than the flat MonitorId workaround though, can we just change protected $Id; to public $Id; on Monitor.php:377? That fixes the same bug at the 
  ▎ source, also fixes the line in parseStreams that you noticed but didn't change, and avoids adding a parallel field that every future serialization site has to remember to populate. The original protected was incidental to the Id() 
  ▎ guard added in e11b700f and not load-bearing.
  
  There's a possible argument for keeping the flat MonitorId as belt-and-braces, but my preference would be the structural fix.

Isaac says: Let's make it public.

IgorA100:
> Isaac says: Let's make it public.

If AI and Isaac don't mind, of course I'll do that!
Actually, I did a quick search to find other places in the code where Monitor.Id would be undefined, but I couldn't find any.

Great, I'll make Id public and hopefully it won't cause any problems later.

IgorA100:
> The protected $Id; declaration was added in https://github.com/ZoneMinder/zoneminder/commit/e11b700fbd3df34df123478d17e13f30acf1dca9 (2023-09-13)

For 2.5 years, no one noticed that the "Edit" button wasn't working :)
Actually, I discovered this a year ago, but then forgot about it and didn't investigate the code and the cause of the error.

> Igor — nice catch on the root cause

Thanks.
Sometimes it's nice to be a little smarter than the AI ​​:)))

I've made the necessary changes to the code.

PR #4858, opened 2026-05-29
fix: widen DefaultScale columns to preserve fit_to_width

### LINKED ISSUE #4750: DefaultScale column truncates fit_to_width to fit_to (CHAR(6))
The `Monitors.DefaultScale` and `MonitorPresets.DefaultScale` columns are defined as `CHAR(6)` but the scale option `fit_to_width` is 12 characters. Saving "Fit to Width" truncates the stored value to `fit_to`, which is not a valid key in the `$scales` array, producing warnings like:

```
WAR [Specified value fit_to not in contents: Array(... [fit_to_width] => Fit to width ...)
```

Repro:
1. Edit any monitor, set Viewing -> Default Scale to "Fit to Width", save.
2. View an event for that monitor.
3. Warning is logged via `web/includes/functions.php:437` (htmlOptions), called from `web/skins/classic/views/event.php` via `htmlSelect('scale', $scales, $scaleSelected, ...)`.

Fix: widen `DefaultScale` to `VARCHAR(16)` in schema and add a migration to widen the column and convert existing `fit_to` values to `fit_to_width`.

AJ0070:
Widen `Monitors.DefaultScale` and `MonitorPresets.DefaultScale` from `CHAR(6)` to `VARCHAR(16)` so `fit_to_width` is stored correctly.

Adds a DB migration to widen both columns and convert existing truncated `fit_to` values to `fit_to_width`, plus a regression test covering the shipped schema and migration.

resolves #4750


AJ0070:
@connortechnology

PR #4862, opened 2026-05-31
If the requested Object isn't found in the database, generate a Warning instead of an Error. (Object.php)

IgorA100:
It seems to me that if the requested Object wasn't found in the database, it's not a fatal error, but simply a warning. Perhaps it should even be logged as a Debug level.
Isaac, I could be wrong, of course. If I'm wrong, just close this PR.

### DIFF
--- a/web/includes/Object.php
+++ b/web/includes/Object.php
@@ -24,7 +24,7 @@ class ZM_Object {
         $table = $class::$table;
         $row = dbFetchOne("SELECT * FROM `$table` WHERE `Id`=?", NULL, array($IdOrRow));
         if (!$row) {
-          Error("Unable to load $class record for Id=$IdOrRow");
+          Warning("Unable to load $class record for Id=$IdOrRow");
           return;
         }
       } else {


connortechnology:
I agree with warning instead of error.  I'd have to think about Debug.  If we are loading by id an object that no longer exists, why do we still have a reference to it?  Deleting an object should remove all references.

IgorA100:
>  If we are loading by id an object that no longer exists, why do we still have a reference to it?

This reference could be anywhere: a bookmark or browser window, email, notepad, etc.
This means that it's usually an outdated reference that was once valid, but now the object has been deleted and the reference is no longer valid.
Handling such situations (when the reference is invalid) should likely be managed in specific modules, not in the global "Object" class.

connortechnology:
This almost never happens on any of the servers I manage.  It should only happen in the case of a crash in which case zmaudit.pl should clean it up.

IgorA100:
> This almost never happens on any of the servers I manage.

My English is probably bad...
You had a link to a saved event http://XXXXX/zm/index.php?view=event&eid=1908498 and you passed it on to someone else. By the time they clicked the link, the event had already been deleted from the database and physically stored on disk.
Or you left a link to a recorded event open in your browser window, and after some time, the event was deleted from the database and disk, and you're trying to open the link again.
So, this problem only occurs when clicking a previously saved link (not a newly generated one by the server) that has become invalid.
So, the database entry will be missing, and the event itself will no longer be physically stored on disk.

PR #4867, opened 2026-06-02
When executing the streamCmdStop() command, always stop the stream regardless of the passed "action" parameter on Watch page

IgorA100:
We only call streamCmdStop() without passing the "action" parameter in one place: when pressing the "STOP" button.
It makes sense to also stop the stream when pressing the "STOP" button. Otherwise, attempts will be made to start playback of the stream, which is not intended.

### DIFF
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -210,7 +210,7 @@ function streamCmdPlay(action) {
   }
 }
 
-function streamCmdStop(action) {
+function streamCmdStop() {
   monitorStream.onplay = false; //Without this line, "onPlay" is triggered immediately due to "if (this.onplay) this.onplay();" in MonitorStream.js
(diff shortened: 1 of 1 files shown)


connortechnology:
I think action was from a previous time.  Seems like it has no use anymore and should just go away. I think the idea was to differentiate between "update the buttons" and "actually stop the stream".  But update the buttons should be a separate function then.

IgorA100:
> Seems like it has no use anymore and should just go away. 

Yes, I wanted to do that too, but then I decided not to touch it and wait for your opinion, Isaac.
Ok, I'll remove `"action"` from the parameters of` streamCmdStop()`

IgorA100:
Done: https://github.com/ZoneMinder/zoneminder/pull/4867/commits/ad18228b2e07c45b63743112d6f2ac5e0a56f545

PR #4884, opened 2026-06-04
Slightly sped up DOM rendering when switching monitors on Watch page

IgorA100:
Clearing via `textContent` is performed without page parsing, which speeds up cleanup. 
When executing `innerHTML`, clearing will first occur by parsing the page, and then assigning new elements. 
This will slightly speed up DOM rendering by quickly clearing.

### DIFF
--- a/web/skins/classic/views/js/watch.js
+++ b/web/skins/classic/views/js/watch.js
@@ -944,6 +944,7 @@ function streamReStart(oldId, newId) {
   el.removeEventListener('mouseleave', handleMouseLeave);
 
   //Change main monitor block
+  monitor_div.textContent = '';
   monitor_div.innerHTML = currentMonitor.streamHTML;
 
   const volumeControls = document.getElementById('volumeControls'+oldId);


connortechnology:
copilot doesn't like your change. I'm not sure.  I'm a little surprised that that would give any noticeable speed increase.  Would be nice if we could just say turn off dom updates until we are ready to redraw....  maybe there is a way. I just don't know it.  

github has gotten stupid about copilot billing.  I'm going to integrate my paid Claude subscription instead.

IgorA100:
> Would be nice if we could just say turn off dom updates until we are ready to redraw....

I don't yet know what criteria to use to determine whether we're ready for a repaint.
99% of the time, we're ready by default because we've switched to the next monitor. Because if the DOM isn't rendered, playback won't start.
In theory, we don't need a DOM repaint only in one case: when we switch between monitors very quickly (less than 1 second).

> copilot doesn't like your change.

I once tested the speed of clearing using `textContent = ''`, and it worked noticeably faster than` innerHTML = ''`.
I haven't tested it yet, but it seems to me that redrawing has become faster.
In any case, `innerHTML = 'ХХХХХХХ'` first completely clears the parsing process and only then adds new elements. Considering that we're removing not just one or two DOM elements, but significantly more, there should be a performance benefit.

> github has gotten stupid about copilot billing. I'm going to integrate my paid Claude subscription instead.

Heh....
Does each check cost money, or are there unlimited plans for a certain period?

IgorA100:
Okay, that was a spontaneous idea, let's cancel it for now.
There's no time to delve deeper into this right now.

PR #4896, opened 2026-06-09
Fix: Correct operation on the Monitor page with "Source Type = Web site"

IgorA100:
- The absence of "#zones-tab" should not result in errors on the Monitor page. (monitor.js)
- When getting the value for "form.elements['newMonitor[RTSPStreamName]']" and "form.elements['newMonitor[Recording]']", check for the presence of the elements themselves. (monitor.js.php)

### DIFF
--- a/web/skins/classic/views/js/monitor.js
+++ b/web/skins/classic/views/js/monitor.js
@@ -313,37 +313,40 @@ function initPage() {
 
   var sourceFormMonitor = $j('#contentForm').serialize();
   // Manage the ZONES Button
-  document.getElementById("zones-tab").addEventListener("click", function onZonesClick(evt) {
-    if ($j('#contentForm').serialize() !== sourceFormMonitor) {
-      evt.preventDefault();
-      const data = {
-        request: "modal",
-        modal: "saveconfirm",
-        key: messageSavingDataWhenLeavingPage
-      };
+  const zonesTab = document.getElementById("zones-tab");
+  if (zonesTab) {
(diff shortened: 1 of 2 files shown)


IgorA100:
Many places contain an analysis of Source Type = Web site, but in zm_monitor.cpp, this monitor type is not among the allowed types.
```
  if (!strcmp(dbrow[col], "Local")) {
    type = LOCAL;
  } else if (!strcmp(dbrow[col], "Ffmpeg")) {
    type = FFMPEG;
  } else if (!strcmp(dbrow[col], "Remote")) {
    type = REMOTE;
  } else if (!strcmp(dbrow[col], "File")) {
    type = FILE;
  } else if (!strcmp(dbrow[col], "NVSocket")) {
    type = NVSOCKET;
  } else if (!strcmp(dbrow[col], "Libvlc")) {
    type = LIBVLC;
  } else if (!strcmp(dbrow[col], "VNC")) {
    type = VNC;
  } else {
    Fatal("Bogus monitor type '%s' for monitor %d", dbrow[col], id);
  }
```
There's a discrepancy somewhere...

connortechnology:
It is an artificial type.  Not implemented c++ side.  We actually should implement it I think.  But the way it works is it just embeds a frame next to other monitors so you embed web-based data alongside video. It's a hack.

IgorA100:
> It's a hack.

But it's useful :)
If you can fix this on the C++ side, I'll fix what's not working in the Web interface.

This PR should at least fix the Web interface bugs first.

Issue #2466, opened 2019-01-24
Log Injection


Loginsoft-Research:
**Describe Your Environment**
- ZoneMinder v1.33.1
- Installed from - ppa:iconnor/zoneminder-master 

**Describe the bug**
This attack targets the log files of the target host. The attacker injects, manipulates or forges malicious log entries in the log file, allowing him to mislead a log audit, cover traces of attack, or perform other malicious actions. The target host is not properly controlling log access. As a result tainted data is resulting in the log files leading to a failure in accountability, non-repudiation and incident forensics capability

Most components of ZoneMinder can emit informational, warning, error and debug messages in a standard format. These messages can be logged in..

Example of these messages is:

`03/01/06 13:46:00.166046 zmpkg[11148].INF [Command: start]`

The first part refers to the date and time of the entry, the next section is the name (or an abbreviated version) of the script, followed by the process id in square brackets, a severity code (INF, WAR, ERR or DBG) and the debug text.

All the mentioned information can be manipulated & inject by an attacker by enticing the victim to visit a specially crafted link, which in turn will inject a custom Log message provided by the attacker. The same will be reflected in the `Log` view page

**To Reproduce**
  Affected URL :
http://localhost/zm/index.php

POST Data - `browser[name]=chrome&browser[platform]=win&browser[version]=41&file=admin&level=INF&line=128&message=User 'admin' Logged in&page=1&request=log&task=create`

- Navigate to the Affected URL, Custom Log will be injected. 
- Navigate to the http://localhost/zm/index.php?view=log, Log will be visible.

<img width="737" alt="log injection" src="https://user-images.githubusercontent.com/47003572/51708285-86eea900-2049-11e9-9b27-9565c3c1798f.png">

**Expected behavior**
- Do not allow tainted data to be written in the log file without prior input validation. Whitelisting may be used to properly validate the
data.

**Debug Logs**
```

None

```

mnoorenberghe:
> The first part refers to the date and time of the entry, the next section is the name (or an abbreviated version) of the script, followed by the process id in square brackets, a severity code (INF, WAR, ERR or DBG) and the debug text.
> 
> All the mentioned information can be manipulated & inject by an attacker by enticing the victim to visit a specially crafted link, which in turn will inject a custom Log message provided by the attacker. 

This is a bit of an overstatement… we already validate the log level, the date/time come from the server, and the component will always be "web_js". Therefore, the only ones that can be spoofed are:
[...]

PR #4912, opened 2026-06-13
fix: strip control characters from client-supplied web_js log fields

SteveGilvarry:
## Summary

Defense-in-depth hardening for the `request=log&task=create` endpoint (`web/ajax/log.php`), relating to the long-standing log injection report #2466.

`createRequest()` passed the client-supplied `message` and `file` straight to `logPrint()`. `logPrint()` only strips *trailing* newlines (`/[\r\n]+$/`), so embedded newlines survived into the flat text log file and could forge what looked like additional log lines. (Notably, `logPrint()` even carries the comment *"Didn't we already replace all newlines with spaces above?"* — the intent was to strip them, but the regex only handles the trailing case.)

This strips control characters (newlines, CR, tabs, etc.) from `message` and `file` before logging:

```php
$file = preg_replace('/[\x00-\x1F\x7F]+/', ' ', $file);
$message = preg_replace('/[\x00-\x1F\x7F]+/', ' ', $_POST['message']);
```

## Context — this is hardening, not a fix for an exploitable hole

The endpoint is already well-guarded, which is why #2466 is being closed as not a current vulnerability:
- Requires an authenticated **System:Edit** user; `ZM_LOG_INJECT` (which would relax that) defaults to **off**.
- Component is hardcoded to `web_js` — daemon entries cannot be spoofed.
- Level is validated, line is `validInt`, file strips the URL scheme/host.
- The Log view (`skins/classic/views/js/log.js` → `processRows()`) escapes `<`/`>`, so the earlier stored-XSS variant (exploit-db 51071, < 1.37.24) no longer applies.

A log message is a single line and the Log view stores it as one `Message` column, so this change has no effect on legitimate use.

## Testing

`php -l web/ajax/log.php` clean. No build required (PHP). Behaviour unchanged for normal single-line messages; embedded control characters are collapsed to spaces.

### DIFF
--- a/web/ajax/log.php
+++ b/web/ajax/log.php
@@ -49,28 +49,36 @@ ajaxResponse($data);
 //
 
 function createRequest() {
-  if (!empty($_POST['level']) && !empty($_POST['message'])) {
-    ZM\logInit(array('id'=>'web_js'));
+  // Every field here is attacker-controlled. Coerce each to a scalar and strip
+  // control characters (newlines, carriage returns, tabs, etc.) so nothing can
+  // forge additional lines in the text log file, and so non-scalar input (e.g.
+  // message[]=x) cannot trip a TypeError in logPrint(). A log message is a
+  // single line; the Log view stores it as one Message column regardless.
+  $sanitize = function($value) {
+    return is_scalar($value) ? preg_replace('/[\x00-\x1F\x7F]+/', ' ', (string)$value) : '';
+  };
 
-    $file = !empty($_POST['file']) ? preg_replace('/\w+:\/\/[\w.:]+\//', '', $_POST['file']) : '';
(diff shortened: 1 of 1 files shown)


SteveGilvarry:
Good catches from the review — all three were valid; pushed a follow-up (`630f778c3`):

1. **Non-scalar input** — every field is now coerced to a scalar before sanitizing, so `message[]=x` can't make `preg_replace` return an array and TypeError in `logPrint()`.
2. **`level` field** — now sanitized too, and the "Unexpected logger level" message logs the sanitized value rather than the raw field (same control-char vector via a different field). Also dropped the `$_POST` mutation in favour of a local var.
3. **`print_r($_POST, true)` dump** — replaced with a fixed message so the invalid-request path no longer emits raw POST (with embedded newlines) into the log.

`php -l` clean.

Issue #2915, opened 2020-04-17
Clean up /var/cache /var/tmp /tmp use


connortechnology:
The /tmp/zm/ directory is somewhat redundant in light of
/var/cache/zoneminder/temp/, and the latter presents none of the
security risks of the former. Why not have all the temp stuff
live in one location under /var?

This comes from debian bug tracker as a wish list item.  

On thinking about it, we really should verify that nothing sensitive goes to /tmp.  At present for exporting and archiving downloads, etc, we write to /var/tmp.  systemd somewhat mitigates this by sandboxing, but we should just do it right in the first place. 

We should also (in debian packaging) be writing our events by default under /var/lib/zoneminder to sync up with redhat.

onlyjob:
That's from the https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=858682

If I remember correctly, _tmpfs_-mounted `/tmp` could provide with advantages of being faster than long term storage `/var/cache/zoneminder/temp/` which might be NFS-mounted.

Perhaps `/tmp` could be protected with systemd's `PrivateTmp=true`?

connortechnology:
There are definite performance advantages, which is why many of our tmp files have gone into /tmp or /run.  However a lot of them can be large video files which can fill up a tmpfs pretty quick, so we have been using /var/tmp.  

Systemd has been helping a lot with the security aspect.  I don't know if PrivateTmp=True is the default but in my ubuntu systems, the apache process is PrivateTmp=true and that helps a lot when exporting video.  

There is a setting that used to default to on for zms, where it would cache video frames to disk to act as a rewind buffer.  This worked ok when saving to /tmp but when we moved it to /var/tmp performance suffered greatly.  I have since set the default to off for this feature.  In future I think we should just keep the jpegs in ram.  I don't see any advantage to writing them out to disk.

onlyjob:
`PrivateTmp=True` is unlikely to be a system default but it is easy to set it up per .service file when required.
I think I've used `/tmp` to buffer event's jpegs because otherwise checking event was too slow...

petterreinholdtsen:
The Debian bug report is still unresolved.

PR #4913, opened 2026-06-13
fix: keep temp/scratch files in ZoneMinder's own temp dir, not system /tmp (refs #2915)

SteveGilvarry:
Addresses the temp-file half of #2915: ZoneMinder scratch files were written into world-writable system temp, and on the Debian/Ubuntu family the temp dir *itself* lived in `/tmp` / `/var/tmp`.

Two commits.

## 1. Scratch files use randomized names in ZM's temp dir, not hardcoded `/tmp`
- `scripts/zmvideo.pl.in`: the ffmpeg concat list was built at a predictable `/tmp/<concat_name>.concat.lst` — open to a symlink/race and leaking monitor/event names. Now created with `File::Temp` (randomized, atomic `O_EXCL`) under `@ZM_TMPDIR@`. The list entries are absolute paths, so its location doesn't affect ffmpeg resolution.
- `web/ajax/training.php`: detection scratch image used `tempnam(sys_get_temp_dir(), ...)`, escaping ZM's tree. Now `tempnam(ZM_DIR_TEMP, ...)`.

## 2. Move `ZM_TMPDIR` out of system temp → `/var/cache/zoneminder/temp`
Only the RedHat build used a ZM-owned temp dir; everyone else used shared system temp:

| Build | before | after |
|-------|--------|-------|
| CMake default | `/var/tmp/zm` | `/var/cache/zoneminder/temp` |
| `distros/ubuntu2004` | `/var/tmp/zm` | `/var/cache/zoneminder/temp` |
| `distros/beowulf` | `/tmp/zm` | `/var/cache/zoneminder/temp` |

Also repoints the CakePHP `api/app/tmp` symlink (cache/sessions/logs) from shared `/var/tmp` to the same dir, matching what RedHat already does. `/var/cache` is correct here because everything written to this dir is regenerable (exports, uploads, concat list, detection scratch, CakePHP cache).

## Upgrade safety
- `ZM_DIR_TEMP` lives in `config.php` under `/usr/share` (a regular file, not a conffile), so it's replaced on upgrade and the new path is active immediately — no dpkg prompt.
- `/var/cache/zoneminder/temp` is already in the packaging `.dirs` (was provisioned but unused), so it exists and is `www-data`-owned on every install — no creation gap.
- No migration: temp contents are regenerable. CakePHP recreates its `cache/sessions/logs` on demand (as it already does under `/var/tmp`); one-time effect is a CakePHP cache/session reset. `ZM_UPLOAD_LOC_DIR` is a DB Config option (default-off FTP upload filter) that self-creates its dir, so existing installs are unaffected.

## Not covered (out of scope, separate follow-ups)
- The official Debian package on salsa (`ZM_TMPDIR=/tmp/zm`, `ZM_CONTENTDIR=/var/cache`) — maintained downstream, needs a Debian-side change.
- Moving events off `/var/cache` → `/var/lib` (the events FHS wart) — separate from temp.

refs #2915

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,8 +151,8 @@ set(ZM_RUNDIR "/var/run/zm" CACHE PATH
   "Location of transient process files, default: /var/run/zm")
 set(ZM_SOCKDIR "/var/run/zm" CACHE PATH
   "Location of Unix domain socket files, default /var/run/zm")
-set(ZM_TMPDIR "/var/tmp/zm" CACHE PATH
-  "Location of temporary files, default: /tmp/zm")
+set(ZM_TMPDIR "/var/cache/zoneminder/temp" CACHE PATH
+  "Location of regenerable temporary files, default: /var/cache/zoneminder/temp")
 set(ZM_LOGDIR "/var/log/zm" CACHE PATH
   "Location of generated log files, default: /var/log/zm")
 set(ZM_WEBDIR "${CMAKE_INSTALL_FULL_DATADIR}/zoneminder/www" CACHE PATH
--- a/distros/beowulf/rules
+++ b/distros/beowulf/rules
@@ -23,7 +23,7 @@ override_dh_auto_configure:
             -DZM_CONFIG_SUBDIR="/etc/zm/conf.d"        \
             -DZM_RUNDIR="/run/zm"                  \
             -DZM_SOCKDIR="/run/zm"                 \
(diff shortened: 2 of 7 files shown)


connortechnology:
/tmp can often be tmpfs-backed. So we made a choice to use /var/tmp since exporting can involve generating quite large files.  

I'm on the fence about whether to use system tmp dirs or our own.  Going to have to hear the arguments a bit more.

PR #4928, opened 2026-06-15
Many Log page improvements

IgorA100:
I also added a new modal window for displaying any messages instead of using `Alert()`.
Display a message to the user: `zmAlert (message, title)`
The `title` is optional.
The number of simultaneously displayed messages is limited to 5 to avoid overloading the web interface. All messages after the 5th will be displayed in the browser console.
All 5 messages are displayed with a diagonal offset of 20 pixels, allowing you to visually see how many messages are being displayed.

This PR fixes an issue where an AJAX table update request takes a long time to complete. The previous request doesn't have time to complete before the page automatically refreshes according to the REFRESH_LOGS value. This loop could continue indefinitely, and the user would never see the process complete.

Additionally implemented:
- Display the number of displayed and unfiltered logs, taking into account the client locale.
- Display the current status of the AJAX request next to the time of the last request (log.php)
Options:
-- `"in process"` - the request is executing
-- `"awaiting"` - the current request was interrupted and the next request, either autoRefresh or user-supplied, is awaited.
-- `"error"` - an error occurred while executing the request
-- `"time spent executing the last request"`
This is especially relevant when requests take a long time to execute (several seconds or tens of seconds).
- If the user has selected a row(s) with a checkbox, disable table auto-refresh and abort the current refresh if one is in progress. This will allow work to continue with the selected rows; otherwise, the table may be refreshed, causing the checkboxes on all selected rows to disappear.

- Don't execute `table.bootstrapTable('hideLoading')` when a new AJAX request arrives and the page is hidden; this misleads the user. The page may be hidden while the previous AJAX request is being executed. We now control the hiding and showing of the "Loading" label elsewhere in the code.
- The `secsToTime()` function can now accept the `"round"` argument.
- Minutes in the `secsToTime()` function are now displayed as integers, not fractional values.
- The "autoRefresh" button now responds to clicks, allowing you to temporarily disable automatic refresh.
D

- Display a linear progress bar when deleting logs.
DВ

- Display numbers at the bottom of the table based on the Locale.
This will separate long number triads for easier comprehension.
DВD


IgorA100:
I've completed this PR.

IgorA100:
We should not display a modal window if a user's AJAX request interrupts the background table update process.

IgorA100:
I hope I have completed this PR

PR #4929, opened 2026-06-15
Fix: Replace "%" with "%25" in the message line (log.js)

IgorA100:
Otherwise, decodeURIComponent() will fail, as "%" without subsequent characters is not allowed in a URL.

### DIFF
--- a/web/skins/classic/views/js/log.js
+++ b/web/skins/classic/views/js/log.js
@@ -76,7 +76,7 @@ function ajaxRequest(params) {
 function processRows(rows) {
   $j.each(rows, function(ndx, row) {
     try {
-      row.Message = decodeURIComponent(row.Message)
+      row.Message = decodeURIComponent(row.Message.replace(/%/g, '%25'))
           .replace(/</g, "&lt;").replace(/>/g, "&gt;") // Replace link tags
           .replace(/event (\d+)/g, "<a href=\"?view=event&eid=$1\">event $1</a>");
     } catch (e) {
--- a/web/skins/classic/views/js/log.js
+++ b/web/skins/classic/views/js/log.js
@@ -76,7 +76,7 @@ function ajaxRequest(params) {
 function processRows(rows) {
   $j.each(rows, function(ndx, row) {
     try {
-      row.Message = decodeURIComponent(row.Message.replace(/% /g, '%25%20'))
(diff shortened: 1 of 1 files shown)


web/skins/classic/views/js/log.js
```diff
@@ -76,7 +76,7 @@ function ajaxRequest(params) {
 function processRows(rows) {
   $j.each(rows, function(ndx, row) {
     try {
-      row.Message = decodeURIComponent(row.Message)
+      row.Message = decodeURIComponent(row.Message.replace(/% /g, '%25%20'))
```
connortechnology: This handles `% ` but still throws on every other malformed percent sequence — e.g. `95%of memory`, a trailing `100%`, or `%5`. Those fall into the `catch` and the message is left undecoded (no `event N` link replacement). Escaping any `%` not followed by two hex digits covers all cases in the same one line:

```suggestion
      row.Message = decodeURIComponent(row.Message.replace(/%(?![0-9A-Fa-f]{2})/g, '%25'))
```

connortechnology: Also... claude flagged that we may not need decodeURIComponent anymore.

connortechnology:
Separate, non-blocking thought on this PR: it's worth questioning whether the `decodeURIComponent()` call should exist at all.

The server side (`web/ajax/log.php:264`) already runs `htmlspecialchars()` on the message and strips non-ASCII before sending it to the client, and log messages aren't deliberately percent-encoded anywhere I can find. The decode appears to be legacy — it was changed from `decodeURI` to `decodeURIComponent` back in b35e27d0b (Feb 2021) without an encoding step ever being added on the producing side.

If messages are never actually percent-encoded, then `decodeURIComponent` is only ever a liability here: at best a no-op, at worst it throws on any literal `%` (which is exactly the bug this PR works around). Dropping the decode entirely might be the real fix.

That's a larger change with more regression risk than this PR, so I'd keep the regex fix here and track the decode question separately. Just flagging it so the workaround doesn't obscure the root cause.

IgorA100:
> That's a larger change with more regression risk than this PR

Yes, I also wondered why `decodeURIComponent()` was needed here, but I didn’t delve deeply into the code for analysis, especially since manually reviewing the entire code is hard work and errors are possible.

Issue #145, opened 2013-09-16
Readme.fedora contains incorrect references


gitsop01:
The Readme.fedora contains incorrect references and a updated file has been submitted to Bugzilla.
https://bugzilla.redhat.com/show_bug.cgi?id=1001825

PR #4930, opened 2026-06-16
fix: compute camera frame buffer sizes in size_t to avoid int overflow

SteveGilvarry:
## Summary

Fixes three CodeQL `cpp/integer-multiplication-cast-to-long` alerts (#282, #145, #85) plus one unflagged sibling of the same defect. The libVNC and libVLC camera backends computed frame-buffer sizes by multiplying 16/32-bit operands and only widening the result to `size_t` afterwards, so the multiplication itself can overflow before the conversion. The VNC framebuffer dimensions are advertised by the remote server (`uint16_t`), so the inputs are externally influenced.

## Changes

Cast the first operand to `size_t` so each product is computed in `size_t`:

- **`src/zm_libvnc_camera.cpp`**
  - `SwScale::Convert` source size `framebufferWidth * framebufferHeight * 4` (alert #282)
  - `SwScale::Convert` dest size `width * height * colours` (alert #145)
  - `resize()` `bufferSize`: changed `int` → `size_t` (not widened, so CodeQL did not flag it, but the same overflow applied); `Debug` format updated to `%zu`
- **`src/zm_libvlc_camera.cpp`**
  - `Image::Assign` size `width * height * mBpp` (alert #85)

No behavioural change for valid dimensions; the cast only affects the overflow case.

## Testing

Both translation units compile clean with `-Werror -fsyntax-only` using the project's compile flags. CI performs the full multi-platform build.

### DIFF
--- a/src/zm_libvlc_camera.cpp
+++ b/src/zm_libvlc_camera.cpp
@@ -80,7 +80,7 @@ void LibvlcUnlockBuffer(void* opaque, void* picture, void *const *planes) {
   LibvlcPrivateData* data = reinterpret_cast<LibvlcPrivateData*>(opaque);
 
   bool newFrame = false;
-  for( unsigned int i=0; i < data->bufferSize; i++ ) {
+  for( size_t i=0; i < data->bufferSize; i++ ) {
     if ( data->buffer[i] != data->prevBuffer[i] ) {
       newFrame = true;
       break;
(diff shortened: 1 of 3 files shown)


SteveGilvarry:
Review feedback addressed in 45099e53a:

- **`zm_libvnc_camera.cpp` `resize()`** — now checks the `av_malloc` result and returns `FALSE` (with an `Error` log) on failure instead of returning `TRUE` with a null `frameBuffer`, which libVNC would write into and crash. More relevant now that `size_t` sizing can request large allocations for server-advertised dimensions.
- **`zm_libvnc_camera.cpp` scale `Debug()`** — the logged buffer sizes are now computed in `size_t` with `%zu`, so they can't overflow `int` and match the values passed to `SWScale::Convert`.
- **`zm_libvlc_camera`** — `LibvlcPrivateData::bufferSize` widened to `size_t` and computed in `size_t` in `PrimeCapture()`, so the allocation itself can't overflow. `Image::Assign` now passes the same stored `mLibvlcData.bufferSize` used for the allocation, so the read size can never exceed the buffer; the compare loop index is widened to `size_t` to match.

Verified locally: both translation units compile with `-Werror -fsyntax-only`. CI runs the full multi-platform build.

Issue #266, opened 2013-12-07
Implement unit testing


fastolfe:
Initially for discussion.  There are essentially no tests to speak of in ZM and no way to verify that things aren't broken after significant code changes than to poke around and wait for feedback on the forums.  This means selecting a unit testing framework, potentially rewriting some code to make it testable, write some tests, and hook those tests into Travis or some other test automation service.  I'm interested in feedback about the value of this and ideas about frameworks and implementation.  I'm familiar with https://code.google.com/p/googletest/ but that's largely it.

PR #4931, opened 2026-06-16
fix: resolve CodeQL DOM-XSS and unsafe-sanitizer findings in skin/monitor JS

SteveGilvarry:
## Summary

Resolves three CodeQL alerts in ZoneMinder-authored JavaScript (the rest of the open JS alerts are in vendored libraries — jQuery UI, Bootstrap, hls.js, bootstrap-table — and are out of scope here).

| Alert | Rule | Location |
|---|---|---|
| #135 | `js/xss-through-dom` | `web/skins/classic/js/skin.js` (`.zmlink` click) |
| #266 / #267 | `js/incomplete-multi-character-sanitization` / `js/polynomial-redos` | `web/skins/classic/js/skin.js` (`strip_html`) |
| #246 | `js/xss-through-dom` | `web/skins/classic/views/js/monitor.js` (`ControlEdit_onClick`) |

## Changes

- **`.zmlink` handler:** navigated to the element's `href`/`data-url` via `location.assign()` without checking the scheme, so a `javascript:`/`data:` URL in the attribute would execute on click. Now skips navigation for `javascript:`/`data:`/`vbscript:` schemes.
- **`strip_html()`:** replaced the `/<[^>]+>/g` tag-stripping (an incomplete sanitizer that can also backtrack) with `DOMParser` + `textContent` — correct and linear. The five callers (`devices`/`reports`/`snapshots`/`controlcaps`/`events.js`) use it to extract a plain `Id` from a rendered table cell; behaviour is preserved.
- **`ControlEdit_onClick`:** `encodeURIComponent()` the select value before building the `controlcap` URL.

## Testing

`eslint` clean on both files; `node --check` passes. The alerts auto-resolve on the next CodeQL scan after merge.

### DIFF
--- a/web/skins/classic/views/js/monitor.js
+++ b/web/skins/classic/views/js/monitor.js
@@ -954,7 +954,10 @@ function ControlId_onChange(ddm) {
 function ControlEdit_onClick() {
   const ControlId = document.getElementById('ControlId');
   if (ControlId) {
-    window.location = '?view=controlcap&cid='+encodeURIComponent(ControlId.value);
+    const cid = parseInt(ControlId.value, 10);
+    if (Number.isInteger(cid) && cid > 0) {
+      window.location = '?view=controlcap&cid=' + cid;
+    }
   }
 }
 
--- a/web/skins/classic/views/js/monitor.js
+++ b/web/skins/classic/views/js/monitor.js
@@ -954,7 +954,7 @@ function ControlId_onChange(ddm) {
 function ControlEdit_onClick() {
   const ControlId = document.getElementById('ControlId');
   if (ControlId) {
-    window.location = '?view=controlcap&cid='+ControlId.value;
+    window.location = '?view=controlcap&cid='+encodeURIComponent(ControlId.value);
   }
 }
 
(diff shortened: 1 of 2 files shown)


web/skins/classic/views/js/monitor.js
```diff
@@ -954,7 +954,7 @@ function ControlId_onChange(ddm) {
 function ControlEdit_onClick() {
   const ControlId = document.getElementById('ControlId');
   if (ControlId) {
-    window.location = '?view=controlcap&cid='+ControlId.value;
+    window.location = '?view=controlcap&cid='+encodeURIComponent(ControlId.value);
```
connortechnology: @copilot @claude we can go further here, ControlId should only be a positive integer

SteveGilvarry:
Review feedback addressed in 23710249a (on top of the Copilot autofix e2ca8ee7b):

- **`monitor.js` `ControlEdit_onClick`** — `ControlId` is now parsed with `parseInt(..., 10)` and only navigated when `Number.isInteger(cid) && cid > 0`, so the URL is built from a typed positive integer rather than the raw select value.
- **`skin.js` `.zmlink` handler** — dropped the unused binding from the `catch` clause (`catch (e)` → `catch`).

Verified locally: `eslint` clean on both files and `node --check` passes.

PR #4938, opened 2026-06-17
fix: guard zms buffer_level against zero buffer count

### LINKED ISSUE #4936: zms crash (SIGFPE) in MonitorStream::processCommand: division by zero on temp_image_buffer_count
## Summary

`zms`/`nph-zms` crashes with signal 8 (SIGFPE, floating point exception) in `MonitorStream::processCommand`.

Symbolized backtrace (from `eu-addr2line`):
```
zm_die_handler(int, siginfo_t*, void*)            src/zm_signal.cpp:214
MonitorStream::processCommand(StreamBase::CmdMsg const*)  src/zm_monitorstream.cpp:306
StreamBase::checkCommandQueue()                   src/zm_stream.cpp:138
```

## Root cause

`src/zm_monitorstream.cpp:306`:
```cpp
if (playback_buffer > 0)
  status_data.buffer_level = (MOD_ADD( (temp_write_index-temp_read_index), 0, temp_image_buffer_count )*100)/temp_image_buffer_count;
```

The guard checks `playback_buffer`, but the divisor (and the `MOD_ADD` modulo, `((var)+(limit)+(delta))%(limit)`) is `temp_image_buffer_count`. They are different members.

`processCommand()` runs on the `command_processor` thread, started in `runStream()` (around line 537). `temp_image_buffer_count = playback_buffer` is not assigned until later (around line 552). In the window between thread start and that assignment, `temp_image_buffer_count` is still 0 (constructor default) while `playback_buffer` is already > 0. A command arriving in that window passes the guard and divides by zero -> SIGFPE.

Sites with many monitors and frequent zms restarts hit this regularly because the browser issues a command immediately on connect.

Observed on 1.39.16 (noble build).

## Fix

Guard on the actual divisor `temp_image_buffer_count > 0` instead of `playback_buffer`.

SteveGilvarry:
Fixes #4936.

## Root cause

`nph-zms`/`zms` crashed with SIGFPE (signal 8, integer divide-by-zero) in `MonitorStream::processCommand` (`src/zm_monitorstream.cpp:306`). The `buffer_level` calculation divides by `temp_image_buffer_count` (and takes a `MOD_ADD` modulo by it), but the surrounding code only guarded on a *different* variable, `playback_buffer`.

`processCommand` runs on the `command_processor` thread, which `runStream()` starts **before** it assigns `temp_image_buffer_count = playback_buffer`. In that startup window `temp_image_buffer_count` is still `0` while `playback_buffer` is already `> 0`, so a stream command arriving in that window divides by zero and raises SIGFPE, killing the streamer.

Backtrace confirmed via `eu-addr2line`: `zm_signal.cpp:214` → `zm_monitorstream.cpp:306` → `zm_stream.cpp:138`.

## Fix

Extract the percentage math into a pure inline helper `MonitorStreamBufferLevel(write_index, read_index, buffer_count)` in `src/zm_monitorstream.h` that returns `0` when `buffer_count <= 0`, and replace the crashing expression in `src/zm_monitorstream.cpp` with a call to it. The old code crashed on a zero count; the helper now returns 0.

## Tests

Added `tests/zm_monitorstream.cpp` (registered in `tests/CMakeLists.txt`) covering zero-count, negative-count, empty/full/half percentages, and wrap-around (write behind read).

Ran locally (Debug build, `-DBUILD_TEST_SUITE=ON`):
- `./tests/tests "MonitorStreamBufferLevel"` → **All tests passed (8 assertions in 1 test case)**.
- `ctest --output-on-failure` → **100% tests passed, 0 failed out of 93**.

Build is clean with no warnings introduced by this change.

### DIFF
--- a/src/zm_monitorstream.cpp
+++ b/src/zm_monitorstream.cpp
@@ -530,10 +530,10 @@ void MonitorStream::runStream() {
   }
 
   openComms();
+  // Declared here so it stays in scope for the join() at the end of the
+  // stream loop, but not started until the temporary buffer state below has
+  // been initialised (see the launch after the playback_buffer setup).
   std::thread command_processor;
-  if (connkey) {
-    command_processor = std::thread(&MonitorStream::checkCommandQueue, this);
-  }
 
   if (type == STREAM_JPEG)
     fputs("Content-Type: multipart/x-mixed-replace; boundary=" BOUNDARY "\r\n\r\n", stdout);
@@ -599,6 +599,13 @@ void MonitorStream::runStream() {
     Debug(2, "Not using playback_buffer");
   } // end if connkey && playback_buffer
 
+  // Start the command processor only now that temp_image_buffer_count, the
+  // read/write indices and temp_image_buffer are initialised, so a command
+  // arriving early cannot observe a half-initialised stream (the original
+  // cause of the divide-by-zero in issue #4936).
+  if (connkey) {
+    command_processor = std::thread(&MonitorStream::checkCommandQueue, this);
+  }
 
   while (!zm_terminate) {
     if (feof(stdout)) {
--- a/src/zm_monitorstream.cpp
+++ b/src/zm_monitorstream.cpp
@@ -302,10 +302,8 @@ void MonitorStream::processCommand(const CmdMsg *msg) {
       status_data.analysing = monitor->shared_data->analysing;
       status_data.score = monitor->shared_data->last_frame_score;
 
-      if (playback_buffer > 0)
-        status_data.buffer_level = (MOD_ADD( (temp_write_index-temp_read_index), 0, temp_image_buffer_count )*100)/temp_image_buffer_count;
-      else
-        status_data.buffer_level = 0;
+      status_data.buffer_level =
+          MonitorStreamBufferLevel(temp_write_index, temp_read_index, temp_image_buffer_count);
     }
   } // end monitor_mutex scope
   status_data.delayed = delayed;
--- a/src/zm_monitorstream.h
+++ b/src/zm_monitorstream.h
@@ -22,6 +22,18 @@
 
 #include "zm_stream.h"
 
+// Returns the playback buffer fill level as a percentage (0-100).
+// Guards against buffer_count <= 0, which would otherwise raise SIGFPE via
+// modulo/division by zero. processCommand() can run on the command thread
+// before runStream() assigns temp_image_buffer_count, so the count may still
+// be 0 while playback_buffer is already > 0.
+// See https://github.com/ZoneMinder/zoneminder/issues/4936
+inline int MonitorStreamBufferLevel(int write_index, int read_index, int buffer_count) {
+  if (buffer_count <= 0)
+    return 0;
+  return (((write_index - read_index + buffer_count) % buffer_count) * 100) / buffer_count;
+}
+
 class MonitorStream : public StreamBase {
  protected:
   struct SwapImage {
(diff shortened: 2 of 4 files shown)


SteveGilvarry:
I have a feeling this is fixing/guarding the symptom of stuff happening before it should, running that thought down also at the moment.

SteveGilvarry:
@connortechnology thoughts?
You're right to push on this — the divide-by-zero is one symptom of a broader startup-ordering problem. Here's what the code shows:

  535:  command_processor = std::thread(&MonitorStream::checkCommandQueue, this);  // thread starts
  ...
  550:  temp_image_buffer_count = playback_buffer;     // count set — 15 lines later
  551:  temp_read_index  = temp_image_buffer_count;    // indices set here too
  552:  temp_write_index = temp_image_buffer_count;
  ...
  594:  temp_image_buffer = new SwapImage[temp_image_buffer_count];  // buffer allocated even later

  So the command thread is live and calling processCommand before the stream's buffer state exists. The divide-by-zero was just the one access that happened to be fatal; in that same window temp_read_index/temp_write_index are
  also stale and temp_image_buffer is still nullptr.

  The good news: I read all of processCommand — it's mostly setting scalar flags (paused, scale, replay_rate, x/y) and building status_data. It does not index into temp_image_buffer[...], so there's no null-deref lurking — the
  buffer_level math was the only thing that crashed. So the committed guard does stop the actual crash.

  But the cleaner structural fix is to remove the race at the source: start the command_processor thread after the buffer state is initialized, rather than before. Moving the thread-start (535) to after the init block (~595) is
  safe — nothing between those points needs the thread running, and any commands that arrive meanwhile just sit in the socket buffer until it starts. Then no command can ever observe a half-initialized stream.

  My recommendation: keep both.
  - The MonitorStreamBufferLevel helper stays — it's cheap defense-in-depth and protects against future reorderings or other call paths.
  - Additionally reorder so the thread launches last. Belt and suspenders: the guard makes the math safe, the reorder makes the state coherent.

  One larger thing I'd flag but not fold into this PR: those fields (temp_image_buffer_count, temp_read_index/write_index, paused, scale…) are written by the command thread and read by the main stream loop with no visible
  synchronization (not atomic, no mutex) — a pre-existing data race. Tightening that is a bigger, separate change and riskier; worth its own issue rather than scope-creeping a SIGFPE fix.

  Want me to add the thread-start reorder to PR #4938 (small, keeps the helper), and open a separate tracking issue for the unsynchronized shared-state concern?

connortechnology:
Yeah I am a fan of clean init at the start.  We shouldn't start trying to process commands until we are actually ready to process commands.

SteveGilvarry:
Added a second commit (629842c9) that addresses the root cause rather than just the symptom: the `command_processor` thread was being started in `runStream()` *before* `temp_image_buffer_count`, the read/write indices and `temp_image_buffer` were initialised, so an early command observed a half-initialised stream (that window is where the divide-by-zero came from).

The thread object is still declared up front so it stays in scope for the `join()`, but its launch is deferred until after the `playback_buffer` setup. With both commits: the `MonitorStreamBufferLevel()` helper keeps the math safe (defense-in-depth / other call paths), and the reorder removes the init-window race at the source.

Rebuilt and re-tested: `MonitorStreamBufferLevel` passes (8 assertions), `ctest` 93/93, no new warnings.

The broader concern — `processCommand` mutating shared playback state (`paused`, `replay_rate`, `scale`, buffer indices…) read by the main loop without synchronization — is tracked separately in #4939, as it's a larger change than this crash fix.

PR #4941, opened 2026-06-18
perf: reuse Image buffer in EventStream::sendFrame (modern C++ alternative to #4609)

connortechnology:
## Summary

Reduce per-frame heap allocations in `EventStream::sendFrame` while keeping the code modern C++.

- Add a `reuse_image_` member and decode JPEGs into it via `ReadJpeg`, instead of `new Image(...)`/`delete` every frame. `ReadJpeg`/`WriteBuffer` reuses the pixel allocation when dimensions match (every frame of a given event), eliminating a ~2MB malloc/free per streamed frame. Applied on both the JPEG and `STREAM_MPEG` paths.
- The uncommon ffmpeg (mp4) path uses `std::unique_ptr<Image>` so it frees automatically — no manual `delete`, no `if (image != &reuse_image_)` pointer-identity guard.
- Replace the existence-check `stat()` calls (which filled an unused `struct stat`) with `std::filesystem::exists()` using the `std::error_code` overload, preserving `stat`'s non-throwing "any failure means not present" behaviour.

The per-frame path-string allocation that `stringtf()` incurred was already eliminated on master via the `reuse_filepath_` `std::string` member.

## Relationship to #4609

This is an alternative to #4609, achieving the same goal (minimise per-frame heap allocations in the streaming hot path) using modern C++ — a reused `Image` member and `unique_ptr` rather than a fixed `char[PATH_MAX]` buffer + `string_view` + `snprintf`/`access`. Supersedes the image-reuse portion of #4609.

## Testing

- Built `zms` cleanly (`cmake --build . --target zms`), no new warnings.

### DIFF
--- a/src/zm_eventstream.cpp
+++ b/src/zm_eventstream.cpp
@@ -27,6 +27,7 @@
 #include "zm_storage.h"
 #include <algorithm>
 #include <arpa/inet.h>
+#include <memory>
 #include <sys/stat.h>
 
 #include <filesystem>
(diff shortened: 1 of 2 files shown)


SteveGilvarry:
Reviewed this end-to-end (correctness + an empirical allocation check). 👍 — correct and a real win.

**Correctness — the things that could have made buffer reuse subtly wrong, all check out:**
- `prepareImage` never mutates its input in place: for any scale/zoom it `Assign()`s into a `static Image copy_image` and mutates that, returning the original otherwise — so `reuse_image_` keeps its full-resolution content across frames.
- `ReadJpeg` calls `WriteBuffer(new_w, new_h, …)` *before* `jpeg_start_decompress`/scanlines, so a frame with different dimensions reallocates before any byte is written — no OOB, and even a mid-decode `longjmp` leaves dims/buffer consistent.
- Ownership is right: `unique_ptr` on the ffmpeg path, member reuse on the JPEG path; the removed `delete image` is matched by the `unique_ptr`. No leak/double-free.
- `std::filesystem::exists(p, ec)` is behaviour-equivalent to the old `stat() < 0` (non-throwing, any error → "not present", follows symlinks), and `<sys/stat.h>` is still needed by the other `stat()` further down.

**Empirical check** — heaptrack on a micro-benchmark linking `libzm` and decoding a real event frame (2160×3840 → 24.94 MB RGB buffer), old-vs-new pattern over 200 frames:

| 200-frame stream | before (`Image` per frame) | after (reused) |
|---|---|---|
| 24.94 MB decode-buffer allocations | **200** | **1** |
| total allocation calls | 4639 | 4026 |
| peak heap | 25.52 M | 25.52 M (unchanged) |

The per-frame decode buffer goes from allocated-every-frame to allocated-once, with **identical peak memory** (the reused buffer is the same footprint as one transient).

**Two minor, non-blocking notes:**
1. `ReadJpeg`'s return value is ignored, so a corrupt-but-*existing* frame silently leaves `reuse_image_` holding the previous frame (arguably more graceful than the old blank, but a persistent corruption now repeats the last good frame with no error). Might be worth a `Warning` / `sendTextFrame` on `!ReadJpeg(...)`.
2. The trailing `image = nullptr;` is now dead and could be dropped.

SteveGilvarry (approved):

PR #4958, opened 2026-06-27
Feat: Add "except INF" to the "Level" filter drop-down list on the Logs page

IgorA100:
Sometimes you only need to display warnings and errors without informational messages. You have to select "WAR," "ERR," "FAT," and so on separately, since the informational messages are the most numerous. This may not be the best solution, but it's fast.
Ideally, I think you should create a multi-section for the "Level" filter.

### DIFF
--- a/web/ajax/log.php
+++ b/web/ajax/log.php
@@ -190,17 +190,17 @@ function queryRequest() {
     $query['values'][] = $_REQUEST['level'];
   }
 */
-  $L = (isset($_REQUEST['level']) && !empty($_REQUEST['level']) && is_scalar($_REQUEST['level'])) ? (string) $_REQUEST['level'] : '';
+  $level = (isset($_REQUEST['level']) && !empty($_REQUEST['level']) && is_scalar($_REQUEST['level'])) ? (string) $_REQUEST['level'] : '';
   $level_codes = array_flip(ZM\Logger::$codes);
   $levelValueAllowed = false;
-  if (!empty($L)) {
-    if (isset($level_codes[$L])) {
+  if (!empty($level)) {
+    if (isset($level_codes[$level])) {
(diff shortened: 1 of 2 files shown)


IgorA100:
I don't know how correct the phrase "except INF" is, but I tried to use the shortest form of the phrase.

web/ajax/log.php
```diff
@@ -192,10 +192,23 @@ function queryRequest() {
 */
   $L = (isset($_REQUEST['level']) && !empty($_REQUEST['level']) && is_scalar($_REQUEST['level'])) ? (string) $_REQUEST['level'] : '';
```
connortechnology: I realise this is pre-existing... but I really try to use single letter variable names... if you are up for it, maybe we could rename is to... level_code? or just level?

connortechnology:
I really don't like the except idea.  Would much rather we just made level a multi-select.  It is more powerful.

web/ajax/log.php
IgorA100: Done: https://github.com/ZoneMinder/zoneminder/pull/4958/changes/fde4ab450e848cf2846057f09c690870edc291bc

IgorA100:
> I really don't like the except idea.

I agree that this isn't the best solution, and I've already stated this.
But right now, I don't have time to convert select to multi-select, as it's not a priority.
Perhaps later, say in the fall (if someone doesn't do it first), I'll convert to multi-select.
Isaac, if you don't like this temporary solution, you can simply close this PR without merging.

IgorA100:
I'm closing this PR because multiselect is implemented in https://github.com/ZoneMinder/zoneminder/commit/66299ae7cf9c9c91e7e33e6463c070adf76fb2b9

PR #4963, opened 2026-07-01
[1.38] Feat: Allow national Unicode characters in monitor name

IgorA100:
Similar to the Master #4785


IgorA100:
I've migrated all the changes from the Master branch.
Everything should work, but I'm unable to test this PR in version 1.38.
I consider this PR complete, but I need enthusiasts to test it.

connortechnology:
Would it not have been possible to just cherry pick #4785?   In future, to minimise if you do a piece of work for master and think it should go in 1.38, use the milestone feature on the right to tag it for 1.38 inclusion.

IgorA100:
> use the milestone feature on the right to tag it for 1.38 inclusion.

Unfortunately, this isn't available to me :(

IgorA100:
@connortechnology 
Thanks for the invitation. I accepted it, but it seems nothing has changed.

PR #4965, opened 2026-07-04
feature: Allow use of secondary stream for analysis/motion detection

madcamel:
It's really hot outside and my NVR was melting down to zmc CPU usage. Here's a fun fix with some help from the slop machine. This is an absolutely massive improvement, my machine went from flamethrower to mostly idle.

-snip-
## What

Uses SecondPath (should be configured as a lower resolution stream from the camera) for frame analysis when AnalysisSource is set to Secondary.

AnalysisSource already exists in the schema and UI but was a no-op. This wires it up.

## Why

In Modect, `zmc` software-decodes the entire primary stream just to feed motion
detection, even with no live viewers. On a 5MP/10fps monitor that is roughly a
full CPU core (measured: ~59% libavcodec + ~13% libswscale). Most cameras
already publish a low-res substream; analysing that instead cuts idle decode
cost, while motion events still record the primary at full resolution.

## How it works

- New `SecondStreamThread` (one per monitor, started only when `SecondPath` is
  set and `AnalysisSource=Secondary`) owns its own `FFmpeg_Input` on the
  substream, reconnects with exponential backoff, and publishes each decoded
  frame into a mutex-guarded mailbox at the substream's native resolution. It
  never enqueues packets into the packetqueue, so the single-video-stream
  invariants there are untouched.
- `Monitor::getMotionSourceImage()` feeds `Analyse()`. In secondary mode it
  returns the mailbox image, upscaled to camera dimensions only when a frame is
  actually scored (a cheap sequence/age peek gates the copy + upscale). Zones
  stay in primary coordinates and need no changes, because the analysis image is
  scaled to the same `camera_width x camera_height` the primary path already
  produces.
- The monitor's `AnalysisImage` setting (YChannel vs FullColour) drives whether
  the sidecar extracts the Y plane or swscales to colour.
- `effective_decoding` downgrades `DECODING_ALWAYS` to on-demand while secondary
  analysis is active, so the primary decodes only for live viewers. Alarm/event
  flow is unchanged: scores attach to the primary packet, so pre-event buffering
  and recording use the primary as before.
- The sidecar forces software decoding (`FFmpeg_Input::set_no_hwaccel()`): the
  D1 substream is cheap on the CPU, hardware-format frames cannot be consumed by
  the mailbox, and this avoids contending with the primary over the VAAPI device.
- `FFmpeg_Input` gains an `Open(filename, AVDictionary**)` overload so the
  sidecar can pass `rtsp_transport` and a socket read timeout (`timeout` for
  ffmpeg >= 5.0, `stimeout` for older). The existing single-argument `Open()`
  delegates to it unchanged.

## Interaction with existing `SecondPath` use

`SecondPath` is already consumed elsewhere, so reviewers should be aware:
- `FfmpegCamera` opens the substream as an audio source (`mSecondInput`) when
  the primary has no audio stream; go2rtc and RTSP2Web restreaming also open it.
- The sidecar adds one more in-process connection. On a monitor with
  `AnalysisSource=Secondary` and no primary audio, `zmc` will hold two
  connections to the substream (audio + analysis); with restreaming enabled the
  camera may see more. This is additive load only — the sidecar owns a separate
  `FFmpeg_Input` and shares no state with `mSecondInput`, so the existing
  teardown paths are unaffected. Cameras with tight session limits are the
  constraint here.

## Behaviour when the substream is unavailable

- No frame yet, or newest frame older than 10s: treated as "no motion data" —
  no scoring, no forced state change, no alarm storm while blind.
- Read failure: the sidecar closes and reconnects with backoff (1s -> 30s),
  which also keepalives the RTSP session so it does not sit in CLOSE-WAIT.

## Upgrade note

Because `AnalysisSource` was previously a no-op, any monitor already set to
`Secondary` in the database will change behaviour after upgrade: it will begin
opening and analysing the substream. A monitor with a stale or misconfigured
`SecondPath` will log reconnect errors and report "no motion data" rather than
analysing the primary. Worth a mention in release notes.

## Testing

- Builds clean (`cmake --build . --target zmc`), no warnings.
- Dropped CPU load by 80%
- Works for me.jpg

## Limitations

- Analysis inherits the substream's spatial precision (a D1 frame upscaled to
  primary dimensions is blocky). Zone areas and pixel-count thresholds stay
  valid because the analysis image is at primary resolution.
- If the substream aspect ratio differs from the primary, the upscale stretches
  it; zones still function.

### DIFF
--- a/src/zm_ffmpeg_input.cpp
+++ b/src/zm_ffmpeg_input.cpp
@@ -101,6 +101,12 @@ int FFmpeg_Input::Open(const char *filepath, AVDictionary **options) {
     std::list<const CodecData *>codec_data = get_decoder_data(input_format_context->streams[i]->codecpar->codec_id, "auto");
     for (auto it = codec_data.begin(); it != codec_data.end(); it ++) {
       const CodecData *chosen_codec_data = *it;
+#if HAVE_LIBAVUTIL_HWCONTEXT_H && LIBAVCODEC_VERSION_CHECK(57, 107, 0, 107, 0)
+      if (no_hwaccel and (chosen_codec_data->hwdevice_type != AV_HWDEVICE_TYPE_NONE)) {
+        Debug(1, "Skipping hardware codec %s (software decoding forced)", chosen_codec_data->codec_name);
+        continue;
+      }
+#endif
       Debug(1, "Found codec %s", chosen_codec_data->codec_name);
 
       streams[i].codec = avcodec_find_decoder_by_name(chosen_codec_data->codec_name);
--- a/src/zm_ffmpeg_input.cpp
+++ b/src/zm_ffmpeg_input.cpp
@@ -38,10 +38,6 @@ int FFmpeg_Input::Open(
   return 1;
 }
 
-int FFmpeg_Input::Open(const char *filepath) {
-  return Open(filepath, nullptr);
-}
-
 int FFmpeg_Input::Open(const char *filepath, AVDictionary **options) {
   int error;
 
--- a/src/zm_ffmpeg_input.cpp
+++ b/src/zm_ffmpeg_input.cpp
@@ -39,10 +39,14 @@ int FFmpeg_Input::Open(
 }
 
 int FFmpeg_Input::Open(const char *filepath) {
+  return Open(filepath, nullptr);
+}
+
+int FFmpeg_Input::Open(const char *filepath, AVDictionary **options) {
   int error;
 
   /** Open the input file to read from it. */
-  error = avformat_open_input(&input_format_context, filepath, nullptr, nullptr);
+  error = avformat_open_input(&input_format_context, filepath, nullptr, options);
   if ( error < 0 ) {
     if (std::string(filepath).find("incomplete") != std::string::npos) {
       Warning("Could not open input file '%s' (error '%s')",
--- a/src/zm_monitor.cpp
+++ b/src/zm_monitor.cpp
@@ -2334,10 +2334,10 @@ bool Monitor::Analyse() {
             Event::StringSet zoneSet;
 
             // Resolve the motion-detection source.  In secondary-analysis mode
-            // this is the substream sidecar image (already scaled/converted to
-            // camera dimensions), independent of whether the primary was decoded.
-            // do_score is false when there is no fresh/valid substream frame to
-            // score this round.
+            // this is the substream sidecar image at its native substream
+            // resolution (zones are rebuilt at that size), independent of whether
+            // the primary was decoded.  do_score is false when there is no
+            // fresh/valid substream frame to score this round.
             bool do_score = true;
             Image *motion_image = getMotionSourceImage(packet, do_score);
 
--- a/src/zm_monitor.cpp
+++ b/src/zm_monitor.cpp
@@ -182,7 +182,6 @@ Monitor::Monitor() :
   analysing(ANALYSING_ALWAYS),
   recording(RECORDING_ALWAYS),
   decoding(DECODING_ALWAYS),
-  effective_decoding(DECODING_ALWAYS),
   secondary_analysis(false),
   RTSP2Web_enabled(false),
   RTSP2Web_type(WEBRTC),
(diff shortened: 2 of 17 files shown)


src/zm_monitor.cpp
```diff
@@ -3626,6 +3605,64 @@ void Monitor::closeEvent() {
...
+      return nullptr;
+    }
+    last_secondary_sequence = sequence;
+    do_score = fresh and !stale;
+
+    // Upscale the native-resolution substream image to camera/zone dimensions so
+    // that DetectMotion and the reference image match the primary coordinate
+    // space and zones need no changes.
+    secondary_image.Assign(secondary_image_native);
+    if ((secondary_image.Width() != static_cast<unsigned int>(camera_width)) or
```
connortechnology: It doesn't make sense to upscale.  Only makes sense to downscale.  Zone coordinates are percentages now.  Motion detect results scan be scaled to match either image.

Also, we should still honour y-image for motion detection.

connortechnology: by that I mean we can use the y-image on the secondary stream for even LESS cpu.

src/zm_monitor.cpp
```diff
@@ -3905,6 +3942,20 @@ int Monitor::PrimeCapture() {
...
+  if (second_stream) {
+    second_stream->Stop();
+    second_stream->Join();
+  }
+
+  // AnalysisSource=Secondary: analyse the low-res substream instead of the
+  // full-res primary.  When active, the primary is only decoded on demand (for
+  // live viewers), so DECODING_ALWAYS is downgraded to viewer-driven decoding.
+  secondary_analysis = (analysis_source == ANALYSIS_SECONDARY) and !second_path.empty();
+  effective_decoding = decoding;
```
connortechnology: What if we are encoding in videostore?  This isn't enough.   We should honour the setting.  If the user wants decoding=ondemand, they can set that.  We will just always decode the secondary.

connortechnology:
THanks for taking this on, we have been slowing working towards it but you have jumped right in, which should force us to move faster.

That being said, the reason it was taking so long is it is a bit more complicated than this initial effort addresses.  I have concerns about doing the open and capture outside of the Camera object.  Not opposed to it being it's own thread, but I think it should be injecting into the packet queue.  I'm not convinced that this implementation will stay synced between the streams.  We need to sync and link the two streams, likely in wallclock units instead of pts/dts.  

And now I will sic the AI on it.

src/zm_second_stream_thread.h
```diff
@@ -0,0 +1,70 @@
...
+  std::thread thread_;
+
+  FFmpeg_Input *input_;
+  SwsContext *convert_context_;
+
+  std::mutex mutex_;
+  Image latest_image_;
+  bool have_image_;
+  uint64_t sequence_;
+  SystemTimePoint capture_time_;
```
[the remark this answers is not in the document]
connortechnology: Just to add some further info here:  If we use  a steady clock, we should start it at capture, and then base timestamps for both streams off that start point.

src/zm_ffmpeg_input.cpp
```diff
@@ -39,10 +39,14 @@ int FFmpeg_Input::Open(
 }
 
 int FFmpeg_Input::Open(const char *filepath) {
+  return Open(filepath, nullptr);
+}
+
+int FFmpeg_Input::Open(const char *filepath, AVDictionary **options) {
```
connortechnology, line 41: I think we could just use AVDictionary **options = nullptr in the declaration to avoid two ::Open functions.

madcamel:
Thanks for the review and feedback. I'm just flying by the seat of my pants, I'm not very familiar with zm's non-perl code. To be frank I'm avoiding the packet queue because that would add a lot of complexity and touches a lot of things I don't understand. It's probably over my head.

I will make a decent attempt at addressing the other issues you've raised as time permits, it would be really nice to have this functionality in mainline.

madcamel:
I've addressed many of the issues in your code review. Admittedly 90% vibe coded. Keeping the extra monotonic timestamps is somewhat janky but it works. Because it's not using the packet queue I've tried to keep this more a bolt-on than a fully integrated solution.

I have been running this on my instance for a couple days now and it seems to be stable and perform correctly.

connortechnology:
THere is a lot in here that I like, and I will likely cherry-pick some of it to reduce the size.  

I like having a second thread doing capture of the second stream, and my code in zm_ffmpeg_camera is kindof a hack that i would like to replace with your approach.  But I really feel we need to be injecting from both capture threads into the packet queue sorted by wall clock.  I'm not sure we need the steady clock stuff, I think the existing system clock stuff will do.  Maybe claude can defend itself on that.
 
Anyways, those are some thoughts, I will look it over and think about it some more.

madcamel:
Thanks again for looking it over. The work you do on ZM is very appreciated. It may not be modern and shiny but it's extremely reliable. Even if this code is not up to your standards I hope something can be gained from it!

IgorA100:
I'm following this PR, and it would be great if the existing AnalysisFPSLimit parameter was finally fully implemented. This would further reduce CPU usage. Motion analysis is probably possible with 5 FPS, and if the camera delivers 25 FPS, the gains would be significant. For example, for live viewing of the second stream, a high FPS is desirable, but for analysis, a significantly lower FPS is desirable.

madcamel:
The sidecar should already be honoring AnalysisFPSLimit

PR #4967, opened 2026-07-05
Clear mseSourceBuffer only if the number of buffers is greater than zero. (MonitorStream.js)

IgorA100:


connortechnology:
Can you comment why? Are we just preventing a throw?

IgorA100:
If there are no buffers in sourceBuffers, then there's nothing to delete.
For example, we're trying to play H.265. The Chromium browser in Win 8, for example, tries to start playback, but as soon as the track is added to the buffer, an error occurs. So we stop playback. As a result, the buffer is empty. There will be a warning when clearing the buffer; we don't need unnecessary warnings.
Or, for example, we're quickly switching between monitors, and switching to the next monitor occurs when the buffer already exists, but it's still empty (this, of course, happens very, very rarely).

connortechnology:
Cool, just wanted to know if we were reducing warnings or there was a greater issue.

IgorA100:
> or there was a greater issue.

No. I would definitely mention that.
I'm currently working on more accurate player switching and error handling. And various other minor errors are also being caught...

PR #4984, opened 2026-07-16
feat(decoder): "auto" hardware-decode detection with software fallback

anton-vinogradov:
## What

Adds an `auto` value for the per-monitor `DecoderHWAccelName` option. Instead of naming a specific hwaccel device type, `auto` probes every hwaccel `libav` exposes and uses the first one that both (a) the decoder advertises a matching hw config for and (b) whose device can actually be created — transparently falling back to software decoding when none work.

## Why

Today `DecoderHWAccelName` requires the user to know and name the correct device type (`vaapi`, `cuda`, …). A wrong or unavailable choice only falls back to software if creating *that one* device fails, and picking the right name across heterogeneous hardware is guesswork. This makes hardware decoding safe to turn on out of the box — useful for cutting the CPU cost of decoding high-resolution H.265/H.264 streams for motion analysis.

## How

In `FfmpegCamera::OpenFfmpeg()` the single-type block is replaced by a small candidate loop:

- `auto` → candidate list is every type from `av_hwdevice_iterate_types()`.
- any other value → the single named type (unchanged behaviour).

For each candidate it checks `avcodec_get_hw_config()` for a matching `hw_pix_fmt`, then tries `av_hwdevice_ctx_create()`, and uses the **first** that succeeds. A device-create failure now downgrades to a `Warning` and continues to the next candidate instead of erroring out. If the list is exhausted, `use_hwaccel` is cleared and decoding proceeds in software.

Single file changed (`src/zm_ffmpeg_camera.cpp`), no schema/config-option changes — `auto` is just a new accepted value for the existing string field.

## Compatibility

The default value and every explicit device name behave exactly as before. `auto` is strictly opt-in.

## Testing

Builds cleanly against current `master` (`zmc`).

Runtime-validated with a standalone probe that reproduces the exact selection loop and then decodes real event footage, transferring hw frames back to system memory. On an **Intel Celeron N5105 iGPU** (iHD driver, VAAPI):

| Scenario | Selected | Result |
| --- | --- | --- |
| `auto`, HEVC 1080×1920, iGPU present | `vaapi` | 201/201 frames hw-decoded + transferred, 152.8 fps |
| explicit `vaapi`, HEVC (legacy path) | `vaapi` | 201/201 hw |
| explicit `cuda`, HEVC (unavailable) | software | graceful fallback, 201 frames decoded |
| `auto`, H.264 1080×1920, iGPU present | `vaapi` | 148/148 frames hw-decoded |
| `auto`, HEVC, **no accelerator** (no `/dev/dri`) | software | graceful fallback, 201 frames decoded |
| explicit `none`/software | software | 201 frames decoded |

`auto` iterated `vdpau`/`cuda` (device-create failure → skipped) before selecting `vaapi`, and hardware decode ran ~1.8× the software throughput while offloading the CPU.

### DIFF
--- a/src/zm_ffmpeg_camera.cpp
+++ b/src/zm_ffmpeg_camera.cpp
@@ -27,6 +27,7 @@
 #include "url.hpp"
 
 #include <thread>
+#include <vector>
 
 extern "C" {
 #include <libavutil/time.h>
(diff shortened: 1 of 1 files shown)


connortechnology:
Nice! Thanks for this.  Next we could make it a comma separated list of hwaccels to try...

PR #4997, opened 2026-07-22
fix: only write analysis jpegs for alarm frames

### LINKED ISSUE #4996: Save JPEGs 'Analysis images only' writes every analysed frame since 1.38 instead of only alarmed frames
## Describe the bug

Since 1.38, setting **Save JPEGs** to *Analysis images only (if available)* writes an analysis jpeg for **every analysed frame**, not just alarmed frames. On a 10fps MOCORD-style monitor (Always Recording + Always Analysing) this produces ~6,000 jpegs per 10-minute event (~4GB vs ~400MB on 1.36).

Reported by a user upgrading 1.36.x → 1.38.3.

## Expected behavior (1.36)

Analysis images were only created for PREALARM/ALARM frames with an alarmed zone, and only when savejpegs > 1:

```cpp
// 1.36 zm_monitor.cpp
if (zone.Alarmed()) {
  if (zone.AlarmImage() and (savejpegs > 1) and snap->image) {
    if (!snap->analysis_image)
      snap->analysis_image = new Image(*(snap->image));
    snap->analysis_image->Overlay(*(zone.AlarmImage()));
```

So *Analysis images only (if available)* saved jpegs only for alarmed frames, with the zone overlay.

## Root cause

The 1.38 analysis pipeline creates `packet->analysis_image` unconditionally for every frame that goes through motion detection (added so the live Analysis stream shows full colour instead of greyscale) — `src/zm_monitor.cpp`:

```cpp
// Instead of showing a greyscale image, let's use the full colour
if (!packet->analysis_image)
  packet->analysis_image = new Image(*(packet->image));
```

`Event::AddFrame` (`src/zm_event.cpp`) then writes the analysis jpeg whenever the packet has one — which is now always:

```cpp
if (save_jpegs & 2) {
  if (packet->analysis_image) {
    std::string event_file = stringtf(staticConfig.analyse_file_format.c_str(), path.c_str(), frames);
```

## Fix

Gate the analysis-jpeg write on the frame actually being an alarm frame (`frame_type == ALARM`), restoring 1.36 semantics without affecting the live analysis view.

SteveGilvarry:
Fixes #4996

## Problem

Since the 1.38 analysis pipeline rewrite, `packet->analysis_image` is created for every frame that goes through motion detection (so the live Analysis view shows full colour instead of greyscale). `Event::AddFrame` wrote an analysis jpeg whenever the packet had one — which is now always — so **SaveJPEGs = Analysis images only (if available)** saved every analysed frame of the event instead of only alarmed frames as in 1.36. On a 10fps Always Recording + Always Analysing monitor that is ~6,000 jpegs (~4GB) per 10-minute event vs ~400MB on 1.36.

## Fix

Gate the analysis-jpeg write on `frame_type == ALARM`, restoring 1.36 semantics: analysis jpegs are only written for alarm frames (including scored pre/post-alarm buffer frames). The live analysis view is unaffected.

### DIFF
--- a/src/zm_event.cpp
+++ b/src/zm_event.cpp
@@ -608,7 +608,10 @@ void Event::AddFrame(const std::shared_ptr<ZMPacket>&packet) {
       }
     } // end if is an alarm frame
 
-    if (save_jpegs & 2) {
+    // Only write analysis jpegs for alarm frames. Since 1.38 every analysed
+    // packet carries an analysis_image (for the live analysis view), so
+    // writing whenever one exists would save every frame of the event.
+    if ((save_jpegs & 2) and (frame_type == ALARM)) {
       if (packet->analysis_image) {
         std::string event_file = stringtf(staticConfig.analyse_file_format.c_str(), path.c_str(), frames);
         Debug(1, "Writing analysis frame %d to %s", frames, event_file.c_str());


connortechnology:
I think the if (packet->analysis_image) used to do the gating.  But now we always assign analysis_image.  Instead we should only populate analysis-image if we have motion detection score (See line 2364 of zm_monitor.cpp).  

Thing is, AI comes later and may also create it, but with a zero motion score, so frame type != ALARM.

So I don't think this is the way to go.

PR #4999, opened 2026-07-22
Fix: Do not hide the control button "#mfbpanel" for filter=inline mode, since this panel, in addition to the filter, has other necessary controls on the Watch & Montage page

### LINKED ISSUE #4994: Regression: When using the left side menu, "#mfbpanel" has the class "hidden-shift" and is therefore not displayed.

IgorA100:
Closed https://github.com/ZoneMinder/zoneminder/issues/4994


IgorA100:
I made the necessary changes

PR #5003, opened 2026-07-24
Fix: Avoid duplicate "handleClick" calls on Watch page

IgorA100:
We don't need to additionally assign a listener via `setup_onclick` in `monitorStream`, since we have a global listener:
```
document.addEventListener('click', function(event) { 
  handleClick(event);
});
```


IgorA100:
@connortechnology 
Excuse me, but please roll back this PR.
Such a minor change has a side effect.
I'll try not to make hasty decisions again, but will examine each change in more detail.

PR #5013, opened 2026-07-27
When playing with ZMS MJPEG on the Watch page, do not try to connect to AudioMotion, as this is useless, and clear the visual information about the status of audio and video tracks.

IgorA100:


web/skins/classic/views/js/watch.js
```diff
@@ -1444,9 +1444,13 @@ function controlWhatDisplay(oldId, newId) {
   }
   if (noAudioMotion) {
     destroyAudioMotion(oldId);
-    document.querySelector('.stream-info-status-track').innerText = '';
+    monitorStream.updateStreamInfoStatusTrack("");
   } else {
-    connectAudioMotion(newId);
+    if (monitorStream.activePlayer.indexOf("zms") !== -1) {
```
connortechnology: I'd like to eventually get to a point where we could say something more like if (player.supportsAudio()) .  I think long term, zms will be more like a websocket server, which may actually server video/audio.  but that is a ways off.

IgorA100: > more like if (player.supportsAudio()) . I think long term, zms will be more like a websocket server, which may actually server video/audio.

Yes. I'd even separate audio/video streaming into a separate class. I was just thinking about this yesterday. We already know how to receive both audio and video tracks from streams, but since we've grown to have a huge "zoo" of streams, we need to structure them more and more intelligently.

IgorA100:
Oh, how embarrassing it is to make such typos :(
I've fixed the code.

PR #5016, opened 2026-07-27
Fix: Changed the handling of key presses for PanZoom

IgorA100:
We now assign them using the monitor ID rather than simply using the ".panzoom" namespace.
This allows us to correctly remove the listener for a specific monitor.
Previously, the listener wasn't removed, which resulted in unnecessary code execution.


IgorA100:
@connortechnology 
Although Copilot's comment was hidden, it was helpful.
I followed the recommendations.
I remove existing handlers for the namespace before rebinding the handler.

PR #5022, opened 2026-07-28
fix(tags): Sort tags per user per browser

### LINKED ISSUE #5021: Each user applying tags affects the global tag order
**Describe Your Environment**
- Version of ZoneMinder: 1.39.17 (latest commit)
- How you installed ZoneMinder: Docker - zoneminder-containers/eventserver-base

**Describe the bug**
When multiple users are applying tags to events at the same time, each user's latest applied tag moves to the top which affects the other user's sorted tag list.  When users are applying tags quickly using the buttons to apply the "last" tag, the "last" tag can change from what they applied last (and what they expect).

**To Reproduce**
1. User1 apply tag1 to an event
2. User2 apply tag2 to an event
3. User1 use the button in the tag input to "Apply the last tag, then play the next event" 
4. Notice that tag2 was applied while User1 expected tag1 to be applied

**Expected behavior**
Each user should have their own sorted list of tags that they have applied.

Simpler1:
Fixes #5021

### DIFF
--- a/web/ajax/event.php
+++ b/web/ajax/event.php
@@ -180,6 +180,11 @@ if ( canView('Events') or canView('Snapshots') ) {
     $values = array($_REQUEST['tid']);
     dbFetchAll($sql, NULL, $values);
 
+    // Record this tag against the user's personal recency order so it
+    // sorts to the top of THEIR dropdown next time, on any device.
+    require_once('includes/TagOrder.php');
+    \ZM\TagOrder::recordUsage($user->Id(), validCardinal($_REQUEST['tid']));
+
     ajaxResponse(array('response'=>$response));
     break;
   case 'removetag' :
--- a/web/ajax/tags.php
+++ b/web/ajax/tags.php
@@ -3,8 +3,34 @@
 switch ( $_REQUEST['action'] ) {
   case 'getavailabletags' :
     $sql = 'SELECT * FROM Tags ORDER BY LastAssignedDate DESC';
-    $dbFetchResult = dbFetchAll($sql);
-    ajaxResponse(array('response'=>$dbFetchResult));
+    $tags = dbFetchAll($sql);
+
+    // Re-sort so tags THIS user has applied recently come first, ahead of
+    // the global "most recently applied by anyone" fallback order. The
+    // recency list is stored server-side (User_Preferences), keyed by Tag
+    // Id, so it's correct even on a device this user has never used before.
+    require_once('includes/TagOrder.php');
+    $tagOrderIds = \ZM\TagOrder::load($user->Id());
+    if ($tagOrderIds) {
+      $byId = array();
+      foreach ($tags as $tag) {
+        $byId[$tag['Id']] = $tag;
+      }
+      $personal = array();
+      $used = array();
+      foreach ($tagOrderIds as $tid) {
+        if (isset($byId[$tid]) and !isset($used[$tid])) {
+          $personal[] = $byId[$tid];
+          $used[$tid] = true;
+        }
+      }
+      $rest = array_values(array_filter($tags, function($tag) use ($used) {
+        return !isset($used[$tag['Id']]);
+      }));
+      $tags = array_merge($personal, $rest);
+    }
+
+    ajaxResponse(array('response'=>$tags));
     break;
   case 'createtag' :
     $sql = 'INSERT INTO Tags (Name, CreatedBy) VALUES (?, ?)';
--- a/web/includes/TagOrder.php
+++ b/web/includes/TagOrder.php
@@ -1,0 +1,66 @@
+<?php
+//
+// ZoneMinder per-user tag ordering
+//
+// Stores each user's personal "recently used tags" list in User_Preferences
+// as a JSON array of Tag Ids (never Names -- names are free-text and can
+// collide case-insensitively or shadow Object.prototype keys on the client,
+// so Id is the only safe key). This lives server-side so the ordering is
+// correct on any device/browser the user logs into, not just the one that
+// wrote it.
+//
+namespace ZM;
+
+class TagOrder {
+  const MAX_ENTRIES = 50;
+  const PREF_NAME = 'tagOrder';
+
+  // Returns an array of Tag Ids, most-recently-used first. Always an array,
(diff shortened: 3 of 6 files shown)


IgorA100:
The issue https://github.com/ZoneMinder/zoneminder/issues/5021 does exist, but I think it needs to be addressed in a different way than in this PR.
Sorry, but this is just my humble personal opinion.

connortechnology:
I actually kinda like the way this is done.  I think we need to go further, but it is interesting.  THe one problem is that it won't persist across browsers.  Maybe that's not such a big deal.  I kinda feel like we may need to do a custom dropdown widget with a sorting preference option... because _I_ don't want to sort them this way, but I totally get why you would want to.  Let's discuss a bit more.  I'll ask claude what it thinks.

connortechnology:
(Claude's review, requested above. Checked against the branch, the schema, and the surrounding code; `npx eslint` passes on the branch.)

## Framing: the button and the dropdown are two different things

Issue #5021 is really about the **"Apply the last tag, then play next/prev event"** buttons (`event.php:281,291` → `tagAndNext()` → `addTag(availableTags[0])`), not about dropdown display order. This PR conflates them: it fixes the button by re-sorting the entire dropdown.

Decoupling them dissolves the disagreement in this thread. Track "my last applied tag" separately from dropdown sort order — @Simpler1 gets a correct button, @connortechnology keeps the dropdown sorted the way he prefers, and the sort-preference widget becomes an independent nicety rather than a prerequisite for fixing the bug.

## Persistence isn't just cosmetic here

The "won't persist across browsers" concern is more serious than it first looks, because the failure mode is silent and *writes to the database*. On a browser with no localStorage yet, `applyPersonalTagOrder()` returns the server list unchanged, so `availableTags[0]` is once again the globally-most-recent tag — and #5021 reproduces exactly: the user presses the button and **the wrong tag is applied to the event**. New device, cleared cache, private window, second workstation. A wrong-tag write shouldn't depend on client-side storage being warm.

The server already has the data: `Events_Tags.AssignedBy` is populated on every assignment (`web/ajax/event.php:175`). Two options:

1. **Derive it** — order `Tags` by `MAX(ET.AssignedDate) WHERE ET.AssignedBy = $user->Id()`, falling back to `Tags.LastAssignedDate`. Zero new state, retroactively correct on existing installs. Needs an index (`Events_Tags` PK is `(TagId, EventId)`; nothing on `AssignedBy`), and it decays as events are purged or tags un-assigned — probably acceptable, since recently-used tags sit on events that still exist.
2. **Store it** — a JSON recency list in `User_Preferences` (table exists; `$user->Preference('replayMode')` at `event.php:139` is the established pattern). No migration, no decay, and it's the same data model this PR already implements — just moved from localStorage to a per-user row. A `tagSortMode` preference (`personal` / `global` / `alphabetical`) lives naturally next to it.

Either way the sort belongs in `web/ajax/tags.php:5`, which currently hardcodes `ORDER BY LastAssignedDate DESC`, so the client stays dumb.

## Concrete bugs in the current implementation

Verified by running `applyPersonalTagOrder()` standalone:

**1. Case collision silently drops a tag.** `Tags.Name` is `latin1_bin` (`zm_create.sql.in:1397`), so `Dog` and `dog` are genuinely distinct rows. `byNameLower`/`usedNames` collapse them:

```js
applyPersonalTagOrder([{Id:1,Name:'Dog'},{Id:2,Name:'dog'},{Id:3,Name:'Cat'}], ['Dog'])
// => [dog, Cat]   -- "Dog" is gone
```

`Dog` disappears from the dropdown entirely and becomes unselectable.

**2. A tag named `constructor` throws and takes out the whole tag UI.** `byNameLower['constructor']` resolves to `Object.prototype.constructor`, which is truthy, so `tag.Name.toLowerCase()` throws. The exception escapes the `.done()` handler in `getAvailableTags()`, `availableTags` is never assigned, the dropdown stays empty, and `tagAndNext` then calls `addTag(undefined)`. Tag names are free-text user input via `createtag`. (This is Copilot's second comment — the impact is worse than "incorrect behavior".)

**3. `__proto__` as a tag name** drops that tag from the list. Not a security problem — the mutation is confined to a local object — but the same class of bug.

All three disappear if the recency list is keyed on **Tag Id** instead of name. I'd make that change regardless of where the data ends up living.

**4. Corrupt localStorage never self-heals** (Copilot's first comment — valid). A non-array value makes `saveTagUsage()` throw on `.filter()`; the `catch` swallows it, so the bad value is never overwritten and recency silently stops updating forever. An `Array.isArray()` guard in `loadTagOrder()` fixes it.

## Convention notes

- **`var currentUsername` in an inline `<script>` doesn't belong in `event.php`.** `web/skins/classic/views/js/event.js.php` is this view's established PHP→JS channel (`connKey`, `eventData`). It's also emitted inside `if ($Event->Id())`, so the global is undefined on no-event pages — which is why the `typeof` guard is needed; moving it to `event.js.php` removes that fork.
- **Username is the wrong key** — usernames are renameable; `$user->Id()` is stable and equally available. Also worth noting: with `ZM_OPT_USE_AUTH` off everyone is `defaultUser` = `admin` (`config.php.in:174`), so the per-user split silently doesn't apply.
- The new code is all `var` / `function(){}` while the surrounding tag code uses `const`, arrows and default params. ESLint doesn't complain, but it reads as bolted-on.

## Verdict

The idea is sound and the direction is right, but I wouldn't merge as-is.

Minimum before merge: key on Tag Id (fixes 1–3), add the `Array.isArray` guard, move the user identifier into `event.js.php`.

That said — since the localStorage-only approach still applies the wrong tag on a fresh browser, which is the actual bug in #5021, option 2 above looks like the better target. It's a small diff from what's already written here, needs no migration, and the sort-mode preference falls out of it for free.

Simpler1:
PHP Solution implemented that addresses all concerns.

Simpler1:
@connortechnology Are you wanting me to make the changes recommended in the Suppressed comments from Copilot before merging?

connortechnology:
I don't think so, I'm just too busy to look at anything. Will get to it on weekend

connortechnology:
Ok, so some of those should really be addressesd, but I'll get claude to do it.

PR #5041, opened 2026-08-08
Update CMakeLists.txt cache path

ovi-vansid:
Change default cache path for OpenBSD to /tmp (instead of /dev/shm)


connortechnology:
Hey I don't think this is the right change, as it sets the default for everyone.  I think what you want to do is add a section like what is at 306 for FreeBSD

ovi-vansid:
Sure thing, I can make that change.  Not 100% sure how the PR process works, so don't know if you need to reject it or if I need to cancel/close it.

connortechnology:
If you push changes, they will appear here, and when ready we merge.

PR #5044, opened 2026-08-08
Update CMakeLists.txt for OpenBSD

ovi-vansid:
Added OpenBSD as additional target distro and added condition for /tmp  as default cache path for OpenBSD.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,8 +199,13 @@ set(ZM_PATH_SHUTDOWN "/sbin/shutdown" CACHE PATH
   "Path to shutdown binary, default: /sbin/shutdown")
 
 # Advanced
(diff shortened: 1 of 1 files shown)


ovi-vansid:
Wasn't sure if I should use MATCHES or STREQUAL (I see both are used for distro comparisons).

connortechnology:
MATCHES is for using a regular expression.  So strequals is probably what you want although either will likely do.  I think this can be simplified instead of adding the extra if STREQUALS OpenBSD, at the initial setting of ZM_PATH_MAP, I think you can just add it to the area down below that you have added. So all the OpenBSD stuff is in one place.

ovi-vansid:
I'll group all of these together and change the comparisons to STREQUALS.  You can reject this PR and I'll submit another.

PR #5045, opened 2026-08-09
Update CMakeLists.txt for OpenBSD

ovi-vansid:
Added OpenBSD as additional target distro and added condition for /tmp as default cache path for both FreeBSD and OpenBSD.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,13 +199,6 @@ set(ZM_PATH_SHUTDOWN "/sbin/shutdown" CACHE PATH
   "Path to shutdown binary, default: /sbin/shutdown")
 
 # Advanced
-if(ZM_TARGET_DISTRO MATCHES "OpenBSD")
-  set(ZM_PATH_MAP "/tmp" CACHE PATH
-  "Location to save mapped memory files, default: /dev/shm")
-else()
-  set(ZM_PATH_MAP "/dev/shm" CACHE PATH
-  "Location to save mapped memory files, default: /dev/shm")
-endif()
 set(ZM_PATH_ARP "" CACHE PATH
   "Full path to compatible arp binary. Leave empty for automatic detection.")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -199,8 +199,6 @@ set(ZM_PATH_SHUTDOWN "/sbin/shutdown" CACHE PATH
   "Path to shutdown binary, default: /sbin/shutdown")
 
 # Advanced
-set(ZM_PATH_MAP "/dev/shm" CACHE PATH
```
connortechnology, line 202: This has to remain as the default

connortechnology: Oh I see you added an else down at the bottom.  But we are setting all the defaults up here, so let's just leave it here.

connortechnology:
Hey don't both closing then opening a new pr.  Just add your change to the existing branch. The PR will auto-update.

CMakeLists.txt
ovi-vansid: Yes, you said to keep all the BSD customization in one place, so I didn't see a better way to fit the default configuration without adding an `else()` at the bottom of that portion.  Do you want to move it back to the top of that section?  I think doing so would need another `if` block.

ovi-vansid:
> Hey don't both closing then opening a new pr. Just add your change to the existing branch. The PR will auto-update.

Sorry, this is me: [https://xkcd.com/1597/](https://xkcd.com/1597/)

connortechnology:
No worries, just trying to turn you into a lean, mean gitting machine.   If we leave the line at the top, your if OpenBsd down in the 300's will override it.  Don't need an if at the top.  Maybe we should move them all down to your else, but that's a bigger cutnpaste.  I need to get around to adding Debian/Ubuntu defaults as well.

connortechnology:
That looks great!

PR #5055, opened 2026-08-16
fix: make MonitorStream playback state shared with the command thread atomic

### LINKED ISSUE #4939: zms: MonitorStream shares mutable playback state across threads without synchronization
### Summary

`MonitorStream` runs two threads: the `command_processor` thread (`checkCommandQueue` → `processCommand`) and the main streaming loop in `runStream()`. `processCommand` mutates a set of plain (non-atomic, unguarded) member fields that the main loop reads on every iteration, with no mutex or atomics between them. This is a data race (undefined behavior under the C++ memory model), independent of any single symptom.

### Affected fields (written by the command thread, read by the stream loop)

- `paused`
- `replay_rate`
- `scale`
- `zoom`, `x`, `y`
- `temp_read_index`, `temp_write_index`, `temp_image_buffer_count`
- the seek/`send_frame`-related state

`processCommand` sets these (e.g. `paused = true`, `replay_rate = …`, `scale = …`) on the command thread, while `runStream()` reads them to decide what/when to send.

### Why this matters

In practice it usually "works" on x86 because aligned word reads/writes are atomic-ish and the values are individually small, but it is still UB: there are no ordering or visibility guarantees, and a torn or stale read of the buffer indices can produce wrong playback behavior. It also means tools like TSan will flag it.

### Context

This came out of #4936 / #4938. That work fixed the concrete crash:
- a divide-by-zero (SIGFPE) when `temp_image_buffer_count` was read as 0 during startup — guarded with a `MonitorStreamBufferLevel()` helper, and
- the command thread being started before the buffer state was initialised — now deferred until after init.

Those address the **startup window**. This issue tracks the **ongoing concurrent access** to the shared playback state, which is broader and riskier to change, so it is intentionally split out rather than folded into the SIGFPE fix.

### Suggested direction

- Make the simple scalars (`paused`, `replay_rate`, `scale`, `zoom`, `x`, `y`) `std::atomic`, and/or
- Guard the buffer indices/count as a unit under the existing stream mutex, since they must be read consistently together.
- Run the streamer under ThreadSanitizer (`-DTSAN=ON`) to enumerate the full set of racing accesses before deciding on the locking strategy.

AJ0070:
Fixes #4939.

`processCommand()` runs on the command thread and mutates playback state that `runStream()` reads every iteration, with nothing between them. That is a data race and therefore undefined behaviour, regardless of whether it currently appears to work on x86.

**Fields made atomic**

In `StreamBase`: `replay_rate`, `scale`, `zoom`, `x`, `y`, `paused`, `stopped`, `step`, `send_twice`, `maxfps`, `frame_type`. In `MonitorStream`: `delayed`, `temp_read_index`, `temp_write_index`, `temp_image_buffer_count`.

The `last_*` companions (`last_scale`, `last_zoom`, `last_x`, `last_y`, `last_crop`) stay plain. Only the streaming loop touches those.

**Verified under ThreadSanitizer**

The issue asked for this to be run under TSan, so it was. `tests/zm_stream_races.cpp` drives a minimal `StreamBase` subclass from two threads, one writing the fields the way `processCommand()` does and one reading them the way `runStream()` does. It needs no database, shared memory or capture daemon, so it runs anywhere the test suite builds.

```
cmake -DTSAN=ON -DBUILD_TEST_SUITE=ON .. && cmake --build .
./tests/tests "[races]"
```

Against master, TSan reports 7 data races, on `paused`, `replay_rate`, `zoom`, `x`, `send_twice`, `maxfps` and `frame_type`:

```
WARNING: ThreadSanitizer: data race (pid=2354883)
  Read of size 1 by thread T18:
    #0 readPlaybackState tests/zm_stream_races.cpp:67
  Previous write of size 1 by thread T17:
    #0 applyCommands tests/zm_stream_races.cpp:58
```

With this change, 0. The full suite under TSan is also clean: 122 test cases, 10023 assertions, no reports.

One environment note for anyone reproducing: TSan aborts with `unexpected memory mapping` under the default ASLR settings on recent kernels, so the binary needs `setarch $(uname -m) -R ./tests`.

**Why atomics rather than the mutex for the buffer indices**

The issue suggested guarding the index triple under the existing mutex. I did not, and it is worth saying why. `runStream()` owns every write to those three, and they are read across threads in exactly one place: `processCommand()` computing a buffer-level percentage for the status reply. The indices are otherwise read and written throughout the streaming hot loop, so taking a lock around them would put a mutex acquisition in the per-frame path to protect a status number.

Atomics remove the undefined behaviour, which is the actual defect. The remaining looseness is that the three can be read from slightly different moments. That is harmless here: each index individually stays within `[0, count)`, so their difference stays within `(-count, count)`, and the `+ count` before the modulo in `MonitorStreamBufferLevel()` covers that whole span. The result is always a sane 0-100. There is a test section that exhaustively checks every index pairing for a 64-slot buffer to pin that down.

If you would rather have the strict version, say so and I will move it under `monitor_mutex`, but it costs a lock per frame.

**Consistency snapshots**

Making these atomic exposed a subtler problem: several expressions read the same field two or three times, so they could mix an old value with a new one. Those now snapshot once into a local:

- `prepareImage()` derives a crop rectangle from `zoom`, `x`, `y` and `scale` together. Re-reading mid-calculation could pair a new zoom with an old click position and produce a crop matching neither. It also now records `last_scale`/`last_zoom`/`last_x`/`last_y` from the snapshot, so they describe what was actually rendered rather than whatever arrived since.
- `updateFrameRate()` re-read `maxfps` on every iteration of its halving loop.
- The sleep-time calculation in `runStream()` read `maxfps` three times and `replay_rate` twice in one expression.

**Not addressed**

`now`, `last_frame_sent` and `last_fps_update` are `TimePoint`s that also cross the two threads. Making them atomic means explicit loads at every arithmetic site, which is a much larger and more invasive diff than this one, so I left it out rather than bundle it in. The new harness does not cover them either, since it exercises the fields directly rather than the timing logic. Worth a follow-up issue if you agree.

**Testing**: full suite passes, 122 test cases and 10023 assertions, in both a normal build and a TSan build. `zms`, `zmc` and `zmu` all build warning-free.

### DIFF
--- a/src/zm_eventstream.cpp
+++ b/src/zm_eventstream.cpp
@@ -497,10 +497,10 @@ void EventStream::processCommand(const CmdMsg *msg) {
     paused = false;
     replay_rate = (((unsigned char)msg->msg_data[1]<<8)|(unsigned char)msg->msg_data[2])-VARPLAY_RATE_OFFSET;
     if (replay_rate > 50 * ZM_RATE_BASE) {
-      Warning("requested replay rate (%d) is too high. We only support up to 50x", replay_rate);
+      Warning("requested replay rate (%d) is too high. We only support up to 50x", replay_rate.load());
       replay_rate = 50 * ZM_RATE_BASE;
-    } else if (replay_rate < -50*ZM_RATE_BASE) {
-      Warning("requested replay rate (%d) is too low. We only support up to -50x", replay_rate);
+    } else if (replay_rate.load() < -50*ZM_RATE_BASE) {
+      Warning("requested replay rate (%d) is too low. We only support up to -50x", replay_rate.load());
       replay_rate = -50 * ZM_RATE_BASE;
     }
     break;
@@ -532,7 +532,7 @@ void EventStream::processCommand(const CmdMsg *msg) {
       replay_rate = 50 * ZM_RATE_BASE;
       break;
     default :
-      Debug(1,"Defaulting replay_rate to 2*ZM_RATE_BASE because it is %d", replay_rate);
+      Debug(1,"Defaulting replay_rate to 2*ZM_RATE_BASE because it is %d", replay_rate.load());
       replay_rate = 2 * ZM_RATE_BASE;
       break;
     }
@@ -587,10 +587,10 @@ void EventStream::processCommand(const CmdMsg *msg) {
   case CMD_ZOOMIN :
     x = ((unsigned char)msg->msg_data[1]<<8)|(unsigned char)msg->msg_data[2];
     y = ((unsigned char)msg->msg_data[3]<<8)|(unsigned char)msg->msg_data[4];
-    Debug(1, "Got ZOOM IN command, to %d,%d", x, y);
+    Debug(1, "Got ZOOM IN command, to %d,%d", x.load(), y.load());
     zoom += 10;
     send_frame = true;
-    if (paused) {
+    if (paused.load()) {
       step = 1;
       send_twice = true;
     }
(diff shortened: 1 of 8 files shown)


AJ0070:
Followed up on the TSan request in the issue rather than leaving it as reasoning. Added `tests/zm_stream_races.cpp`, which drives a minimal `StreamBase` subclass from two threads: one writing the fields the way `processCommand()` does, one reading them the way `runStream()` does. No database, shared memory or capture daemon needed, so it runs anywhere the suite builds.

Result: against master TSan reports 7 data races, on `paused`, `replay_rate`, `zoom`, `x`, `send_twice`, `maxfps` and `frame_type`. With this change, 0. The full suite under TSan is clean as well, 122 cases and 10023 assertions.

Two things worth knowing if you reproduce it:

- TSan aborts with `unexpected memory mapping` under default ASLR on recent kernels. `setarch $(uname -m) -R ./tests` works around it.
- Catch2 runs the binary at build time for test discovery, which trips the same thing, so the discovery step needs skipping or the same wrapper in a TSan build.

The harness covers the scalar playback state. It does not cover the `TimePoint` fields (`now`, `last_frame_sent`, `last_fps_update`), which are still racing and still out of scope for this PR.

PR #5057, opened 2026-08-17
fix: do not treat EPERM as missing hard link support on FreeBSD

AJ0070:
#5056 treats EPERM as "filesystem has no hard links". That's right on Linux (exFAT, FAT) but wrong on FreeBSD, where link(2) reports missing link support as EOPNOTSUPP and uses EPERM only for a directory source, an immutable source, or an immutable target dir. All real errors, so right now FreeBSD silently renames on them and drops the log from Error to Debug. FreeBSD is in CI.

So: skip EPERM on FreeBSD, Linux unchanged.

Also stopped the rename path logging "Filesystem does not support hard links", since on Linux EPERM is usually protected_hardlinks or an immutable source. Logs the errno instead.

### DIFF
--- a/src/zm_event.cpp
+++ b/src/zm_event.cpp
@@ -238,9 +238,12 @@ Event::~Event() {
       // hard links there is no way to have the file under both names at once,
       // and a brief window beats every event on the system being named as
       // though it never finished recording.
+      const int link_errno = errno;
       if (rename(video_incomplete_path.c_str(), video_path.c_str()) == 0) {
-        Debug(1, "Filesystem does not support hard links, renamed %s to %s",
-              video_incomplete_path.c_str(), video_path.c_str());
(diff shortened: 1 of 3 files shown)


connortechnology:
Your AI is out of control!  We don't need yet another file with that much code just to rename a file.

AJ0070:
Fair, cut it back to just the bug. +36/-17 across 3 files now, no new files, helper stays inline in zm_event.h where you had it.

The bit worth keeping is FreeBSD: EPERM there only ever means a directory source, an immutable source, or an immutable target dir, and missing link support comes through as EOPNOTSUPP. So #5056 turns three real errors into a silent rename. Linux untouched.

Also dropped the "Filesystem does not support hard links" log line, it's usually wrong on Linux. Logs the errno now.

The rest was me over-building to satisfy an automated review comment. Gone.

PR #5059, opened 2026-08-20
The "rand" parameter change for SRC has been separated into a separate function, and a function for replacing SRC has been added.

IgorA100:
- We will use `Date.now()` instead of `Math.random()`.
- We also now use `refreshStreamSrc()` when stopping stream playback to avoid retrieving a cached image from the browser.

For example, when viewing a stream on the Watch page for a long time and then clicking `stop`, Firefox uses the cached image, not the current one. For example, we're viewing a stream and the camera switches to black-and-white mode. We click `stop`, and we get a color image cached XX minutes ago. It's ugly.

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -2048,8 +2048,18 @@ function initThumbAnimation() {
   }
 }
 
-function setSrc(elem, src) {
-  elem.src = src + (src.includes('?') ? '&' : '?') + '_=' + Date.now();
+function addOrUpdateRandParam(src) {
+  const rand = Date.now();
+  if (/[?&]rand=\d+/i.test(src)) {
+    return src.replace(/([?&])rand=\d+/i, '$1rand=' + rand);
+  }
+  return src + (src.includes('?') ? '&' : '?') + 'rand=' + rand;
+}
+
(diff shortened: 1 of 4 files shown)


web/skins/classic/js/skin.js
```diff
@@ -2048,6 +2048,10 @@ function initThumbAnimation() {
   }
 }
 
+function setSrc(elem, src) {
+  elem.src = src + (src.includes('?') ? '&' : '?') + '_=' + Date.now();
```
connortechnology: There should already be a rand= in the query string.  THis is a little redundant. Maybe just update the rand=, or add it if not present.

IgorA100: > or add it if not present.

OK.

IgorA100:
@connortechnology 
I wanted to make a minimal change, but I also had to refactor the code :)
I think I've completed this PR.

I haven't tested the code 100% after refactoring, but there shouldn't be any issues.

web/skins/classic/js/skin.js
IgorA100: Done https://github.com/ZoneMinder/zoneminder/pull/5059/changes/32f0e37583a391864e095613d2e4a37bba8d8256

Issue #3301, opened 2021-06-21
handle out of ram in events list


connortechnology:
Trying to list a large result set like say the month colu,m easily exhausts the default 128Mb ram limit.  

We should detect and give a decent error message or somehow not require that much ram.Basically if the ajax fails we need to display a useful message instead of just loading, please wait.

SteveGilvarry:
**Root cause:** the events query endpoint (`web/ajax/events.php` → `queryRequest`) has no SQL `LIMIT` for pagination. It `SELECT`s every matching row, materializes them all into `$unfiltered_rows`, then `array_slice`s for the requested page. It has to materialize the whole set because `total`/`totalNotFiltered` are computed as `count()` of those arrays (events.php:374–378), so the `offset`/`limit` from the client are only applied in PHP, not in SQL — a 25-row page of a large month still loads the whole month into memory.

**Scale:** ~3.35 KB per materialized row (raw fetch); with `$filtered_rows`, storage/disk lookups and `json_encode`'s transient copy, realistic peak is ~7–13 KB/row. On the default 128M limit that's an OOM somewhere around 8–15k events in a single result set, which a busy multi-camera month can reach. (Won't reproduce on a small instance or one with `memory_limit = -1`.)

Two fixes, matching the "give a decent error message **or** don't require that much ram":

### 1. Front-end: stop the endless spinner and show a useful message

The current `ajaxRequest` error handler is a no-op (just `console.log`), so on a 500 the table stays on "Loading...". `params.error()` is bootstrap-table's contract to stop the spinner (it calls `hideLoading()` and fires `load-error.bs.table`):

```js
    error: function(jqXHR) {
      // the previous request is aborted on every reload; ignore that
      if (jqXHR.statusText == 'abort') return;

      // tell bootstrap-table the load failed so it stops the "Loading..."
      // spinner instead of hanging forever
      params.error(jqXHR);

      // a 500/0 on a large result set is almost always the server running out
      // of memory building the response
      const likelyTooBig = (jqXHR.status === 500 || jqXHR.status === 0);
      alert(likelyTooBig
        ? 'Failed to load events. The result set may be too large for the server to build at once — narrow the filter (e.g. a shorter date range) or reduce the page size.'
        : 'Failed to load events (HTTP ' + jqXHR.status + ').');
    }
```

This helps every failure mode, not just OOM, and is independent of #2.

### 2. Server-side: bound the memory

Push pagination into SQL so PHP never materializes the whole set:
[...]

PR #5069, opened 2026-08-26
fix: clear all logs when no rows are selected

AJ0070:
Fixes #4727, diagnosed by @SteveGilvarry.

The button stays disabled until rows are selected, so pressing it on a freshly loaded Log view does nothing at all, with no error anywhere to explain why. No selection now means clear all, with the button reading Clear ALL Logs so the difference shows before the confirmation does. The backend takes an explicit `all` and clears in one DELETE rather than 100 ids per request.

Tests: `node tests/js/clear-logs.test.js` (10), `php tests/php/test_log_clear_all.php` (12).


IgorA100:
I really like this idea!
But I think if no rows are selected, the button label would be more accurate: "Clear ALL Logs." And when at least one row is selected, the button label should remain as it is now: "Clear Logs."
This is just my suggestion. We'll see what Isaac @connortechnology has to say.

AJ0070:
Good call, done. Button reads Clear ALL Logs with nothing selected and Clear Logs once rows are picked, tooltip follows it.

web/ajax/log.php
```diff
@@ -31,6 +31,13 @@
       $ids = array_map('intval', (array)$_REQUEST['ids']);
       $placeholders = implode(',', array_fill(0, count($ids), '?'));
       dbQuery('DELETE FROM Logs WHERE Id IN (' . $placeholders . ')', $ids);
+    } else if (!empty($_REQUEST['all'])) {
+      // Deleting by id needs one request per 100 rows, which is unusable at the
+      // hundreds of thousands of rows people accumulate, and the table's
+      // check-all only ever selects the current page. This clears the lot in one
+      // statement. It is a separate parameter rather than "no ids given" so a
+      // malformed request that lost its ids cannot empty the table by accident.
+      dbQuery('DELETE FROM Logs');
```
connortechnology: What id we had some filters populated, then this could just clear all results matching that filter.  

Can we do truncate if there are none?

AJ0070: Both done. logFilter() is now shared by the query and the delete, so a filtered clear removes exactly the rows on screen, and with nothing filtered it truncates.

Two things worth flagging from doing it. A filter value that fails to parse (a typed-in bad date, an unknown level) used to leave an empty WHERE, which the delete would have read as "nothing filtered" and truncated the lot, so it now refuses instead. And the ctrl-click confirmation bypass could reach the clear-everything path once the button stopped being disabled, so that is limited to selected-row deletes again.

Note TRUNCATE resets AUTO_INCREMENT where DELETE does not, so log ids restart at 1.

connortechnology:
Verified on the branch: 19 JS assertions pass, all 20 PHP cases pass, eslint and `php -l` clean.

This is good work. The core design is right in a way that matters: `rememberFilter(params.data)` captures the exact payload the last query sent, and posting it back with `all: 1` through the same `logFilter()` builder gives parity by construction rather than by two code paths agreeing to be careful. That is the correct shape for a destructive operation.

One thing to fix before merge.

## 1. The three new failure messages can never reach the user

`ajaxError()` (`web/includes/functions.php:1812`) answers HTTP **200** with `{result:'Error'}`, and `deleteLogs`'s `success` handler never checks `data.result`. So all three of the new messages — "Cannot clear logs: the filter could not be interpreted", "Failed to delete the filtered log entries", "Failed to clear the log table" — land in the success path: progress bar to 100%, table refreshed, alert closed, nothing said.

Concretely: a user types a half-formed date, the button says "Clear Filtered Logs", they confirm, and they get a successful-looking progress bar with every log still there and no explanation. The `unparseable-date` and `delete-fails` cases prove the server refuses correctly; nothing tests that the browser says so.

This is the same class of bug as #3301 and the same fix you just landed in #5081 — check `data.result == 'Error'` in `success` and `zmAlert(data.message, translate['AJAXRequestError'])`. Worth a test case for it too.

## 2. `count($advsearch)` fatals on a non-JSON filter

`json_decode($_REQUEST['filter'], JSON_OBJECT_AS_ARRAY)` returns null for junk, and `count(null)` is a TypeError on PHP 8 (checked on 8.4.24). Pre-existing in `queryRequest`, but the delete path reaches it now too. `$advsearch = is_array($advsearch) ? $advsearch : array();` is a one-liner worth doing while you are in the function. It fails closed either way, so this is robustness rather than a hole.

## 3. Smaller things

- `in_array($col, $columns)` is non-strict and `$col` is concatenated into SQL at `$col.' LIKE ?'`. PHP 8 semantics make it safe today, but a `true` third argument is free and documents the intent.
- TRUNCATE resets `AUTO_INCREMENT` on `Logs.Id` and the DELETE fallback does not. Cosmetic for logs, but the comment presents them as interchangeable — worth half a sentence.
- `clearLogsBtn.setAttribute('title', text)` will not move the tooltip. The button has `data-toggle="tooltip"` and Bootstrap caches the title into `data-original-title` at init, so the tooltip keeps saying "Clear Logs" while the label says "Clear ALL Logs".

## What I checked that holds

Worth stating, since these are the parts that could have been dangerous:

- `canEdit('System')` gates both the ids path and the new all path (`web/ajax/log.php:26-28`).
- `queryRequest` uses `$filter['where']` verbatim and appends only ORDER/LIMIT, so the delete's WHERE is byte-identical to the query's.
- No JS/PHP divergence on "is a filter active" — the failure mode I went looking for, where the browser promises "Clear Filtered" and the server sees no filter and truncates. It cannot happen: all seven `FILTER_KEYS` are only ever set non-empty, `level`/`Component` are guarded by `levels && levels.length` (`log.js:83-92`), and bootstrap-table only emits `filter` when the object is non-empty (`bootstrap-table.js:10736`), so it can never arrive as `'{}'`.
- The `1=0` result for a fully-rejected advanced search is the right choice, and `$requested` correctly separates "asked to filter" from "produced SQL". That distinction is what stops a mistyped date from truncating the table, and it is the heart of this PR.
- Requiring a selection for the ctrl-click bypass was a good catch on your part — that shortcut reaching clear-everything would have been nasty.
- The button starts `disabled` in the HTML and is enabled on `load-success.bs.table` (`log.js:445-448`), so a clear-all cannot fire before `rememberFilter` has run.

Fix 1 and I will merge this.

PR #5070, opened 2026-08-26
fix: lock to stop a second zmcontrol server per monitor

### LINKED ISSUE #4423: zmcontrol needs locking
When zmcontrol is run given a command, if there isn't a zmcontrol process running, we start a new daemon process.  But there is no locking, we can run many of these and since they talk on a socket... it's a mess.

If the socket exists, this should be an error and the daemon process should exit.

AJ0070:
Fixes #4423.

The server branch unlinked the socket before binding it, so a second control server for the same monitor took it over, leaving the first listening on an unlinked inode while both still held the camera's control connection. It now flocks `zmcontrol-<id>.lock` before touching the camera or the socket and exits if another server holds it. The lock file carries the holder's pid, and one that cannot be opened is reported separately from one that is held, so a file left behind by a hand run as root does not read as a duplicate server.

This covers the socket server only. `--protocol` direct mode returns before the lock is taken and can still contend with a running server for the camera.

Tests: `perl tests/perl/test_exclusive_lock.pl`, 20 passed.

### DIFF
--- a/scripts/ZoneMinder/lib/ZoneMinder/General.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/General.pm
@@ -35,6 +35,7 @@ our %EXPORT_TAGS = (
       daemonControl
       parseNameEqualsValueToHash
       hash_diff
+      acquireExclusiveLock
       ) ]
     );
 push( @{$EXPORT_TAGS{all}}, @{$EXPORT_TAGS{$_}} ) foreach keys %EXPORT_TAGS;
@@ -56,6 +57,29 @@ use ZoneMinder::Logger qw(:all);
 use ZoneMinder::Database qw(:all);
 
 use POSIX;
+use Fcntl qw(:flock);
+
+# Take an exclusive, non blocking lock on $path, creating it if needed. Returns
+# the open filehandle on success and undef if someone else already holds it or
+# the file cannot be opened. The lock lives for as long as the caller keeps the
+# returned handle, so store it somewhere that outlives the work it guards. It is
+# released when the handle goes out of scope or the process exits, including on
+# a crash, which a lock built out of a pid file or the presence of a socket
+# cannot manage.
+sub acquireExclusiveLock {
+  my $path = shift;
+
+  my $fh;
+  if (!open($fh, '>', $path)) {
+    Error("Can't open lock file $path: $!");
+    return undef;
+  }
+  if (!flock($fh, LOCK_EX|LOCK_NB)) {
+    close($fh);
+    return undef;
+  }
+  return $fh;
+}
 
 # For running general shell commands
 sub executeShellCommand {
@@ -707,6 +731,7 @@ of the ZoneMinder scripts
       systemStatus
       parseNameEqualsValueToHash
       hash_diff
+      acquireExclusiveLock
       ) ]
 
 
--- a/scripts/ZoneMinder/lib/ZoneMinder/General.pm
+++ b/scripts/ZoneMinder/lib/ZoneMinder/General.pm
@@ -36,6 +36,7 @@ our %EXPORT_TAGS = (
       parseNameEqualsValueToHash
       hash_diff
       acquireExclusiveLock
+      lockHolderPid
       ) ]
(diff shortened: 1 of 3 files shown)


connortechnology:
Thanks for this. I fetched the branch, ran `perl tests/perl/test_exclusive_lock.pl` (9 passed, 0 failed) and syntax-checked `General.pm`.

The core is right. The lock is taken before `$control->open()` and before `unlink($sock_file)`, which is exactly where it has to be for #4423. Using a separate `.lock` file rather than flocking the socket is the right call — a socket left behind by a SIGKILLed server can't wedge startup forever. `$lock_fh` is scoped to the whole server block so the kernel releases it on any exit path including a crash, and the comment saying so is accurate. The test covers the property that actually matters (a killed holder must not block the next server).

Three things before I merge it.

**1. A lock file created by a different user permanently bricks control for that monitor.**

`misc/zoneminder-tmpfiles.conf.in:1` creates the sock dir as `0755 www-data:www-data`, not sticky. `acquireExclusiveLock` uses `open($fh, '>', $path)`, which needs write permission on the *file*. Today a root-owned `zmcontrol-1.sock` is harmless — the server unlinks it and the directory is www-data-writable. A root-owned `zmcontrol-1.lock` is not: it lands 0644 root:root and www-data can never open it for write again.

That's a realistic path — running `zmcontrol.pl --id N` as root to debug a control module is something we suggest in the tree already (see the comment in `Control/Trendnet.pm:166`). After that, every start logs `Can't open lock file ...: Permission denied` followed by `Another zmcontrol server is already running for monitor N, exiting`, and PTZ stays dead until someone deletes the file by hand. The second message is the bad part — it sends you hunting for a process that doesn't exist.

Please do both: create it world-writable (`sysopen($fh, $path, O_WRONLY|O_CREAT, 0666)` plus an explicit `chmod`, since umask 022 will otherwise strip it), and have `acquireExclusiveLock` distinguish "cannot open" from "held by someone else" so the caller doesn't assert a duplicate server that isn't there. A test for the two-outcome distinction would be welcome alongside the existing ones.

**2. The lock file is truncated and then left empty.**

`open '>'` truncates and nothing is ever written, so `cat zmcontrol-1.lock` tells the operator nothing and `fuser`/`lsof` is the only way to find the holder. Write `$$` into the handle after the flock succeeds — it costs nothing and lets the failing caller read the pid back and log "held by pid N", which makes the error in (1) actionable.

**3. Under zmdc, `exit(1)` becomes a permanent restart loop — worth a comment in the code.**

zmdc starts children with `keepalive=>1` (`scripts/zmdc.pl.in:498`), so the reaper reschedules any non-zero exit with exponential backoff capped at `ZM_MAX_RESTART_DELAY`. The realistic duplicate case — zmdc restarted and lost its pid table while an orphaned control server kept the socket — now becomes an `Error` every <=15 min forever instead of a socket takeover. That's the better failure and I'm fine with it, but nothing self-heals and the log doesn't tell anyone what to do. Fixing (2) mostly covers it; a line in the comment noting that the exit is expected to be retried by zmdc would help the next reader.

One note, not a blocker: `--protocol` direct mode doesn't take the lock. Fine for the socket half of the bug, but the PR description also cites "both still held the camera's control connection" — direct mode can still contend with a running server for the camera. Out of scope for #4423, just don't let the description imply it's covered.

Things I checked that are *not* problems: taint under `-wT` is fine, since the existing `unlink($sock_file)` already exercises a taint-checked op on the same `$Config{ZM_PATH_SOCKS}`. `Fcntl`'s `:flock` doesn't collide with the existing `use POSIX`. The 100644 mode on the new test matches the other four files in `tests/perl/`.

AJ0070:
All three done. acquireExclusiveLock returns ($fh, $status, $pid) now, sysopens without O_TRUNC and chmods 0666, so a root-created lock file no longer wedges the monitor and the holder's pid survives a failed acquire. zmcontrol names the pid for a held lock and points at the file when it cannot open one at all.

Also corrected the description, since direct --protocol mode does not take the lock. Tests are 20 now.

PR #5076, opened 2026-08-27
feat: derive zone detection settings from a drawn rectangle

### LINKED ISSUE #5075: Zone editor: derive detection thresholds from a drawn rectangle
Setting up a zone means putting numbers into Min Alarmed Area, Min Filtered Area and Min Blob Area. There's no good way to know what a person at your gate is actually worth in pixels, so it ends up being guesswork followed by a lot of trial and error against real events.

It would be much easier to draw a rectangle around the smallest thing you care about and have the thresholds worked out from that.

The docs already give the rule for it — definezone.rst [J] suggests a minimum Alarmed Area of "25% of the area that an object of interest takes up in the Zone region" — but there's nothing in the UI that applies it for you.

pliablepixels:
Fixes #5075

Adds a measuring tool to the zone editor. Arm the crop button, drag a box around the smallest object you want to detect, and the thresholds are filled in from its area.

A few notes:

- It switches the zone to Blobs, because MinBlobPixels is the only threshold checked per blob. The other two are zone-wide sums, so a handful of leaves can add up and trip them even though nothing person-sized moved.
- The Max fields are deliberately left blank. They set overload_count or drop the blob entirely, and one rectangle can't tell you a sensible ceiling when the same person is much bigger up close.
- MinPixelThreshold and the filter size are left alone if you've already set them, since they're properties of the camera rather than the object. Pick a sensitivity preset first and measuring won't undo it.
- You can load a still image instead of the live view, which is handy for measuring against a frame you already have rather than waiting for someone to walk past. It's read in the browser, nothing gets uploaded.


connortechnology:
Cool feature.  Looks great.

connortechnology:
@pliablepixels I've opened pliablepixels/ZoneMinder#7 against `zone-object-size-helper` with the four review points `eb0ad029f` left open, so merging it here updates this PR rather than replacing it.

- **Ratio persistence** — `zone.php` reloads those cookies through `validInt()`, which strips non-digits, so `70.5` was stored and returned as `705` while the live calculation read it as `70.5` through `parseFloat`. Rounded and range-checked before storing, and written back into the field.
- **Page Reset** — cleared the tool state, which Reset was leaving behind. Wrapped `resetBtn.onclick` rather than adding a listener, because it has to run before `resetChanges` calls `updateArea`.
- **Zone area changes** — wrapped `updateArea` to re-derive from the box still on screen, so the stored percentages keep describing the measured object after a vertex moves.
- **Touch and keyboard** — pointer events with capture (plus `pointercancel` and `touch-action`), and an object size in capture pixels beside the ratios as the keyboard-operable equivalent. It runs the same derivation and raises the same filter warning, and a finished drag writes its size back into those fields.

One extra: the `Type` listener covers the select, but `applyPreset` writes `Type` directly and fires no change event, so a preset making the zone Inactive or Privacy left the tool armed. `pointerdown` now re-checks and stands down.

Tests gained coverage of the two new pure functions — 46 assertions pass, ESLint and `php -l` clean. The DOM wiring is source reading, so the drag, the Reset path and the typed size still want a browser check, especially on touch.

connortechnology:
Moved this to #5119 in the main repo, which carries both of your commits unchanged plus a third finishing the review items that were still open (ratio persistence through `validInt`, page Reset leaving tool state, re-deriving when the zone area changes, and pointer/touch/keyboard operation). The fork PR I opened earlier is closed in favour of it.

The DOM wiring is browser-tested there — details on #5119. Nothing about your work changed; #5076 can be closed once #5119 goes in.

Issue #5074, opened 2026-08-26
Event export fails silently — per-event export subdirectory created without execute permission (missing chmod)


techrockedge:
Describe the bug
When exporting events (Console → select events → Export), the export can fail to include images, video, and the generated HTML detail/frame pages for an event, with no clear error shown in the UI. The underlying PHP log shows repeated cp: cannot stat ... Permission denied errors for files that do exist and are readable.
Root cause
In web/skins/classic/includes/export_functions.php, inside exportEvents(), the top-level export directory is created and explicitly locked down with a chmod:
$export_dir = ZM_DIR_EXPORTS.'/'.$export_root.($connkey?'_'.$connkey:'');
...
chmod($export_dir, 0700);
But the per-event subdirectory created just below it is not:
$event_dir = $export_dir.'/'.$event->Id();
if (!(@mkdir($event_dir) or file_exists($event_dir))) {
  ZM\Error("Can't mkdir $event_dir");
}
// no chmod() call here
Since no explicit mode is passed to mkdir(), the resulting permissions on $event_dir depend entirely on the umask in effect for the PHP process at that moment. On systems where that umask is restrictive enough to remove all execute bits (e.g. resulting in 0660/drw-rw----), the directory becomes untraversable — even for root. This is a documented Linux kernel behavior: CAP_DAC_OVERRIDE does not bypass the directory search/execute check when a directory has zero execute bits set for anyone, so root-owned PHP processes (e.g. running under php-cgi as root) are blocked exactly the same as any other user.
The result is every subsequent file operation into $event_dir (copying JPEGs, writing zmEventDetail.html, zmEventFrames.html, etc.) fails with Permission denied, and the export silently produces an incomplete or empty archive.
To Reproduce
1. Have a PHP-FPM/php-cgi environment with a umask that removes execute bits from newly created directories with default 0777 mode (e.g. a strict umask such as 0117). 2. Export any event from the ZoneMinder web UI (Console → Export, or the Download button). 3. Check /var/tmp/zm/zmExport_<connkey>/<event_id> — its permissions will lack execute bits entirely. 4. Check the web server / PHP error log for repeated cp: cannot stat ...: Permission denied lines referencing files inside that directory.
Expected behavior
The per-event export subdirectory should have the same explicit, predictable permissions as the top-level export directory, regardless of the PHP process's umask.
Suggested fix
Add an explicit chmod() call after creating $event_dir, mirroring the existing handling of $export_dir:
[...]

PR #5077, opened 2026-08-29
fix: keep the per-event export dir to the owner and off the flat path

AJ0070:
refs #5074 rather than fixes, since per your analysis this is not what causes the reported symptom.

The per-event export directory was the only mkdir in the export path with no chmod after it, so its mode came from the umask: 0755 under the usual 0022, leaving event footage world readable behind a 0700 parent. It is 0700 now like the other two here and all three in download_functions.php, and the three chmods report failure instead of failing silently. The mkdir and its continue moved inside the non-flat branch, since a flat export copies into $export_dir directly and never uses that directory.

Tests: `php tests/php/test_export_dir_permissions.php`.

### DIFF
--- a/tests/php/test_export_dir_permissions.php
+++ b/tests/php/test_export_dir_permissions.php
@@ -1,13 +1,23 @@
 <?php
-// Exporting events created the per-event subdirectory with a bare mkdir(), so
-// its mode came from the PHP process umask. A strict umask leaves it with no
-// execute bit, and a directory with no execute bit cannot be traversed even by
-// root, because the search check is the one thing CAP_DAC_OVERRIDE does not
-// bypass. Every copy into it then failed with permission denied and the export
-// came out incomplete with nothing useful in the UI. Issue #5074, reported with
-// this diagnosis by techrockedge.
+// Exports hold event footage, and the export path keeps its directories to the
+// owning user: ZM_DIR_EXPORTS and the per-export directory are both chmod 0700,
+// as are all three directories in download_functions.php. The per-event
+// directory was created with a bare mkdir(), so its mode came from the umask
+// instead, leaving it 0755 under the usual 0022. The footage was world readable
+// and only the 0700 parent above it was keeping anyone out. Issue #5074.
+//
+// Note on the umask: parent permissions do not propagate to a child directory,
+// only a POSIX default ACL would do that and ZM sets none, so the umask alone
+// decides the mode.
+//
+// The source checks below read the file with token_get_all() rather than
+// matching text. Regex versions of these passed on the very regressions they
+// name: one matched any chmod near a mkdir regardless of which directory it
+// applied to, one skipped any mkdir that carried a mode argument, and one never
+// looked at the mode at all, so reverting 0700 to 0755 went unnoticed.
 // Run: php tests/php/test_export_dir_permissions.php   (from the repo root)
 
+$root = __DIR__.'/../..';
 $failures = 0;
 function check($label, $ok, $detail = '') {
   global $failures;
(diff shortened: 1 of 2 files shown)


connortechnology:
Thanks, and thanks to @techrockedge for the report. I fetched the branch: `php -l` clean on both files, test passes 6/6 as a normal (non-root) user.

I want the chmod — but for a different reason than the one in the comment, and there's a regression to fix first.

## 1. The `continue` regresses flat exports

`$event_dir` is used only at `:943-945` and `:954`, nowhere else in the file. Under `$exportStructure == 'flat'` (`:960`) every copy goes to `$export_dir.'/'.$file` or `$export_dir.'/'.$event->Id().'_'.$file` — never into `$event_dir`. So a flat export never needed that directory, and after this change a failed mkdir skips an event that would have exported perfectly well.

Either move the mkdir/chmod into the non-flat branch, or only `continue` when `$exportStructure != 'flat'`.

## 2. The CAP_DAC_OVERRIDE explanation is wrong, and it's being written into the source

The new comment says the directory is "not traversable even by root: the search check is the one thing CAP_DAC_OVERRIDE does not bypass". That isn't so. In `generic_permission()`, for `S_ISDIR` CAP_DAC_OVERRIDE grants access unconditionally; the "at least one exec bit must be set" rule applies only to non-directories. Tested against a `drw-rw----` directory:

```
euid=0  write into 0660 dir: OK (1 bytes)
scandir: .,..,probe.txt
```

Please drop that paragraph from both the code comment and the test header rather than shipping it as documentation.

**The justification I would put there instead is Isaac's:** exports contain event footage, and the policy in this codebase is that export directories are owned-user-only. `chmod(..., 0700)` at `:915` and `:923` is *tightening*, not enabling — under the default umask 0022 those directories would be 0755. `web/includes/download_functions.php:46,50,60` locks all three of its directories the same way. `$event_dir` is the only directory in the export path without it, so today it is 0755: world-readable footage, saved in practice only by the 0700 parent above it. That is worth closing on its own merits, whatever the reporter's umask turns out to be.

Two related notes, since they affect whether this actually closes #5074:

- Parent directory permissions never propagate — I checked, parent 0700 and parent 0777 both give a child of 0755 under umask 022. Only a POSIX *default ACL* on the parent would do that, and nothing in ZM sets one. So umask alone decides, and for any ordinary umask (0022 -> 0755, 0077 -> 0700) the directory is traversable and this chmod is not what fixes the reported symptom.
- @techrockedge: `cp` stats the destination as well as the source, so `cp: cannot stat ...: Permission denied` has identical wording for both. The path inside the message is the discriminator. If it names `/var/tmp/zm/zmExport_<connkey>/<eid>/...` the diagnosis holds; if it names `/var/lib/zoneminder/events/...` this is a source-side problem and the export chmod won't help. Could you post one full line? And given the above, if php-cgi really is running as root then a 0660 export directory would not have blocked it, so it would be worth confirming the effective uid at the point of failure: `<?php printf("umask=%04o uid=%d euid=%d\n", umask(), posix_getuid(), posix_geteuid());` served through the real web path. On Void, `grep -i umask /proc/<pid>/status` for the php and `runsvdir` pids will show which layer sets the odd mask; `0117` in particular is the classic "remove execute from files" mistake, since umask cannot distinguish files from directories.

## 3. The source-scanning guard has a false negative

The window is `array_slice($lines, $i, 12)` starting at the mkdir line, and `strpos($window, 'chmod(')` does not check which mkdir the chmod belongs to. `mkdir(ZM_DIR_EXPORTS)` at `:912` and `chmod($export_dir, ...)` at `:923` are eleven lines apart, so I deleted `chmod(ZM_DIR_EXPORTS, 0700)` at `:915` and the check still reported PASS. It does correctly fail when the `$event_dir` chmod is removed, so it protects this fix — just not the other two it claims to cover. Pairing each mkdir with a chmod on the same variable would close it.

## 4. Smaller things

- The second source check is a literal-text regex, `/Can\'t mkdir \$event_dir"\);\s*\n\s*continue;/`. Any rewording of the error string or reflowing of those lines breaks it, and it then fails with a message about export permissions that has nothing to do with what changed.
- The "fixed" half of the behavioural test does its own `mkdir` + `chmod` in a temp dir rather than calling into ZoneMinder, so it asserts that PHP's `chmod` works. The reproduction half is genuinely valuable; the fix half isn't. Understandable given `exportEvents()` needs a DB, config constants and `exec` — worth a comment saying so, so nobody later mistakes it for behavioural coverage.
- `chmod()`'s return is unchecked. If it fails, the export fails in exactly the way being fixed, silently. The two existing chmods have the same gap so you are consistent, but a `ZM\Error` on failure is what would make the next report of this diagnosable.

## Not this PR

The "fails silently" half of the issue title is still open: `exec($cmd, $output, $return)` at `:972` routes `$return` only into `ZM\Debug`, so cp failures never surface and `exportEvents()` still hands the user an archive. Separately `$output` is never reset between iterations and PHP's `exec()` appends to a non-empty array, so that Debug line grows with every file in the export. Both pre-existing — worth a follow-up issue.

Fix 1 and 2 and I'm happy to merge this.

AJ0070:
Both fixed: the mkdir and continue are inside the non-flat branch, and the CAP_DAC_OVERRIDE paragraph is gone from the code and the test header, replaced with the footage-permissions reason.

Your guard finding was worse than the one case you hit. It also passed when a mkdir grew a mode argument and the chmod was deleted, and it never looked at the mode at all, so reverting 0700 to 0755 went unnoticed. Rewrote the checks on token_get_all instead of matching text, and mutation tested each. chmod returns are checked on all three now, and #5085 covers the exec output.

PR #5081, opened 2026-08-29
fix: stop the events table loading forever when the query fails

AJ0070:
refs #3301.

Covers the half you asked for: "if the ajax fails we need to display a useful message instead of just loading, please wait". The events table's error handler only console.logged and never called hideLoading, so a query that 500s on the memory limit left the table saying "Loading, please wait" for good. It now hides the indicator and goes through logAjaxFail, same as the log table. Aborts still stay silent, since every reload aborts the request in flight.

Not the memory side. @SteveGilvarry's root cause there is that the events query has no SQL LIMIT and materialises every matching row, which is a bigger change.

Tests: `node tests/js/events-ajax-error.test.js`, 3 assertions driving the real ajaxRequest.


connortechnology:
Thanks. I fetched the branch: `node tests/js/events-ajax-error.test.js` passes 3/3 and `npx eslint` is clean on both changed files.

Right diagnosis and the right place, small diff. `table` at `events.js:69` is the module-level const from `events.js:10`, so no stray re-query. The abort early-return is correct — the superseding request owns the overlay, which is the reasoning `console.js:181-183` already spells out. Good to see the dead commented-out code go. The test is a real one too: it drives the shipped `ajaxRequest` through `vm` rather than reimplementing the handler, and it separates abort from failure.

Two things I'd like before merging.

**1. Nothing is actually displayed to the user, so the ask in #3301 isn't met yet.**

The issue asks to "display a useful message instead of just loading, please wait", and the PR body quotes that line as the half this covers. But `logAjaxFail` (`skin.js:1163`) is console-only. `log.js:107-115`, which you cite as the model, does three things — `zmAlert(...)`, `hideLoading`, `logAjaxFail`. This copies the last two and drops the one the user can see. Net effect: the spinner vanishes and the table sits there unchanged with nothing saying why.

It's worse in the exact scenario from the issue. `web/ajax/events.php:2` sets `ini_set('display_errors', '')`, so a memory-limit fatal returns a 500 with an empty body. `logAjaxFail` then takes its `!jqxhr.responseText` branch at `skin.js:1170` and prints "Ajax request failed. No responseText." — so for the one case this is aimed at, even the console output is useless.

Please add `zmAlert(...)` the way `log.js:112` does. That needs an `ErrorUpdatingEventTable` string in `web/lang/en_gb.php` (reuse the existing `AJAXRequestError`) and a `translate` block in `events.js.php`, which doesn't have one yet — `log.js.php:1-8` is the pattern.

**2. The sibling `success` path has the identical spinner-forever bug and isn't touched.**

Three lines above your fix, `events.js:56-59`:

```js
if (data.result == 'Error') { alert(data.message); return; }
```

That returns without calling `params.success()`, so bootstrap-table never clears loading — same "Loading, please wait" forever. And it is reachable: `ajaxError()` (`web/includes/functions.php:1812`) defaults to HTTP **200** with `{result:'Error'}`, and `web/ajax/events.php` uses it for permission failures (line 11) and unrecognised tasks. One `table.bootstrapTable('hideLoading')` in that branch plus a fourth test case and #3301 is closed on both paths instead of one.

**Nit:** `logAjaxFail(jqXHR)` drops its other two arguments — the signature is `(jqxhr, textStatus, error)`, so the console line reads "Request Failed: undefined, undefined". jQuery hands the error callback all three, so forwarding them is free. `log.js:115` has the same slip, so you're at least consistent with the model.

Not for this PR, but worth a separate issue: `console.js:183-187` routes a 401 to login rather than leaving a dead table behind an expired session. The events list is exactly where sessions expire.

AJ0070:
Both fixed. Added the zmAlert the way log.js does it, with an ErrorUpdatingEventTable string and a translate block in events.js.php, and the success path's result=Error branch now hides the loading overlay too. Fifth test case covers that one.

Also forwarded textStatus and errorThrown. Left log.js alone since #5069 is still open in that file. Filed #5083 for the 401 case.

PR #5091, opened 2026-09-02
Fix: Do not hide the control button "#mfbpanel" for filter=inline mode, since this panel, in addition to the filter, has other necessary controls on the MontageReview page

IgorA100:
Similar to #4999

### DIFF
--- a/web/skins/classic/views/montagereview.php
+++ b/web/skins/classic/views/montagereview.php
@@ -308,18 +308,12 @@ getBodyTopHTML();
     <div id="header">
 <?php
 $filter_inline = filterSettingsInline();
-$html = '';
-// In inline mode the filter panel sits at the top of the page and this flip
-// icon is what hides/shows it. In sidebar mode the panel lives in the sidebar
-// extruder, which has its own show/hide control, so the flip icon is omitted.
-if ($filter_inline) {
-  $html .= '<a class="flip" href="#"
+$html = '<a class="flip" href="#"
            data-flip-control-object="#mfbpanel"
            data-flip-control-run-after-func="applyChosen drawGraph"
            data-flip-control-run-after-complet-func="changeScale">
              <i id="mfbflip" class="material-icons md-18" data-icon-visible="filter_alt_off" data-icon-hidden="filter_alt"></i>
            </a>'.PHP_EOL;
-}
 $html .= '<div id="mfbpanel" class="'.($filter_inline ? '' : 'hidden-shift ').'container-fluid">'.PHP_EOL;
 echo $html;
 // The monitor attribute filters pick which cameras are shown and are rarely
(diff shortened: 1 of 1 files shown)


connortechnology (commented):
Thanks for this — the diagnosis is right, and the master code is mine to answer for.

## First hunk: correct, and fixes a real bug

`#mfbpanel` is not just the filter panel on this page. It closes at `</div><!--flipMontageHeader-->` (montagereview.php:418), so `#ScaleDiv`, `#SpeedDiv`, `#ButtonsDiv` (pan / zoom / 24h / 8h / 1h / All Events / Live / Fit / Download Video) and `#timelinediv` are all inside it, and they stay behind when `insertControlModuleMenu()` (skin.js:2505) lifts the filter out into the extruder — it only relocates `#mfbpanel .controlHeader` and `#fieldsTable`.

So in sidebar mode master rendered the panel `hidden-shift` (`position:absolute; left:-999em`) with no flip icon, and every one of those controls was off screen with nothing to bring it back. Exactly as your title says.

53f6a67fd is the culprit: it flipped `if (!$filter_inline)` to `if ($filter_inline)` when the right answer was no condition at all. I checked that inline mode regained its toggle and didn't notice the panel carries non-filter controls. #4999 had already settled this for montage.php and watch.php.

Worth noting you kept `($filter_inline ? '' : 'hidden-shift ')` rather than hardcoding `hidden-shift` the way #4999 did — that's the right call.

## Second hunk: this one loses the monitor attribute filters in sidebar mode

`insertControlModuleMenu()` builds the sidebar filter block from exactly two nodes:

```js
const filterOne = document.querySelector('#mfbpanel .controlHeader');
const filterTwo = document.querySelector('#fieldsTable');
```

`filterOne` is the `<div class="controlHeader">` that `buildMonitorsFilters()` emits (_monitor_filters.php:192) — i.e. the contents of `#monitorFilterBar`. Not rendering it in sidebar mode doesn't move it to the sidebar, it deletes it: sidebar users lose the Group / Name / Source / Status / Server / Storage filters on montage review. The Monitor dropdown survives only because it is spliced into `#fieldsTable`.

Two things mask that:

- `#fieldsTable` itself carries `class="filterTable controlHeader"` (Filter.php:1186), so `#mfbpanel .controlHeader` silently falls through to it — `filterOne === filterTwo`, the double `prepend` is a harmless no-op, and nothing errors.
- If `count($filter->terms()) == 0` — reachable with an explicit `?filter=` in the request, which skips the default-term block at montagereview.php:181-195 — both are null, and `filter.prepend(null)` stringifies into a literal `"null"` text node in the extruder, twice. One of those already exists on master.

The motivation is sound though: with the first hunk the panel becomes visible, exposing a MonitorFilters toggle over an emptied container. The fix is to gate only the `<a class="flip">` anchor and keep the `<div id="monitorFilterBar">` unconditional, so the function still has something to relocate.

## Follow-up

I've opened #5094 with your first hunk unchanged plus that reworked second hunk, and two small hardenings in `insertControlModuleMenu()` — naming `#monitorFilterBar .controlHeader` directly so the selector can't silently fall through to `#fieldsTable`, and skipping an absent node rather than handing `null` to `prepend()`.

Neither is browser-verified yet, so a look in both filter modes would be welcome. Happy to close that in favour of an updated #5091 if you'd rather carry it yourself.

PR #5094, opened 2026-09-04
fix: reach the montage review panel controls in sidebar filter mode

connortechnology:
Supersedes #5091, keeping its first hunk and reworking the second. Thanks @IgorA100 — the diagnosis there is right and the master code was mine to fix.

## The bug

`#mfbpanel` is not only the filter panel on montage review. It closes at `</div><!--flipMontageHeader-->`, so `#ScaleDiv`, `#SpeedDiv`, `#ButtonsDiv` (pan / zoom / 24h / 8h / 1h / All Events / **Live** / **Fit** / **Download Video**) and `#timelinediv` are all inside it, and they stay behind when `insertControlModuleMenu()` lifts the filter out into the sidebar extruder.

In sidebar mode (`ZM_WEB_FILTER_SETTINGS_POSITION` not `inline`, left navbar) the panel is rendered `hidden-shift` — `position:absolute; left:-999em` — with no flip icon, so every one of those controls was off screen with nothing to bring it back.

53f6a67fd flipped `if (!$filter_inline)` to `if ($filter_inline)` when the right answer was no condition at all; it verified inline mode regained its toggle and missed that the panel carries non-filter controls. #4999 had already settled this for montage.php and watch.php.

## Changes

**Render the `#mfbpanel` flip icon in both modes** — same as montage.php and watch.php. This is #5091's first hunk unchanged.

**Emit the MonitorFilters toggle inline only, but keep `#monitorFilterBar` in both modes.** With the panel now reachable in sidebar mode, that toggle would sit over a container whose contents have been moved to the extruder, so it is inline-only. The container itself has to stay: it is where `insertControlModuleMenu()` collects the attribute filters from, so omitting it does not move them to the sidebar, it drops the Group / Name / Source / Status / Server / Storage filters from montage review entirely. With no toggle in sidebar mode it carries the `hidden-shift` the toggle would otherwise have applied, keeping the emptied container out of the visible panel.

This is where this PR differs from #5091, which gates the container along with the toggle.

**Name the monitor filter bar directly in `insertControlModuleMenu()`.** `#fieldsTable` also carries `class="filterTable controlHeader"` (`Filter.php:1186`), so the previous `#mfbpanel .controlHeader` silently matches it when the bar is absent — `filterOne === filterTwo`, the double `prepend` is a no-op, and losing the attribute filters looks like it worked.

**Skip an absent node instead of passing it to `prepend()`**, which stringifies `null` into a literal `"null"` text node in the extruder. Reachable today: a request-supplied `filter` with no terms skips the default-term block and renders no `#fieldsTable`.

## Testing

`php -l` clean on the view; `npx eslint web/skins/classic/js/skin.js` clean. The rest is source reading — the local instance here serves an installed copy rather than this checkout, so **this still wants a look in the browser in both filter modes**: inline (panel visible, both toggles work) and sidebar (panel toggleable, scale/speed/buttons/timeline reachable, attribute filters present in the extruder).

https://claude.ai/code/session_01Nr76CednxtDt2nPuq6WrbL

### DIFF
--- a/web/skins/classic/js/skin.js
+++ b/web/skins/classic/js/skin.js
@@ -2507,10 +2507,16 @@ function insertControlModuleMenu() {
     filter = document.createElement('div');
     filter.setAttribute("id", "filterMontagereview");
 
-    const filterOne = document.querySelector('#mfbpanel .controlHeader');
+    // Named precisely: #fieldsTable also carries .controlHeader, so a looser
+    // '#mfbpanel .controlHeader' silently matches it instead when the monitor
+    // filter bar is absent, which hides the fact that the attribute filters
+    // never made it into the extruder.
+    const filterOne = document.querySelector('#monitorFilterBar .controlHeader');
     const filterTwo = document.querySelector('#fieldsTable');
-    filter.prepend(filterTwo);
-    filter.prepend(filterOne);
+    // Either can be absent - a filter with no terms renders no #fieldsTable -
+    // and prepend() stringifies null into a literal 'null' text node.
+    if (filterTwo) filter.prepend(filterTwo);
+    if (filterOne) filter.prepend(filterOne);
   } else if (currentView == 'watch') {
     destroyChosen();
     filter = document.querySelector('.controlHeader form');
--- a/web/skins/classic/views/js/montagereview.js
+++ b/web/skins/classic/views/js/montagereview.js
@@ -1309,13 +1309,18 @@ function loadEventData(e) {
       const name = el.attr('name');
 
       if (name) {
-        const found = name.match(/filter\[Query\]\[terms\]\[(\d)+\]\[val\]/);
+        const found = name.match(/filter\[Query\]\[terms\]\[(\d+)\]\[val\]/);
         if (found) {
           const attr_name = 'filter[Query][terms]['+found[1]+'][attr]';
-          const attr = this.form.elements[attr_name];
           const op_name = 'filter[Query][terms]['+found[1]+'][op]';
-          const op = this.form.elements[op_name];
-          if (attr) {
+          // Looked up by name rather than through this.form: in sidebar filter
+          // mode insertControlModuleMenu() moves #fieldsTable out of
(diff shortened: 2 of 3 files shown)


connortechnology:
@IgorA100 what do you think?

IgorA100:
> @IgorA100 what do you think?

I'll look into it more thoroughly and let you know in a few days. I'm really busy with work right now :(

IgorA100:
Everything is fine, but you also need to edit the `montagereview.js` file.
Replace the code
```
      if (name) {
        const found = name.match(/filter\[Query\]\[terms\]\[(\d)+\]\[val\]/);
        if (found) {
          const attr_name = 'filter[Query][terms]['+found[1]+'][attr]';
          const attr = this.form.elements[attr_name];
          const op_name = 'filter[Query][terms]['+found[1]+'][op]';
          const op = this.form.elements[op_name];
          if (attr) {

```
With:
```
      if (name) {
        const found = name.match(/filter\[Query\]\[terms\]\[(\d+)\]\[val\]/);
        if (found) {
          const attr_name = 'filter[Query][terms]['+found[1]+'][attr]';
          const op_name = 'filter[Query][terms]['+found[1]+'][op]';
          const attr = document.querySelector('[name="' + attr_name + '"]');
          const op = document.querySelector('[name="' + op_name + '"]');

          if (attr && op) {
```

Old code doesn't work when using the filter in the left sidebar because `<FORM>` is missing.
The new code is more versatile and works seamlessly in both cases.

In the regular expression, `(\d)+` was replaced with `(\d+)`. This is more accurate, as we get a single group of digits, not several groups, which we later use as `found[1]`.

I didn't create a separate PR, but published the necessary changes here. I believe all changes should be implemented in one PR, not in separate ones.

IgorA100:
@connortechnology 
Yes, I missed another place in 'montagereview.js' that needed some changes.
Cloud found it.
I'll check this PR tomorrow. I think everything should work correctly.

PR #5098, opened 2026-09-05
fix: name the six event audit columns that render as raw keys

singhharsh1708:
Six column headers in the event audit report call `translate()` with keys that no lang file carries, so the table shows `FirstEvent`, `LastEvent`, `MinGap`, `MaxGap`, `MissingFiles` and `ZeroSize` beside the translated `Id`, `Name`, `Server` and `Events`.

Adds the six to `en_gb.php` rather than renaming the keys to spaced English, because `ru_ru`, `es_la` and `ja_jp` already translate all six under these names and a rename would drop those translations.


connortechnology:
Alternatively, we could adjust the event audit to add a space in each so that no translation is necessary for english.

singhharsh1708:
Done, spaced keys in the view instead, en_gb untouched.

singhharsh1708:
These six are already translated in ru_ru, es_la and ja_jp, so renaming the keys drops those. Went back to adding the English entries.

PR #5107, opened 2026-09-06
fix: escape monitor names emitted into javascript

singhharsh1708:
Seven of the per-view JavaScript files interpolate the monitor name straight into a quoted JS string, so a name containing an apostrophe emits `'name': 'Bob's Garage',` and the file stops parsing. Montage, watch, cycle, zone, zones, montagereview and user all build their monitor object that way, and montage does the same for saved layout names.

Such a name cannot come from the monitor form, which strips the apostrophe through the `filter_regexp` in `Monitor.php`, but the API applies no character rule, and layout names have no filter at all.

These files already escape their neighbours, `validJsStr` on Path and SecondPath in `watch.js.php` and `validHtmlStr` on the name at `montagereview.js.php:159`, so this uses `validJsStr` to match. The emitted snippet fails `node --check` today and parses once escaped.

### DIFF
--- a/web/skins/classic/views/js/cycle.js.php
+++ b/web/skins/classic/views/js/cycle.js.php
@@ -16,7 +16,7 @@ foreach ( $monitors as $monitor ) {
 ?>
 monitorData[monitorData.length] = {
   'id': <?php echo $monitor->Id() ?>,
-  'name': '<?php echo $monitor->Name() ?>',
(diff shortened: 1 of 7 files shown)


singhharsh1708:
Correction to the example above: the monitor form strips the apostrophe, so a name like that has to arrive through the API, which only checks Name is non-blank. Layout names have no filter either.

PR #5108, opened 2026-09-06
fix: escape names and paths written into html

singhharsh1708:
Five places write a user-editable name or path into HTML without escaping: the monitor name in the event view and twice in the timeline key, the report name and the storage name, path and url inside `value="..."`, and the control name in the control caps table, whose next line already escapes Protocol.

A name containing a double quote closes the attribute, and the browser then parses what follows as further attributes. Fed `Front" onmouseover="alert(1)` to `DOMDocument`, the input tag today parses as two attributes; with `validHtmlStr` it parses as one whose value is the literal text.

Same helper and same shape as 47d7e0d37, which escaped Zone, Event and Server names.

### DIFF
--- a/web/ajax/modals/storage.php
+++ b/web/ajax/modals/storage.php
@@ -59,15 +59,15 @@
           <tbody>
             <tr>
               <th class="text-right " scope="row"><?php echo translate('Name') ?></th>
-              <td><input type="text" name="newStorage[Name]" value="<?php echo $newStorage->Name() ?>"/></td>
+              <td><input type="text" name="newStorage[Name]" value="<?php echo validHtmlStr($newStorage->Name()) ?>"/></td>
             </tr>
             <tr>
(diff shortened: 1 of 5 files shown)


IgorA100:
Monitor names are checked in `/web/ajax/monitor.php` using the regular expression `'Name' => array('type'=>'text','filter_regexp'=>'/[^\w\p{L}\p{M}\p{N}\-\.\(\)\:\/ ]/u', 'default'=>'Monitor')'`, from `/web/includes/Monitor.php`.
I think monitor name filtering should only happen in one place, otherwise it'll be a bit of a mess

singhharsh1708:
That filter is in ZM_Object, so it covers the web path only: the API model just requires Name non-blank, no character rule. Report and Control names and Storage Url carry no filter at all, and all three are in this diff.

PR #5120, opened 2026-09-07
fix: stop orphaned zms processes accumulating, without killing streams on tab hide

connortechnology:
Fixes the orphaned `zms` accumulation reported in #4706, without killing streams on tab hide — which @connortechnology ruled out there, since killing takes the status polling and alarm sounds down with it.

**Reproduced and verified on a live instance.** Two monitors on a montage page, one simulated tab hide/show cycle each time:

| | before | after 1 cycle | after 2 cycles | processes |
| --- | --- | --- | --- | --- |
| stock `zms` + stock `MonitorStream.js` | 2 | **4** | **6** | replaced each cycle, old ones orphaned |
| with this branch | 2 | **2** | **2** | **the same two, resumed** |

The pre-fix run is exactly the accumulation #4706 describes. With the branch the process ids do not change at all across the cycles, and both pictures are live afterwards.

## The two defects

**A stopped `zms` could never exit** (`src/zm_monitorstream.cpp:670`). The stopped branch slept and `continue`d without writing anything, so the socket was never touched and a client that had gone away was never discovered: no write, no `EPIPE`, no `SIGPIPE`. The `continue` also skipped the `ttl` check at the bottom of the loop, so even a caller's deadline did not apply. The only way out was an explicit `CMD_QUIT`, and a stream that missed one stayed until the machine was restarted. This is why they accumulate rather than merely appear.

The last captured frame is now held on entering the stopped state and re-sent every five seconds, as the paused state already does, and `ttl` applies here too. Confirmed in the logs of a running stream:

```
DB1 [Got STOP command]
DB1 [Saving stopped image from index 0]
DB2 [Sending keepalive frame while stopped]      <- every 5s thereafter
```

`setLastViewed` is deliberately still not called: capture and decoding should not be held active for a stream that is not playing.

**`select_zms()` replaced the connkey without quitting the process it addressed** (`web/js/MonitorStream.js`). Once the key is gone nothing can reach that `zms` again, so no `CMD_QUIT` can ever be delivered to it. `getStreamCmdResponse()` already learned this — its reload path calls `quitConnKey()` first, with a comment saying exactly why — but the path every ordinary `start()` takes did not. `quitConnKey` had precisely two references in the file: its own definition and that one use.

## Resuming instead of rebuilding

The third commit makes the hide/show path *resume* the stopped `zms` with `CMD_PLAY` rather than build a new one — what `CMD_STOP` is designed for, and what the discussion on #4706 assumed should work. It did not, for three separate reasons, none of which were visible from reading:

1. `stop()` clears `activePlayer`, which is the only thing the resume branch tests, so after a stop it was unreachable.
2. `srcAuthCurrent` required a non-empty `zmAuth.hash`, which is `''` whenever auth is off — so those installs always rebuilt, for no reason, since there is no hash that can go stale.
3. `streamCommand()` drops anything sent while `!started`, and `started` is not set until the end of `select_zms()`. The resume issued its `CMD_PLAY` into nothing. Resuming after a *pause* worked only because `pause()` leaves `started` set.

With those fixed the `zms` status reports `stopped: 0` and ~15 fps — and the picture still looked frozen, because the rebuild path clears the `"Loading..."` info block from `img_onload`, which cannot fire when `src` never changes. The resumed stream was playing underneath its own overlay. Clearing the block on the resume path is the last piece.

@IgorA100 reported on #4706 that a `CMD_PLAY` after `CMD_STOP` does not restart the video. That was correct, and points 3 and 4 above are why.

## Testing

- `zms` builds clean, no new warnings; full C++ suite **138/138 pass**; ESLint clean on `MonitorStream.js`.
- Live before/after process counts as tabled above, on a 2-monitor montage page.
- Keepalive behaviour confirmed in `zms` debug logs.
- Resume confirmed end to end: same pids across two cycles, `stopped: 0` and ~15 fps from the stream status query, and the burnt-in camera timestamp advancing in the browser afterwards.

## Relationship to #4706

Supersedes it. Same diagnosis of the symptom, different mechanism: nothing is killed on tab hide, and the process that does get replaced is now told to quit rather than left behind.

### DIFF
--- a/src/zm_monitorstream.cpp
+++ b/src/zm_monitorstream.cpp
@@ -562,6 +562,10 @@ void MonitorStream::runStream() {
   Image *paused_image = nullptr;
   SystemTimePoint paused_timestamp;
 
+  // Same, for the stopped state. See the stopped branch in the loop below.
+  Image *stopped_image = nullptr;
+  SystemTimePoint stopped_timestamp;
+
   if (connkey && (playback_buffer > 0)) {
     // 15 is the max length for the swap path suffix, /zmswap-whatever, assuming max 6 digits for monitor id
     const int max_swap_len_suffix = 15;
@@ -668,10 +672,49 @@ void MonitorStream::runStream() {
       continue;
     }
     if (stopped) {
-      // In stopped state, do nothing except wait for a new command.
-      // Don't call setLastViewed() so we don't keep capture/decoding active unnecessarily.
+      // Stopped halts playback but keeps the connection, so that a later
+      // command can resume on the same process. Hold the last frame and
+      // re-send it every few seconds, as the paused state does.
+      //
+      // The write is the point: with nothing written, this branch never
+      // touches the socket, so a stream whose client has gone away is never
+      // told about it and never sees EPIPE. Since it also skipped the ttl
+      // check at the bottom of the loop, such a process had no way to exit at
+      // all and stayed until the machine was restarted. refs #4706
+      //
+      // setLastViewed is deliberately still not called: capture and decoding
+      // should not be held active for a stream that is not playing.
+      if (!stopped_image
+          && monitor->shared_data->valid
+          && (monitor->shared_data->last_write_index != monitor->image_buffer_count)) {
+        int index = monitor->shared_data->last_write_index % monitor->image_buffer_count;
+        Debug(1, "Saving stopped image from index %d", index);
+        stopped_image = new Image(*monitor->ReadShmFrame(index));
+        stopped_timestamp = SystemTimePoint(zm::chrono::duration_cast<Microseconds>(monitor->shared_timestamps[index]));
+      }
+      if (now - last_frame_sent > Seconds(5)) {
+        if (stopped_image) {
+          Debug(2, "Sending keepalive frame while stopped");
+          if (!sendFrame(stopped_image, stopped_timestamp)) zm_terminate = true;
+        } else if (sendTextFrame("Stopped") <= 0) {
+          // Nothing captured yet to hold, but we still have to write something
+          // to notice a client that is no longer there.
+          Debug(2, "Failed to send stopped keepalive text frame");
+          zm_terminate = true;
+        }
+      }
+      // The ttl check lives at the bottom of the loop, which this branch never
+      // reaches, so a stopped stream would otherwise outlive its own deadline.
+      if (ttl > Seconds(0) && (now - stream_start_time) > ttl) {
+        Debug(2, "Stopped and now - start > ttl. break");
+        break;
+      }
       std::this_thread::sleep_for(MAX_SLEEP);
       continue;
+    }
+    if (stopped_image) {
+      delete stopped_image;
+      stopped_image = nullptr;
     }
     monitor->setLastViewed();
     if (frame_type == FRAME_ANALYSIS)
--- a/src/zm_monitorstream.cpp
+++ b/src/zm_monitorstream.cpp
@@ -558,12 +558,14 @@ void MonitorStream::runStream() {
   std::string swap_path;
   bool buffered_playback = false;
 
-  // Last image and timestamp when paused, will be resent occasionally to prevent timeout
-  Image *paused_image = nullptr;
+  // Last image and timestamp when paused, will be resent occasionally to prevent timeout.
+  // unique_ptr because the loop below has break paths that leave the function
+  // while one of these is still held.
+  std::unique_ptr<Image> paused_image;
   SystemTimePoint paused_timestamp;
 
   // Same, for the stopped state. See the stopped branch in the loop below.
-  Image *stopped_image = nullptr;
+  std::unique_ptr<Image> stopped_image;
   SystemTimePoint stopped_timestamp;
 
   if (connkey && (playback_buffer > 0)) {
@@ -689,13 +691,13 @@ void MonitorStream::runStream() {
           && (monitor->shared_data->last_write_index != monitor->image_buffer_count)) {
         int index = monitor->shared_data->last_write_index % monitor->image_buffer_count;
         Debug(1, "Saving stopped image from index %d", index);
-        stopped_image = new Image(*monitor->ReadShmFrame(index));
+        stopped_image = zm::make_unique<Image>(*monitor->ReadShmFrame(index));
         stopped_timestamp = SystemTimePoint(zm::chrono::duration_cast<Microseconds>(monitor->shared_timestamps[index]));
       }
       if (now - last_frame_sent > Seconds(5)) {
         if (stopped_image) {
           Debug(2, "Sending keepalive frame while stopped");
-          if (!sendFrame(stopped_image, stopped_timestamp)) zm_terminate = true;
+          if (!sendFrame(stopped_image.get(), stopped_timestamp)) zm_terminate = true;
         } else if (sendTextFrame("Stopped") <= 0) {
           // Nothing captured yet to hold, but we still have to write something
           // to notice a client that is no longer there.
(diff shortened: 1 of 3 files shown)


connortechnology:
Tested on a live instance and **dropped one of the three commits** as a result.

Before/after, two monitors on a montage page, one hide/show cycle each time:

| | start | 1 cycle | 2 cycles |
| --- | --- | --- | --- |
| stock `zms` + stock `MonitorStream.js` | 2 | 4 | 6 |
| this branch | 2 | 2 | 2 |

The keepalive works as intended — `Got STOP command`, `Saving stopped image from index 0`, then `Sending keepalive frame while stopped` every 5s, process alive and reachable throughout.

**What I got wrong:** the third commit made hide/show resume the stopped `zms` with `CMD_PLAY` rather than rebuilding. On paper that is what `CMD_STOP` is for. In practice the connkey and process are preserved, `started` returns to true — and the picture stays frozen on the keepalive frame. The streaming process logs `viewing fps: 0.00` and goes on sending keepalives; no `CMD_PLAY` ever reaches it.

@IgorA100 said exactly this on #4706 and I dismissed it as the client re-creating the stream. It wasn't. Resuming on an existing connkey needs the command delivery path investigated first, which is a separate piece of work and not a `MonitorStream.js` change. That commit is gone; the branch now rebuilds as before, but quits the old process instead of abandoning it.

connortechnology:
Update: the resume commit is back, and now actually works. My earlier comment said resuming a stopped `zms` was not achievable from `MonitorStream.js` — that was wrong, I stopped one layer short.

Three things were preventing the `CMD_PLAY` from ever arriving:

1. `stop()` clears `activePlayer`, the only thing the resume branch tests.
2. `srcAuthCurrent` demands a non-empty `zmAuth.hash`, which is `''` when auth is off — so those installs always rebuilt, though there is no hash that can go stale.
3. **`streamCommand()` drops anything sent while `!started`**, and `started` is not set until the end of `select_zms()`. The resume was issuing its `CMD_PLAY` into nothing. Resuming after a *pause* works only because `pause()` leaves `started` set.

With those fixed the stream status came back `stopped: 0`, ~15 fps — and the picture *still* looked frozen. That last part was a red herring of the same kind: the rebuild path clears the `"Loading..."` info block from `img_onload`, which cannot fire when `src` never changes, so the resumed stream was playing underneath its own overlay.

Live result on a 2-monitor montage page, two hide/show cycles:

- the same two `zms` pids throughout — not replaced, not orphaned
- both pictures live afterwards, camera timestamp advancing

So hide/show now costs nothing to return from: no new process, no restart latency, and the status polling that drives alarm sounds is never interrupted. @IgorA100 — your observation on #4706 that `CMD_PLAY` after `CMD_STOP` does not restart the video was right; points 3 and 4 are why.

connortechnology:
Both review points are real, and both are broader than flagged. Fixed in 13b0ed1ca.

**The `stopped_image` leak.** Correct, and `paused_image` immediately above it has had the same defect for as long as the paused keepalive has existed — freed only by a later iteration observing that the state changed, so any `break` leaves the function with it allocated. Both are now `std::unique_ptr<Image>`, which makes the exit path stop mattering rather than enumerating it.

**The suppressed comment about `break` leaving `zm_terminate` false.** This one is the more serious of the two and it is not limited to the line it was raised on. `StreamBase::checkCommandQueue()` (src/zm_stream.cpp:124) loops on `!zm_terminate` around a `recvfrom` with a one second `SO_RCVTIMEO`, so it returns only once that flag is set, and `runStream()` joins it unconditionally whenever `connkey` is set. Any break that leaves the flag false sits in `join()` indefinitely.

The stopped-state break this branch added was the third instance, not the first. The `ttl` and `frames_to_send` breaks at the bottom of the loop have always had it:

```cpp
if (ttl > Seconds(0) && (now - stream_start_time) > ttl) {
  ...
  break;                    // zm_terminate still false
}
if (frames_to_send > 0 && frame_count >= frames_to_send) {
  break;                    // likewise
}
```

So a stream that ended on its own deadline or its own frame count did not exit — which is another mechanism for exactly the accumulation this PR is about, independent of the stopped-state one it set out to fix. All three now set the flag, matching what the `feof`/`ferror` and `sendFrame`-failure breaks already did.

**Testing.** `zms` builds clean with no new warnings, C++ suite 133/133 on this branch.

I could not reproduce the join hang end to end: no `zmc` is running on this machine, and a stream against a monitor with no capture takes the not-initialised path, which sets the flag correctly on its own 60 second timeout and so exits normally. The verification here is by reading — `checkCommandQueue`'s sole loop condition is the flag, and `join()` is unconditional — rather than by observation, which is a weaker result than the live before/after process counts in the PR description and I would rather say so than imply otherwise.

connortechnology:
Rebased onto master and now reproduced, which I could not do when I posted the fix — there was no `zmc` running on that machine at the time.

Same build of `zms` before and after the change in f46cc1bcd, against a live capturing monitor, run as `www-data` under `timeout`:

| request | before | after |
| --- | --- | --- |
| `mode=jpeg&monitor=3&frames=1` | **exit 124 at 20.5s** (killed) | **exit 0 at 1.2s** |
| `mode=jpeg&monitor=3&ttl=5` | **exit 124 at 25.7s** (killed) | **exit 0 at 5.3s** |

Both produced their output correctly in both builds — one JPEG part and ~27 KB for `frames=1`, ~1.37 MB of stream for `ttl=5`. The difference is entirely in whether the process then exits. Before the change it finished its work and sat in `command_processor.join()` until something killed it.

`frames=1` is worth dwelling on: that is the console thumbnail request. Every one of them left a `zms` behind that never exited on its own. That is a second mechanism for the accumulation in #4706, entirely separate from the stopped-state one this branch set out to fix, and it predates the branch.

Rebase is clean, no conflicts. `811010ee6` dropped from the branch since it landed on master separately. Re-verified on the new base: `zms` builds with no new warnings, C++ suite 146/146, ESLint clean on `MonitorStream.js`.

connortechnology:
Second review: the suppressed comment is a real bug and is fixed in 1361f9380, with the JS tests it asked for. The C++ test request I am pushing back on.

**The dropped `CMD_PLAY` on the initial `mode=paused` path.** Correct, and it is the same defect as point 3 in the PR description — I fixed the resume branch and missed the branch directly below it, which has the same shape:

```js
} else if (srcAuthCurrent && (-1 != stream.src.indexOf('mode=paused'))) {
  this.streamCmdTimer = setInterval(...);
  this.streamCommand(CMD_PLAY);     // started is still false here
}
```

`streamCommand()` returns early on `!this.started` (MonitorStream.js:1751) and the tail of `select_zms()` does not set it until after this branch has run. With auth on, reaching it needed a still-valid hash, so it was intermittent. With auth off the `srcAuthCurrent` change on this branch makes it the path *every* initial load takes, so this branch turned an intermittent bug into a reliable one. Fixed the same way as the resume branch.

**JS tests.** Added `tests/js/monitorstream-resume.test.js`, in the style of the existing `monitorstream-kill.test.js`. It asserts what reaches the wire rather than which branch ran, which is what would have caught this:

- resume sends exactly one `CMD_PLAY` on the existing connkey, and leaves `src` and `connKey` untouched
- initial `mode=paused` sends exactly one `CMD_PLAY` on its connkey — this is the regression
- auth off takes the resume path instead of rebuilding
- a stale auth hash still rebuilds and quits the process the old connkey addressed

Removing just the one-line fix fails the initial-paused case and leaves the other three passing, so it is pinned to the defect rather than to the code shape. Full JS suite passes, ESLint clean.

**The C++ test request I am not doing.** `tests/zm_monitorstream.cpp` covers `MonitorStreamBufferLevel`, a free function. The stopped-state keepalive and TTL exit live inside `MonitorStream::runStream()`, which needs a `Monitor` with attached shared memory, a capture daemon writing frames, and a bound command socket. Testing it at that level means either restructuring `runStream()` around a seam that exists only for the test, or standing up most of a ZoneMinder install inside Catch2 — and the second would land in the suite CI actually runs, where neither is available.

What I did instead is measure it, which I posted above: `frames=1` and `ttl=5` against a live capturing monitor, before and after, exit 124 at 20.5s/25.7s versus exit 0 at 1.2s/5.3s. I would rather say the process-exit guarantee is covered by a reproducible measurement than add a test that needs a seam invented for it.

PR #5130, opened 2026-09-12
fix: qualify Approx as Catch::Approx in audio detector tests

SteveGilvarry:
`tests/zm_audio_detector.cpp`, added in c7deb3faa, uses the bare `Approx(...)` matcher. That is the Catch2 v2 spelling — v3 moved it into the `Catch` namespace — so the file does not compile against Catch2 3.x:

```
tests/zm_audio_detector.cpp:47:70: error: use of undeclared identifier 'Approx'; did you mean 'Catch::Approx'?
```

Four call sites, all in that one file. The rest of `tests/` is already on v3 idiom — `zm_config.cpp` writes `Catch::Approx`, and `zm_catch2.h` includes `catch2/catch_all.hpp` — so this was the only holdout.

### Why CI missed it

No workflow passes `-DBUILD_TEST_SUITE=ON`, so `tests/` is never configured or compiled in CI. `ci-build-werror.yml` builds with `ENABLE_WERROR=ON` but leaves the test suite off, which means a test file can stop compiling without anything going red. Worth considering as a separate change — adding the test suite to one CI job would have caught this at the source.

### Verification

Built against Catch2 3.15.2 on macOS (arm64, Homebrew). Before the change the object fails to compile; after it, it builds clean and the full suite passes:

```
./tests/tests "~[notCI]"
All tests passed (12480 assertions in 144 test cases)
```

Note the suite has to be run from the build's `tests/` directory — the font tests load `data/fonts/*` by relative path.

https://claude.ai/code/session_01B5KL9Xbi7K5aGsauLtd8tG

### DIFF
--- a/tests/zm_audio_detector.cpp
+++ b/tests/zm_audio_detector.cpp
@@ -44,7 +44,7 @@ TEST_CASE("Audio RMS of signed 16-bit samples") {
 
   SECTION("a full scale square wave is full scale") {
     const auto samples = SquareS16(32767, 64);
-    REQUIRE(AudioDetector::RmsS16(samples.data(), samples.size()) == Approx(1.0).margin(0.001));
+    REQUIRE(AudioDetector::RmsS16(samples.data(), samples.size()) == Catch::Approx(1.0).margin(0.001));
   }
 
   SECTION("the most negative sample does not exceed full scale") {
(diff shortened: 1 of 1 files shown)


SteveGilvarry:
Closing as a duplicate of #5129, which landed first and is a strict superset: it carries the identical `Catch::Approx` hunk plus the `tests/main.cpp` removal, `find_package(Catch2 3 REQUIRED)`, and a `ci-cpp-tests.yml` job so this cannot rot unnoticed again.

Verified #5129 on macOS (arm64, Homebrew Catch2 3.15.2) before closing: clean build, `./tests "~[notCI]"` passes 12480 assertions in 144 test cases.

PR #5137, opened 2026-09-13
fix: make the Perl scripts' taint-safe PATH configurable

SteveGilvarry:
Follow-up to the note in #5136, which documented a workaround for this rather than fixing it.

The Perl scripts run under `-T`, so they cannot trust the caller's `PATH` and set their own. Sixteen of them hardcode:

```perl
$ENV{PATH}  = '/bin:/usr/bin:/usr/local/bin';
```

That assumes everything they shell out to lives under `/usr` or `/usr/local`. `ZoneMinder::General::findDbCommand` looks for a database client on that `PATH`, and `zmupdate.pl` and `zmcamtool.pl` run whatever it finds — so an install whose client sits anywhere else cannot apply schema changes:

```
Upgrading DB to 1.39.34 from 1.39.34
sh: mysql: command not found
Command 'mysql -u'zmuser' -p'zmpass' -hlocalhost zm < .../zm_update-1.39.34.sql' exited with status: 127
```

even though both `mariadb` and `mysql` are on the caller's `PATH`. Homebrew on Apple Silicon is the case that surfaced it — the client is in `/opt/homebrew/bin` — but a `--prefix=/opt` install on Linux is the same shape, as is anything keeping its database client outside the FHS locations. It is the same class of assumption as `FindGSOAP` only looking under `/usr/share` and `/usr/local/share`.

### The change

`@ZM_SCRIPT_PATH@` in place of the literal, defaulting to the same three directories plus wherever cmake actually found a client:

```
-- Perl script PATH: /bin:/usr/bin:/usr/local/bin:/opt/homebrew/bin
```

Overridable with `-DZM_SCRIPT_PATH=` for packagers who want to pin it exactly. Configure now also warns when no client is found at all, since that failure otherwise turns up much later and reports something unrelated to the cause.

**Nothing changes for a conventional install.** The directory is appended only when it is not already in the list, so a client under `/usr/bin` leaves the default byte-for-byte as it was.

`Memory.pm` is deliberately untouched. It has its own narrower `/bin:/usr/bin`, and the only command it runs is `uname` through an absolute path from `ZM_PATH_UNAME`, so widening it would buy nothing.

Worth saying explicitly: every directory on this list has to be one only root can write to, or perl's taint checks refuse to run anything at all. That is a reason to keep it overridable rather than derive it from something like the install prefix, which a packager may not control. The doc string says so.

### Verification

Three cases on macOS 26, Apple Silicon:

| | result |
|---|---|
| client in `/opt/homebrew/bin` | appended |
| `-DZM_SCRIPT_PATH=/bin:/usr/bin` | respected verbatim |
| detection pointed at `/usr/bin/mariadb` | default unchanged |

Build clean, Catch2 suite 146 cases. I also ran the equivalent change against a live install before writing this — patching the installed `zmupdate.pl` by hand took it from the failure above to `Database successfully upgraded to version 1.39.34`, and from there to a running system with events recording.

https://claude.ai/code/session_01B5KL9Xbi7K5aGsauLtd8tG

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -877,6 +877,40 @@ if(WITH_SYSTEMD)
   endif()
 endif()
 
+# The taint-safe PATH the Perl scripts run external commands with.
+#
+# They run under -T, so they cannot trust the caller's PATH and set their own.
+# That was hardcoded to the FHS locations, which assumes everything they shell
+# out to lives under /usr or /usr/local. ZoneMinder::General::findDbCommand
+# looks for a database client that way, and zmupdate.pl and zmcamtool.pl run it,
+# so an install whose client sits elsewhere - Homebrew on Apple Silicon puts it
+# in /opt/homebrew/bin, and a --prefix=/opt build is the same shape - fails with
+# "sh: mysql: command not found" even though the client is on the caller's PATH.
+#
+# Default to the FHS locations plus wherever a client was actually found, and
+# let a packager override the lot. Every directory named here has to be one only
+# root can write to, or perl's taint checks will refuse to run anything at all.
+set(_zm_script_path "/bin:/usr/bin:/usr/local/bin")
+find_program(ZM_DB_CLIENT_EXECUTABLE NAMES mariadb mysql)
+if(ZM_DB_CLIENT_EXECUTABLE)
+  get_filename_component(_zm_db_client_dir "${ZM_DB_CLIENT_EXECUTABLE}" DIRECTORY)
+  string(FIND ":${_zm_script_path}:" ":${_zm_db_client_dir}:" _zm_db_client_found)
+  if(_zm_db_client_found EQUAL -1)
+    set(_zm_script_path "${_zm_script_path}:${_zm_db_client_dir}")
+  endif()
+  mark_as_advanced(ZM_DB_CLIENT_EXECUTABLE)
+  unset(_zm_db_client_dir)
+  unset(_zm_db_client_found)
+else()
+  message(WARNING
+    "No mariadb or mysql client found. zmupdate.pl applies schema changes by "
+    "running one, so database upgrades will fail until one is on ZM_SCRIPT_PATH.")
+endif()
+set(ZM_SCRIPT_PATH "${_zm_script_path}" CACHE STRING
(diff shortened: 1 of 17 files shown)


SteveGilvarry:
Pushed a follow-up commit after checking the security properties of what I'd written, and I'd got this wrong.

Auto-appending wherever a client was found widens who can decide what the daemons execute, and **perl will not catch it**. Its taint check refuses a world writable directory on `PATH` but permits a group writable one:

```
world writable 0777 -> DIED: Insecure directory in $ENV{PATH} while running with -T switch
group writable 0775 -> ran OK
```

That is not theoretical. Homebrew's `bin` is `0775` owned by group `admin`:

```
drwxrwxr-x  stevengilvarry  admin  /opt/homebrew/bin
```

So as originally written this PR would have silently handed every member of `admin` the ability to place a `mysql` binary that the ZoneMinder daemons then run — exactly the thing the hardcoded `PATH` existed to prevent. On a single-admin workstation those users can `sudo` anyway, so it is not a boundary crossing there, but it is a real weakening on any multi-user machine, and a default that does it without saying so is the wrong shape.

The new commit checks the directory before adding it, leaves it out when it is writable by group or world, and says why:

```
CMake Warning: Found a database client at /opt/homebrew/bin/mariadb, but
/opt/homebrew/bin is writable by group or world and has not been added to
ZM_SCRIPT_PATH: anyone who can write there would choose what the Perl daemons
run. Tighten its permissions, or set ZM_SCRIPT_PATH explicitly if you accept that.
```

An explicit `-DZM_SCRIPT_PATH=` still wins, so anyone who accepts the risk can opt in — the difference is that it is now a decision rather than a default.

Verified: the real Homebrew `bin` is refused with that warning, a `0755` directory is still appended, `-DZM_SCRIPT_PATH=` overrides both, build clean.

Worth flagging one consequence for review. On Intel macOS, and anywhere else `/usr/local/bin` is owned by a user rather than root, the *existing* default already contains a group writable directory — that predates this PR and I have not touched it, but it is the same exposure and may deserve its own look.

https://claude.ai/code/session_01B5KL9Xbi7K5aGsauLtd8tG

